PR #1, opened 2010-09-21
Add support for earlier version of PHP5

j4mie:
This just removes the one place where a closure is used in the main Slim class, and provides an example bootstrap-php5.php file. For those of us not yet using PHP5.3.

### DIFF
--- a/bootstrap-php5.php
+++ b/bootstrap-php5.php
@@ -1,0 +1,45 @@
+<?php
+
+require('slim/Slim.php');
+
+/* INIT SLIM */
+
+Slim::init();
+
+/* CALLBACKS */
+
+Slim::before('example_before');
+function example_before() {
+	Slim::response()->write('Before!');
+}
+
+Slim::after('example_after');
+function example_after() {
+	Slim::response()->write('After!');
+}
+
+/* ROUTES */
+
+Slim::get('/', 'get_example');
+function get_example() {
+	Slim::render('index.php');
+}
+
+Slim::post('/post', 'post_example');
+function post_example() {
+	echo '<br/><br/>Here are the details about your POST request:<br/><br/>';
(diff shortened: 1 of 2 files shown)


codeguy:
Just merged in your changes. After looking through the new code, I figure Slim should now work with PHP 4 >= 4.3, but I'm not entirely sure on this until someone can confirm. So for now I'm leaving the official version requirement as PHP >= 5.0. Thanks for your help on this!

j4mie:
Thanks for the merge! This is definitely PHP5 only, as PHP4 doesn't support `public`/`private`/`protected` visibility on methods/properties, doesn't support constructors named `__construct`, etc.

codeguy:
Yeah I don't know why I was thinking PHP 4.x would work. But I've since updated the class headers to clarify that PHP 5 is required. My next big step is probably going to be setting up PHPUnit and writing tests to improve test coverage.

kolanos:
I don't know if I would worry about <5.3 incompatibility. One of the things that would make Slim interesting is that it is 5.3+ and that it makes full use of the 5.3 features. Just my $0.02.

codeguy:
@Kolanos Thus far Slim works great with 5.3, making good use of anonymous functions. So if you are a 5.3 user, Slim will work great. However, if you are using < 5.3, Slim will still work but with a slightly less elegant syntax. So currently Slim will work either way. I think this is a good thing especially for those on shared hosts that are hesitant (or unable) to upgrade to 5.3 just yet. The only big thing Slim doesn't get from 5.3 is namespaces... but I'm not really convinced Slim needs them since Slim is only really intended for smaller web services. Down the road I may introduce a 5.3-only branch. But for now my goal is to let as many people as possible enjoy Slim. Thanks for your input! Let me know if you have any other questions.

PR #3, opened 2010-09-22
Multiple Router pattern support

lukemorton:
I was playing with slim and straight away wanted the same callable in use for two patterns. First I solved this with:
    function get_page($page = 'index')
    {
    }

```
Slim::get('/', 'get_page');
Slim::get('/:page', 'get_page');
```

Then I realised an array of patterns would be much more DRY:

```
Slim::get(array('/', '/:page'), function() {
});
```

But sometimes the arguments may be swapped about or completely different, so I added two methods to Router, Router::params() and Router::param(). This allows you to get hold of the params from within the callable, Router::param even lets you define a default.

Also I could not see the purpose in have two init functions, why not just have the view class definition as optional within the one init? So I have made that change also.

Great work so far though, seriously useful!!
Luke

### DIFF
--- a/slim/Router.php
+++ b/slim/Router.php
@@ -128,6 +128,26 @@ class Router {
 		}
 	}
 	
+	/**
+	 * Return the parameter names and values for this Route
+	 *
+	 * @return array
+	 */
+	public function params() {
+		return $this->params;
+	}
+
+	/**
+	 * Return a parameter or default if needed
+	 *
+	 * @return string
+	 */
+	public function param($key, $default = NULL) {
+		if (isset($this->params[$key])) {
+			return $this->params[$key];
+		}
+		return $default;
+	}
 }
 
 /**
@@ -189,7 +209,7 @@ class Route {
 	 */
 	public function __construct( $pattern, $method, $request, $callable, $conditions = array() ) {
 		
-		$this->pattern = ltrim($pattern, '/');
+		$this->pattern = $pattern;
 		$this->method = $method;
 		$this->request = $request;
 		$this->callable = $callable;
@@ -197,21 +217,34 @@ class Route {
 		$this->params = array();
 		$this->matched = false;
 		
-		//Extract URL params		
-		preg_match_all('@:([\w]+)@', $this->pattern, $paramNames, PREG_PATTERN_ORDER);
-		$paramNames = $paramNames[0];
+		// If pattern not an array, make it an array to work in foreach
+		if ( !is_array($this->pattern) ) {
+			$this->pattern = array($this->pattern);
+		}
 		
-		//Convert URL params into regex patterns, construct a regex for this route
-		$patternAsRegex = preg_replace_callback('@:[\w]+@', array($this, 'convertPatternToRegex'), $this->pattern);
-		$patternAsRegex = '@^' . $patternAsRegex . '$@';
-				
-		//Cache URL params' names and values if this route matches the current HTTP request
-		if( $this->method == $this->request->method && preg_match($patternAsRegex, $this->request->resource, $paramValues) ) {
-			array_shift($paramValues);
-			foreach( $paramNames as $index => $value ) {
-				$this->params[substr($value, 1)] = urldecode($paramValues[$index]);
+		foreach ($this->pattern as $_pattern) {
+			
+			// Pattern
+			$_pattern = ltrim($_pattern, '/');
+			
+			//Extract URL params
+			preg_match_all('@:([\w]+)@', $_pattern, $paramNames, PREG_PATTERN_ORDER);
+			$paramNames = $paramNames[0];
+	
+			//Convert URL params into regex patterns, construct a regex for this route
+			$patternAsRegex = preg_replace_callback('@:[\w]+@', array($this, 'convertPatternToRegex'), $_pattern);
+			$patternAsRegex = '@^' . $patternAsRegex . '$@';
+	
+			//Cache URL params' names and values if this route matches the current HTTP request
+			if( $this->method == $this->request->method && preg_match($patternAsRegex, $this->request->resource, $paramValues) ) {
+				array_shift($paramValues);
+				foreach( $paramNames as $index => $value ) {
+					$this->params[substr($value, 1)] = urldecode($paramValues[$index]);
+				}
+				$this->matched = true;
+				break;
 			}
-			$this->matched = true;
+		
 		}
 		
 	}
@@ -257,7 +290,6 @@ class Route {
 	public function params() {
 		return $this->params;
 	}
-	
 }
 
 ?>
--- a/slim/Slim.php
+++ b/slim/Slim.php
@@ -91,31 +91,25 @@ class Slim {
 		$this->settings = array();
 	}
 	
-	/***** INITIALIZERS (Choose One) *****/
+	/***** INITIALIZER *****/
 	
 	/**
 	 * Initialize Slim
 	 *
 	 * This instantiates the actual Slim app and sets a default
 	 * 404 Not Found handler.
+	 *
+	 * Optionally can set a view class on initialisation
 	 */
-	public static function init() {
+	public static function init($viewClass = NULL) {
 		self::$app = new Slim();
 		self::notFound(array('Slim', 'defaultNotFound'));
+		
+		if ( !is_null($viewClass) ) {
+			self::view($viewClass);
+		}
 	}
-	
-	/**
-	 * Initialize Slim with View
-	 *
-	 * Along with initializing Slim (see above), this also sets the View
-	 * class used to render templates.
-	 *
-	 * @param string $viewClass The name of the view class Slim will use
-	 */
-	public static function initWithView($viewClass) {
-		self::init();
-		self::view($viewClass);
-	}
+
 	
 	/***** ROUTING *****/
 	


codeguy:
Hi. Thanks a lot for your suggested changes. I agree 100% about the simplified Slim::init() function. An oversight on my part. As for the multiple patterns for one callable... I see what you mean but I want to think on it more tonight before I pull in the changes. Will comment here again once I pull. Thanks!

codeguy:
So here's what I'm thinking... Let's say you have two routes '/one' and '/two/:foo' where the second route expects a url param named "foo" as the second segment. Both should map to a single callable. Currently, URL params are passed into the callback function as parameters, in order of appearance. I would like to keep it this way. I also don't like the syntax of passing an array as the first parameter. It may be DRY but it just seems dirty from a syntax point of view. One of my goals with Slim is to keep the syntax clean and simple. What are your thoughts on this?

```
Slim::get('/one', '/two/:foo', function ($foo = NULL) {
    //Do something here
});
```

This way both routes map to the same callable, URL params are still available as callback parameters, and the syntax is cleaner. To accomplish this, you would only need to pop off the last parameter of Slim::get(), assuming it is the callable object. All previous parameters to Slim::get() are assumed to be route patterns. All of your other changes are great.

I also like your changes to Router, specifically having both Router::params and Router::param($key, $default)... I assume you intend these to be used like:

```
Slim::get('/one', '/two/:foo', function ($foo = NULL) {
    $fooFromFunction = $foo;
    $fooFromRouter = Slim::router()->param('foo', 'defaultFoo');
});
```

Correct? Let me know what you think! Thanks again for your help.

lukemorton:
I like it, it looks cleaner! It's still DRY with your solution the only problem that could arise is that if I held an array of routes somewhere I could not easily pass them to Slim::get(). Arrays are easy to throw more data into the as and when needed, even in a different location. However I think I'm being pedantic now and since what I was trying to achieve only needed two patterns it won't really affect me too much, it's just an array allows you to be slightly more dynamic.

As for the param() and params() method, they would allow you to do something like this in the routes:

```
Slim::get('/one/:foo', '/two/:bob', function () {
    Slim::router()->param('foo', 'defaultFoo');
    Slim::router()->param('bob', 'defaultFoo');
});
```

Two differently named route variables, and it simplifies working out the callable arguments.

Overall I'm happy with your vision of my suggestions so feel free to pull and modify, I'll take another look tonight.

Regards,
Luke

j4mie:
One possible thing to consider when implementing this:

I have been thinking about the idea of adding optional named routes and reverse routing to Slim for generating URLs in a DRY way.

This would work as follows. When defining your routes, you would pass an optional name for the route as the last argument:

```
Slim::get('/hello/:name', function ($name) {
    echo $name;
}, 'hello');
```

This name would be stored in the router. To use this, you could call:

```
Slim::reverse('hello', array('name' => 'Jamie');
```

This would substitute each of the values in the second parameter into the pattern associated with the route name. So it would return the string `/hello/Jamie`.

This means if you modify the pattern associated with your  functions, any URLs generated in your application would automatically change.

What do you think of this idea? Do you think it would be a useful addition? If so, how would this be impacted by the idea of mapping multiple patterns to the same view function? I can see several options here:
1. Simply disallow reversing patterns when multiple patterns are mapped to the same function.
2. Pass in an index into the (proposed) `reverse` function to reverse the patterns in order. So if your function is mapped to two patterns, `Slim::reverse('hello', array('name' => 'Jamie), 0)` would reverse the first, and `Slim::reverse('hello', array('name' => 'Jamie'), 1)` would reverse the second. Obviously, different arrays of arguments could be passed in to each one.
3. Name each _pattern_ separately, instead of each mapping. I'm not sure how the syntax for this would work yet. Perhaps (using codeguy's suggested syntax) the patterns passed as the first arguments to the mapping functions could (optionally) be arrays, containing both the function and the pattern:
   
   `Slim::get(array('/hello/:name', 'hello'), array('/hi/:name', 'hi'), function () { echo $name; });`

The last two options seem a bit too complex to me, so (assuming you like this proposal) maybe we should just go with the first one. Does anyone else have any better ideas?

More generally (and in response to the original idea), in the real world, how often are people likely to map the same function to multiple URLs? What are the use cases for this?

Jamie

lukemorton:
I like the idea of Slim::reverse, although perhaps it should be accessible through the Router object rather than Slim itself, i.e. `Slim::router()->reverse()`, although that is a bit more verbose it makes sense that it is part of the routing logic.

I would propose that reverse routing would work exactly like it does the right way round that is of this:

```
Slim::get('/foo/:bar', '/foo/:bar/:jar', function() {
    // ...
}, 'foo');
```

`Slim::reverse` would try to match the patterns in the order they are defined. So for no defined parameters the first route is always matched `Slim::reverse('foo')`. Where as `Slim::reverse('foo', array('foo' => 'something', 'jar' => 'something else'))` would match the second.

Real world cases are having varying routes use the same logic such as my initial use:

```
Slim::get('/', '/:page', function() {
    Slim::render(Slim::router()->param('page', 'index').'.php');
});
```

But this is also useful in many other applications where you have indexes, or optional levels of routing such as:

```
Slim::get('/:controller', '/:controller/:action', '/:controller/:action/:id', function() {
    // ...
});
```

Luke

j4mie:
That looks like a good solution. What would happen if two patterns had the same set of arguments?

I understand those use cases now. I think the basic problem is that anonymous functions are by definition not reusable, so you have to do everything you need to do with them right where they are defined. I suppose to keep Slim simple this would be the right approach.

One other quick idea that might be a possible solution here - if Slim allowed internal requests (invoking the routing system without making an actual HTTP request) then you could have a simple set of cascading routes that would accomplish what you need:

```
Slim::get('/', function () {
    Slim::internalRequest('/index');
});

Slim::get('/:page', function () {
    // Do your page processing here
});
```

Each route could then still have a separate name.

This would also have the added advantage of allowing HMVC (like Kohana 3).

lukemorton:
Two patterns with the same set of arguments? Then there should only be one pattern, unless I'm misunderstanding the question.

I think somewhat internal requests might be providing too much of a solution. Although a handy feature I'm not sure they are required at the moment, you could always redirect the request.

I love Kohana don't get me wrong, but this is a lightweight RESTful solution. If you need Kohana features, use Kohana.

Luke

j4mie:
I was thinking more along the lines of having the same function at two different URL prefixes - `/hello/:name` and `/hi/:name` for example. Maybe that wouldn't be very common though.

I agree 100% that Slim should be kept lightweight. I'm just trying to help think of the cleanest/simplest solution. The simple one-to-one mapping between callables and patterns does seem like a benefit that needs to be carefully thought about before changing things.

Sinatra takes the approach of using an array of routes: http://www.sinatrarb.com/faq.html#multiroute

If Slim does go down this road, I actually think codeguy's solution is cleaner in PHP's syntax. To address your concerns (passing an array of routes), Slim could support both approaches. If the first argument to the mapping functions is an array, Slim could iterate over it and map each pattern to the callable. If the first argument(s) are strings, it could pop off the callable and then map to each string. 

I can think of one possible problem (assuming reverse routing is added). The mapping function would just receive an array of arguments, all of which might be strings. The last argument might be a route name, or it might be a callable. How would it distinguish between the two? We can't use `is_callable` because it's quite possible that the route name might also be a valid callable..

lukemorton:
My pull request allowed an array of routes as Sinatra does, Josh requested that they each be an extra argument to the method which might look slightly cleaner but is less extensible. However I do not personally mind which way this decision is taken.

As for your reverse routing issue, I am not so sure... an array for patterns would solve this issue, however it is an issue on a piece of code that does not exist yet. I'm sure Josh will have an opinion :)

codeguy:
All nice ideas. Heading into work, but I'll think about this today and post my thoughts later tonight.

codeguy:
Hey guys,

After thinking a while on this... first, I completely agree about adding in named routes. This would make templating much easier; even if the routes themselves change, the URLs in the templates will remain consistent when using a helper method such as "url_for()" or something similar as you have already demonstrated.

However, I'm not entirely satisifed yet with the suggested implementations of having multiple routes tied to a single callback, all within a single Slim::get() call. It's getting too complicated, which is something I want to avoid at all costs. I don't intend for Slim to be the most advanced framework, but instead the most approachable and the easiest to use for 80% of use-cases. I'm ok if the extra 20% of the time you have to write a little extra code.

I would rather the burden of multiple routes with a single callback be placed on defining the multiple routes rather than condensing everything into a single Slim::get() call. It is very easy to define a separate function (regular or anonymous) which can be passed into each separate route definition. This would be DRY with regards to the single definition of the callback function. But each route would need to be defined separately, like this:

```
$callback = function ($name) {
    //Do something here
};
Slim::get('/hello/:name', $callback);
Slim::get('/hi/:name', $callback);
```

Of course, there is some duplication of effort here, but I think you retain readability and ease of use at the expense of a little extra effort. I of course will continue to think on this and if a better solution presents itself in the future I'll be sure to re-think this.

Now, for named routes. My near-future plans were to add an additional $opts parameter to Slim::get, Slim::post, Slim::put, and Slim::delete. This $opts paramater would be an array that, at the very least, could be used to define conditions for route parameters, like this:

```
Slim::get('/book/:id', $callback, array('name' => '\d+'));
```

This $opts parameter could of course be converted like this for further extensibility:

```
Slim::get('/book/:id', $callback, array(
    'conditions' => array('name' => '\d+')
);
```

So this leaves us with several options. We could always add 'name' as a key in the $opts array, like this:

```
Slim::get('/book/:id', $callback, array(
    'name' => 'show_book',
    'conditions' => array('name' => '\d+')
);
```

But this seems overly verbose if you only need to specify a name. Or we could generate a new first-class static function in Slim specifically for named routes, like this:

```
Slim::named('GET', 'show_book', '/book/:id', $callback);
```

The only additional requirement for this approach is the need to specify the HTTP request type as the first parameter and the name as the second parameter. What do you think?

lukemorton:
Perhaps to keep the simplicity of the REST specific methods you could allow the name to be specified as a string, however if you want to add more complex configuration you pass it as an array:

```
Slim::get('/book/:id', $callback, 'book');
Slim::get('/book/:id', $callback, array(
    'name' => 'book',
    'conditions' => array('id' => '\d+')
));
```

This allows basic named functionality, along with more complex configuration which would be fairly complex anyway. Since how would you specify advanced options to the Slim::named()? Yet another parameter? IMO it breaks away from a RESTful orientation to have another function not named after it's HTTP request type i.e. Slim::named().

Alternatively you could return an object from Slim::get() so that you could call methods against it for further configuration, i.e.

```
Slim::get('book/:id', $callback)
    ->name('book')
    ->conditions(array('id' => '\d+'));
```

That seems cleaner, but obviously adds more logic to the core workings of Slim.

codeguy:
Great idea! I really like this idea:

```
Slim::get('book/:id', $callback)
    ->name('book')
    ->conditions(array('id' => '\d+'));
```

And it is more readable and still extendable. We could simply return the Route object from the Slim::get() method, and expose the new methods on the Route object. Anyone have any objections?

codeguy:
I'll mockup some code tonight on a separate branch and push it up when ready.

j4mie:
That's a lovely solution. I'm a big fan of fluent interfaces :)

This allows the basic case to be kept very simple, but allows as much extensibility for the future as we need. Great idea.

codeguy:
Hi guys. Just pushed a new branch called "named-routes". This is a major refactoring of the Router and Route code. Small changes were necessary in Slim so the route is returned from Slim::get, Slim::post, Slim::put, and Slim::delete, but nothing else. These changes also let me decouple Request from Route. I've added Route::name() and Route::conditions() as public instance methods. They work and are callable, but currently don't do anything just yet. My goals tomorrow evening are to get these two methods functional. Thanks!

codeguy:
I have pushed several commits which enable support for named routes and route regular expression conditions. See the bootstrap file for an example.

lukemorton:
Nice additions!

j4mie:
Looks great. My original attempt at implementing named routes was almost identical. 

Do you have a roadmap or some rough ideas for future additions? I'm guessing that increasing the test coverage will be the next priority..

codeguy:
Thanks! I'm going to manually update the View class as mentioned at the top of this thread rather than pull in the remote branch with a bunch of other changes that are no longer needed. Also, this is my first open source project, so I'm not really knowledgable with open source etiquette. I want to make sure credit is given where credit is due. So feel free to toss in some of your own @author doc comments whenever some of your own code is pulled in. I'm going to go ahead and close out this thread.

codeguy:
@j4mie Just missed your comment when closing the thread. My _very_ next step is to setup an account with some sort of online ticket system so we can organize what is getting done and by whom. Do you have any suggestions for online (hopefully free) ticketing software? Once that is setup, I'll put together a roadmap. And yes, test coverage is probably my next big task since the routing engine is fairly stable at this point. I've got the foundation in place with PHPUnit in the tests/ directory.

j4mie:
Why not just use Github Issues? Or maybe set up a wiki page for the roadmap, then promote each item to an issue when it comes time to implement it? As this thread shows, the issues system seems to work well as a discussion board for design/architecture ideas.

codeguy:
ha. Forest for the trees. Also my first time using GitHub for a big project, obviously :) Sounds good. I'll put together a road map tonight (you guys can do the same with features you'd like to see in Slim) and we can compare notes when ready. Thanks!

codeguy:
I have opened a new issue at http://github.com/codeguy/Slim/issues/issue/5 where we can start discussing Slim's RoadMap.

PR #7, opened 2010-09-28
Quick cleanup from using a local install - this helped when running from a subdirectory

christianchristensen:


codeguy:
Thanks! I'll pull in these changes during lunch.

codeguy:
Just pulled in the changes.

PR #18, opened 2010-09-29
HaangaView

jrivero:
Require Haanga.php and Haanga folder from the project of cesar:

http://github.com/crodas/Haanga/tree/master/lib/

Haanga is a excellent template engine like Twig but more optimal and have a Django syntax

I reuse the test of named routes for example of haangaview in bootstrap_haanga.php


codeguy:
Thanks! I will be out of the country until Monday, but I will pull in your code when I get back. Ultimately, my goal is not to include the actual third-party library code with Slim, but to simply provide example custom View class files for each third party library. The end user could then download the third party library and then integrate it using the example custom View files provided in the Slim download.

But I'll be sure to pull your code and run through your tests and add your HaangaView to the Slim repo. Thanks!

jrivero:
I like your ideas

You know flask project? creation and distribution extensions is very simple, and the framework core is very robust.

codeguy:
I am not familiar with Flask. Do you have a link?

jrivero:
Yes! Is a great micro-framework python based.
http://flask.pocoo.org

codeguy:
Flask looks cool. Seems very similar to Sinatra, but in Python. I think its intentions are the same as Slim, and it has a few features that I intend to add to Slim soon, such as Logging, Sessions, Debug mode, etc.

jrivero:
Now I think it is better to create a slim-haanga project as an extension following the idea of flask.  Later will eliminate the code and put it on new project

PR #22, opened 2010-10-06
Changed regexp for matching URIs that contain various punctuation marks

l15n:
Makes the router mostly conform to rfc2396 "Uniform Resource Identifiers (URI): Generic Syntax"

http://www.ietf.org/rfc/rfc2396.txt

Includes tests. The tests don't cover the case of the "+" character because PHP's urldecode converts "+" characters to spaces.


codeguy:
Thanks! Just pulled in your changes.

PR #25, opened 2010-10-07
Indentation problems

j4mie:


codeguy:
Thanks! Just pulled your changes in. I'll fix the 404 issue later today after work. My next goal is to write tests for Request and Response. I'm also thinking of changing Slim::raise() to Slim::halt()... semantically, I think that makes more sense. And that should wrap up development for 1.0.

j4mie:
Great. Should we create a "Contributing to Slim" page in the wiki with style guidelines etc?

codeguy:
Absolutely. The preferred style is:
1. Tabbed indentation
2. Each method should be well documented using PHPDoc comments
3. Submitted changes should include 100% PHPUnit test coverage (or as close as is reasonably possible)

I think it would be beneficial to go into more details regarding PHP coding style. A great example is the Zend Framework's coding guidelines at http://framework.zend.com/wiki/display/ZFDEV/PHP+Coding+Standard+(draft)#PHPCodingStandard%28draft%29-Indentation. I certainly won't use all of those guidelines, nor do I agree with all of them, but that would be a great resource while writing these up. I'll start this on the Wiki tonight.

PR #33, opened 2010-10-12
Added TwigView class.

jwpage:
Points for discussion:
1. Should `TwigView` live within the `slim` directory, or within a
   `slim/views` (or something) directory?
2. How should `CustomView`s handle options? Currently, two options
   (`twigDirectory` and `twigOptions`) simply exist as public vars
   that can be overridden. Was originally using `Slim::config` but
   dropped it, as it doesn't fit with the rest of the framework.
3. Should `TwigView::getTwigEnvironment()` be a singleton function, as
   it currently is? This means if someone renders a template, then
   wants to change the `twigOptions` and re-render, it won't take
   effect. I assume that's a fairly rare use-case, so haven't worried
   about it.
4. Should the Slim repository include the templating libraries it includes
   `CustomView`s for? I'm guessing not, so I haven't included it here.

Any feedback?

### DIFF
--- a/slim/Request.php
+++ b/slim/Request.php
@@ -112,7 +112,7 @@ class Request {
 		}
 		$this->headers = $this->getHttpHeaders();
 		$this->cookies = $_COOKIE;
-		$this->isAjax = isset($request->headers['X_REQUESTED_WITH']) && $request->headers['X_REQUESTED_WITH'] == 'XMLHttpRequest';
+		$this->isAjax = isset($this->headers['X_REQUESTED_WITH']) && $this->headers['X_REQUESTED_WITH'] == 'XMLHttpRequest';
 		$this->checkForHttpMethodOverride();
 	}
 
--- a/slim/TwigView.php
+++ b/slim/TwigView.php
@@ -35,13 +35,17 @@
  *
  * The TwigView is a Custom View class that renders templates using the Twig 
  * template language (http://www.twig-project.org/).
+ *
+ * Two fields that you, the developer, will need to change are:
+ * - twigDirectory
+ * - twigOptions
  */
 class TwigView extends View {
 
     /**
-     * @var TwigEnvironment The Twig environment for rendering templates.
+     * @var array The path to the Twig code directory.
      */
-    private $twigEnvironment = null;
+    public  $twigDirectory = null;
 
     /**
      * @var array The options for the Twig environment, see 
@@ -50,10 +54,9 @@ class TwigView extends View {
     public  $twigOptions = array();
 
     /**
-     * @var array The location of the Twig code directory.
-     * Defaults to __DIR__/../lib/Twig/lib/Twig in TwigView::getTwigDirectory().
+     * @var TwigEnvironment The Twig environment for rendering templates.
      */
-    public  $twigDirectory = null;
+    private $twigEnvironment = null;
 
     public function render( $template ) {
         $env = $this->getEnvironment();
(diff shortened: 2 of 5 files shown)


codeguy:
1. Only Slim classes will be in the `slim/` directory. I know I haven't made this too clear... I'm in the process of writing a Contributing to Slim guide which clarifies all of these details.
2. Because this class will likely be written/created/used by the Slim user, I think compartmentalizing the TwigView settings within the TwigView class would be fine.... so class vars or class constants should work. Or you could always use Slim::config('foo','bar'). Not so good for decoupling, but I don't see this CustomView existing without the Slim class anyway. If you do embed the vars in the class itself, be sure to prominently feature these at the very top of the class definition so they are super easy to find.
3. Yeah I wouldn't worry too much about it. If it becomes an issue, we can refactor later.
4. No. Definitely not. I want to keep the repository as light-weight as possible with little to no dependencies. We can just include the CustomView class with Slim, or we can publish the CustomView class in the Wiki documentation with hyperlinks to the respective template library's website (my vote).

Nice work! Let me know your thoughts and we'll figure out the best way to move forward.

jwpage:
1. Gotcha.
2. Opting for class vars at this stage. The rest of the slim/ classes are decoupled from Slim, so we should keep it that way.
3. Yep.
4. Agree with you there about not including Twig in the Slim repo. 
   Keeping CustomViews with the repo (perhaps in a slim/plugins dir?) or within a separate repo (slim-plugins?) would be my vote, since then we can include unit tests for it to prevent breaking the CustomViews (and other plugins) with future API changes.

I'm relatively happy with the commits at this stage, so you're welcome to pull it into the repository or put it up on the wiki if you like.

codeguy:
Sounds great. I'll take a look this weekend and pull your changes into a `plugins/` or `extras/` directory as you suggested. Been super busy this week with client work. Hoping to get back to Slim this weekend. Thanks for your help!

codeguy:
Just pulled in your custom TwigView class. I moved it out of the `slim/` directory and into the `views/` directory. I ran your unit tests locally and it worked great. However, because the Twig library will not be distributed with the Slim Framework, I removed the TwigView unit tests so that the core tests would pass without the Twig library.

What do you think about making $twigDirectory and $twigOptions public static variables? This way one could set these in `bootstrap.php` rather than in the TwigView class definition. I'm fine with TwigView as it is, just throwing the idea out there. Thanks!

jwpage:
Definitely correct, they should be static. I'll fix that and do another pull req.

codeguy:
Do you think it should be static? This would limit the functionality of the custom View class per HTTP request, but it does keep things nice and simple. Or would you rather I tweak the Slim::init() method so that it can A) accept the name of the custom View class as it currently does, and B) also accept an object of type View. Option B would allow you to instantiate and customize the custom View before initializing the Slim app, and then pass the custom View object into the Slim::init() method. Instantiating the custom View with option B would then allow you to avoid static variables and potentially utilize various instances of the custom View throughout the Slim app. Either way is fine. One is simple, less flexible. The other a bit more complex but more flexible. Thoughts?

jwpage:
Generally you don't need to render multiple templates per HTTP request anyway, and if you wanted to do multiple ones with different functionality in the class you'd just have to set TwigView::$twigDirectory to something else before the second `render()`.

As for the the Slim::init() call, it possibly does make sense to accept either a class name and a View class. It'd allow for much more flexibility in case the developer requires it.

PR #36, opened 2010-10-15
more whitespaces ;)

hexa2k9:
so. here we go again. I guess it's not finished yet, but as far as I can see there shouldn't be anything else.
- Chris

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,60 @@
+user www www;
+worker_processes 5;
+error_log logs/error.log;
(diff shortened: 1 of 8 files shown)


codeguy:
Thanks for the white space fixes. That's something I've been meaning to do, especially once I publish the Contributor's Guide this weekend. Good call regarding `require` vs. `include` in the View. I removed the `nginx.conf` file, though. Not sure why that would need to be included in the main repository. Thanks!

Issue #35, opened 2010-10-15
isAjax won't return true even if HTTP Header "X-Requested-With" is set to "XMLHttpRequest"


hexa2k9:
I'm not totally sure, if this is an Issue really or if here's going something wrong, but I'm POSTing data to Route using a jQuery.ajax call. For debugging I added an Output-Field which should contain "Slim::request()->isAjax;". This unfortunately is always set to "false".

Firebug anyway tells me that the Request Header Field "X-Requested-With" is set to "XMLHttpRequest" so it should be recognized as Ajax-Call.

for the records: I'm on 5.2.6 via apache2 on that machine but happens on my usual nginx, php 5.3.3 via fastcgi as well.

jwpage:
I've just added a fix, and a pull request for it in Issue #37

codeguy:
Just pushed @jwpage's fix to the repo.

PR #37, opened 2010-10-15
Isajax

jwpage:
Fixes issue #35


codeguy:
Pulled. Thanks for catching this.

PR #39, opened 2010-10-16
Changed README to Markdown format.

jwpage:
Call me picky, but I like READMEs to be properly formatted in Markdown so they appear all pretty in Github.

If you're in the same boat, feel free to pull in these formatting changes.


codeguy:
Nice. I didn't realize GitHub would do that. Just pulled in your changes.

PR #42, opened 2010-10-25
Added MustacheView class

jwpage:
As mentioned in the roadmap, here's a MustacheView class for rendering Mustache templates in Slim.

Like my TwigView class, you'll probably want to drop the lib and test after pulling in the changes.


codeguy:
Awesome! Thanks for this. Will pull in your changes tonight.

codeguy:
Pulled in

PR #49, opened 2010-10-29
SmartyView implementation

josedasilva:
Just implemented SmartyView custom view. Feedback is welcomed.


josedasilva:
Yes, i had the same doubt myself. I believe that having an accessor method to expose the Smarty object would be the cleanest way to do this, in terms of customizing the Smarty object, because you have more needs, when performing a complex development, from  Smarty than render(), just think of fetch() , Cache or Filters usage. I don't think that making the view class more complex, meaning adding more Smarty class method wrappers would be the way, this would kill the Simple Framework purpose of Slim.

codeguy:
Sounds good, and I agree completely. Go ahead and make that one small change and I'll pull in your work. Thanks!

PR #82, opened 2011-01-15
Hooks, filters and minor tweaks

bobdia:
Combined hooks/filters in Slim.php, added priority argument, default hook invocations changed to applyHook().

Minor tweaks in Route.php and Router.php.


bobdia:
An idea for a default filter to add: urlFor()

(My bad, I closed this by mistake.)

PR #88, opened 2011-01-26
Slim doesn't handle URLs that contain language specific characters like umlauts etc

yaph:
Slim cannot handle a URL like example.com/jugador/Kaká as it contains an accented á. With this change the regular expression returned by Route::matches allows for Unicode letters, see http://www.php.net/manual/en/regexp.reference.unicode.php for more info

### DIFF
--- a/slim/Route.php
+++ b/slim/Route.php
@@ -230,13 +230,12 @@ class Route {
         //Extract URL params
         preg_match_all('@:([\w]+)@', $this->getPattern(), $paramNames, PREG_PATTERN_ORDER);
         $paramNames = $paramNames[0];
-
         //Convert URL params into regex patterns, construct a regex for this route
         $patternAsRegex = preg_replace_callback('@:[\w]+@', array($this, 'convertPatternToRegex'), $this->getPattern());
         if ( substr($this->getPattern(), -1) === '/' ) {
             $patternAsRegex = $patternAsRegex . '?';
         }
-        $patternAsRegex = '@^' . $patternAsRegex . '$@';
+        $patternAsRegex = '@^' . $patternAsRegex . '$@u';
 
         //Cache URL params' names and values if this route matches the current HTTP request
         if ( preg_match($patternAsRegex, $resourceUri, $paramValues) ) {
@@ -264,7 +263,7 @@ class Route {
         if ( array_key_exists($key, $this->conditions) ) {
             return '(?P<' . $key . '>' . $this->conditions[$key] . ')';
         } else {
-            return '(?P<' . $key . '>[a-zA-Z0-9_\-\.\!\~\*\\\'\(\)\:\@\&\=\$\+,%]+)';
+            return '(?P<' . $key . '>[\w\pL\-\.\!\~\*\\\'\(\)\:\@\&\=\$\+,%]+)';
         }
     }
 
--- a/slim/Route.php
+++ b/slim/Route.php
@@ -105,6 +105,19 @@ class Route {
         return self::$defaultConditions;
     }
 
+    /**
+     * Check whether PCRE extension was compiled with unicode support
+     * 
+     * @return bool
+     */
+    public static function checkPcreUnicodeSupport() {
+        static $pcreCheck;
+        if ( !isset($pcreCheck) ) {
+            $pcreCheck = (@preg_match('/\pL/u', 'a') == 1) ? true : false;
+        }
+        return $pcreCheck;
+    }
+
     /***** INSTANCE ACCESSORS *****/
 
     /**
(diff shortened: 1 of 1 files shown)


bobdia:
Is using only \p{L} (letters) possibly too restrictive? How about \p{N} (numbers), \p{P} (punctuation) or \p{S} (symbols)? I understand why you wouldn't want to allow all Unicode characters, but does {L} cover what's needed?

bobdia:
Also, I think \w match unicode characters when using the 'u' modifier at the end.

yaph:
Maybe the regular expression should be even less restrictive, I can well imagine numbers and symbols as parts of URLs that are not covered by the existing regular expression. For example the € sign in a URL still leads to a 404 page.

yaph:
Only adding the u modifier does not suffice, \w does not match á if the \pL sequence is missing.

bobdia:
Had to reread what the 'u' modifier does. It's used when you want the regex pattern to include Unicode chars.

So we just need the \p{class} stuff, you can remove the 'u' modifier.

Edit:

Never mind about that, just tested it. We do need the 'u'. The docs were slightly confusing for me.

yaph:
Just tried it without the u modifier and Slim fails to load the correct page. The manual page linked above also empathizes that UTF-8 has to be selected.

codeguy:
@yaph Thanks for your contribution. I want to make sure this works as expected, so give me a day or two to test this on my end. Another reference is http://www.regular-expressions.info/php.html (see paragraph 4 beneath "The preg Function Set").

Also, we need this to degrade gracefully if PCRE was not compiled with unicode support.

yaph:
Regarding your point about missing unicode support I found this forum discussion https://bbs.archlinux.org/viewtopic.php?pid=838794
What do you think of using the @ operator for a support check?

codeguy:
@yaph After a quick search, that test seems to be the consensus for detecting PCRE unicode support. Sounds good to me.

yaph:
Well, I haven't dived very deep in the Slim code yet so I am not sure which would be the best place to perform such a check. It should happen only once during a request life cycle and set as an object property that is easily accessible in the Route class.

codeguy:
@yaph You could add a new public static method to the Route class and memoize/cache the result. The method would return the already set value if available, or calculate/set/return the value if not.

yaph:
Sounds like a good approach.

codeguy:
@yaph Thanks. I'm going to wrap up a new MimeType handler first, then I'll pull this in afterward..

yaph:
@codeguy great and thanks to you for Slim!

tobsn:
URLs are written only with the graphic printable characters of the US-ASCII coded character set. The octets 80-FF hexadecimal are not used in US-ASCII, and the octets 00-1F and 7F hexadecimal represent control characters; these must be encoded.

http://www.ietf.org/rfc/rfc1738.txt

yaph:
@tobsn I don't see what's wrong with unicode letters in URLs as browsers can handle them.

tobsn:
its just not RFC. everything that is not US ASCII has to be encoded... /Kak%c3%a1
tough i don't really mind. just wanted to point out that it doesnt have to support it.

bobdia:
Tobsn is right.

Not all browsers, webservers, middleware, etc handle Unicode the same way. It doesn't seem like a bad idea to have Slim handle both regular and Unicode URLs, but we may be proven wrong in the future.

yaph, I think it's very likely you are seeing a decoded URL in your browser(s), not a true unicode one.

yaph:
@bobdia I generate links in a Slim based app that contain unicode characters without URL encoding them. Links that contain these characters lead to an error and thus break the app.
Of course, I could encode the URLs before generating the HTML links, but a user might still type unicode chars directly in the address bar, which would also break the Slim app.
Try typing special characters like á or ä  directly in the address bar to change a Google search query or go to a certain Wikipedia page. This will work.

tobsn:
yeah, not really. if you look into the source of google etc. the href links are all encoded. wikipedia of course can receive such urls directly and unencoded and maybe slim should support it but should always return them encoded.

codeguy:
I'm willing to update Slim so that it can parse unicode URLs. That may not be RFC, but I see no harm in this additional functionality assuming it does not break the framework's existing functionality. Of course, all URLs generated or returned by Slim should be URL encoded and RFC compliant.

codeguy:
After a lot of thought on this, I've decided not to accept non-ASCII characters in route URLs. I found an excellent discussion on StackOverflow about this. See:

http://stackoverflow.com/questions/2015546/allowing-non-english-ascii-characters-in-the-url-for-seo

Wikipedia is known for using URLs with non-ASCII characters. But based on that discussion thread, and after examining the source code of this Wikipedia page:

http://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller

...you can see that even though the UTF8 characters are used in anchor text, the anchor _links_ are still url encoded. So even though you may see UTF8 URLs displayed by the browser, the browser is still sending URL-encoded requests to the server.

I'm going to close this out and keep Slim in accordance with the RFC3986 spec.

MatthewCallis:
For what it's worth, the first sentence of the answer on that StackOverflow answer is completely wrong "Firstly, search engines really don't care about the URLs." and it might be worth it to make a note since the change is so simple in a "Slim Hacks" section or some other title on how to achieve this, seeing the URL in the title correctly makes all the difference in a users experience, as some browsers will actually show the URL correctly, be it `http://list4.auctions.yahoo.co.jp/jp/テレビゲーム-ゲーム-おもちゃ-ゲーム/22844-category.html` or `http://på.ws/` rather than http://xn--p-2fa.ws/

codeguy:
The first sentence is irrelevant really... but I do agree with you. SEO has nothing to do with my decision to omit this feature though. I was more looking at the fact that browsers will make a best effort to show the decoded URL to the end user while still submitting the encoded URL with the request.

And even if you paste a URL with UTF8 characters in your browser address bar, the HTTP request sent to the server is still URL encoded before it is sent. I'm certainly willing to have a "Slim Hacks" page on the Wiki. Not a problem.

codeguy:
https://github.com/codeguy/Slim/wiki/Slim-Unofficial-Hacks

yaph:
@codeguy as an application developer you can control the links generated by your app so it's easy to make sure they work, but you cannot control what users type in the address bar and in many cases it's unfeasible to control the content editors write.

PR #95, opened 2011-02-14
Haanga view

iromli:
This is the HaangaView code that we've talked about previously, please have a look


MatthewCallis:
This would be really useful for people coming from Django, good work!

codeguy:
Pulled.

PR #99, opened 2011-03-06
Parent class of HaangaView

iromli:
I noticed that View class name has been changed to Slim_View, so this is a quickfix for HaangaView class


codeguy:
Thanks. Pulled in.

PR #123, opened 2011-04-10
Slim::flash casts to string

### LINKED ISSUE #120: Do not cast Slim::flash / Slim::flashNow values to strings
By not typecasting flash values, we can pass arrays instead of strings.

jkaflik:
Removes Slim::flash casting to string. Now we can store any type of data in flash messages.
fixes #120


codeguy:
This has been pulled into 1.3.3. Not sure why it didn't auto-close. But this is done. Thanks!

PR #124, opened 2011-04-23
uninitialized app->flash makes stop() fail

fgalassi:
Basically, i tried to log to an unwritable path. Setup of logging is before setup of session/flash. The exception got caught and ended up calling stop() which failed() trying to app->flash->save(). This is a quick fix. Maybe you should strive to make stop() bulletproof (try/catch ?). Wherever an exception happens, stop() should not fail.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -881,7 +881,7 @@ class Slim {
      * @return  void
      */
     public static function stop() {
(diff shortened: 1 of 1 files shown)


codeguy:
Thanks for catching and fixing this. Willl pull in this weekend.

PR #126, opened 2011-04-28
Autoload breaks when Slim.php isn't placed in default directory.

jkaflik:
Let's try to change default "Slim" directory name to ex. "slim", then autoload fails, because "Slim_" in classes names points to exacly "Slim" directory.

I've fixed it by simple substr.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -152,7 +152,7 @@ class Slim {
         if ( strpos($class, 'Slim') !== 0 ) {
             return;
         }
-        $file = dirname(__FILE__) . '/../' . str_replace('_', DIRECTORY_SEPARATOR, $class) . '.php';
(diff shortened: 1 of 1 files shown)


codeguy:
I'm running into issues trying to pull this in, so I've just copied and pasted it instead. I also changed `substr($class,4)` to `substr($class,5)` so that it trims `Slim_` rather than `Slim` from the full class name. Thanks for your help! I will have this change pushed up shortly to both branches.

jkaflik:
No problem :)

mpjdotdev:
I am having this issue with current stable version from official site, shouldn't it be updated?

If I try with the latest development release it works like a charm ;)

codeguy:
hmm I thought I had merged the fix into both branches. I'll check again this weekend.

1994rstefan:
Still not working.

If you name the folder "Slim" the autoloader trys to load "web_files/Slim/View.php". If you name the folder "slim" the autoloader trys to load "web_files/slimSlim/View.php" and fails.

PR #127, opened 2011-05-03
Strip PHP closing tags

munro:
A few PHP files had whitespace after the closing tag, which was breaking the ability to send headers on my machine.  So let's just remove them completely :D


tobsn:
wut? seriously?

PR #136, opened 2011-06-19
wtf do this bro

haileys:


silentworks:
Can you please change the title of this as it isn't explanatory as to what your pull contains.

agush22:
Hahahaha best title ever

ffflabs:
So, how about echoing 

```
<?php 
<p><a href="https://github.com/codeguy/Slim-Extras" target="_blank">Browse 'custom Views'</a></p>
?>
```

if you close a PHP tag to output an HTML block, how would you echo '<?php' ? The parser would go back to PHP mode and throw a fatal error due to invalid syntax. Heredocs let you echo whatever you want, including php tags.

PR #147, opened 2011-07-16
Unsupported http method crashing Slim

trachalakis:
Stop Slim from crashing when the request is made useing an unsupported http method. Instead return a 404 page


codeguy:
Appreciate the help, but changes I am preparing to push to the development branch tonight make this error a moot point; routing has been re-built in a much more flexible way. Details and new code coming tonight!

PR #168, opened 2011-08-25
Fixed tests

ghost:
Fixed random failure (+1 second more) in testNewCookie and testStripSlashesIfMagicQuotes if magic_quotes_gpc is off.


codeguy:
Awesome. Will pull this in tonight.

ghost:
Sorry, I put sleep function (to test delay between create $hourFromNow and create $c2) in previous commit in wrong place and I too fast used wrong test function (I don't know PHPUnit well, yet). You can est this fix by putting sleep function between create two variables $hourFromNow and $c2, sometimes there times were different in your code.

codeguy:
Haven't forgotten about this. I'll roll this into a hotfix in the next day or so.

PR #177, opened 2011-09-20
New option 'routing.basepath'

acelot:
New option 'routing.basepath'. With this option you can add a prefix to all routes. This can be useful when you do the administrator part of the site so that every time do not write route '/admin/[more]'.


acelot:
I'm from Russia, excuse me for my English :)

PR #186, opened 2011-10-15
ClassName::methodName" Routes support for PHP 5.3

olivierphi:
Hello !

In order to be able to map Controllers instances to Slim Routes, I added "ClassName::methodName" Routes support for PHP 5.3 (an instance of $className is automatically created instead of calling the method statically).
It shouldn't change anything on PHP < 5.3, because `is_callable()` will then return `false`.

As you can see, it's a very minor addition... but it can be a very handy one :-)

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -389,7 +389,14 @@ class Slim_Route {
         }
         //Invoke callable
         if ( is_callable($this->getCallable()) ) {
-            call_user_func_array($this->callable, array_values($this->params));
+            
+            if( is_string($this->callable) && strpos($this->callable, '::' )!==false ) {
+                list($className, $methodName) = explode('::', $this->callable);
+                call_user_func_array( array(new $className(), $methodName), array_values($this->params));
+            } else {
+                call_user_func_array($this->callable, array_values($this->params));
+            }
+            
             return true;
         }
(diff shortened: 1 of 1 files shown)


maikg:
According to [PHP's documentation](http://www.php.net/manual/en/language.pseudo-types.php) this syntax is already reserved for calling static methods on a class (PHP 5.2.3 and up).

``` php
// Type 4: Static class method call (As of PHP 5.2.3)
call_user_func('MyClass::myCallbackMethod');
```

This change will therefore break existing Slim applications. Maybe use something like `ClassName#methodName`?

olivierphi:
Hmmm... PHP developers are used to the `::` syntax, it may be better to keep it ?
I can a new PR with a check of the `static` nature of the called method ; this way, users would be able to use static methods or instance methods, with the same syntax.

codeguy:
Closing this out as this change would intercept and alter the expected native behavior of a static method callable.

olivierphi:
What if we use "#" as a separator between a class name and its instance method name ? (i.e. `myClass#myMethod`)
This is not "PHP standard", but we are used to this notation (in documentations, for exemple), and it can be intuitive.

What's your opinion about it ? :-)

PR #195, opened 2011-12-21
Templates path assignement to view refactoring.

raphael-riel:
Was required for TwigView(slim-extra) to add filters on Twig.

Explanation:
I needed to add a filter to Twig before render().
But SlimView->getEnvironment() needed its templates path to initialize Twig.
Problem is that the path are given to the view on render...
Extracting some code was necessary; and now, the paths are passed to the view upon assignement.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -574,9 +574,22 @@ class Slim {
             } else {
                 $this->view = new $viewClass();
             }
+            $this->assignPathsToView();
             $this->view->appendData($existingData);
         }
         return $this->view;
+    }
+
+    /**
+     * Feed to current view with the templates path from the configuration.
+     */
+    public function assignPathsToView(){
+        $templatesPath = $this->config('templates.path');
+        //Legacy support
+        if ( is_null($templatesPath) ) {
+            $templatesPath = $this->config('templates_dir');
+        }
+        $this->view->setTemplatesDirectory($templatesPath);
     }
 
     /***** RENDERING *****/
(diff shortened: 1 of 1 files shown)


codeguy:
I previously merged a hotfix that moved template path assignment into the Slim constructor, but I prefer your approach much better as it will also apply when new Views are added later during runtime, too. I don't think I want a separate method just for template path assignment, though. I'd rather toss out the legacy code and use only `$this->view->setTemplatesDirectory($this->config('templates.path'));`.

I can make these changes on my end after I pull down what you have. Will try to merge this later today.

codeguy:
This is done in v1.5.1.4 now available in the master branch.

raphael-riel:
Looking at the end result, function view() seems to have duplicates...

```
//[...]
$this->assignPathsToView();
//[...]
$this->view->setTemplatesDirectory($this->config('templates.path'));
```

These two does the same...

Also, paths are no longer set on render(). This makes me think that if template path changes between view-assign and render, it won't be taken in account.

codeguy:
I'm looking at both branches and I don't see this. Perhaps your local branch is not up to date with Slim's official branches? I still need to adjust the development branch so that it acts the same as master.

The primary issue addressed by this change was to ensure the View object knew its template path upon app instantiation. I can assign the template path in both locations, instantiation and render, to accommodate both scenarios. Does that seem reasonable? It may be redundant in most cases, but I can see it being helpful in some situations.

EDIT: Be sure you look at the official Slim repo's branches and not the raw pull requests. Some other changes were likely made after the pull request's changes were pulled in.

raphael-riel:
Seems like I messed with my branches... 

As for point B, I prefer to configure once and only once the app; so not updating on render makes sense for me.

Thanks!

PR #221, opened 2012-01-11
Readme update

ghost:
Added quick note for the DSL abbreviation.


codeguy:
Pulled this into the develop branch (and the 1.6.0 release branch on my local machine). I tweaked it slightly so that only the abbreviation itself was linked, and I linked to the Wikipedia entry rather than Fowler's page that pushes his book. I felt only a quick definition of "DSL" was necessary rather than a big list of specific patterns.

ghost:
Sounds great., thanks.

PR #229, opened 2012-01-24
Added test mediatype no param

fgalassi:
I added a test to cover content-type extraction when no parameters are present. It's a possibility in RFC and the code wasnt clear to me (preg_split returns the whole thing as an array when pattern doesnt match)


codeguy:
Pulled into develop. Looks like GitHub's post-push hooks are failing.

PR #230, opened 2012-01-25
Better isFormData

fgalassi:
I made slim consider form data, POST requests without a content type. Same as in rack implementation. Useful for testing (and maybe broken client), otherwise you have to add content type form urlencoded in the mocked environment to make it work. It took some time to figure it out.


codeguy:
Not sure why GitHub did not auto-close this. I've pulled this into the develop branch.

PR #236, opened 2012-02-02
Limit 'global' footprint of Slim

KampfCaspar:
Hi!

Please excuse me if I'm not doing it right - but I assume this is the way of letting you know about suggestions?

As it is, Slim/develop has a rather large 'global' footprint, like insisting on its own autoloader upon require, installing a custom error handler upon construction, overwriting global configuration etc.

Please find attached my suggestions to minimize this footprint, like:
- only enabling the custom error handler during Slim runtime
- optionaly disable the custom Slim autoloader

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -132,11 +132,14 @@ class Slim {
      */
     public function __construct( $userSettings = array() ) {
         //Setup Slim application
+        $this->settings = array_merge(self::getDefaultSettings(), $userSettings);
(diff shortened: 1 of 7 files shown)


codeguy:
Thank you very much for your contributions. I'll merge these into the develop branch so they are included in the 1.6.0 release.

KampfCaspar:
As the error handler now is active only while the app is 'run' I disabled the respective unit test.

KampfCaspar:
got its own branch

PR #237, opened 2012-02-03
Encapsulate 'Stream' handling

KampfCaspar:
Hi!

I was missing the possibility to stream plain PHP streams... And as I saw, the possible streaming sources where a closed lot.

I suggest defining the stream interface and letting people add their own streaming sources.

HP

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1070,6 +1070,15 @@ class Slim {
 
     /***** STREAMING *****/
 
+    /** Stream something to the client
+     *
+     * @param Slim_StreamInterface $handler
+     */
+    public function stream( Slim_StreamInterface $handler, $options = array() ) {
+        $this->response = new Slim_Http_Stream($handler, $options);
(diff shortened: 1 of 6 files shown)


codeguy:
Agreed. What are your thoughts, though, on sending the `Slim_Http_Stream` response object back up through the onion, as opposed to a string object (as Slim currently does)? Is this avoidable? I think it introduces additional complexity for unsuspecting middleware. I'm not sure it is avoidable, though, short of exiting the middleware stack and streaming directly to the client. This has no bearing on your commits... I agree on what you've done. I'll make sure these are pulled in and introduced for v1.7.0.

KampfCaspar:
By sending a Slim_Http_Response object 'down the onion', NOTHING has to be sent back up. One can inspect the Response object which will expose the current status/headers/body.

codeguy:
Closing for now. Will re-visit in the future. See road map.

PR #239, opened 2012-02-03
Limit 'global' footprint of Slim

KampfCaspar:
'Duplicate' with its own branch...

I did not know that pull requests get appended to ex post...


codeguy:
Merged into the develop branch and will be in the 1.7.0 stable branch. I've already frozen the 1.6.0 release code and wanted to avoid a complicated merge.

PR #241, opened 2012-02-08
Added "ClassName#instanceMethodName" Routes callable support.

olivierphi:
Because it is sometimes very useful to be able to create a Class instance rather than calling a static method of this Class (it allows custom common initialization in the Class constructor, for example), this simple and intuitive notation allows this process, without breaking anything.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -392,6 +392,12 @@ class Slim_Route {
             call_user_func_array($this->callable, array_values($this->params));
             return true;
         }
+        // Do we have "className#instanceMethodName" pattern ?
+        if (is_string($this->callable) && strpos($this->callable, '#')!==false) {
+            list($className, $instanceMethodName) = explode('#', $this->callable);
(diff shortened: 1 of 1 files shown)


codeguy:
I'll look at pulling this in for the 1.7.0 release. This only appears to be a bit of sugar on top of:

```
$app->get('/foo', array(new Foo(), 'instanceMethod'));
```

I can appreciate the simpler syntax, for sure. Will get back with on this later after 1.6.0 stable is released.

olivierphi:
Great !
Yes, the syntax is a little bit simpler than `array(new Foo(), 'instanceMethod')` ; but more of all, it allows some lazy-loading (the `Foo` instance will be created only when we need it)

codeguy:
I'm closing this out in favor of this similar issue:

https://github.com/codeguy/Slim/pull/311

The alternative implementation is more DRY and better handles exceptions if the class or method does not exist.

PR #245, opened 2012-02-15
If gzip is enabled the Content-length header will be incorrect

georgyangelov:
The flush(); call will force the headers to be sent so the webserver can't change it. This causes a connection closed error in chrome and other browsers. Experienced on windows with Apache 2.2.19 and PHP 5.3.6

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -303,7 +303,7 @@ class Slim_Http_Response {
         }
 
         //Flush all output to client
-        flush();
+        //flush();
     }
 
     /**


jbrower:
I'm curious as to why the content-length is being set by PHP?  If you are using php-fpm (running php externally via cgi or fcgi) this will cause a lot of issues.  Does anyone know why content-length is being specified by php?

codeguy:
I'll take a look at this soon and have a fix added to 1.6.1 if necessary.

jbrower:
If you'd like, I can assist with testing/bugshooting.

codeguy:
@jbrower Definitely. Let me do some digging. Once I have something for you to look at, I'll push to a remote branch on GitHub for you to test.

jbrower:
I look forward to it.  Also, do you know why content-length was being set? It might be ok to just let the content-length not be included in the response.

codeguy:
So after a lot of reading, I can't find a reliable answer on whether or not Apache/nginx/etc. will automatically add the `Content-Length` header on its own (with or without gzip compression). I assume that an HTTP/1.1-compliant web server _will_ set the `Content-Length` header... but that's only an assumption.

I looked at Symfony's commits, and they had a very similar discussion... I believe they decided to remove the `Content-length` header for the time being until they could find a reliable solution later.

What are your thoughts?

codeguy:
Also, make sure you are testing this with the `develop` or `release-1.6.0` branch. Those do not `flush()` output. And because they do not `flush()` output, the web server should have the opportunity to manipulate the `Content-Length` header before the response is sent to the client, even if the header is already set in PHP.

Let me know if you still have issues with the newer branches.

jbrower:
It seems to me like setting the content-length is something that falls squarely in the realm of the webserver.  By leaving it up to the webserver, it can either send things with whatever transfer encoding it feels best works for the connecting client.  Chunked for example.  If you are forcing a content-length to be sent, a lot of that usefulness is lost.  When I commented out the lines setting the content-length, my php-fpm issues disappeared and I saw things transfering properly without prematurely killed connections.

codeguy:
@jbrower I'm coming to the same conclusion. I'll remove the `Content-Length` header from version 1.6.0 and forward.

PR #246, opened 2012-02-15
Improve Encryption

ehaas:
Previously were using md5($expiration) for the cookie encryption IV.
Downsides to this approach:
1) IV is predictable if attacker knows system time
2) Same for all systems with same system time
3) Only 128 bits of entropy (IV for default AES-256 is 256 bits)

The new approach solves these problems by combining the server secret
with the expiration to generate two HMACSHA1 messages, truncating the
result to 32 bytes, and using that for the IV.


codeguy:
Thanks! I'll have a look and get this pulled in this weekend.

PR #252, opened 2012-02-26
Added support for Backbone.js's X-HTTP-Method-Override header

wheresrhys:
Would be a great addition to the frameowrk I think as it means SLIM will work with Backbone's emulateHTTP setting. At present it only works when you turn on both emulateHTTP and emulateJSON (unless you have successfully configured a RESTful server, which as I've found out lately, is less than easy, particularly for a front-end developer like myself)

...in emulateHTTP mode)

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -393,13 +393,18 @@ class Slim_Http_Request {
      * @return  void
      */
     protected function checkForHttpMethodOverride() {
-        if ( isset($this->post[self::METHOD_OVERRIDE]) ) {
-            $this->method = $this->post[self::METHOD_OVERRIDE];
+        $changed = false;
+    	if ( isset($this->post[self::METHOD_OVERRIDE]) ) {
+			$this->method = $this->post[self::METHOD_OVERRIDE];
             unset($this->post[self::METHOD_OVERRIDE]);
-            if ( $this->isPut() ) {
-                $this->put = $this->post;
-            }
-        }
+            $changed = true;
+		} else if(isset($this->headers['x-method-override'] )) {
+			$this->method = $this->headers['x-method-override'];
+			$changed = true;
+		}
+		if ( $changed && $this->isPut() ) {
+			$this->put = $this->post;
+		}
     }
 
 }


zekenie:
I've implimented the code you provided and I'm still getting the 403 error. Any advise?

wheresrhys:
Do you have emulateJSON turned on in backbone? It should be off. Other than that I can't offer any advice really, apart from maybe to check against my htaccess and httpd.conf settings http://stackoverflow.com/questions/9422210/setting-up-a-restful-service-for-backbone-js-with-apache-and-windows

zekenie:
Yeah, emulate JSON is turned off. I think there's something not working right with apache.

jakub-id:
Guys, any move on this, would be very good to get it to the development branch since it's such a tiny thing. I think 'wheresrhys' patch has a problem though -- the header is x-http-method-override not x-http-method

wheresrhys:
@skoczko cheers for the positive response

The patch is correct though: x-http-method-override gets converted (by SLIM?) to x-method-override by the time it arrives at this function, which is why I check for x-method-override

jakub-id:
Yes -- I checked Slim's source and it seems to be both translating "-" into "_" and stripping any "HTTP_" (not even a leading one!!) from the key's value in the Request#header method. This is simply not gonna fly as the header is literally 'X-HTTP-Method-Override'. I hate when frameworks do such things behind your back. Anyhow, you can read the original header from $env and in the devel version it's possible to specify 'middlewares' a sort of filtering classes. Here's what I did for the x-http-method-override/x-method-override:

class HeaderMethodOverride implements Slim_Middleware_Interface {
  public function __construct($app, $settings = array()) {
    $this->app = $app;
    $this->settings = $settings;
  }

  public function call( &$env ) {
    if (isset($env['REQUEST_METHOD']) && $env['REQUEST_METHOD'] === 'POST' ) {
      $method = null;
      if (isset($env['X_HTTP_METHOD_OVERRIDE']))
        $method = $env['X_HTTP_METHOD_OVERRIDE'];
      else if (isset($env['X_METHOD_OVERRIDE']))
        $method = $env['X_METHOD_OVERRIDE'];
      if ($method) {
        $env['slim.method_override.original_method'] = $env['REQUEST_METHOD'];
        $env['REQUEST_METHOD'] = strtoupper($method);
      }
    }
    return $this->app->call($env);
  }
}

$app->add('HeaderMethodOverride');

codeguy:
Hi guys. I'll investigate and get a fix added to the develop branch soon.

codeguy:
Ok. Sorry for the wait on this. I've just pushed a fix to both the `develop` and `release-1.6.0` branches.

First, Slim removing `HTTP_` from header names, even if said substring was not the prefix, was a bug. Now, Slim will only strip `HTTP_` from header names if that substring is indeed the prefix.

Second, I've updated the `Slim_Environment` class to parse all `X-` prefixed custom headers. Previously, Slim would only parse a whitelist of headers.

Third, I've updated Slim's method override middleware to respect the `X-Http-Method-Override` header.

Thanks for bringing this to my attention!

wheresrhys:
Great - thanks :)

gpoole:
Working with Slim dev-master I've noticed that the MethodOverride middleware appears to expect that HTTP_ will be stripped off $_SERVER vars, but it looks like in Environment the part that strips HTTP_ off certain vars has been commented out or replaced somehow. As a result it doesn't seem to be correctly interpreting X-HTTP-Method-Override anymore; it's looking for $_SERVER['X_HTTP_METHOD_OVERRIDE'] but the variable is in $_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']. Can anyone confirm this or am I looking at the wrong thing?

PR #255, opened 2012-03-07
Fix URI port parsing

benpjohnson:
Hi,

I've been playing with the new PHP5.4 web server running on a non-standard port and noticed the getHostWithPort() method returning host:port:port so I had a dig around and found the current methods don't handle the case of Host including a port which seems to be part of the spec: 

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

I've knocked up a couple of tests to reproduce the issue along with a fix. I fixed this in Request rather than Environment as I imagined you would want Environment itself to reflect the actual server environment.

Best Regards,

Ben

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -392,6 +392,9 @@ class Slim_Http_Request {
      */
     public function getHost() {
         if ( isset($this->env['HOST']) ) {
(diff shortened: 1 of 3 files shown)


codeguy:
Thanks! I'll get this pulled in shortly after 1.6.0 is out the door later this week.

Issue #213, opened 2012-01-08
Allow wildcard "*" characters in route patterns


codeguy:
All URL segments captured by each "*" will be an array argument.

```
$app->get('/books/:author/*/category/:category', function ($author, $splat, $category) {});
```

If the route above is requested with URI **/books/Clive+Cussler/and/in/category/adventure**, the `$splat` argument will equal `array('and', 'in')`.

PR #265, opened 2012-03-16
Develop

StJudeWasHere:
Pull request for Issue #213

Allow wildcard \* characters in route patterns, it accepts route patterns like this:

```
/books/:author/*/category/:category
/hello/*/:world
/hello(/:world(/*))
/hello/*/world/*/foo/bar
```

The wildcard will be passed as an array parameter so requesting /hello/foo/bar/world with this pattern

```
$app->get('/hello/*/world', function ($wildcard) {});
```

Would result in the $wildcard variable being an array like array('foo', 'bar')

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -7,4 +7,4 @@ RewriteEngine On
 # RewriteBase /
 
 RewriteCond %{REQUEST_FILENAME} !-f
-RewriteRule ^(.*)$ index.php [QSA,L]
+RewriteRule ^ index.php [QSA,L]
--- a/README.markdown
+++ b/README.markdown
@@ -68,13 +68,13 @@ Ensure the `.htaccess` and `index.php` files are in the same public-accessible d
 
     RewriteEngine On
     RewriteCond %{REQUEST_FILENAME} !-f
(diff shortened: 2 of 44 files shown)


codeguy:
Awesome work. It looks like all of the important changes are in commit 603d7a5, correct? Want to make sure I don't miss anything. I'll likely cherry-pick that in so I don't have to worry about conflicts with all of the other commits in this pull request.

I'll get 1.6.1 out in a day or so. Wildcard support will be going into 1.7.0 which I'll start on as soon as 1.6.1 is out.

StJudeWasHere:
@codeguy, the commit 603d7a5 contains all the modifications needed to allow wildcards in route patterns.

codeguy:
I made several tweaks... I changed your `preg_match` to `strpos`... it's more naive, but I feel it does what is needed with much less overhead. I also avoided an array lookup by using an already established variable.

StJudeWasHere:
Cool, thank you.

On 24 June 2012 22:34, Josh Lockhart
reply@reply.github.com
wrote:

PR #270, opened 2012-03-27
Remove error_reporting call, should be left up to the implementers

nortron:


codeguy:
This has already been completed in the develop branch and will be introduced in version 1.7.0.

PR #277, opened 2012-04-10
replaced image with base64 ecoded image tag for demo page.

lukeholder:
This means I can more quickly use clone slim's repo and get to work without having to delete the image, but just the demo page. A cleaner starting point. Love slim by the way!


codeguy:
Makes sense to me. I'll get this pulled in for 1.6.1.

codeguy:
I just pulled this in. Not sure why GitHub did not mark it as complete.

PR #287, opened 2012-04-20
Update README: Do not use `if` in a `location` block for nginx

hexa2k9:
As `IfIsEvil` (http://wiki.nginx.org/IfIsEvil#What_to_do_instead) it should be avoided in this case. `try_files` basically does the same (tries to request the file, then a directory and falls back to a rewrite to /index.php in case nothing of the previous does match). This should be a lot smarter.


hexa2k9:
Sorry, for putting that into `master` branch by the way.

codeguy:
No worries. I'll get this pulled into the next point release.

Issue #233, opened 2012-01-31
Support for capturing template result


Tieno:
I'd be nice if the Slim app instance could provide direct support for capturing the render result too, instead of directly outputting it. (I use some twig templates for sending emails).

So far I've done this myself by adding a method to the TwigView extension class, which is accessed like so $app->view()->getRender('template.twig', $data);

pendexgabo:
This feature already exists. Check the method on https://github.com/codeguy/Slim/blob/master/Slim/View.php#L153 you just need to call (AFAIK)

``` php
$app->view()->setData('color', 'red');
$result = $app->view()->render('template.twig');
```

codeguy:
Closing. This was solved in a previous pull request. Will be available on the View object in 1.6.3.

PR #291, opened 2012-04-27
Add fetch method to return rendered template without echoing it

eocdev:
Allow a rendered template to be returned as follows:

```
$rendered_template = $app->view()->fetch('template.tpl');
```
## Reasoning

This is useful for rendering a template to a string. That string can then be passed on to a function that sends email, for example.

### DIFF
--- a/Slim/View.php
+++ b/Slim/View.php
@@ -145,6 +145,18 @@ class Slim_View {
     }
 
     /**
+     * Fetch rendered template
+     *
+     * This method returns the rendered template as a string
+     *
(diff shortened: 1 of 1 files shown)


urosgruber:
isn't this a dupe of #233 ? Render method of a View outputs rendered template. Why double return it?

codeguy:
I believe that `fetch` is a more intuitive method name compared to `render`. However, this was already addressed in an earlier pull request. I've added a deprecation warning to `Slim_View::render` (it will be made protected in a future release). Thanks!

PR #293, opened 2012-05-01
Updated composer.json with better keywords

silentworks:
I removed the Slim-Extras keywords and added in keywords that describe Slim more.


codeguy:
Thanks! Pulled into the release-1.6.3 branch. Will roll into master later this week.

PR #296, opened 2012-05-12
Fix for `applyHook()` to pass `$hookArg` to all registered callbacks

ghost:
I addressed an issue in applyHook() where only the first callback was receiving the specified $hookArg parameter. 

Also, applyHook() no longer returns anything.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1002,7 +1002,7 @@ class Slim {
      * Invoke hook
      * @param   string  $name       The hook name
      * @param   mixed   $hookArgs   (Optional) Argument for hooked functions
-     * @return  mixed
+     * @return  void
      */
     public function applyHook( $name, $hookArg = null ) {
         if ( !isset($this->hooks[$name]) ) {
@@ -1016,11 +1016,12 @@ class Slim {
             foreach( $this->hooks[$name] as $priority ) {
                 if( !empty($priority) ) {
                     foreach($priority as $callable) {
-                        $hookArg = call_user_func($callable, $hookArg);
+                        call_user_func($callable, $hookArg);
                     }
                 }
             }
-            return $hookArg; } }
+        }
+    }
 
     /**
      * Get hook listeners


Slim/Slim.php
```diff
@@ -1002,7 +1002,7 @@ public function hook( $name, $callable, $priority = 10 ) {
      * Invoke hook
      * @param   string  $name       The hook name
      * @param   mixed   $hookArgs   (Optional) Argument for hooked functions
-     * @return  mixed
+     * @return  void
```
ghost, line 1005: Removed `return $hookArg;` so now the method no longer returns anything.

Slim/Slim.php
```diff
@@ -1016,11 +1016,12 @@ public function applyHook( $name, $hookArg = null ) {
             foreach( $this->hooks[$name] as $priority ) {
                 if( !empty($priority) ) {
                     foreach($priority as $callable) {
-                        $hookArg = call_user_func($callable, $hookArg);
+                        call_user_func($callable, $hookArg);
                     }
                 }
             }
-            return $hookArg; } }
```
ghost, line 1023: Removed this line per our discussion:

`There should be no return value for`applyHook()`. If someone really wants each registered callback for a given hook to update the argument, that can still be  accomplished by passing an object reference.`

Slim/Slim.php
```diff
@@ -1016,11 +1016,12 @@ public function applyHook( $name, $hookArg = null ) {
             foreach( $this->hooks[$name] as $priority ) {
                 if( !empty($priority) ) {
                     foreach($priority as $callable) {
-                        $hookArg = call_user_func($callable, $hookArg);
+                        call_user_func($callable, $hookArg);
```
ghost, line 1019: Removed the `$hookArg` assignment and simply call the function.

codeguy:
Thanks! Cherry-picked this into the release-1.6.3 branch. Will roll into master sometime this week.

ghost:
Awesome, glad I could help!

Issue #299, opened 2012-05-15
Cookies with the same name "bug"


fgalassi:
When slim receives 2 cookies with the same name (e.g. with different paths) it silently builds an array of values. That's unexpected and can easily lead to bugs since the 2 cookies condition could happen at a later time.

Rack and PHP both keep the leftmost cookie only, that should be the more specific pathwise according to RFC2109:
https://github.com/rack/rack/blob/master/test/spec_request.rb
http://www.php.net/manual/en/function.setcookie.php#91140

I think slim should do the same and update Slim_Request::cookies() accordingly, or at least document this specific behavior in documentation. If you want me to fix this just ask.

See you
Federico

codeguy:
Interesting. If you can send a pull request, that'd be great. Otherwise I can get to this in the next point release after 1.6.3 is released.

PR #301, opened 2012-05-18
Fix same name cookies

fgalassi:
Fixed as i said in issue #299. While i was there i noticed comma is a valid cookie delimiter and fixed that as well.


codeguy:
Pulled into release-1.6.3. Thanks!

PR #302, opened 2012-05-22
Add templates system for SlimView

kmarcisz:
Allows to have templates for Slim App. If set up on Slim start, supplied template is rendered by default, and can be overwritten when necessary. Fully backwards compatible.


codeguy:
This is something that should be abstracted into its own custom View... this is really the entire purpose of custom Views (e.g. TwigView, SmartyView, and so on).

kmarcisz:
Should it? Every major framework has a support for some basic templating system. While using "require_once 'header.php';" might be fine for smaller projects, having a optional default layout speeds up web development by a lot. There's less clutter and code is way more organized.
My changes are fully backwards compatible, with tests coverage and provide per view / global layouts. I can't think of a way of setting up custom view to use layout that would not break XML / JSON compatibility. Is there perhaps something I am missing on how to achieve it with custom Views?

PR #305, opened 2012-06-01
CSRF protection

miraage:
Here is CSRF protection for Slim framework.

Sorry for spelling mistakes in docs/code/etc .. =)


miraage:
Whooooops class#method support should not be attached to this pull request :(

PR #311, opened 2012-06-15
Rails-like callables class#method

miraage:
Here is another way to process class#method callables.
I found this solution being pretty nice.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -385,16 +385,43 @@ class Slim_Route {
         $req = $this->router->getRequest();
         $res = $this->router->getResponse();
         foreach ( $this->middleware as $mw ) {
+            $this->railizeCallable($mw);
             if ( is_callable($mw) ) {
                 call_user_func_array($mw, array($req, $res, $this));
             }
         }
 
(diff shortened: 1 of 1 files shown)


Baachi:
This is a bad idea.
If your method `static`, then you can call it with this:`SomeClass::method`

codeguy:
I've been on the fence about this feature for a while. I recently closed a duplicate issue in favor of this one for its more DRY code. However, I've decided not to include this feature for these reasons:
1. It seems useful to, and has been requested by, only a small handful of developers
2. It duplicates native PHP syntax already available (e.g. "Class::method")
3. It may cause confusion or arguments among developers familiar with other language syntaxes

Closing.

olivierphi:
Hello Josh !

Still no will to create such a feature in Slim ? It may be useful only to a small handful of developers, as you said, but I still think it's a simple feature which doesn't break anything in current Slim features. Maybe with another syntax, clearly documented ? Like the Laravel's `Route::get('welcome', 'home@index');``, for instance ?
I love Slim, but having only static Controllers is a pain... :-/

PR #312, opened 2012-06-15
Fixed issue that saw any echo'd output before the $app->run() call get duplicated in the output stream.

briannesbitt:
If you had any echo'd output before the $app->run() call it would get duplicated in the page output when output_buffering is turned on.  In PHP 5.4+ (maybe before?) output_buffering is set to 4096 by default.

Ex.

``` php
<?php
$app = new Slim();

printf("%s:%d<br/>\n", __FILE__, __LINE__);

$app->get('/hello/:name', function ($name) {
    echo "Hello, $name!<br/>\n";
});

$app->run();
printf("%s:%d<br/>\n", __FILE__, __LINE__);
?>
```

Output =>

``` html
B:\dev\slimapp\index.php:5
Hello, brian!
B:\dev\slimapp\index.php:5
B:\dev\slimapp\index.php:12
```

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1128,6 +1128,11 @@ class Slim {
      * @return void
      */
     public function run() {
+        //Clear any existing output buffer
+        if (ob_get_length() > 0) {
+            $this->response->write(ob_get_clean());
+        }
+
         set_error_handler(array('Slim', 'handleErrors'));
 
         //Apply final outer middleware layers
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -106,6 +106,16 @@ class Slim {
     );
 
     /**
+     * @var int
+     */
+    protected $obLevel;
+
+    /**
+     * @var string
+     */
+    protected $preContent;
+
+    /**
      * Slim autoloader
      *
      * Lazy-loads class files when a given class is first referenced.
@@ -155,6 +165,9 @@ class Slim {
         if ( is_null(self::getInstance()) ) {
             $this->setName('default');
         }
+
+        //Capture current ob level to be used when a Stop is caught
+        $this->obLevel = ob_get_level();
 
         //Set default logger that writes to stderr (may be overridden with middleware)
         $logWriter = $this->config('log.writer');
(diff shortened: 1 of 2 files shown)


briannesbitt:
Hmm... actually it seems this only fixes some of the issues.  Right now if you do a :

``` php
$app->get('/', function () use ($app) {
   echo "Hello World!<br/>\n";
   $app->stop();
});
```

You get the "Hello World!" string repeated twice in the output when output buffering is on.

Gotta think about this some more.

briannesbitt:
I have a better fix that works for all scenarios with output_buffering off and on.  I'll commit later tonight.

codeguy:
Cool. Spending most of today on Slim. Will look for your updated code later today.

briannesbitt:
I actually want to add some ob init code at the top of the run() call.

``` php
$this->obContextLevel = ob_get_level();
```

This is necessary to conditionally close the ob in the Stop catch block later as sometimes it gets closed by the ob_get_clean() but during stop(), halt() etc that code doesn't run and the Stop catch must do it.  Unfortunately a lot of the tests don't go through the run() but rather call the call() method directly and forces the Stop catch block to check for isset($this->obContextLevel) which I don't like.

Easy thing I think is to initialize the $this->obContextLevel in the constructor and then grab it again at the start of the run() incase the user called ob_start() in between the Slim object creation and run() call.... doubt it will happen but might as well.

As I said, I'll post later tonight :-)

briannesbitt:
The reason the test "testStopWithSubsequentOutput()" passes right now is because it is calling $s->call() rather than $s->run() which is what a real application would do.  When you change the test case to execute $s->run() then the test fails with the double output I mentioned earlier:

``` html
1) SlimTest::testStopWithSubsequentOutput
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Foo'
+'FooFoo'
```

Anyway, this patch should (when I commit in a few minutes) fix both cases with output buffering on or off and fix the stop(), halt() and regular code path.

codeguy:
I'm not entirely satisfied with this implementation. Nothing wrong with the code, I just think that it's too messy (as a result of pre-existing implementation) and needs to be simplified, even if that means refactoring elsewhere. Let me think on this some.

codeguy:
Take a look at https://github.com/codeguy/Slim/commit/148cd1b52024fc0415775d38ffea59b9e434b73c and see if that does what you need. It works for me and has minimal changes to the codebase.

PR #315, opened 2012-06-17
getCurrentRoute()

pablofierro:
fix for calling getCurrentRoute() in a middleware


codeguy:
See https://github.com/codeguy/Slim/commit/1719b40112bed2f847115826ae1743cc866aaf4f

Your change caused a lot of errors, primarily `Only variables should be passed by reference`.

So I changed it slightly, but the result is still the same.

PR #321, opened 2012-06-22
Feature augment output buffer test coverage

briannesbitt:
Augment testing with output buffer on for runner, stop() and halt().  The phpunit expectOutputString() function uses output buffering so that will mimic OB being set in the INI.  The tests also always check the output buffer and not just the response body value, hence run() is necessary vs. just call().

### DIFF
--- a/tests/SlimTest.php
+++ b/tests/SlimTest.php
@@ -817,20 +817,32 @@ class SlimTest extends PHPUnit_Framework_TestCase {
     }
 
     /**
-     * Test stop with subsequent output with output buffer on and pre content
+     * Test stop with output buffer on and pre content
      */
-    public function testStopWithSubsequentOutputWithOutputBufferingOnAndPreContent() {
-        $this->expectOutputString('1.2.Foo.3');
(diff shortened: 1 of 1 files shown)


briannesbitt:
Thought about this today and I don't like my current implementation.  I don't like how they rely on the phpunit functionality.  They also need to assert that the `ob_get_level()` is 0 since that was the issue from before... stranded output buffer holding the repeated data that was flushed but the underlying php engine and hence wasn't caught by phpunit tests.

briannesbitt:
Ok so decided to leave the the current tests as is using phpunit `expectOutputString()`.  I still was missing the real issue that caused all of this.  There were output buffers left open as the stop exception code bypassed the `ob_get_clean()` in the `call()` method and only made a call to `called ob_get_contents()` rather than closing the output buffer.  In the spirit of simple unit tests that try to only test one feature I added new tests to ensure there were no open output buffers caused by `run()`, `stop()` and `halt()`.  This should provide adequate test coverage of this issue.

codeguy:
Thanks! Pulled into the release-1.6.4 branch.

PR #324, opened 2012-06-28
Make PUT and DELETE params available

ChristianVermeulen:
Hi, I've updated the put and delete methods in the request class to make sure they parse the input found in "slim.input". Otherwise the data is lost upon initialisation of the Slim class.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -244,7 +244,11 @@ class Slim_Http_Request {
      * @return  array|mixed|null
      */
     public function put( $key = null ) {
-        return $this->post($key);
+        parse_str( $this->env['slim.input'], $input );
+        if( $key )
+            return $input[ $key ];
+        else
+            return $input;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


miraage:
Aren't they look for slim.input like a post() current version?
I wrote unit tests for Slim, and my slim.input data in Slim_Environment::mock was correctly recognized on PUT request.

ChristianVermeulen:
For me they didn't I var dumped the php://input before Slim is initialised with succes.
When I var dump it on the next line after init it comes up empty.
Also when I var dump the $app->request()->parameters(); or put() or post() it remained empty.
However, when I var dumped $app->request() I do see the values in slim.input.
So I used that to recover the params that for some reason were lost...

Note: I send PUT requests to my localhost using the RESTconsole extension for Chrome (https://chrome.google.com/webstore/detail/cokgbflfommojglbmbpenpphppikmonn) Maybe the problem remains in the extension and the way it sends its request values?

miraage:
Tested now. Really strange. Maybe I have performed tests on develop branch (don't remember).
Seems to be really buggy.

codeguy:
This feature is already available in v1.6.4. If you are having issues, please open a new thread at http://help.slimframework.com/.

Issue #323, opened 2012-06-24
Move route dispatch into Slim_Router


codeguy:
Move route dispatching responsibility into `Slim_Router`. Currently, it is in `Slim_Route`. This allows us to soon provide custom router dispatchers, and may lay the groundwork for invoking named routes from another route without a separate HTTP request.

peters:
+1

codeguy:
Closing. See #332.

PR #331, opened 2012-07-09
Move route dispatch into Slim_Router

StJudeWasHere:
#323 Move route dispatch into Slim_Router

This allows to invoke named routes from another route.

```
    echo 'hello';
    $app->router()->dispatch($app->router()->getNamedRoute('world'));
});

$app->get('/world', function() use ($app) {
    echo ' world.';
})->name('world');
```


StJudeWasHere:
This is supposed to be a pull request for the release-1.6.5 branch.

PR #332, opened 2012-07-09
Move route dispatch into Slim_Router

StJudeWasHere:
#323 Move route dispatch into Slim_Router

This allows to invoke named routes from another route.

```
    echo 'hello';
    $app->router()->dispatch($app->router()->getNamedRoute('world'));
});

$app->get('/world', function() use ($app) {
    echo ' world.';
})->name('world');
```


StJudeWasHere:
This commit only moves the dispatcher method from the Slim_Route to the Slim_Router.

However, with this change I don't think the route needs to know about the router at all and it could be completely removed from Slim_Route.

I'll create a new pull request with that modifications too for evaluation purposes.

codeguy:
Closing in favor of #335.

PR #337, opened 2012-07-15
Remove template var

StJudeWasHere:
The template var in the Route class doesn't seem to be needed at all.

The first commit removes the var only, keeping the setter and getter to avoid breaking apps using this methods. The second commit removes them too, the pattern setter/getter can be used instead.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -44,11 +44,6 @@ class Slim_Route {
     protected $pattern;
 
     /**
-     * @var string The route template (e.g. "/books/:id") used with the `urlFor()` helper method
-     */
-    protected $template;
-
-    /**
      * @var mixed The route callable
      */
     protected $callable;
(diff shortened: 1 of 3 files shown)


codeguy:
The template var is actually used for wildcard support in v1.6.5. For example, suppose you have a route with pattern `/foo/*/*/bar`. This will not work with the `urlFor` helper method. So route objects will be given a new `setTemplate` method to store an override of the route pattern to be used with the `urlFor` helper (e.g. `foo/:paramName/bar`). This was the simplest way I could think of. If you have an alternative suggestion, let me know.

StJudeWasHere:
Righ, I didn't see that one.

codeguy:
It's entirely flexible before it's merged. I like your suggestion, but not a fan of the long names (e.g. `slim_route_wildcardN`). What if we went with "splat" instead (like Sinatra). That'd be easier on the fingers. But I do like your solution better than what's in place.

codeguy:
You want to tackle this? Else I can get to it later this evening or mid-week. Moving into a new home tomorrow and Tuesday, so not sure if I'll be able to code much until Wed.

StJudeWasHere:
I'll post a pull request.

PR #338, opened 2012-07-15
urlFor update

StJudeWasHere:
This branch uses hardcoded values for the wildcard params in the urlFor method instead of the template var, it also removes the template var and its setter/getter methods from Route.

Please review the code and see if this is more convenient than what's in place.


codeguy:
Ah, must have been tired last night ;) Thanks. Will get this pulled in today.

PR #346, opened 2012-07-18
Revision to suggested .htaccess content

collegeman:
On my local test system (MacBook Pro, OS 10.7.4, running built-in Apache 2.2.21 and built-in PHP 5.3.10), I use symlinks from a ~/www folder to my repository folders, and something about the symlink part of this (I think) breaks with the .htaccess as specified by your docs, but doesn't break using the .htaccess content above (which is from WordPress). This .htaccess snip also works in my production environment, which is (for now) a free slice on Pagoda.

### DIFF
--- a/docs/routing-indepth-with-rewrite.markdown
+++ b/docs/routing-indepth-with-rewrite.markdown
@@ -16,8 +16,12 @@ Here is an example directory structure:
 The **.htaccess** file in the directory structure above contains:
 
     RewriteEngine On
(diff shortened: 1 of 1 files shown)


codeguy:
I'm not an expert with Apache rewrite rules. Can you explain what your changes do line by line? I think what I have satisfies most cases though.

codeguy:
Closing. Send new PR to the latest `release-*` branch. Include explanation of rewrite rules there.

PR #347, opened 2012-07-18
Allow flash object to be iterated over

DexterTheDragon:
Typically when I display flash messages I like to do something like:

``` php
<?php
foreach ( $flash as $key => $message) {
    echo "<div class='{$key}'>{$message}</div>";
}
```

This patch adds IteratorAggregate to Slim_Middleware_Flash to allow iterating on the messages. I also added a test for it.

### DIFF
--- a/Slim/Middleware/Flash.php
+++ b/Slim/Middleware/Flash.php
@@ -44,7 +44,7 @@
   * @author     Josh Lockhart
   * @since      1.6.0
   */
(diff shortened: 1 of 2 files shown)


adamnbowen:
:+1: Makes templates a lot less stupid to write.

codeguy:
No idea why this wasn't in there already :) Thanks. Will definitely get this pulled into 1.6.5.

codeguy:
Pulled into release-1.6.5 branch.

PR #352, opened 2012-07-28
Prevent E_NOTICE when session already started

nortron:
Using strict equality


nortron:
Recreated pulling into release branch instead of master

PR #360, opened 2012-08-01
Fix #359 embedded PHP 5.4 webserver support

### LINKED ISSUE #359: routes with dot don't work when using PHP 5.4 embedded webserver
Whenever the URL includes a dot the PATH_INFO is always "/" and the correct value is located in SCRIPT_NAME... This makes Slim's routing fail.

```
<?php
require 'Slim/Slim.php';
$app = new Slim();
$app->get('/', function( ){
    echo "FRONT PAGE";
});
$app->get('/hello/:name', function ($name) {
    echo "Hello, $name!";
});
$app->get('/he.llo/:name', function ($name) {
    echo "HE.LLO $name!";
});
$app->run();
?>
```

Launch:
`php -S 0.0.0.0:3000 index.php`

`http://localhost:3000/hello/Joe` will display `Hello, Joe!`

`http://localhost:3000/he.llo/Joe` will display `FRONT PAGE`

`http://localhost:3000/hello/Joe.Smith` will display `FRONT PAGE`

mateusza:
Fix #359 embedded PHP 5.4 webserver support as discussed in email.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -467,7 +467,13 @@ class Slim_Http_Request {
      * @return string
      */
     public function getPathInfo() {
-        return $this->env['PATH_INFO'];
+	if ( substr( @$_SERVER['SERVER_SOFTWARE'], 0, 3 ) === "PHP" ){
+            return ( $this->env['SCRIPT_NAME'] === "" )
+                ? $this->env['PATH_INFO']
+                : $this->env['SCRIPT_NAME'];
+        } else {
+            return $this->env['PATH_INFO'];
+        }
     }
 
     /**


briannesbitt:
As I mentioned in the other....

I don't think this is a Slim issue but rather a PHP dev webserver issue.

If you access `http://127.0.0.1/he.llo/df`
$_SERVER['REQUEST_URI'] = "/he.llo/df"
$_SERVER['SCRIPT_NAME'] = "/he.llo/df"

Where as if you access `http://127.0.0.1/hello/df`
$_SERVER['REQUEST_URI'] = "/hello/df"
$_SERVER['SCRIPT_NAME'] = "/index.php"

That smells of server issue not Slim.  Slim will then see this as a url rewrite happening and `$env['PATH_INFO'] = substr_replace($_SERVER['REQUEST_URI'], '', 0, strlen($env['SCRIPT_NAME']));` will proceed to set PATH_INFO to `\`

Not to mention if you have a route of `$app->get('/he.llo/:name', function ($name) {}` that will also match `/heAllo/brian` as the period is a wilcard in a regular expression so I'd say its generally not a great idea anyway.

codeguy:
Closing this PR. Please re-submit with fixes applied to Slim_Environment. I'll research this more here, too, running the built-in PHP server. Also, I recommend you URL-encode periods as they may be construed with a different meaning if left un-encoded. Try that first.

boypt:
I encounter this problem, too. but I add a hook before route to do the hack, instead of changing Slim core code.

``` php
//in the 'development' mode
    $app->hook('slim.before.router', function () use ($app) {
        if ( substr( @$_SERVER['SERVER_SOFTWARE'], 0, 3 ) === "PHP" ){
            $env = $app->environment();
            if($env['PATH_INFO'] === '/' && $env['SCRIPT_NAME'] !== basename(__FILE__)) {
                $env['PATH_INFO'] = $env['SCRIPT_NAME'];
            }
        }
    });
```

The dev-server has to be run this way ( not perfect, your static files won't be served):

```
php -S 0.0.0.0:8081 index.php
```

You need another route file to do this, according to the PHP Wiki: https://wiki.php.net/rfc/builtinwebserver

I think @codeguy should add a caveat on the document about the strange about php-dev-server, or add a midware or a switch to support it.

PR #363, opened 2012-08-10
Changes in Slim::getInstance and (optional) htaccess

ffflabs:
.htaccess tested in Apache with VirtualDocumentRoot and Mod Alias
Slim::getInstance offers the option to create the instance if there isn't one

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -6,5 +6,13 @@ RewriteEngine On
 #
 # RewriteBase /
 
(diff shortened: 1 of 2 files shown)


sirlancelot:
looks like the line-endings have been changed.

codeguy:
Yep. Make sure you use UNIX-style LF (line feed) as EOL character.

ffflabs:
You're right. My mistake. I'll close this request and pull another one with the right formatting,

PR #365, opened 2012-08-11
VirtualDocumentRoot support and optional create instance when there isn't (fixed LF)

ffflabs:
**.htaccess** for [VirtualDocumentRoot](http://httpd.apache.org/docs/2.2/mod/mod_vhost_alias.html#virtualdocumentroot) and or [mod_alias](http://httpd.apache.org/docs/2.2/mod/mod_alias.html)  (in case you don't know which path you'll run on)

**Slim.php** adds optional _$create_ parameter, default false. If true, it will create a new Slim instance if there isn't one.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -5,6 +5,10 @@ RewriteEngine On
 # absolute physical path to the directory that contains this htaccess file.
 #
 # RewriteBase /
-
+# The first rewritecond will catch the relative path while the %1 added to the rewrite rule will apply it.
+# It assumes you're using only letters and numbers as your route
+#RewriteCond %{REQUEST_URI} (.+)/(\w+)$
+#RewriteCond %{REQUEST_FILENAME} !-f
+#RewriteRule ^   %1/index.php [QSA,L]
 RewriteCond %{REQUEST_FILENAME} !-f
 RewriteRule ^ index.php [QSA,L]
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -189,10 +189,11 @@ class Slim {
     /**
      * Get Slim application instance by name
      * @param   string      $name The name of the Slim application to fetch
+     * @param   boolean $create determines wether it should create a new slim instance when there is no current one
      * @return  Slim|null
      */
-    public static function getInstance( $name = 'default' ) {
-        return isset(self::$apps[$name]) ? self::$apps[$name] : null;
+    public static function getInstance( $name = 'default', $create=false ) {
+                return isset(self::$apps[$name]) ? self::$apps[$name] : ($create ? new Slim($name) :null);
     }
 
     /**


codeguy:
I'm not a rewrite wizard. Can you explain your .htaccess changes a bit more?

The `Slim::getInstance` method was never intended to construct. Instead, it was for injecting a Slim instance within a route callback for those using PHP < 5.3. For PHP >= 5.3, I recommend currying the `$app` instance variable into route callbacks with `use`. I really want to de-emphasize the `getInstance` method anyway at this point.

Closing. Please re-submit separate PRs for each change.

ffflabs:
Im using a lot of custom hooks as mini plugins, Some of them are not called/invoked on the regular workflow, therefore all they get is an empty slim instance. My first workaround was to always create a new Slim instance before invoking a plugin hook, but it seemed like a waste of footprint in those cases when a Slim instance was already in memory, My solution covers both situations.

 Regarding changes on htaccess, let's say you are on a shared host and each customer has a vhost in the form of:

```
<Virtualhost>
    DocumentRoot /home/user1/htdocs
    ServerName user1.server.com
</Virtualhost>
```

Calling a Slim route from that vhost is completely straightforward because htaccess knows where is beign called from.

But, for shared hosts, your site might be running fon a _VirtualDocumentRoot_ environment.  This allows for Apache to serve an unlimited number of virtual hosts and subdomains, without any specific vhost entry and without restarting the webserver to add more vhosts. Any user will have its own subdomain as they have an _htdocs_ in their home folder.

```
<Virtualhost>
     VirtualDocumentRoot /home/%0/htdocs
      ServerName thetimebilling.com
      ServerAlias *.server.com
<Virtualhost>
```

If, let's say, user2 (http://user2.server.com/) has his Slim RESTful API on a subfolder called api (http://user2.server.com/api/) and he makes the example route books

```
$app->get('/books/:id', function ($id) {
//Show book identified by $id
});
```

the regular .htaccess doesn't really know where are you calling it from, so it will make the webserver try and get a file from the global document root:

```
  /var/www/html/books
```

Which will throw a 404. They usually say you can get around this by adding

   RewriteBase /

To .htaccess, but in that case, you'll be pointing to

   /home/user2/htdocs/books

Which also is a 404. The manual solution is to add

   RewriteBase /api/

But this will only work if you are on a folder whose name is _api_.  If you want slim to run in whatever folder avoiding manually changing the RewriteBase parameter, this mod does it for you, as in htaccess you are able to know where you are regarding your document root (even if it's virtual or just a mod_alias) so it can dinamically prepend it  to slim's index   thus getting the relative index.php you see from tour document root.

```
#sets the rewrite relative to your document root
RewriteBase /
# strips the last part of the URI (which we assume is the route name)
RewriteCond %{REQUEST_URI} (.+)/(\w+)$
# prepends the first part of the URI (the subfolders) to the Slim script you're running
RewriteRule ^    %1/index.php [QSA,L]
```

If you are to deploy 10 instances then is small potatoes, but I have 2000 app custombers and 12 Slim routes in each wich broke when I used  VirtualDocumentRoot.

PR #371, opened 2012-08-16
Simplify router and delay uri pass

briannesbitt:
- Simplified the router and its interface.
- Delayed the passing of the URI until matching is required which allows hooks to modify the requested URI.
- Added route as parameter for route middleware.  Added test.  I think not sending request and response is ok as well since nothing else (routes, hooks etc) do that.  The user can always use ($app) and get it themselves.

Note the request URI is passed into the `router::dispatch()` method.  I am not a fan of this but I left it as is and will chat with you tomorrow with 2 changes that will easily remedy this.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -155,7 +155,7 @@ class Slim_Router {
         //Invoke middleware
         foreach ( $route->getMiddleware() as $mw ) {
             if ( is_callable($mw) ) {
-                call_user_func($mw);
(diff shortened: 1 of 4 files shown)


codeguy:
Available this afternoon to chat about this. Want to make sure we find the best solution before merging. Hop online whenever.

briannesbitt:
Going to close this one as https://github.com/codeguy/Slim/pull/418 is the newer.

PR #376, opened 2012-08-20
Content Negotiation support for Slim

njh:
Hello,

I have been waiting for Content Negotiation support in Slim for a while and then needed it again for a new project, so decided to go ahead with implementing it. I know there is a good chance that you won't accept this pull request, but I have already made good use of it and hopefully you can at least make use of some of the code.

Usage example:
http://gist.github.com/3398258

Implementation notes:
- I decided to make a new class 'Slim_Negotiator' to keep the functionality contained, testable and avoid bloating other classes
- I decided to create a respondTo() method on the Slim app object, rather than adding properties to the route, because the code was more readable and also avoids parameter confusion.

There is a test suite with 100% coverage and testing a good range of scenarios.

If you decide to accept the pull then there is still some work required to write some documentation.

nick.


codeguy:
Thanks for your work on this. I'm not going to pull it in as there are a few things I want to do differently, but I do like some of the ideas you're using. I'll likely cherry pick some things from this soon and provide credit where/if necessary.

willdurand:
@codeguy any news on the content negociation topic? I'm not sure to use Slim because of that missing feature, and it seems this topic is on the roadmap for more than a year..

This PR was nice, better than nothing IMO

gcatlin:
I'm also wondering about the status of content negotiation in Slim.

codeguy:
No news to report yet. I've been swamped at the day job so haven't had much time to do much of anything this year so far.

gcatlin:
Have you had any further thoughts on the design? Perhaps I and others could help out.

codeguy:
No, I haven't had any time to sit down and play with the design of it. I'm perfectly happy to delegate that out to others willing and able to implement it well.

If you'd like to spearhead the effort, maybe in conjunction with @njh above, that'd be great. Please discuss and solidify your approach here in writing before diving into code though. But that'd be great if you're interested.

gcatlin:
Sounds good. Could you identify the things you like about @njh's design and which things you'd change, so we can have some direction?

willdurand:
By the way, I wrote a tiny library to solve this issue:
https://github.com/willdurand/Negotiation.

## 

William Durand | http://www.williamdurand.fr

codeguy:
Nice! I don't see why we should reinvent the wheel to parse the raw headers. The only new thing this introduces are dependencies. But I'm fine with that.

faabiosr:
Anything new?

mewejo:
Any update on this?

codeguy:
This is available in the Slim-Middleware repository here:

https://github.com/codeguy/Slim-Middleware

mewejo:
Thanks Josh

mewejo:
I'm not sure if you provide support here, but a push in the right direction is all I'm after..

composer.json:

{
    "require": {
        "slim/slim": "2._",
        "slim/middleware": "_"
    }
}

Installing with that, doesn't actually seem to install the ContentNegotiation class, at all? This is after clearing cache + deleting the vendor dir. Stuck in a rut, can't seem to make any progress with this. Not used Composer before, so hopefully I'm missing something simple.

lalop:
Did you tried "slim/middleware": "dev-master" ?

marleyiam:
Hi! Does anybody could show me an example about how to set and use Content Negotiation using SlimExtras+ ? Cheers!

PR #387, opened 2012-08-26
Handle object returned by route handlers as string and impove exceptions handling

jcubic:
Data returned by handlers added by Slim::error Slim::notFound and routes are processed as strings and append at the end for the body (because return is always after someone put echo in route handler).

Slim::error is now exception handler and accept Exception as argument (Forget to put that information into comements)

I removed Slim_Middleware_PrettyExceptions middleware put that code into a function (in exception.php maybe it need to be put elsewhere or change to a method in Slim class). 

Exceptions are handled in this way
- if user don't define error handler via Slim::exception it will exceute defaultError or prettyException depending on config('debug')
- if user define handler it will overwrite this check.
  Also exceptions from user exception handler are cached and the default handler is exectued (defaulError or prettyException).

Also in order to intercept data returned from handlers I needed to put the code that was in Router::dispatch (it's now dead code, I din't remove it) into Slim::call where it was executed, this probably fix also the issue with `slim.before.dispatch` and  `slim.after.dispatch` which was executed even if handler was not exected (dispatched).


jcubic:
Hey I'm new to this, why there are 74 commits here?

jcubic:
I mark it as PR for master instead of develop

PR #389, opened 2012-08-26
handle return values from handlers and improve exceptions handling

jcubic:
Data returned by handlers added by Slim::error Slim::notFound and routes are processed as strings and append at the end for the body (because return is always after someone put echo in route handler).

Slim::error is now exception handler and accept Exception as argument (Forget to put that information into comements)

I removed Slim_Middleware_PrettyExceptions middleware put that code into a function (in exception.php maybe it need to be put elsewhere or change to a method in Slim class). 

Exceptions are handled in this way
- if user don't define error handler via Slim::exception it will exceute defaultError or prettyException depending on config('debug')
- if user define handler it will overwrite this check.
  Also exceptions from user exception handler are cached and the default handler is exectued (defaulError or prettyException).

Also in order to intercept data returned from handlers I needed to put the code that was in Router::dispatch (it's now dead code, I didn't remove it) into Slim::call where it was executed, this probably fix also the issue with `slim.before.dispatch` and  `slim.after.dispatch` which was executed even if handler was not exected (dispatched).

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1174,16 +1174,32 @@ class Slim {
             foreach ( $this->router as $route ) {
                 if ( $route->supportsHttpMethod($this->environment['REQUEST_METHOD']) ) {
                     try {
-                        $this->applyHook('slim.before.dispatch');
-                        $dispatched = $this->router->dispatch($route);
-                        $this->applyHook('slim.after.dispatch');
-                        if ( $dispatched ) {
+                        if ( substr($route->getPattern(), -1) === '/' &&
+                             substr($this->resourceUri, -1) !== '/' ) {
+                            throw new Slim_Exception_RequestSlash();
+                        }
+                        //Invoke middleware
+                        foreach ( $route->getMiddleware() as $mw ) {
+                            if ( is_callable($mw) ) {
+                                call_user_func($mw);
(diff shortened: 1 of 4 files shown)


Slim/Slim.php
```diff
@@ -31,6 +31,15 @@
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+if ( !defined('MCRYPT_RIJNDAEL_256') ) {
+    define('MCRYPT_RIJNDAEL_256', 0);
+}
+if ( !defined('MCRYPT_MODE_CBC') ) {
+    define('MCRYPT_MODE_CBC', 0);
+}
```
jcubic, line 39: The file don't look as good as before, but those constans where needed, and giving Notices.

codeguy:
Closing this PR as it's unusable. Consider it a learning experience :) I went through the same thing. As a project maintainer, it's easiest for me to receive a separate PR for each unique change you want to make. This means I can pull in a PR for one thing, and potentially reject or delay a PR for another thing.

For each change you want to contribute, please create a separate branch just for that one change. Then send a PR based on that separate branch.

Does this make sense? If not let me know and I'll try to clarify further.

PR #392, opened 2012-08-26
Treat objects returned from handlers as strings

jcubic:
Separed PR for using values returned by routes, notFound and Error as String like object for body of the response. the result is appended after the php ob buffer

``` php
$app->get('/', function() {
     echo '{{ debug }}';
     return new Template('page.template', array('title' => 'hello'));
});
```

will add to body

```
{{ debug }}
<html>
  <title>hello</title>
...
```

Bigger change is moving code from Router::dispatch to Slim::call where it was called. So now Router::dispatch is dead code. This change was nessesary to get value from invoation of route callback. IMHO it was weird that router return routes and then those routes was passed back to router, if router know which routes to dispach why he don't just dispach them without returing them, or if routes are returned from router then those routes should be handled in that place.

PS: 1 commit for this change and 2 commits because I put dispach hooks in wrong place.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -200,7 +200,7 @@ class Slim_Http_Response implements ArrayAccess, Countable, IteratorAggregate {
      */
     public function write( $body, $replace = false ) {
         if ( $replace ) {
-            $this->body = $body;
+            $this->body = (string)$body;
         } else {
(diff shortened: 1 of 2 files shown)


jcubic:
I try to use debug false and my error handler and it turns out that I forget to add  string handling in error handler,

PR #399, opened 2012-08-28
Remove error handlers from Router

StJudeWasHere:
Remove error handlers form the Router class and let the Slim class to handle it.

Error callbacks were stored in the Router class but it seems more appropriate to have them in the Slim class instead. This simplifies the Router code.


codeguy:
Sorry. Getting merge conflicts. Can you rebase against the latest code in the develop branch?

StJudeWasHere:
Ready, it should merge without conflicts now.

PR #400, opened 2012-08-28
request: Send back exceptions in JSON if negotiated

chelmertz:
This patch exposes the accept-header, of the environment, to the
request. If I ask for json (like popular js-frameworks do), let them
not have to get a styled and very hard to parse HTML exception in the
response, but rather a well formed json-string.

With sensible messages in uncaught exceptions (if that ever can be
considered sensible..), this method let's all calling javascript reach
the error message in response.message on their end.

Signed-off-by: Carl Helmertz helmertz@gmail.com

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -322,6 +322,18 @@ class Slim_Http_Request {
     }
 
     /**
+     * Get Accept header
+     * @return string
+     */
+    public function getAccept() {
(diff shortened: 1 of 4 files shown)


codeguy:
I think this concept can be abstracted further for multiple content types. Content negotiation and accept-header parsing is on the todo list. Will revisit this suggestion in the near future.

Issue #398, opened 2012-08-28
Make Slim a DI container with Pimple


ziadoz:
It would really cool as a future enhancement to make the main Slim class a dependency injection container using Pimple (https://github.com/fabpot/Pimple). The Silex micro framework does this and it makes storing and using your application dependencies really simple. It could also be useful for accessing the request, response, environment and log objects too:

```
$app = new Slim;

$app['config'] = Yaml::parse(__DIR__ . '/config.yml');

$app['pdo'] = $app->share(function($app) {
    $config = $app['config'];
    $pdo = new PDO($config['dsn'], $config['user'], $config['pass']);
});

$app->get('/', function() use ($app) {
    $app['response']['X-Powered-By'] = 'Awesome Sauce';

    $post = $app['request']->post();
    $app['pdo']->query('...');

    $app['log']->debug('foobar');
});
```

codeguy:
This is definitely something I will consider. Up to this point, Slim has supported PHP 5.2 and newer so Pimple was not an option. Slim 2.0 will move to 5.3 and newer which opens up all sorts of new possibilities. Leaving open and marking for 2.0 milestone.

ziadoz:
Thanks for considering the feature, I think it would be really nice for Slim to have.

Another thing that might be useful is binding any route closure objects to the Slim application object (PHP 5.4 only) using the `Closure::bindTo()` method:

```
$app = new Slim;

$app->get('/', function() {
    // No need to write: use ($app)
    print_r($this->request());
});
```

No idea if you'd want to include this in Slim now (using a PHP version check) or later in the 2.0 milestone, but I thought I'd just throw it out there anyway.

chelmertz:
I use Pimple and Slim now, it's not that many extra letters to write + the dependency is optional.

```
$app = new Slim;
$di = new Pimple;

$app->get('/', function() use ($app, $di) { .. });
```

I really like use of `Closure::bindTo()` but I think DI can be mimicked in 20 lines or less and included in Slim without that extra dependency of Pimple.

ziadoz:
Yes, mimicking Pimple probably makes sense since there isn't much to it really.

StJudeWasHere:
I just created a pull requested to experiment with this idea.

pyrsmk:
Really interesting, hope it will come soon ;)

[...]

PR #403, opened 2012-08-30
Array access

StJudeWasHere:
This is an initial implementation for #398.

So far what it does is implement ArrayAccess in the Slim class and make the Request, Response, Environment and Router array accessible.

```
$app = new \Slim\Slim();
$app['response']['status'] = '500';
```

Maybe adding the Slim settings and the application middleware to the array access would be cool too, something like the Pimple's share method might be also useful.

I kind of like this DI container idea and I think it would allow Slim to evolve in a much more modular framework, what do you guys think?

### DIFF
--- a/Slim/MiddlewareContainer.php
+++ b/Slim/MiddlewareContainer.php
@@ -1,0 +1,139 @@
+<?php
+/**
+ * Slim - a micro PHP 5 framework
+ *
+ * @author      Josh Lockhart <info@slimframework.com>
+ * @copyright   2011 Josh Lockhart
+ * @link        http://www.slimframework.com
+ * @license     http://www.slimframework.com/license
+ * @version     1.6.7
+ * @package     Slim
+ *
+ * MIT LICENSE
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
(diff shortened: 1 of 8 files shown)


ziadoz:
Looks great.

Adding the Slim settings to the DI container would be really nice. Maybe then the `Slim::config()` method could be removed altogether?

It might also be worth implementing the `IteratorAggregate` interface as well, this way you could iteratate over the Slim object itself and access all the values in `$this->container`:

```
class Slim extends \ArrayAccess, \IteratorAggregate
{
    // ArrayAccess Methods...

    public function getIterator()
    {
        return new \ArrayIterator($this->container);
    }
}

$app = new Slim\Slim;
$app['foo'] = 'bar';

foreach ($di as $key => $value) {
    echo $key . ': ' . $value . "\n";
}

// foo: bar
```

I think the `share` and `protect` methods from Pimple would be a good addition too. The share method in particular would be really handy for storing something like a shared `PDO` object for an application.

StJudeWasHere:
I added the settings and middleware to the ArrayAccess too. Some of the code in the Slim class could certainly be removed now.

codeguy:
Thanks for your work on this. Let me play with it some this weekend and see what I think.

ziadoz:
I think only the Slim object needs to be an array accessible DI container (like Pimple). Not so sure about settings or middleware.

codeguy:
I agree only the Slim class should implement ArrayAccess and Pimple-like features. Applying to settings and middleware is just too much over-engineering. I don't think we need everything that Pimple provides, just basic parameter and object injection, and sharing/singletons.

StJudeWasHere:
That was my feeling too, the last two commits would over-complicate the code. How would you go to make the settings and middleware accessible via ArrayAccess?

StJudeWasHere:
I'm closing this one in favor of #433. It takes it from commit  7e5ed46 so ArrayAccess is only implemented in the Slim class.

PR #413, opened 2012-09-08
Defines the application name through the configuration array

wiltave:
Defines the application name through the configuration array (just in __construct). For example:

``` php
$app = new Slim(array('application-name' => 'Foo'));
$app->getName(); // outputs "Foo"
```

This would avoid a setName call after instantiation. Maybe a __construct parameter would suit better, I'm not sure.

Thank you.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -165,8 +165,13 @@ class Slim {
         //Setup view
         $this->view($this->config('view'));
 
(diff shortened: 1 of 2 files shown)


codeguy:
Closing. Please resubmit against the develop branch. I also prefer to use "name" instead of "application-name".

PR #416, opened 2012-09-09
Autoload classes in include path

vanting:
Replace the file_exists() with stream_resolve_include_path() such that classes in include path can be loaded. 

Project Setting:
+-- lib/Slim/Slim.php
 |-- index.php

In the top of index.php, add this line:
set_include_path('lib');

The original file_exists() doesn't handle this situation properly.


codeguy:
Closing. Please resubmit against the develop branch if this is still valid. Be sure you rebase against the latest code in the develop branch. The built-in Slim autoloader uses the suggested PSR-0 autoloader so I'm hesitant to make any changes.

PR #418, opened 2012-09-10
Simplify slim call and router v2

briannesbitt:
As per discussion!

The only "backtrack" I did was update the phpdocs in a later commit for `Router::getCurrentRoute()` and `Router::getMatchedRoutes()`... sorry!


StJudeWasHere:
If you consider removing the Slim::pass method you can simplify it even more. See #432.

silentworks:
@briannesbitt will this return a string or a object? I so want this to return a string also.

briannesbitt:
Will what return a string or object?

PR #428, opened 2012-09-21
Router constructor doesn't take arguments

StJudeWasHere:
Router constructor doesn't take arguments, the resource uri is set with the Router's setResourceUri() method instead.


StJudeWasHere:
You'r right, I'll close this one. Thanks.

PR #432, opened 2012-09-22
Simplify and remove pass

StJudeWasHere:
- Remove the Slim::pass method and dispatch the first matched route. 

Starting with #418  briannesbitt/simplifySlimCallAndRouterV2 branch remove the Slim::pass method and dispatch the first matched route. This simplifies the Slim::call method.
- Moves dispatch method to the Route.

Move the dispatch method back to the Route class, it seems more appropriate to have the route handle its own middleware instead of relying on the Router to do so.
- Create the routes in Slim instead of Router.

Create the routes in Slim::mapRoute and inject them into the Router instead of creating them in the Router itself.


briannesbitt:
I am actually a fan of pass() and would like to augment it (at some point) to pass to a specific named route.  I use it for my personal blog when the post is not found... https://github.com/briannesbitt/nesbot.com/blob/master/routes.php

briannesbitt:
However I have as a to do!!  "Improve router to return early with first matched route ... get another matched route when/if necessary. (ex. on pass())".  I was leaving it until this initial PR got pulled in... too many changes at once and too hard to make it a separate PR since it would just conflict so much.

StJudeWasHere:
I'm not that fan of the pass method and I wouldn't miss it a bit. In fact, most of the time it can be avoided by just declaring the routes in the appropriate order and making use of the route conditions.

I haven't checked all your blog code and I'm probably missing something, but it seems like you could get rid of it and just throw a notFound instead. The '/:year/:month/:day/:slug' route is the only one you declared with four parameters so $app->pass() won't return any new route at all.

In case you had more routes with four parameters you could still avoid the $app->pass() method doing something like this:

```
// urls with same structure declared before the generic one
$app->get('/2012/7/7/foo-bar', function() {
    echo 'foo bar';
});

// generic url
$app->get('/:year/:month/:day/:slug', function ($year, $month, $day, $slug) use ($app, $posts) {
   $post = $posts->findBySlug($slug);

   if ($post == null) {
      //$app->pass();
      $app->notFound();
   }

   $app->render('post.php', array('post' => $post, 'next' => $posts->next($post), 'prev' => $posts->prev($post)));
})
->name('post')
->conditions(array('year' => '(19|20)\d\d'))
->conditions(array('month' => '(\d|\d\d)'))
->conditions(array('day' => '(\d|\d\d)'));

// this could also be declared before the generic structure, but if you want to declare it here 
// you can use the route conditions in the generic one.
$app->get('/a/b/c/d', function() {
    echo 'abcd';
});
```

Also, instead of augmenting the pass() method to pass to a specific route you can just dispatch it, which is already supported.

```
$app->router()->getNamedRoute('world')->dispatch()
```

briannesbitt:
While I agree that pass() is not required and, similar to a lot of functionality in Slim, there are many ways to tackle the problem.  I prefer the use of pass() here rather than notFound() as it describes what the code is doing which makes it easier to read.  I actually don't match since I couldn't find the resource required, so I am passing up this route for someone else to handle rather than just assuming its not found.  No condition is able to make that decision and I'd rather not rely on the order of the routes being defined.

Also I don't think its as simple as just running dispatch() with the new route as it was not necessarily matched (particularly with your proposed changes of only matching one route) and it would not have its route parameters setup.

StJudeWasHere:
I see your point and I understand you have different coding preferences but, in my opinion, if the resource is not found using the notFound() method is more descriptive than using pass(). Plus, relying on the order of the routes and route conditions is quite powerful too and can do the job in most cases if not all.

The route dispatching thing is already implemented in the current Slim version and you can use it, though you have to dispatch it through the Router instead, but of course if the route needs parameters you will have to set them up first.

```
$route = $app->router()->getNamedRoute('coolroute');
$route->setParams(array('foo'=>'bar'));
$route->dispatch();  
```

I guess you would need to do something like that even with an augmented pass() method as not all matched routes may take the same number of parameters and someone has to decide which params get in an which doesn't.

briannesbitt:
I was waiting for the "but" !  haha  If pass() wasn't there I would probably just throw a notFound.  Really I could go either way since its all doable one way or another.  I guess Josh will be the tie breaker!

I made a small additional commit to my original branch, but actually I don't think it affects this since you only match one route and set currentRoute accordingly.  This change maintains the reason for my update, which was to make the currentRoute available in the slim.before.dispatch hook.

Also I think you can remove the "protected $matchedRoutes;" from Router since its not being used.

StJudeWasHere:
I'm closing this in favour of #436.

PR #433, opened 2012-09-23
DI container

StJudeWasHere:
This comes from #403.

It makes the Slim settings array accessible by adding them to the DI. It also moves the Log to the DI container to make it accessible via $app['log'].

### DIFF
--- a/Slim/Middleware/PrettyExceptions.php
+++ b/Slim/Middleware/PrettyExceptions.php
@@ -66,11 +66,10 @@ class PrettyExceptions extends \Slim\Middleware
         try {
             $this->next->call();
         } catch (\Exception $e) {
-            $env = $this->app->environment();
-            $env['slim.log']->error($e);
+            $this->app['log']->error($e);
             $this->app->contentType('text/html');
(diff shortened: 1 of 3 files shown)


thomas-0816:
AccessAccess is slower than member access (http://bit.ly/PPvkvx).
Auto-completion in an IDE is only possible for members, not for array keys.
AccessAccess makes protected members public which is really bad for router, request, response and environment.

It would be nice to have offsetGet() calling closures automatically, e.g. lazy database connects:
$app['db'] = function() {
  return new mysqli(...);
};
$app['db']->query();

or another example with memcache:
$app = new \Slim\Slim();
$app['mc'] = function(){
&nbsp; return memcache_connect('memcache_host', 11211);
};

$app->get('/books/:id', function ($id) use ($app) {
&nbsp; $some_key = $app['mc']->get('some_key');
});

Instead of $this->config('mode'); there would be better $this->settings['mode']; or $this->settings->mode;

StJudeWasHere:
I'm closing this in favour of #436.

PR #441, opened 2012-10-08
Using use statement to include namespace

silentworks:
This is so that users of older version of Slim can easily upgrade to the new version without worrying about typing the \Slim namespace each call to a Slim class.


codeguy:
Doesn't this nullify using namespaces? It's easy enough for the end user to manually do this if they don't want to type \Slim so much. Not sure this is an assumption we should make.

silentworks:
No this doesn't nullify the use of namespace at all. I think its useful for users transitioning from older version of Slim.

PR #444, opened 2012-10-09
••CORRECTED•• Implemented late static binding

dannykopping:
Implemented late static binding on Slim class so that public/protected static functions can be overridden in deriving classes


codeguy:
Can you rebase against the latest code? This commit assumes the 405 handler was added.

PR #445, opened 2012-10-09
••CORRECTED•• Added "405 Method Not Allowed" error handler

dannykopping:
Added "405 Method Not Allowed" error handler - for overall parity with pre-existing 404 & 500 error handlers


codeguy:
An earlier PR has made this unnecessary. Removed the entire concept of a 405 response in #418.

PR #447, opened 2012-10-12
IIS Web.config for routing added

marchenko1985:
If, for some reasons, you are trying to get Slim working Windows IIS here is example of working Web.config


codeguy:
I don't want to include the actual Web.config file in the repo. Your update to the README file should be sufficient. Can you resubmit this PR with only the README change?

PR #450, opened 2012-10-18
move dispatch method to \Slim\Route

StJudeWasHere:
This way the route completely takes care if itself.


StJudeWasHere:
Updated to make it work with the current develop branch.

PR #453, opened 2012-10-20
move autoloader to its own file

StJudeWasHere:


codeguy:
This will be a breaking change that will require apps to be updated to continue working. So this will likely wait until the  next major point release.

lalop:
Isn't this issue closed on the develop branch https://github.com/codeguy/Slim/blob/develop/Slim/Autoloader.php ?

codeguy:
Yep this is done.

PR #454, opened 2012-10-27
Implemented late static binding

dannykopping:
Implemented late static binding on Slim class so that public/protected static functions can be overridden in deriving classes

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -176,7 +176,7 @@ class Slim
     public function __construct($userSettings = array())
     {
         // Setup Slim application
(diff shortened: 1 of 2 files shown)


dannykopping:
Seems like the build is failing due to something else other than what I'm committing...
Maybe when I rebased I erroneously resolved a conflict?

codeguy:
Did you rebase against the latest develop branch?

dannykopping:
Yup

codeguy:
Ok I'll merge locally and see what's up. Thanks!

PR #465, opened 2012-12-18
Added support for custom route subclasses to \Slim\Router

elazar:
At some point in the past, I threw up [this gist](https://gist.github.com/1832757) as an example of implementing custom routes in Slim. The goal of this particular example was to allow larger Slim-based codebases to separate individual route callbacks into their own files, partly for easier code navigation and partly so that minimal code would be loaded to service any given request.

Right now, using custom routes isn't possible without also implementing a custom router. This pull request removes that need by modifying `\Slim\Router` to accept the name of a subclass of `\Slim\Route` via a new `route()` method and to have its `map()` method create instances of that subclass. These modifications use `\Slim\Route` as the default to maintain backward compatibility.

If this pull request is merged, a developer would only need to subclass `\Slim\Route` rather than having to do both that and subclass `\Slim\Router` as in my gist example.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -64,6 +64,11 @@ class Router
     protected $matchedRoutes;
 
     /**
+     * @var string Name of the class to use for route objects
+     */
+    protected $routeClass = '\Slim\Route';
+
+    /**
      * Constructor
      */
     public function __construct()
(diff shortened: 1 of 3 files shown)


silentworks:
Please send this PR to develop branch instead.

codeguy:
Closing. Please re-submit to the develop branch.

PR #466, opened 2012-12-18
Added support for custom router implementations to \Slim\Slim

elazar:
This pull request is intended to complement [this one](https://github.com/codeguy/Slim/pull/465). While it's no longer necessary to implement custom routers if the latter is merged, the former makes it easier to do so anyway.


codeguy:
Thanks! Can you please re-submit to the develop branch?

PR #469, opened 2012-12-18
Added support for custom router implementations to \Slim\Slim

elazar:
Applied changes from [this pull request](https://github.com/codeguy/Slim/pull/466) to the develop branch as requested.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -180,7 +180,7 @@ class Slim
         $this->environment = \Slim\Environment::getInstance();
         $this->request = new \Slim\Http\Request($this->environment);
         $this->response = new \Slim\Http\Response();
-        $this->router = new \Slim\Router();
+        $this->router = $this->router();
         $this->middleware = array($this);
         $this->add(new \Slim\Middleware\Flash());
         $this->add(new \Slim\Middleware\MethodOverride());
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -607,11 +607,25 @@ class Slim
     }
 
     /**
-     * Get the Router object
+     * Get and/or set the Router object
+     *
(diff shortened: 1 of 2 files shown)


elazar:
Ah, I think I actually intended to change that constructor line to call out to `$this->router()` (to allow for subclasses to override the behavior). I'll try to get that fixed today. Thanks for the feedback. :)

codeguy:
Slim will be getting an IoC container really soon which should make it super simple to define custom router and route objects immediately after app instantiation. This way you can simply instantiate your subclass inside of the router or route's Closure definition.

Thoughts?

elazar:
@codeguy Any chance we could see some example code showing usage of this new feature as you've described it?

codeguy:
Sure. You would define a custom Router for Slim to use like this:

```
<?php
$app = new \Slim\Slim();
$app->container->singleton('router', function ($c) {
    return new \My\Router();
});
// Define routes
$app->run();
```

This assumes that `\My\Router` implements the same interface as `\Slim\Router`, but I think that's a safe assumption to make.

elazar:
I think I would codify that assumption with an actual interface:
1. Create an interface (e.g. `\Slim\RouterInterface`) with the same API as `\Slim\Router`
2. Have `\Slim\Router` explicitly implement that interface in its class definition
3. Change any existing typehint references to `\Slim\Router` in Slim to instead reference that interface

At that point, anything implementing the interface can be provided and PHP internals for checking interface implementation will provide helpful error messages if the interface implementation is flawed.

codeguy:
I agree. Good point. But do you otherwise approve of my approach?

elazar:
Otherwise, sure, this looks like a good approach.

codeguy:
Closing in favor of https://github.com/codeguy/Slim/issues/398

PR #471, opened 2012-12-24
Wrong return type

kusmierz:
getInstance returns \Slim\Slim or Slim (in namespace \Slim), not Slim.


codeguy:
Please send all PRs to the develop branch.

PR #475, opened 2013-01-13
PATH_INFO in url rewritten applications with public directory

xjunior:
PATH_INFO was being set to "/" on most of the situations, leading to misrouting.

Example app would be:

/
-- .htaccess
-- webroot
   -- index.php

.htaccess file:

RewriteEngine On

RewriteRule ^.htaccess$ - [F]

RewriteCond %{REQUEST_URI} ^/(javascript|stylesheet|images)/._$
RewriteRule ^(._)$ /webroot/$1 [L]

RewriteCond %{REQUEST_URI} !^/webroot/._$
RewriteRule ^(._)$ /webroot/index.php [QSA,L]

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -145,7 +145,10 @@ class Environment implements \ArrayAccess, \IteratorAggregate
             } else {
                 $env['SCRIPT_NAME'] = str_replace('\\', '/', dirname($_SERVER['SCRIPT_NAME']) ); //With URL rewrite
             }
-            $env['PATH_INFO'] = substr_replace($_SERVER['REQUEST_URI'], '', 0, strlen($env['SCRIPT_NAME']));
+            $env['PATH_INFO'] = $_SERVER['REQUEST_URI'];
+            if (strpos($_SERVER['REQUEST_URI'], $env['SCRIPT_NAME']) === 0) {
+                $env['PATH_INFO'] = preg_replace($env['SCRIPT_NAME'], '', $_SERVER['REQUEST_URI'], 1);
+            }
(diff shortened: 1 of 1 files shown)


StJudeWasHere:
I think this would be something good to have in Slim but it doesn't pass the tests.

xjunior:
You're probably right. I fixed this cause it was an undocumented pain that I had. I had to code PHP once and Slim was my choice.

Fixed the issue I had but really didn't even ran the tests.

xjunior:
Just checked the build output and the issues seem to be pretty straight forward to fix. Mainly (probably) version compatibility on preg_replace and a missing "/" in the end of a path (I'd have added that if I've ran the tests).

doublemarked:
I believe that your ticket is basically the same as the one I just researched and submitted: https://github.com/codeguy/Slim/issues/521

...my feeling is that the test may be broken. Can you read through my ticket and share your thoughts?

doublemarked:
BTW - I ran your branch through phpunit. It seems kinda broken. Your preg_replace call is missing a regex, and it's erroring out on a bunch of tests.

Maybe you mean to call substr_replace instead of preg_replace? I believe your code will not work if mod_rewrite is disabled.

xjunior:
I really haven't tested at all =/ I was scratching my own itch and was really in a rush (code assessment). I just needed the framework to work properly on a situation I believed that should be the default one.

Sorry to disappoint you. I really have no time/interest to work on this anymore. Wish you the best to fix this issue.  

## 

Carlos Palhares
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

doublemarked:
No worries!

PR #476, opened 2013-01-17
Handle rewrites that route all requests to a subfolder (e.g. via .htaccess & mod_rewrite)

mozmorris:
I very often structure my code in a way to keep my 'app' or any other sensitive folders out of the Document Root (webroot). Some of the hosting environments I deploy don't allow me to change the webroot. In these instance, I use mod_rewrite to route all requests to a sub folder, in essence moving the webroot.

After using Slim, it became apparent this approach wasn't possible with the current code base. I've also followed and read the discussion [here](http://help.slimframework.com/discussions/problems/769-url-routing-breaks-if-rewrite-rules-dont-match-underlying-structure) and [here](http://help.slimframework.com/discussions/problems/769-url-routing-). It appears others have the same predicament.

I've tested these changes with various set-ups. I haven't been able to test with Nginx or lighttpd.


codeguy:
I've decided to leave this as is for now. I may revisit in the future.

PR #477, opened 2013-01-23
Nicely formatted error messages when using cURL

timdavies:
I have been building an API for a website I built using Slim and, whilst testing, have been using cURL to post data to the API. Whenever I made a mistake in the code, the error messages I received were hard to read and obfuscated by HTML tags, no linebreaks, etc.

To help my testing, I added a check into Slim/Middleware/PrettyExceptions.php to detect whether the user is connecting with cURL and, if they are, output the error message and debugging information as plaintext.

This has been quite useful to me and I predict that others will find it useful too as I suspect that a popular use of Slim may be developing simple APIs. If it's something you'd rather not include however, that's ok too :-)

### DIFF
--- a/Slim/Middleware/PrettyExceptions.php
+++ b/Slim/Middleware/PrettyExceptions.php
@@ -82,33 +82,56 @@ class PrettyExceptions extends \Slim\Middleware
      */
     protected function renderBody(&$env, $exception)
     {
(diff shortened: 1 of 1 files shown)


timdavies:
Here's a preview (although I've had to remove large chunks of the error - sorry, I'm at work): 
screenshot

PR #478, opened 2013-01-28
AUTOMATED: WhiteSpace Removal

WhiteSpace-Strippers:
This is an automated request to remove excessive whitespace from source code. You can find more about this project and opt-out at [whitespacestrippers.com](http://whitespacestrippers.com/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
+	- 5.3
+	- 5.4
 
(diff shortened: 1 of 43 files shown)


smagch:
-1
You are just replacing spaces to tabs. As you can see [here](/codeguy/Slim#style-guide), Slim follows [PSR-2](/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) standards which MUST use 4 spaces for indenting, not tabs.

Issue #55, opened 2010-11-17
nginx Configuration


nielsen:
I've been having some trouble getting Slim to play nice with nginx; the site at the root is implementing Slim, but the sub-folders have other apps (Wordpress for example).

This is what I have so far; any thoughts?

```
    location / {
        index bootstrap.php;
        try_files $uri $uri/ /bootstrap.php; 
       # the working WP configuration takes ?q=$1 parameters; should Slim as well?
       # try_files ~ if ( !-e $request_uri ), only less taxing.
    }
```

codeguy:
I am unfamiliar with nginx configuration, which is why I've recommended Apache for Slim. I do know that one of the Slim contributor's (hexa2k9) had created a working nginx configuration file for Slim. See this commit:

https://github.com/codeguy/Slim/commit/4747f4c84e3ef682bd6571df7fb68f6be218c7a1

I'm closing this out since it's not really an issue with Slim. But feel free to leave additional comments if you have any further questions.

dmtrs:
@nielsen :+1:

PR #483, opened 2013-02-14
Added query params support to nginx

hjr3:
The documentation for the nginx configuration did not send query parameters in
the URL to the calling code. This means something like /test?hello=world would
fail to set the $_GET superglobal with the 'hello' key.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -66,7 +66,7 @@ should contain this code:
 
 Your nginx configuration file should contain this code (along with other settings you may need) in your `location` block:
 
-    try_files $uri $uri/ /index.php;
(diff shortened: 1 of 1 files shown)


smagch:
+1
I'm currently doing exactly the same way in my `nginx.conf`. The first time I set up nginx, it takes a hour to solve this problem.

#55

tyd:
+1 ... this got me the first time I used Slim with nginx. Using the same solution now.

codeguy:
Please send all PRs to the develop branch.

hjr3:
Opened #500

PR #487, opened 2013-03-06
Add basic support for php-cli

pgasiorowski:
Allows running a Slim app from comman line.
Converts command line params into Slim GET routes
php index.php hello WooDzu -> $app->get('/hello/:who', ... ) ->
/hello/WooDzu

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -117,49 +117,70 @@ class Environment implements \ArrayAccess, \IteratorAggregate
         if ($settings) {
             $this->properties = $settings;
         } else {
-            $env = array();
 
-            //The HTTP request method
-            $env['REQUEST_METHOD'] = $_SERVER['REQUEST_METHOD'];
+            // Virtual environment for php-cli (ignore for phpunit)
+            if( PHP_SAPI == 'cli' && $_SERVER['SERVER_NAME'] != 'slim' )
+            {
 
-            //The IP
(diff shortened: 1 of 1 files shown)


CodeBrauer:
@codeguy - is there any option to run it on cli (for cronjobs)

mnapoli:
in that case use a CLI framework like [Symfony Console](http://symfony.com/doc/current/components/console/introduction.html) or [Silly](http://mnapoli.fr/silly/)

CodeBrauer:
I already use slim for a project that now needs a cronjob - switching is not so easy .. If there is no support I will build a pure PHP standalone script

mnapoli:
You don't need to switch framework, you can use one alongside the other. Yes you would have to write another PHP file (e.g. `console.php` or whatever) in addition to the existing `index.php`, but the content of that file should be pretty easy if you have structured your application correctly.

pgasiorowski:
I appreciate this pr is somewhat a hack but it's a little pain that slim won't run from cli by default.
The method for cron jobs I use is to invoke curl eg.
`* * * * * curl http://127.0.0.1/api/v1/tasks/gc`.

akrabat:
Maybe http://akrabat.com/run-a-slim-2-application-from-the-command-line/ can help?

CodeBrauer:
Of cause a standalone file would be easy but I just thought it would be possivle directly. Running the cron with cURL would be possible also, but could bring new problems because it's running on your webserver and not in your commandline.

PR #489, opened 2013-03-12
Support 'ANY' Route Method

gmanricks:
## ANY Routes

This adds a 'short circuit' for routes with the ANY method, so that they will match all HTTP methods (even made up ones).

The main functionality comes from the change in the router class. Instead of this:

``` php
if (!$route->supportsHttpMethod($httpMethod)) {
```

I changed it to:

``` php
if (!$route->supportsHttpMethod($httpMethod) && !$route->supportsHttpMethod("ANY")) {
```

Other additions include a helper method inside the Slim class (named <code>any</code>) and a unit-test in the SlimTest.php file

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -100,7 +100,7 @@ class Router
         if ($reload || is_null($this->matchedRoutes)) {
             $this->matchedRoutes = array();
             foreach ($this->routes as $route) {
-                if (!$route->supportsHttpMethod($httpMethod)) {
+                if (!$route->supportsHttpMethod($httpMethod) && !$route->supportsHttpMethod("ANY")) {
                     continue;
                 }
 
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -483,6 +483,18 @@ class Slim
     }
 
     /**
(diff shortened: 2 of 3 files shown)


pgasiorowski:
+1 for this

smagch:
-1
`ANY` is not HTTP method. It's inconsistent with current API. I can't see the use case for that. You can use `->map()` and `->via()`. http://docs.slimframework.com/#Custom-HTTP-Methods

``` php
$app->map('/any', function() {
    echo 'call me via any HTTP method';
})->via('GET', 'POST', 'DELETE', 'PUT', 'OPTIONS');
```

gmanricks:
You are right it is not an official HTTP method, but that wasn't the point. This addition doesn't look for a methods named <code>ANY</code>. It responds to any other method even custom ones that a user may implement, so I don't think it goes against the current API.

smagch:
Ah, I hadn't run into a situation that I want to map all method. But it seems reasonable in a certain case. I've just looked into [Respect/Rest](/Respect/Rest) which has an [any feature](https://github.com/Respect/Rest#class-controllers).

PR #490, opened 2013-03-20
Added support for $app->forbidden(), with accompanying unit tests

mdlayher:
Added support for an `$app->forbidden()` handler which functions identically to `$app->notFound()`, but for HTTP 403 Forbidden errors.  Unit tests have also been added, and are passing.

Travis build can be found here:
https://travis-ci.org/mdlayher/Slim/builds/5666420

Also tweaked a few minor lines in `$app->notFound()` which did not conform to PSR-2.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -107,6 +107,11 @@ class Slim
     protected $error;
 
     /**
+     * @var mixed Callable to be invoked if user forbidden
+     */
+    protected $forbidden;
+
+    /**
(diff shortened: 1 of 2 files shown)


codeguy:
I think there's a more elegant solution to be had (this applies to my current implementation). My own current implementation is too tightly coupled to "not found" and "error" handlers and can be abstracted to better support other  scenarios, such as "forbidden", etc. If you want to explore how to do this that'd be great.

PR #492, opened 2013-03-23
Fixed Tests for PHP 5.3

gmanricks:
There were a few function dereferencing in the tests which caused the tests to fail in PHP 5.3


codeguy:
Thanks! I'm 5.4 only now so appreciate you catching these.

PR #495, opened 2013-03-28
Added slim.before.render default hook

natestarner:
This hook is applied directly before the template render call.

A common use case for this hook would be:

``` php
$app->hook('slim.before.render', function () use ($twig) {
  $twig->addGlobal("session",$_SESSION);
});
```

So the latest possible values in the $_SESSION will be available to the view template regardless of where they are set in the app life cycle.


codeguy:
I think there is a better solution to assign session data to a view... this process will become much easier very soon, too, with a new Slim interface to session data.

PR #497, opened 2013-04-02
New Feature - Route Groups

gmanricks:
# Route Groups

This pull-request adds the ability to declare route groups. It accepts the same parameters as the route functions do (i.e. pattern, mw1, mw2, etc... callable), and will work with placeholders in the pattern.
#### Basic Use

``` php
$app->group('/v1.0/', $mw, function () use ($app) {
    $app->get('/users', function() {
        echo "This is a route for '/v1.0/users';
    }
});
```
#### Using Placeholders

``` php
$app->group('/users/:name', function () use ($app) {
    $app->get('/posts/:id', function ($name, $id) {
        echo "Showing post id #" . $id . " for user " . $name; 
    });
});
```
#### Nesting Groups

``` php
$mw = function () { echo "Middleware 1 - "; };
$app->group('/v1.0/', $mw, function () use ($app) {

    $mw = function () { echo "Middleware 2 - "; };
    $app->group('/users/', $mw, function () use ($app) {

        $mw = function () { echo "Middleware 3 - "; };
        $app->get('/:name', $mw, function ($name) {
            echo "Hello " . $name;
        });

    });
});
```

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -64,11 +64,17 @@ class Router
     protected $matchedRoutes;
 
     /**
+     * @var array Array containing all route groups
+     */
+    protected $routeGroups;
+
+    /**
      * Constructor
      */
     public function __construct()
     {
         $this->routes = array();
+        $this->routeGroups = array();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


StJudeWasHere:
This is a good idea and something like it could be used to distribute Slim plugins like a blog engine or a user management system.

Silex does this in an interesting way that might be worth exploring too:

http://silex.sensiolabs.org/doc/organizing_controllers.html

PR #498, opened 2013-04-04
allow prepending to routes

pvwoods:
The ability to prepend to existing routes is ideal for api's that may have multiple versions (i.e. appending api.example.com/1.4/users where the version is the prepend portion).

the ability to pass a setting called 'prependToRoutes' will now allow users to add data such as a version number to the front of all roues for a given app.


codeguy:
Closing out in favor of https://github.com/codeguy/Slim/pull/497

PR #499, opened 2013-04-04
Blade like templates

Gufran:
Blade like templating. Now developers can format the template using blade syntax like 
{{ $variable }}
@if...@elseif...@else, 
@foreach etc.

Templating is not full blown Blade but just a little part of it. only a limited set of methods is available.

Echoing variables in curly braces e.g. {{ $some_variable }} and {{{ $escaped_string }}}
blade like comments e.g. {{-- this is a comment --}}
including other view file using @include('some_other_view'), 
and these control structures:
@if .. @endif
@if .. @elseif .. @endif
@while .. @endwhile
@for .. @endfor
@foreach .. @endforeach
are implemented

### DIFF
--- a/Slim/View.php
+++ b/Slim/View.php
@@ -212,6 +212,14 @@ class View
      */
     public function render($template)
     {
+        $template = $this->getTemplatesDirectory() . '/' . ltrim($template, '/') . 'tpl.php';
+        
+        if(!file_exists($template)) {
+            throw new \Exception("Cannot render non existing view file " . $template);
+        }
+
+        $template = file_get_contents($template);
+
         // Make sure we include all the required views first
         
         $template = $this->compile_includes($template);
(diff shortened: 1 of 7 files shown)


StJudeWasHere:
It would be great to have Blade templating support in Slim but in my opinion this should go in Slim-Extras, just like we do with Twig, Smarty and all the others.

miraage:
You are welcome to create new extra view.
https://github.com/codeguy/Slim-Extras/tree/master/Views

Gufran:
I completely understand the point here. If you can just spare some time to have a look into diff you'll find just a bunch of functions to perform some basic parsing to make template look cleaner and feature rich as well. I dont think we need to roll out a separate library for those tiny functions.
Anyway, I implemented this for my personal use because I was missing my Laravel way of managing html. Thought of sharing with everyone so here it is.

codeguy:
This is not something that will be added to the core repo. You are welcome to create a custom view and submit a PR to the Slim-Extras repo.

PR #500, opened 2013-04-06
Added query params support to nginx

hjr3:
The documentation for the nginx configuration did not send query parameters in
the URL to the calling code. This means something like /test?hello=world would
fail to set the $_GET superglobal with the 'hello' key.


codeguy:
Thanks!

PR #502, opened 2013-04-11
adding dispatchContext support (for PHP 5.4+)

janeklb:
_I'm not sure if this implementation is ideal, however I think the idea itself is quite powerful and I figured it would be best illustrated via a pull request._

When building Slim apps I've often found it desirable to have access to the \Slim\Slim object inside my route callables. Of course, you can always use PHP's `use` keyword:

``` PHP
$slim = new \Slim\Slim();
$slim->map('/foo', function() use($slim) {
  $slim->forward('/bar');
});
```

However, this can be a bit tedious after declaring a lot of routes.
As a shortcut, I propose a _"dispatchContext"_ that will be accessible via `$this` inside the callables.

The pull request implements this in a way that would be used as follows:

``` PHP
$slim = new \Slim\Slim();
$slim->setDispatchContext($slim);
$slim->map('/foo', function() {
  $this->forward('/bar');
});
```

Of course, the dispatch context can be set to anything, which could make it a very powerful controller delegate.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -419,12 +419,19 @@ class Route
 
     /**
      * Dispatch
-     * @return mixed The return value of the route callable, or FALSE on error
+     * @param  mixed  $context the object context in which the callable should be invoked
+     * @return object          The return value of the route callable, or FALSE on error
      */
-    public function dispatch()
+    public function dispatch($context = null)
     {
         foreach ($this->middleware as $routeMiddleware) {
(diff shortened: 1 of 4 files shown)


janeklb:
<del>On further thought, instead of adding `setDispatchContext($context)`, it might be better to add a `$dispatchContext` parameter to `run()` and `call()` on `\Slim\Slim` and use that instead.</del>

## 

Scratch that. This would rely on middleware passing the context parameter around.

PR #505, opened 2013-04-14
On some environments fopen('php://stderr', 'w') can produce a PHP Warning

dmitry-lomakin:
We'd like to get rid of that


codeguy:
Please resubmit PR to the develop branch.

PR #507, opened 2013-04-17
Wouldn't it be nice if class dependencies were (somewhat) configurable?

al-the-x:
Then I wouldn't have to subclass `\Slim\Slim`, `\Slim\Router`, and `\Slim\Route` just to change some behavior in `\Slim\Route`...! I could _only_ subclass `\Slim\Route` and inform `\Slim\Router` of the dependency change, e.g.

``` php
<?php // index.php

class CustomRoute extends \Slim\Route
{
    function supportsHttpMethod($method)
    {
        // Respond to all HTTP methods by default...
        if ( !$this->methods ) return true;

        return parent::supportsHttpMethod($method);
    }
}

\Slim\Router::$CLASS_ROUTE = 'CustomRoute';

$app = new \Slim\Slim();

// Route "/some/route" for _any_ HTTP method...
$app->map('/some/route', function(){
    echo 'Do Something Awesome';
});

$app->run();
```

I thought about using `\Slim\Slim::config()` but not all classes in the package receive the instance of `\Slim\Slim` to call `config()` on, so I thought better to encapsulate dependencies with static variables acting as psuedo-constants.
- PRO: Each class is responsible for its own dependencies.
- CON: The dependency can only be overridden for the entire class.

Discussion?

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -53,6 +53,20 @@ class Request
     const METHOD_OVERRIDE = '_METHOD';
 
     /**
+     * @var string classname to use for $headers instance
+     * @see \Slim\Http\Request::$headers
+     * @see \Slim\Http\Request::__construct()
+     */
+    public static $CLASS_HTTP_HEADERS = '\Slim\Http\Headers';
+
+    /**
+     * @var string classname to use for $cookies instance
+     * @see \Slim\Http\Request::$cookies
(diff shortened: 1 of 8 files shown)


codeguy:
I'd prefer to let each existing class change its behavior by injecting an appropriate callable or closure.

al-the-x:
Example? Discussion?

PR #508, opened 2013-04-17
Shouldn't `\Slim\Route::via()` call `\Slim\Route::appendHttpMethods()` instead of duplicating code?

al-the-x:
Feel free to call me an idiot, it just seems that `via()` was unnecessarily repeating the code in `appendHttpMethods()` to avoid `call_user_func_array()`... Why can't invoking methods with variable arguments be easier in PHP? Like Ruby, `$this->__send('appendHttpMethods', func_get_args());` would be nicer...

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -277,8 +277,7 @@ class Route
      */
     public function via()
     {
-        $args = func_get_args();
-        $this->methods = array_merge($this->methods, $args);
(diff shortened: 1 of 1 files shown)


codeguy:
I agree it's messy and not very DRY short of using the call_user_func_array function. But the solution is just as messy. I'm going to leave as is if only for obviousness and legibility. Thanks for bringing this up though. It is an issue, but one I'd rather live with for now.

PR #513, opened 2013-04-28
Option to set default values to Request parameters

webpatser:
Added the option to set a default value for the post, get, cookie and delete methods if it's not passed in the request.

Now it defaults to NULL which is sometimes not the value you need it to be :-)


codeguy:
Instead of this:

```
if (!empty($default)) {
    return $default;
} else {
    return null;
}
```

Why not just do this:

```
return $default;
```

If you want to make this change (or post why you would not do this), I'll get this pulled in. Also, please resubmit to the develop branch.

PR #514, opened 2013-04-28
new feature on Router: list all defined routes with method

t1gr0u:
Hi,

we are using Slim on our RESTful APIs and one thing that we do is to list all our available routes, so who ever needs our API they know which routes are available.

$app->router->listRoutes();
This will return an array of routes:
array(
  'GET::/foo',
  'POST::/bar'
);

You can also change the separator or don't show the method
$app->router->listRoutes( '->' );
array(
  'GET->/foo',
  'POST->/bar'
);

$app->router->listRoutes( '::', false );
array(
  '/foo',
  '/bar'
);

Just thought that maybe someone else would be interested in this feature.

T1gr0u

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -281,7 +281,7 @@ class Router
      * Get all the defined routes
      * @param string $separator separator between the method and the route ( default: "::" )
      * @param bool $showMethods show the method of the route ( default: true )
-     * @return string
+     * @return array
      */
     public function listRoutes( $separator = '::', $showMethods = true ) {
       $allRoutes = array();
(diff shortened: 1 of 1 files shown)


codeguy:
I like where you're going with this. But I feel like this functionality is best kept outside of the core framework. This would be a good addition to the Slim-Extras repo, perhaps in a separate introspection lib. The core framework, as you've demonstrated, already provides the interface necessary to extract and display this information.

If you want to approach it from this angle and submit a PR to the Slim-Extras repo, I'd be glad to pull that in for you.

Also, make sure your submitted code is PSR compliant.

PR #516, opened 2013-05-03
Added possibility dynamically add new routes, and pass to them

justgook:
```
    \Slim\Environment::mock(array(
        'SCRIPT_NAME' => '/foo', //<-- Physical
        'PATH_INFO' => '/name/Frank/test', //<-- Virtual
    ));
    $s = new \Slim\Slim();
    $s->get('/name/:name+', function () use ($s) {
        echo "Fail"; //<-- Should not be in response body!
        $s->get('/name/Frank/test', function () use ($s) {
            echo "Frank";
        });
        $s->pass();
    });
    $s->call();
    $this->assertEquals('Frank', $s->response()->body());
```


silentworks:
Not sure where the get() inside of the get() method works, is that what you are implementing? if yes we currently have group() method in develop branch which accomplish this.

Also when sending PR try and send from develop to develop, your master would be missing some of the updates that develop branch has.

justgook:
ill rebuild repush it, but main idea is to be able dynamically add new routes after cycle of routes is started, so if I i add some new route, and use "$app->pass()", than this (just added) route newer fires.
it is usefully when u wont to create lazy loading and define routes form some kind of "controller"

``` php

class Controller {
    private $app,
            $prefix;
    public function __constructor(\Slim\Slim $app) {
        $this->app = $app;
        $routes = $this->register($app);
    }
    protected function register(\Slim\Slim $app) {}
    public function setPrefix($prefix) {
        $this->prefix = $prefix;
    }
    protected function getPrefix() {
        return $this->prefix;
    }
    /**
     * Register all routes to main App
     * @param \Slim\Slim $app
     * @param array      $routes prefix_path:class pairs expl: wiki:z0\modules\Wiki
     */
    public static function defineRoute(\Slim\Slim $app, array $routes) {

        foreach ($routes as $value) {
            list($prefix, $className) = explode(':', $value);
            $app->map($prefix .'/:sub+', function($sub) use($app ,$className, $prefix) {
                $curentController = new $className();
                $curentController->setPrefix($prefix);
                $curentController->register($app);
                $app->pass();
            })->via('GET','POST','PUT', 'DELETE');
        }
    }
}
```

main problem that i dont wont load all unnecessary controllers until i need them.
Disadvantage a cant use those for url generating.

silentworks:
This seems like a bit of a overkill. I prefer the current setup where you define your routes before the run loop completes. I don't think being able to add routes after the run loop is a good idea (well not for Slim at least).

justgook:
btw it increase performance by destroying addition ram memory leak when you every time create new instance of array.. so it is not only possibility dynamically add new routes, but performance increase. so after few updates there can be move all router filtering to that instance (RouterIterrator)

PR #518, opened 2013-05-07
Feature/getflash

silentworks:
Allow to get and set flash with the same method, instead of accessing the $_SESSION directly.


codeguy:
Closing. Andrew, I'd like to better decouple Flash messages from the Slim object.

PR #523, opened 2013-05-18
Test for mcrypt in composer.json

webpatser:
Since mcrypt is needed to for the encrypt/decrypt methods and the unit test will fail if mcrypt is not present, a check in composer would be handy.

Problem is that if the mcrypt library is not present NO encryption is being performed without warning, thus returning the string from the \Slim\Http\Util::encrypt method the same as it was passed.

Another possibility is to alter the test to not fail if mcrypt is not present, but then an alternate encryption method is needed that not depends on mcrpyt.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0"
+        "php": ">=5.3.0",
+        "ext-mcrypt": "*"
     },
     "autoload": {
         "psr-0": { "Slim": "." }


evert:
I agree with @silentworks . mcrypt should be optional, as slim will function without mcrypt as well.

webpatser:
@evert For security reasons I would go for the require statement.

Or add some fallback encryption method to Slim and then remove the requirement.

evert:
Well if you're only using Slim for routing.. it's just a bit cumbersome if that requirement exists. You can still throw an exception if mcrypt-related functionality is called, and mcrypt is not installed.

webpatser:
I understand your specific case. But since other people are also using it for more then just routing :-D

silentworks:
I hope you noticed I mentioned that mcrypt is installed on my production server, but I don't want it on my local. I would prefer if the framework not dictate to me about what I need to have installed as a requirement and more suggest to me. So this would be better off in a `suggest` rather than a `require`. Also make it more prominent in the docs that you should never put Slim into production without mcrypt setup.

I am coming from the point of Slim has always kept its requirements low and not complicate the setup for anyone, hence why we have a Simple DI built in rather than using Pimple (which would become a dependency). I think we should keep it simple as always.

evert:
@webpatser of course. Making it optional would cover both use-cases.

webpatser:
Like I said. I totally get your setup's etc. It's for @codeguy to decide to write a fallback, remove mCrypt from the encryption functions or to live by the fact that if you have it not installed you don't get encryption, when you think it does.

until then I would say, leave this requirement. And if you know what you are doing, remove the requirement manually.

designermonkey:
I'd just like to point out that Laravel does the same and fails without mcrypt installed, even though for some projects, you may not be using encryption. This has caused me massive headaches on a server that I can't get mcrypt installed on (don't ask, it's irritating), and so I would agree that this is a bad move.

tuupola:
Missing mcrypt should not be a showstopper. [Documentation](http://docs.slimframework.com/) says _"The mcrypt extension is required only if you use encrypted cookies."_ Either this patch should be reverted or documentation updated to state that mcrypt extension is mandatory.

tuupola:
Here is a [workaround](https://github.com/tuupola/slim-fake-mcrypt). Add the following to your composer.json

``` json
{
  "repositories": [{
      "type": "vcs",
      "url": "https://github.com/tuupola/slim-fake-mcrypt.git"
  }],
  "require": {
    "tuupola/slim-fake-mcrypt": "dev-master",
    "slim/slim": "2.3.*"
  }
}
```

josegonzalez:
@tuupola you are amazing. <3

josegonzalez:
@tuupola (please add to composer as `fake-mcrypt`)

josegonzalez:
For those using this workaround: `composer update` will not wish to run locally if you already have ext-mcrypt installed somehow. I'd still prefer the mcrypt requirement be dropped (maybe placed into middleware if anything) or this commit reverted at worst, but uninstalling mcrypt locally before running seems to work.

tuupola:
Can you paste the composer output to pastie.org and post link here? I will install ext-mcrypt somewhere to debug it.

PR #529, opened 2013-06-05
Use DIRECTORY_SEPARATOR instead of '/'

lalop:


codeguy:
Please separate into different PRs by using different branches for each new feature. Also submit to the develop branch.

PR #530, opened 2013-06-05
View exception

lalop:
If an exception is throw while rendering a view the error view wasn't sent.


codeguy:
Can you separate your changes into different pull requests?

lalop:
Yes I can :)
sorry for that

PR #531, opened 2013-06-05
OPTIONS HTTP method now returns all HTTP methods available for current URI

vlakarados:
RESTful web services should provide a header that shows all methods available for current URI. This sends the header as well as adds two helper methods to Router class that may prove useful for more purposes.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -95,6 +95,41 @@ class Router
     }
 
     /**
+     * Return all route objects that match the given URI
+     * @param  string               $pattern      The pattern to match against
+     * @return array[\Slim\Route]
+     */
+    public function getAllRoutes($pattern = null) 
+    {
+        if(null === $pattern) {
+            return $this->routes;
+        } else {
+            $routes = array();
+            foreach($this->routes as $route) {
+                if($route->getPattern() == $pattern) {
+                    $routes[] = $route;
+                }
+            }
+            return $routes;
+        }
+    }
+    
+    /**
+     * Return array of methods avaliable for the current pattern
+     * @param  string               $pattern      The pattern to match against
+     * @return array
+     */
+    public function getMethodsAvailable($pattern) 
+    {
+        $methods = array();
+        foreach($this->getAllRoutes($pattern) as $route) {
+            $methods = array_merge($route->getHttpMethods(), $methods);
+        }
+        $methods[] = "OPTIONS"; // Force options method as available as it must return this method's return value or self
+        return array_unique($methods);
+    }
+
+    /**
      * Return route objects that match the given HTTP method and URI
      * @param  string               $httpMethod   The HTTP method to match against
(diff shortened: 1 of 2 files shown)


vlakarados:
This change may be pushed to master branch as I can't see how it can break someones functionality.

nickshanks:
This doesn't add HEAD if GET is available.

vlakarados:
Considering CORS it doesn't require HEAD method, but I will implement this in the nearest future and commit this

nickshanks:
Have also just noticed that you set "Access-Control-Allow-Methods" but don't set the more basic Allow response header!

In a non-Slim project I use this:

``` php
function allow_methods($methods)
{
    // HEAD and TRACE are handled automatically
    // the tunnelling method CONNECT, the WebDAV methods PROPFIND, PROPPATCH, MKCOL, COPY, MOVE, LOCK and UNLOCK, as well as any proprietary methods like DEBUG and TRACK, are prohibited
    // I can see theoretical use cases for COPY, MOVE, LOCK and UNLOCK in the app, but have not needed them thus far. COPY and MOVE can be broken down into GET, PUT and DELETE methods anyway, which are better supported by intermediaries. LOCK would need a timeout facility

    // in a fully RESTful service, GET should be supported by *every* resource, and so should not need specifying
    //  we're not there yet (/action and /ajax are primary offenders; /action is needed for UAs not supporting PUT, PATCH and DELETE)

    $methods = trim($methods);
    if(empty($methods))
        return false;

    $methods = explode(' ', strtoupper($methods));
    $methods = array_intersect($methods, array ('GET','PUT','PATCH','LINK','UNLINK','DELETE','POST'));
    if(in_array('GET', $methods))
        array_unshift($methods, 'HEAD');    // HEAD is automatically supported by PHP, so long as we treat it as a GET

    // TRACE is handled by Apache
    $allow_response_header = 'Allow: '.implode(', ', $methods).', OPTIONS, TRACE';
    if($_SERVER['REQUEST_METHOD'] == 'OPTIONS') {
        // for issues with this, see http://www.mnot.net/blog/2005/04/03/options
        header('HTTP/1.1 200 OK');
        header($allow_response_header);
        exit;
    }
    if(!in_array($_SERVER['REQUEST_METHOD'], $methods)) {
        header('HTTP/1.1 405 Method Not Allowed');
        header($allow_response_header);
        exit;
    }
}
```

vlakarados:
Yes, I'm half-way in my project to get full CORS compatibility in Slim, will update this

akrabat:
This needs to be against the new code in the develop branch.
Also, would the additional code in run() be better off in Middleware?

akrabat:
@vlakarados any updates on this?

akrabat:
Closed to do lack of response. Please reopen if required.

nickshanks:
Codeguy, any word on where you're at with this?

codeguy:
I haven't touched this at all yet.

PR #534, opened 2013-06-06
 redirect to error template if there is an error whlie rendering the view

lalop:


Slim/View.php
```diff
@@ -209,8 +209,13 @@ public function render($template)
         $this->setTemplate($template);
         extract($this->data);
         ob_start();
-        require $this->templatePath;
-
-        return ob_get_clean();
+        ob_start();
```
nubs: It looks like you got an extra ob_start in here.

lalop:
Oo 
effectively, thank you

codeguy:
Can you clarify what issue you are trying to fix?

lalop:
I will try.
Today if an exception is thrown in the view the output buffer isn't turned off so the beginning of the view is send then it's the exception.
With my fix only the exception is send.

Is it more clear ?

lalop:
If you want i can add some test on this feature ?

lalop:
Do you need anything more on this point ? Information, tests ... ?

lalop:
Hello,
what happens for this issue ?

silentworks:
This seem to have been fixed now. Closing off.

PR #535, opened 2013-06-06
replace Container with Helper\Set

StJudeWasHere:
I've been playing with the new ioc branch and I like it a lot, I hope this makes it into the next release.

On the other hand, I was wondering if it would make sense to use the \Slim\Helper\Set class for the container instead of creating a new one. I gave it a try and it passes all the tests too.


codeguy:
There's certainly some overlap and I think they definitely could be merged. Will take a look at your suggestions tonight. Thanks!

PR #537, opened 2013-06-09
Helper\Set property overloading, restores compatibility with Mustache

zomble:
Not sure if wanted __set and __unset. So left them out.

Also doesn't seem to affect IoC changes, but obviously that could be an issue?


codeguy:
You may as well overload __set while you're at it. Go ahead and do that and I'll merge. It should not conflict with anything since property overloading and array access are separate things that will accomplish the same goal.

PR #545, opened 2013-06-23
File streaming

gmanricks:
# File Streaming

This PR adds two functions $app->sendProc() and $app->sendFile() as-well as a modified call loop which is backwards compatible.
### $app->sendProc($command, $contentType = "text-plain")

This function accepts a command and optionally a content type and will transfer the output to the client.
### $app->sendFile($file, $contentType = false)

This function accepts a URI either a local file or a remote URL, you can also specify a content type as the second parameter. If left blank the function will attempt to process and find an appropriate file type.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -136,6 +136,11 @@ class Response implements \ArrayAccess, \Countable, \IteratorAggregate
         $this->headers = new \Slim\Http\Headers(array('Content-Type' => 'text/html'));
         $this->headers->replace($headers);
         $this->cookies = new \Slim\Http\Cookies();
+        $this->body = (object) array(
+           "text" => "",
+           "isStream" => false,
+           "handle" => null
+        );
         $this->write($body);
     }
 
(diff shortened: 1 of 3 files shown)


codeguy:
Nice work!  Will get this into 2.4.

codeguy:
What do you think about changing the method name `sendProc` to `sendProcess`... I'm less concerned about brevity and more about obviousness.

gmanricks:
I renamed the function, also I forgot to mention I also added a function called setDownload([$filename]);

Which will set the header to force a download instead of the browser trying to open it. You can optionally pass in a filename to change the download's name to the specified string.

for example if you have a route for '/file'

and you run:

``` php
$app->setDownload();
```

it will force a download named 'file', whereas:

``` php
$app->setDownload("other.txt");
```

will download the contents of the route under the name 'other.txt'

codeguy:
Haven't forgotten about this. This is next on my todo list. Hoping to get it pulled into the develop branch either tonight or tomorrow.

PR #552, opened 2013-07-09
Added method to fix user's ip behind load balancer

SgtK:


codeguy:
I'd prefer not to overwrite the `REMOTE_ADDR` server variable. Also, `\Slim\Http\Request` already compensates for IPs behind load balancers. See:

https://github.com/codeguy/Slim/blob/master/Slim/Http/Request.php#L572

PR #559, opened 2013-07-14
make PHP aware that we are implementing the PSR logger interface, fix up...

aktau:
... method signatures so the interpreter doesn't throw a hissy fit

### DIFF
--- a/Slim/Log.php
+++ b/Slim/Log.php
@@ -58,7 +58,7 @@ namespace Slim;
  * @author  Josh Lockhart
  * @since   1.0.0
  */
-class Log
+class Log implements \Psr\Log\LoggerInterface
 {
     const EMERGENCY = 1;
     const ALERT     = 2;
@@ -187,7 +187,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function debug($object, $context = array())
+    public function debug($object, array $context = array())
     {
         return $this->log(self::DEBUG, $object, $context);
     }
@@ -198,7 +198,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function info($object, $context = array())
+    public function info($object, array $context = array())
     {
         return $this->log(self::INFO, $object, $context);
     }
@@ -209,7 +209,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function notice($object, $context = array())
+    public function notice($object, array $context = array())
     {
         return $this->log(self::NOTICE, $object, $context);
     }
@@ -220,7 +220,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function warning($object, $context = array())
+    public function warning($object, array $context = array())
     {
         return $this->log(self::WARN, $object, $context);
     }
@@ -232,7 +232,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function warn($object, $context = array())
+    public function warn($object, array $context = array())
     {
         return $this->log(self::WARN, $object, $context);
     }
@@ -243,7 +243,7 @@ class Log
      * @param  array       $context
      * @return mixed|bool What the Logger returns, or false if Logger not set or not enabled
      */
-    public function error($object, $context = array())
+    public function error($object, array $context = array())
     {
         return $this->log(self::ERROR, $object, $context);
     }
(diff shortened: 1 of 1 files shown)


codeguy:
Slim's log implements all of the PSR-3 methods. It does not technically implement `LoggerInterface` because it is not meant to be shared or used with other frameworks. It is only for Slim. You are more than welcome to override the Slim app's default logger with your own third-party logger, either via the app setting or using the new Resource Locator to inject your custom logger.

As is, your changes break unit tests since `LoggerInterface` is not available. I don't want to burden Slim with third-party dependencies since I still want to make it easy to use and install Slim manually, without Composer. You could copy the `LoggerInterface` definition into Slim's repo, but that creates a hard-coded dependency that will be obsolete should the fig-standards repo be updated in the future.

Leaving this open in case others want to add to the discussion.

aktau:
It is not about sharing or using with other frameworks, it is about interacting with components that have not been written with Slim's idea of a logger in mind. When I instantiate one of my own (or someone else's) components and that component requires a logger, I will generally do something like this:

``` php
$comp = new \Some\Namespaced\Component($someVariables, $app->log);

// the definition of \Some\Namespaced\Component, which I may or may not have control over:
class Component {
     public __constructor(array $variables, \Psr\Log\LoggerInterface $log) {
            // ...
     }
}
```

That is, I don't want to rewrite all components I want to use and log to accept a Slim logger instead of a PSR-3 logger. For now I was using a wrapper around the Slim logger which was ugly but doable. It seems I'll still have to use a wrapper, though it be slightly less ugly.

In this way, the PSR logger unifies logging capability. I'm normally not a fan of unity-soup (i.e.: design by committee, pandering to everyone, etc) but I believe logging to be one of those things that is exempt because it is (or should be) so ubiquitous. Everybody has the same logging problem, and if it could be fixed once and not duplicated all over the place that would be nice, real nice. This is the reason Monolog adopted PSR-3, so you can pass in a Monolog logger to something expecting a generic PSR-3 logger and have it just work. This is why log4j got so popular in the java world and copied to many other languages et cetera.

I can understand the dependency argument but it does hurt a little bit for me to have to shim/wrap all loggers. It's not impossible of course and I'll live (especially with 2.3 it's easy to replace), but meh.

codeguy:
Thanks for the feedback. I do understand your situation, but you still imply that Slim's default logger is a separate component intended to be shared among other third-party components. Which is not what it is meant to be.

In all honesty, I'd rather remove Slim's built in logger and defer to more specialized, third-party components that you can inject into Slim using the new resource locator. Currently, Slim's core code does not rely on the existence of a logger. The current \Slim\Log class only exists for your convenience while building Slim apps. I'd honestly recommend that you use a third-party component for logging.

I think, if anything, your (perfectly valid) argument has helped me figure out where I want to steer Slim and how to further simplify its concerns.

Does this make sense?

aktau:
That makes sense, in that way I can steer my mind away from having to force Slim to do my bidding (the forcing part is something I escaped by converting a legacy app from Symfony) and just pretend the Slim logger does not exist. As it stands then, Slim's logger is only useful when these conditons are met:
- you don't use any external logging component, such as Monolog
- you don't want any external component to log stuff, or (alternatively) are prepared to pass in your own logging class whilst ignoring Slim's. Or you use only self-made components that can use Slim's logger directly.
- you dont' want or can't use composer

Maybe the people that want to do this are actually more numerous than I expected and I didn't foresee that. At any rate, I think I've just though of a nice way of solving my problem, thanks for that and for listening to my feedback!

designermonkey:
Personally, I would have thought that interoperability with other components, at least from the perspective of easily using something like Monolog in place of the default, would be a preferable situation to be in. I wouldn't put any questions like "Slim's log should be useable by other frameworks" in place here. It seems more a question of making it easy to let developers add other components to Slim and not vice versa.

Addvilz:
It is extremely easy to wrap anything using logger system currently present in Slim. I, for example, use Apache Log4PHP, it took me literally 20 lines of code and 5 minutes of work to do so, in a reusable way. 

Current implementation is as flexible as it would need to be. Implementing PSR-3 would not help either way, it would merely limit possible interaction between non-PSR loggers and custom implementations out there. Slim's main strength is ability to customize and modify in a way to adapt "standard" and non-"standard" implementations to interact with Slim. It should stay that way. Otherwise we will end up with another PSR-ed framework out there, and there is already plenty. 

Also, I would prefer that Slim would not follow any PSR blindly, nor should it implement vendor functionality to do so, even in form of dependency, Monolog or other in this case, as it would only limit possible interaction between multiple implementations. 

Dependencies here mean a) use composer b) waste hours of maintenance if you do not wish to do so.

aktau:
@Addvilz I'm not so sure it would limit the flexibility, as it is just implementing an interface which doesn't even preclude it from implementing another one if so desired (though it would be messy). I also did not advocate following standards blindly, but I have noted that this specific PSR standard is gaining adoption and I don't see a reason why it shouldn't. I'm not the biggest fan of the PSR anyway, I personally don't like the coding style, for example, and happily ignore it. 

My current solution is to just not use the Slim logger, but directly use the Monolog logger I instantiate at the top of my file. This shaves off a tiny bit of overhead and allows me to delete ~15 lines of configuration, which is great to me.

That said, I would like to know how you have solved the problem in a small amount of time. Can you clarify? Thanks

gmanricks:
What are you trying to plug Slim's logger into ?

aktau:
Anything that needs logging and is called from Slim routes

PR #564, opened 2013-07-16
Missing __isset and __unset

Addvilz:
Fixes critical issue with missing __isset and __unset magic methods. 

Because IoC container is now used, all implementations before 2.3 which explicitly relied upon setting properties directly to Slim object instance are now broken, because isset() returns false although value is present.

Needless to say I just debugged some 12k lines of code to find out that Slim's __isset is missing.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -234,6 +234,10 @@ class Slim
     public function __isset($name){
     	return isset($this->container[$name]);
     }
+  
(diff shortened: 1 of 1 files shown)


codeguy:
Thanks. I've cherry-picked these into the release-2.3.1 branch. Will be merged into master this week.

https://github.com/codeguy/Slim/commits/release-2.3.1

Issue #568, opened 2013-07-18
‘Expires’ field isn't parsed by `strtotime` on serialization of Secure Cookies


shello:
With Slim 2.3.0, when `cookies.encrypt` is `true` (i.e., Secure Cookies are active), and `setCookie` (from `\Slim\Slim`) is called with `$time` (or `cookies.lifetime`) being a _string_, the serialization of the cookie (`\Slim\Http\Util::serializeCookies`) sets the `$expires` argument of `\Slim\Http\Util::encodeSecureCookie` as the string without parsing it with `strtotime`.

This causes the secure cookie value to have the expires string before the first `|` separator, instead of having a timestamp.

When this cookie value reaches `\Slim\Http\Util::decodeSecureCookie` it's not accepted when the (supposed) timestamp is checked against `time()`.

I wrote a test for this, but it depends on `\Slim\Http\Cookies` and `\Slim\Http\Headers`: https://gist.github.com/shello/6026460

codeguy:
Did you want to send a PR, or would you prefer I do this? Just want to make sure you have a chance to receive credit for the fix. Let me know. Thanks!

PR #570, opened 2013-07-18
Fix issue #568: Parse string 'expires' before encoding secure cookie.

shello:
The `\Slim\Http\Util::encodeSecureCookie` function requires the
expires parameter to be an int. A string typed 'expires' or
'cookie.lifetime' would pass a string in this case, which causes the
response cookie to be broken when parsed by `decodeSecureCookie`.


shello:
Sorry, I made the PR too soon, and didn't select the correct branch.

PR #571, opened 2013-07-18
Issue 568 fix serialization secure cookie

shello:
The `\Slim\Http\Util::encodeSecureCookie` function requires the
expires parameter to be an int. A string typed 'expires' or
'cookie.lifetime' would pass a string in this case, which causes the
response cookie to be broken when parsed by `decodeSecureCookie`.

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -186,9 +186,15 @@ class Util
     {
         if ($config['cookies.encrypt']) {
             foreach ($cookies as $name => $settings) {
+                if (is_string($settings['expires'])) {
+                    $expires = strtotime($settings['expires']);
+                } else {
+                    $expires = (int) $settings['expires'];
+                }
+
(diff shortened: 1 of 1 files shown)


codeguy:
Thanks! I've merged this commit into the release-2.3.1 branch. See:

https://github.com/codeguy/Slim/tree/release-2.3.1

If you have time, can you submit a unit test for this change? If not I can get to that this evening after work. Appreciate your help on this!

shello:
@codeguy: Is the unit test function that I attached to issue #568 good enough? (see: https://gist.github.com/shello/6026460)
I'm not sure it is because it couples `\Slim\Http\Cookies` and `\Slim\Http\Headers` to `\Slim\Http\Util`.

codeguy:
I think it's fine. I'm less concerned about it being "good" and more concerned about it accurately testing this particular scenario.

PR #572, opened 2013-07-18
Unit test for #571: Serialize and Decrypt cookie with string expires

shello:


codeguy:
Great. Merged into https://github.com/codeguy/Slim/tree/release-2.3.1.

PR #585, opened 2013-07-21
Don't send any content when it's a HEAD request.

solarissmoke:
Some servers (e.g., Apache) silently discard any content in response to a HEAD request, but others don't...

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1274,8 +1274,10 @@ class Slim
             }
         }
 
-        //Send body
-        echo $body;
+        //Send body, but only if it isn't a HEAD request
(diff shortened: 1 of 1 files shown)


codeguy:
Can you move your changes into the `\Slim\Http\Response::finalize` method?

solarissmoke:
@codeguy I don't see how I can access the relevant properties of the `Request` (i.e., `isHead()`) from inside the `Response` class?

codeguy:
Sorry. You're right. I was reading through this too fast. I'll get this pulled in today!

codeguy:
Merged into release-2.3.3.

PR #588, opened 2013-07-26
Provides read-only access to Headers and Cookies objects instead of public access

tivie:
Instead of making Headers and Cookies objects publicly accessible, uses __get magic method to provide read only access to them.

This prevents accidental replace of the said objects, while keeping the desired functionality and backwards compatibility.


mattzuba:
0efd38f specifically changed these from protected to public - I'd say the desired functionality is to access them directly.

codeguy:
I'm not terribly concerned about this. I'd favor simpler code instead of trying to prevent public access.

PR #594, opened 2013-08-07
Fix `urlFor` with regex

shello:
Before `urlFor` would return '/path/Hello\.Josh' for a pattern
'/path/(:foo\.:bar)' with parameters
`('foo' => 'Hello', 'bar' => 'Josh')`, when the expected return value
is '/path/Hello.Josh' (this test was introduced in commit cc722cc).

(Pull Request description updated (2013/09/26) because GitHub ate my \)

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -192,8 +192,8 @@ class Router
         }
         $pattern = preg_replace($search, $params, $this->getNamedRoute($name)->getPattern());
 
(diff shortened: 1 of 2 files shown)


shello:
Pull Request description updated because GitHub ate my \ twice, making the description senseless.

codeguy:
Picked into release-2.3.2. Will be merged into master very soon.

PR #596, opened 2013-08-08
Match path items on whole words.

marionebl:
Hi there,

I recently ran into some issues with routing, where the callbacks for a route were executed that had a substring of the requested path in it.

An example:

``` php
$app->get("/me/", function() {
     echo("Nice to meet you.");
});

$app->get("/message/", function() {
     echo("No message for you this time.");
});
```

**GET /message/** would execute the callback for the "me" route in this scenario.

The proposed CR tries to fix this by checking if items in the route definition are candidates for matching on whole words and modifying the given parts of the matching regex.

Cheers,
marionebl

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -349,20 +349,13 @@ class Route
     {
         //Convert URL params into regex patterns, construct a regex for this route, init params
         $patternAsRegex = preg_replace_callback(
-            '#:([\w]+)\+?#',
+            '#/?:?([\w]+)\+?/?#',
             array($this, 'matchesCallback'),
             str_replace(')', ')?', (string) $this->pattern)
         );
         if (substr($this->pattern, -1) === '/') {
             $patternAsRegex .= '?';
         }
-
-        // Match path fragments as whole words
-        $patternAsRegex = preg_replace(
-            '#/([\w]+)/#',
-            '/\b${1}\b/',
-            $patternAsRegex
-        );
 
         //Cache URL params' names and values if this route matches the current HTTP request
         if (!preg_match('#^' . $patternAsRegex . '$#', $resourceUri, $paramValues)) {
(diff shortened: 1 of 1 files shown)


designermonkey:
To me this boils down to specificity. As `me` is shorter than `message`, it should come after `message` in the list.

marionebl:
@designermonkey 
Sorry, I am not sure what you want to say.

Do you want to say the mismatch is acceptable and should be handled by every single dev rearranging the route definitions?
Or do you think the problem should be handled by getting the `\Slim\Slim::Router::routes` array ordered by their string length?

designermonkey:
> Do you want to say the mismatch is acceptable and should be handled by every single dev rearranging the route definitions?

When one defines routes that use regex patterns, specificity comes in to play. It does in every app I've used that has routes. Longer possible matches need to come before shorter ones. It's just the way it works.

I'm not saying your fix isn't valid, it surely is, to catch when it's not done right, but it's one of the things I learned, that's all.

designermonkey:
Here's the first reference I could find as to what I meant.

http://laravel.io/topic/30/routes-first-in-first-out

marionebl:
Ok, got it. Thank you for explaining!

tuupola:
In Sinatra routes are matched in the order they are defined. The first route that matches the request will be executed. I find it confusing if execution order is defined by length of the route. Much less confusing to just set the order in source code.

marionebl:
Had some thought on this and for me it is about wether the Slim::Slim::$request_method is to be thought of taking a regular expression as first parameter or a route string with placeholders matching the arguments in the given callback function. 

If it is the common view that the passed string should be seen as regex, I'll close this pull request. Implementing devs can solve this by passing `/^me$/`and `/^message$/` instead of `/me/`and `/message/`.

In this case the routing docs should get some hint about this, though.

tuupola:
I expect **GET /me/** to match /me/ route and **GET /message/** to match /message/. Any other behavior is just plain wrong. I do not think passed string is a regex unless I specifically make it a regex. For for example:

```
$app->get("/^me*/", function() { ... }); /* Regex */
$app->get("/me/", function() { ... }); /* String */
```

designermonkey:
What I mean when I talk about regex, is not whether you as a developer use a regex pattern when creating your route, it's how the router matches the string you've added, as it uses regex matching in it's processing of a route.

https://github.com/codeguy/Slim/blob/master/Slim/Route.php#L352

marionebl:
BAD me did the first pull request without setting up the unit test env. 
The refactored version of this does the same concerning the the issue I want to solve. 
As a bonus it does not produce any new errors according to my local unit tests.

codeguy:
@marionebl Can you move your regex changes into the `\Slim\Route::matchesCallback` method? I'd like to keep regex preparation isolated in that method. Thanks!

PR #600, opened 2013-08-21
When adding to a set, all other values in that array are overwritten, th...

ersmtp:
...is will give us the ability to add to an existing set, or update existing k/v pairs that are already in a set


codeguy:
I'm not sure I see the problem you are trying to solve. Can you post some code that demonstrates the problem?

Issue #472, opened 2013-01-03
PrettyExceptions Middleware and Ajax don't mix well


paulyg:
A common pattern in my ajax heavy apps is on an error to show the text/html returned by the server in an error notice on the client side. My error handler function, registered via `\Slim\Slim::error()`contains logic like this.

``` php
if ($app->request()->isXhr()) {
    // send a string or JSON response
} else {
    // send a full HTML response
}
```

However when my app throws an exception it is caught and handled by the `\Slim\Middleware\PrettyExceptions` class and does not go through the above logic. `\Slim\Middleware\PrettyExceptions` renders out a full HTML page with some styles embedded in the head. When I inject that response into my error notice jQuery and the browser handle the fact that the response is a full HMTL document gracefully. But because of the styles my CSS goes all funny until I dismiss the notice. There are several ways in which this could be improved, so I wanted to get feedback on which one you favored before I go preparing a pull request.
1. Make `\Slim\Middleware\PrettyExceptions` use the `$app->error()` handler internally.
2. Namespace all of the CSS classes in the error response so they don't trample all over the default CSS. Also this way I could use jQuery to yank out the contents of the error from the full HTML page by doing `$(xhr.reponseText).find('body.slimerror')` or similar.
3. Put a check for `$app->request()->isXhr()` inside like my example above.
4. Allow a user defined middleware to be substituted for `\Slim\Middleware\PrettyExceptions` inside of `\Slim\Slim::call()`.

anho:
Definitely +1 for this proposal.

honzabrecka:
+1
I'm for solution 1. or 4.

dinacel:
+1 too, solution 1 seems to be the best.

smagch:
At the moment, Slim doesn't allow Middlewares to stop application. Let's say there is a `MyMiddleware` that can possibly throw a exception. And you want to stop app with a custom status code and a body.

``` php
class MyMiddleware extends \Slim\Middleware
{
    public function call()
    {
        try {
            // something could possibly go wrong
        } catch (\Exception $e) {
            $app = \Slim\Slim::getInstance();
            // I know 10000 is invalid status code. It's just a example.
            $app->halt(10000, 'Hello. Nice to meet you');
            return;
        }
        $this->next->call();
    }
}

$app = new \Slim\Slim();
$app->add(new MyMiddleware());
$app->run();
```

[...]

PR #602, opened 2013-08-30
Allow replacing PrettyExceptions middleware with your own exception handler.

paulyg:
This commit is an attempt to fix issue #472.
- Introduces a new config option,  'exception.handler'. The value should be
  a FQCN. If set it will replace the Slim\Middleware\PrettyExceptions
  middleware that is added to the stack by Slim\Slim in run().
- Adds a new method to Slim\Slim, remove() that allows you to remove
  middleware by passing a class name. As a safeguard you can not remove
  Slim\Middleware or Slim\Slim.
- Makes generateTemplateMarkup public so PrettyExceptions can call it,
  DRYing up the code a bit.
- Tests are included.

### DIFF
--- a/Slim/Middleware/PrettyExceptions.php
+++ b/Slim/Middleware/PrettyExceptions.php
@@ -32,6 +32,8 @@
  */
 namespace Slim\Middleware;
 
+use Slim\Slim;
+
 /**
  * Pretty Exceptions
  *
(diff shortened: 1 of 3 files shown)


codeguy:
This has become moot. The develop branch has removed the PrettyExceptions middleware (rolling its functionality into Slim's default error handler). With version 3.0 (in development), you should override the default error handler and, considering the current application mode, choose to show your own debug info or not.

paulyg:
Sounds like a good plan.

PR #605, opened 2013-08-31
Able to pass arg to group route callback

garyyeap:
will let us have ability to pass args to callback of group route : 

```
$app->group('/:foo/bar', function($foo) use($app)
{
    echo $foo;
    $app->get('/whatever', ....
});
```

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -230,12 +230,14 @@ class Slim
     {
         $this->container[$name] = $value;
     }
-    
-    public function __isset($name){
+
+    public function __isset($name)
+    {
     	return isset($this->container[$name]);
     }
-  
-    public function __unset($name){
+
+    public function __unset($name)
+    {
     	unset($this->container[$name]);
     }
 
@@ -533,8 +535,15 @@ class Slim
         $pattern = array_shift($args);
         $callable = array_pop($args);
         $this->router->pushGroup($pattern, $args);
+
+        // Get params
+        $route = new \Slim\Route($pattern, $callable);
+        $resourceUri = explode('/', $this->request->getResourceUri());
+        $patternLength = count(explode('/', $pattern));
+        $groupResourceUri = array_slice($resourceUri, 0, $patternLength);
+        $route->matches(implode('/', $groupResourceUri));
         if (is_callable($callable)) {
-            call_user_func($callable);
+            call_user_func_array($callable, $route->getParams());
         }
         $this->router->popGroup();
     }
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -539,8 +539,8 @@ class Slim
         // Get params
         $route = new \Slim\Route($pattern, $callable);
         $resourceUri = explode('/', $this->request->getResourceUri());
-        $patternLength = count(explode('/', $pattern));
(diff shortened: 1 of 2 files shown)


endel:
Hey, the feature wasn't released yet? I'm running version 2.3.5 but this isn't working. Thanks!

melbichon:
Please, can you add this feature in next release ?
Seems branch merging did not include commits.

piotr-cz:
I'd really appreciate this feature.
It's a must when using language prefix as a segment in routes.

codeguy:
Assigned this to @gmanricks. He'll review and pull in when appropriate.

garyyeap:
Hi @piotr-cz , if you want to use with language prefix, here is an idea of middleware implementation that  no longer need to include language prefix param in any routes.

```
class I18N extends \Slim\Middleware
{
    public function call()
    {
        $app = $this->app;
        $paths = explode('/', $app->request()->getResourceUri());
        $locale = $paths[1];
        $currentUrl = '/'.join(array_slice($paths, 2), '/');
        $localeFile = $app->config('localePath') .'/'. $locale .'.yml';

        if(!file_exists($localeFile)) {

            if($locale != '') {

                try{
                    $app->notFound();
                } catch(Exception $e) {}

                return;
            }

            $preLocale = $app->getCookie('locale');
            $locale = $preLocale ? $preLocale : $app->config('defaultLocale');

            try{
                $app->redirect('/'.$locale);
            } catch(Exception $e) {}

            return;
        }

        $app->environment->offsetSet('PATH_INFO', $currentUrl);
        $app->setCookie('locale', $locale, mktime().time() + 60*60*24*100);
        $this->next->call();
    }
}
```

piotr-cz:
Thanks @garyyeap, but besides pages with language prefix, I've create an API that's language agnostic. That's why I'd prefer to use groups, it's the cleanest solution.

piotr-cz:
I wonder if this could go into 2.x release since here's the code and test are passing

lalop:
I think this will not happen for Slim2 since there is no new features added to slim2

piotr-cz:
Depends on the point of view. For me it's a bugfix ;)

codeguy:
Re-assigning to @akrabat for review.

akrabat:
Is the work against the master or against develop branch?

piotr-cz:
Under the pull request descriptions it says `slimphp:develop`

akrabat:
Yes, but the diff is against a class that's only in `master` …

piotr-cz:
I'd say that the context of develop branch has changed since this PR. It was more like minor version develop (2.x) and now it's like major future version develop (3.x)

garyyeap:
Is this getting merged? if does, I will send another PR for 3.x on this weekend

akrabat:
@codeguy do you want this for Slim 2 or 3?

codeguy:
Slim 2 is done. If we're going to pull this in, it'll be for 3 only.

piotr-cz:
If Slim 3 will use [nikic/fast-route](https://github.com/nikic/FastRoute), similar PR won't be necessary as it's already built in there.

However Slim 2 (ATM only stable release) IMHO deserves having this fixed. But I won't bug anymore

PR #606, opened 2013-09-01
Fixes flash messaging/SessionCookie decoding failure when cookies.encrypt = true

jeremykendall:
Without a way to check if a cookie is already encoded, `\Slim\Http\Util::serializeCookies()` will encode an already encoded cookie. This is an issue when `\Slim\Middleware\SessionCookie` sets an encoded cookie value on the response object during `\Slim\Middleware\SessionCookie::saveSession()` (as `\Slim\Http\Util::serializeCookies()` will encode every cookie it gets if `cookies.encrypt` = `true`).  Since the cookie is only decoded once during `\Slim\Middleware\SessionCookie::loadSession()`, the cookie's actual decoded value is never available to the application, and `unserialize()` in `\Slim\Middleware\SessionCookie::loadSession()` will fail.

This PR adds the ability to tell `\Slim\Http\Util::serializeCookies()` to skip encoding a cookie during the serialization process.

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -44,6 +44,14 @@ namespace Slim\Http;
 class Util
 {
     /**
+     * An array of md5 hashed cookie values that will be excluded from
+     * encryption during the serialization process.
+     *
+     * @var array Exclude from encryption in Util::serializeCookies()
+     */
+    protected static $encryptionExclusions = array();
+
+    /**
      * Strip slashes from string or array
      *
      * This method strips slashes from its input. By default, this method will only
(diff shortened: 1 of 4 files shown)


codeguy:
I'm wondering if a simpler solution is to simply not encrypt cookie values in the `\Slim\Middleware\SessionCookie` class and let Slim handle cookie encryption as it does already. What are your thoughts?

jeremykendall:
@codeguy :+1: 

That would also take care of the weirdness of having two sets of cookie settings, one global (except for `SessionCookie`) and one just for `SessionCookie`.

codeguy:
Cool. Feel free to update this PR or close and submit a new one and I'll get it pulled into 2.3.2 to be released soon.

jeremykendall:
Sounds good.  Thanks!

jeremykendall:
@codeguy I'm going to create a new PR for simplicities sake, but it looks like the develop branch `SessionCookie` class no longer exists.  Where should I branch from and submit a PR against?

codeguy:
Submit a PR to the 2.3.2 release branch. This class will be removed from the next major release and replaced by the new Session class. But this change is still very important for apps using the 2.\* codebase.

jeremykendall:
Sounds good.  Thanks for the update.

jeremykendall:
Superseded by #635.

PR #609, opened 2013-09-03
Update Router.php

BradCrumb:
Fixed bug where urlForl remove all characters after an empty optional segment.

Ie.
//Old behavior
testroute: /partA/partB(/:partC)/partD(.html)
$url = $app->urlFor('testroute'); //Would become /partA/partB

//New behavior
testroute: /partA/partB(/:partC)/partD(.html)
$url = $app->urlFor('testroute'); //Would become /partA/partB/pardD.html

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -193,7 +193,7 @@ class Router
         $pattern = preg_replace($search, $params, $this->getNamedRoute($name)->getPattern());
 
         //Remove remnants of unpopulated, trailing optional pattern segments
-        return preg_replace('#\(/?:.+\)|\(|\)#', '', $pattern);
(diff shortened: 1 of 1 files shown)


codeguy:
Can you include a unit test for this?

silentworks:
I don't think this is a bug in the behaviour. If I had a scenario where I had another route declared with only 3 params this would break.

``` php
$app->get('/partA/partB(/:partC)/partD(.html)', function () {});
$app->get('/partA/partB(/:partC)', function () {});
```

My second route could be rendered useless because of this PR. I will be closing this off based on that note and on the note of no Unit Test provided. You can re-open with some tests if you still want this merged.

PR #611, opened 2013-09-04
Fix HTTP method override middleware

gpoole:
It appears that Slim used to strip HTTP_ off $_SERVER variables and the MethodOverride middlware expects that to happen, but it looks like Slim isn't doing that anymore. I've updated MethodOverride to use the correct environment variable and added some tests to confirm this.


codeguy:
Thank you! I've cherry-picked this into https://github.com/codeguy/Slim/tree/release-2.3.2

PR #613, opened 2013-09-11
Support for query parameters in routing

amcsi:
I would like to contribute this bit of code which would let you route $_GET query parameters.

Such a route would look like this: 'urlPart/:anotherUrlPart?obligatoryParam=1&controller=:controller&:action=:action', which would match /urlPart/value?controller=cont&action=act&obligatoryParam=1&unneededGetParam=2. As you can see, the order doesn't matter as long as all the needed parameters are there, and values that do not begin with a colon must equal the value given in the route. Also, any unnecessary parameters are ignored and do not cause the match to fail.

I've also included a few test, although it may be possible that more would be necessary. I also haven't written any documentation.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -65,7 +65,7 @@ should contain this code:
 
 #### Nginx
 
-Your nginx configuration file should contain this code (along with other settings you may need) in your `location` block:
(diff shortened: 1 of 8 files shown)


amcsi:
Please look at #614
For this pull request I accidentally compared with the master branch.

PR #614, opened 2013-09-11
Support for query parameters in routing

amcsi:
I would like to contribute this bit of code which would let you route $_GET query parameters.

Such a route would look like this: 'urlPart/:anotherUrlPart?obligatoryParam=1&controller=:controller&:action=:action', which would match /urlPart/value?controller=cont&action=act&obligatoryParam=1&unneededGetParam=2. As you can see, the order doesn't matter as long as all the needed parameters are there, and values that do not begin with a colon must equal the value given in the route. Also, any unnecessary parameters are ignored and do not cause the match to fail.

I've also included a few test, although it may be possible that more would be necessary. I also haven't written any documentation.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -46,6 +46,11 @@ class Route
     protected $pattern;
 
     /**
+     * @var array The query pattern (e.g. "/books/?id=:id"), but in array form
+     */
+    protected $query;
+
+    /**
      * @var mixed The route callable
      */
     protected $callable;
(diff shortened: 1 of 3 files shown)


designermonkey:
But your query parameters will already be available in the Request object, no?

amcsi:
@designermonkey Yes they would. I did want to use this colon system for 2 reasons though:
1. I need to differentiate between mandatory parameter and mandatory value. If the value doesn't have a colon, it means the query param has to match that specific key and value. If the value is a colon, that just means it has to be set, but can be any value.
2. This will allow you allow for clean urls and query urls at the same time if you use the same callback for both types of routes and only check the params for any dynamic values. Or it allows for easy porting of your current project from using clean urls to query parameter urls if it turns out your live hosting site doesn't allow you to use .htaccess files :D

designermonkey:
Good reasons! :D

codeguy:
I'm not really in favor of this. Mandatory query params are the responsibility of the API and not the framework.

amcsi:
But if you're on an environment where you cannot use pretty URLs, you'd basically be having Slim with the inability to use routing at all.
But it's your framework. I have no voice in this.

codeguy:
If you do not have URL rewriting, you can use Slim with `/index.php/foo/bar?abc=123` instead.

amcsi:
Wow, I never realized you could do that with rewriting off! OK, in that case this merge request really isn't needed.

codeguy:
We all learn something new everyday :)

PR #629, opened 2013-09-18
Add default return value for request data

chrisgedrim:
Allow overloading of the default return value of NULL when
requesting request data.

Example:
  http://host/action?foo=bar&abc=123

  $app->request->get('foo')        -- [bar]
  $app->request->get('bar')        -- [NULL]
  $app->request->get('bar', false) -- [FALSE]


codeguy:
Merged into https://github.com/codeguy/Slim/commits/release-2.4.0

Issue #607, opened 2013-09-02
possible config enhancement


ghost:
It would be easier to use the Slim.config method if it would parse the given key. For example:

```
$app->config(array(
    'model' => array(
        'session' => array(
            'expirationSeconds' => 60*30
        )
    )
));

$app->config('model.session', array(
    'directory' => __DIR__.'../session';
));

$app->config('model.session.mime', 'application/json');

$sessionConfig = $app->config('model.session'); // array(expirationSeconds => 1800, directory => __DIR__.'../session', mime => 'application/json')
$expirationSeconds = $app->config('model.session.expirationSeconds'); // 1800
$sessionDirectory = $app->config('model.session.directory'); // __DIR__.'../session'
$sessionFileFormat = $app->config('model.session.mime'); // 'application/json'
```

designermonkey:
I would also like to see something similar to this. It would make finding values a lot more flexible.

ghost:
Yepp, but implementation is not simple I think.

codeguy:
I'd be on board with this, but I'm worried about the necessary complexity. If this were to happen, I'd prefer that Configuration be isolated into a separate `\Slim\Configuration` class. There are other things I'm focusing on now, but anyone else is more than welcome to take this on.

designermonkey:
Not too difficult IMO, as Laravel has already tackled this with the namespaced config approach.

If you wanted to outline what your ideal approach to a separate `Configuration` class were, I could look at this. I have already got a partial working extended framework from Slim that has a `Configuration` class.

codeguy:
I'd move everything to do with configuration currently in the `\Slim\Slim` class (variables, methods) into a new `\Slim\Configuration` class and then instantiate the `\Slim\Configuration` class with the new DI container as `$app->config`.

designermonkey:
OK, no probs. I will leave the methods on the Slim class for deprecation though, and just reference them out of the DI container.

codeguy:
Sounds good.

designermonkey:
Nearly done here :)

designermonkey:
The way I've implemented the Configuration is to have a parent class with default methods, so it can be re-used elsewhere if required, and the Configuration class extends this.

I was going to put the parent in the Helper namespace, but this is now gone :(

As it's merely a helper class, where should it go now?

ghost:
Slim root, or make a new folder for collections :D

codeguy:
Toss it in the top-level folder with the other classes for now.

[...]

PR #631, opened 2013-09-18
Configuration

designermonkey:
Addresses request in #607 to supply nested array handling for configuration, accessible via a separated string key.

Removes older `config` method from `App` class instead of deprecates, as this will be applied in a major point version. (`3.0.0`)


designermonkey:
I am in the process of learning how to use testing, so sorry for the lack of actual tests. I have tested this manually, and can't find any issues, but like most people, I am known to miss things.

codeguy:
Thanks! Will have a look later this week.

designermonkey:
Just realised that `$this->settings` inside the container should really be `$this->config`. I'll edit tonight.

designermonkey:
Closed in favour of https://github.com/codeguy/Slim/pull/691 which is much simpler.

PR #639, opened 2013-09-22
Controllers

designermonkey:
Controllers allow you to specify a controller class for routes. Based on @fortrabbit's work in SlimController, but a much simpler implementation to begin with.

Routes can be specified as

```
$app->get('/books/:id', function ($id) {
    //Show book identified by $id
});
```

With Controllers, they can be specified as

```
$app->get('/books/:id', 'Books::booksList');
```

and require a class as such

```
class Books extends \Slim\Controller
{
    public function booksList($id)
    {
        //Show book identified by $id
    }
}
```

The \Slim\Controller class currently has a single task of accepting the main Slim App to be accessible in a controller as `$this->app`.

Controllers can have a default namespace specified in the settings with a key of `controller.namespace`, which will prepend the class specified in every controller string.


anho:
Personally I really like the approach Silex provides with one of its service providers.

```
$app->get('/books/:id', "books.controller::booksList");
```

`books.controller` would be a key in the IoC container, so you have much more flexibility when constructing your objects.
Additionally I don't like the necessity of the \Slim\Controller class, restricting controllers in such a way is bad design.

designermonkey:
Every framework I've encountered has a base controller to extend from.

designermonkey:
Sorry, short comment there, had to rush off to look after the kids bedtime...

Why would you consider it bad design? Do you have examples?

The reason for the base controller is to ensure that things like the core app instance are loaded without having to declare that in every new controller.

I do like the idea of having a reference to an IoC instance included, but that would also mean defining every controller in the container first before the app runs proper, which is a lot of overkill IMO.

I will have a look at the Silex implementation of this to see exactly what you reference.

anho:
Symfony does not need a base controller ;)

I think that the need of the app container inside the controller does not justify the necessity of a base controller. That takes a lot of flexibility.

Personally I would use some sort of factory ... probably. The default one would implement your strategy. In addition another factory could be implemented that does it the way Silex is doing it. But I'm not sure if this is too much for a micro framework.

anho:
Hopefully to make my idea more clear: I do think these "controllers" should be seen more as reusable objects and that does not fit very well with the current implementation.

Besides providing an abstract class you might use an interface/trait or doing some introspection of the constructor. I really believe there is a more elegant, flexible way to implement such a feature.

erizocosmico:
Is the access to $app really needed in a controller? If not, as anho stated it should be optional to extend from the base controller.

When you call the method from the specified controller you can check if the controller is an instance of \Slim\Controller and if it is you set the $app. If it isn't you don't do anything.

```
// 'Books::bookList'
// $controller is 'Books'
// $method is 'bookList'

$controllerInstance = new $controller();
if ($controllerInstance instanceof \Slim\Controller)
    $controllerInstance->setApp($app);

call_user_func_array(array($controllerInstance, $method), array($args));
```

That way extending from \Slim\Controller is optional.

anho:
This is what I meant when talking about the usage of an interface/trait (e.g. \Slim\ApplicationAware).

designermonkey:
Interfaces are being discussed and worked on.

> it should be optional to extend from the base controller.

It is, you don't have to use it as it is.

lnpbk:
Controllers or not, what's the issue with passing a fully qualified Namespace::method as the second parameter? This gives users the option of handling their routes in closures, or grouping them into a Class, which would still adhere to the Single Responsibility Principle.

`$app->get('/books/:id', '\Me\MyApp\Books::method');`

tuupola:
IMO controllers do not belong to Slim core. They should be provided as an extension.

designermonkey:
Right then.

As technically this isn't actually really about controllers, I've removed any reference to them in the code. This is really about being able to pass a route trough a class and method as well as any other callable.

There at present is no way to use classes and methods, only functions, and this fixes that.

ffflabs:
:+1: to @tuupola , I really like Slim's current approach over Silex's, but I see no problem if there's an extension to use a custom controller as router, same as there are extensions to use Monolog or Twig for their respective use cases.

Issue #615, opened 2013-09-13
Slim 404s when encounters encoded url (%20 etc.)


AshleyWright:
Whenever there is a non-alphanumeric (or ._-) character in the `$_SERVER['REQUEST_URI']` variable `$env['PATH_INFO']` does not get the expected value on apache.

Fix: replace all instances of `$_SERVER['REQUEST_URI']` with `urldecode($_SERVER['REQUEST_URI'])` in `Environment.php`

StJudeWasHere:
This seems to work fine on both branches now, master and develop. Could you confirm it? maybe we can close this issue already.

silentworks:
Over a year and no reply from original author so closing this one out.

PR #642, opened 2013-09-27
Added urldecode to $_SERVER['REQUEST_URI'] on Environment.php

thiagomfranca:
To solve the issue #615 added url decode to $_SERVER['REQUEST_URI'] on Environment.php


codeguy:
Just so I am completely clear about the problem and how this solves it, can you post a sample `curl` command that demonstrates an HTTP request that causes the problem to occur?

silentworks:
No explanation was given as to what this solves and code is now outdated. Closing off this one.

Issue #640, opened 2013-09-23
issue whith flash middleware


lalop:
Trying to hide the flash messages box if there's no flash message, 
I tested the number of flash messages in my `Slim\Middleware\Flash` with a simple `count` but I get `1` when I've no message.

codeguy:
Can you post a very simple app that demonstrates this?

lalop:
Here's a simple app.php :

``` php
$app = new \Slim\Slim(array(
    'templates.path' => './'
));
$app->get('/',function() use($app){

        return $app->render('view.php');

});

$app->run();
```

and the view.php

``` php
<?php
var_dump(count($flash));
```

that display `int 1`

tuupola:
It is because `$flash` is an instance of `Slim\Middleware\Flash` which contains the flash messages. Slim documentation says:

> Flash messages are persisted with sessions; sessions are required for flash messages to work. Flash messages are stored in $_SESSION['slim.flash'].

So you could do something like `var_dump(count($_SESSION['slim.flash']));`

lalop:
Or maybe Slim\Middleware\Flash can implement Countable ?

lalop:
I can do a pull pull request with that if we are agree on the solution ( and on the issue of course )

tuupola:
+1 on implementing Countable if it is ok with Josh.

codeguy:
Implementing countable is fine. Send PR to the release-2.3.3 branch. Be sure you create your feature branch from the current master branch, NOT the develop branch... a lot has changed in develop that already resolves this.

lalop:
hum... ok.
When do you think merging the develop branch into master ?

codeguy:
Not for a few weeks at least.

PR #643, opened 2013-09-27
make flash middleware implements countable

lalop:
see #640


codeguy:
Pulled into release-2.3.3 branch.

PR #648, opened 2013-10-12
Interfaces

designermonkey:
Lots of work for Interfaces as discussed with the working group.

There are many changes here, there need to be more I think, as the unit tests have errors.

I would also recommend that this be rebased and squashed before inclusion into the release branches as there are a lot of commits, and probably will be more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ language: php
 php:
   - 5.3
   - 5.4
+  - 5.5
 
 script: phpunit --coverage-text
--- a/.vagrant.v1.1380248914
+++ b/.vagrant.v1.1380248914
@@ -1,0 +1,1 @@
+{"active":{"default":"d4f35030-4dba-4dbb-9110-ad346b2bbeb7"}}
--- a/.vagrant.v1.1380248914
+++ b/.vagrant.v1.1380248914
@@ -1,1 +1,0 @@
-{"active":{"default":"d4f35030-4dba-4dbb-9110-ad346b2bbeb7"}}
--- a/README.markdown
+++ b/README.markdown
@@ -80,6 +80,10 @@ Be sure you change the `ServerRoot` setting to point to your Slim app's document
         SourceRoot = /path/to/public/directory
     }
 
+    ServerVariables {
+        SCRIPT_NAME = /index.php
+    }
+
     VirtualHost {
         * {
             Pattern = .*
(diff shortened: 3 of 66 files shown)


codeguy:
Looks like there are some merge conflicts. Can you resolve those first? There shouldn't be many... probably just inline documentation changes. Let me know if it's more significant.

codeguy:
I'm not sure I'm a fan of adding the custom classes as an additional argument to the App constructor. I guess it's more efficient than asking the end developer to inject the custom implementations with `$app->foo = $app->share(function ($c) {});`, but it feels messy to me. Thoughts?

codeguy:
Actually, I think asking the end developer to inject with `$app->foo = $app->share(function ($c) {})` is reasonable and a more flexible solution, because who is to say that the overridden component does not require additional configuration (like the default Request and default Session components) in its own closure?

designermonkey:
To be honest, I would also prefer not to apply it that way, the problem for me is the entire configuration would run by constructor, and then have to be customised afterward. It seems like a little too much overhead for me.

Need to think about this then...

codeguy:
Can you expand on why it would be a problem for you? I'm not clear on your previous comment. I'm open to suggestions. Nothing really happens in the constructor as is beyond setting the default closures to be executed when/if the associated components are requested. The only overhead really is having to define the overrides after you instantiate the app.

designermonkey:
I just get a bit over zealous on efficiency, and it would mean that the app would be configured fully, then parts replaced afterward.

Anyhoo, I've reverted some of the work to remove that aspect, and fixed the merge conflicts. There's still lots to do for the tests.

codeguy:
Cool sounds good. Will merge tonight... have to head out to a high school band competition soon.

codeguy:
I can understand the desire for efficiency, but I'd rather favor configuration over convention. It's a write-once-and-done thing, so I'm not too big on mucking up the constructor for the sake of saving some keystrokes :) Anyway, heading out. More tonight.

designermonkey:
@codeguy I've merged the current `develop` branch into this one on my repo, hence all the extra commits as `feature-interfaces` is behind.

designermonkey:
A few issues have arisen rebasing develop into this, so I'm re-applying the interfaces manually to a fresh branch.

codeguy:
Cool sounds good. I've let that fall a bit behind. Ok if I remove it?

designermonkey:
Yeah, can do, I'll PR to develop

PR #650, opened 2013-10-16
remove fragements (#) from the uri

stephenmittag:


codeguy:
If it is not relevant to the server, can you give me an example for when your change would be necessary?

stephenmittag:
It is not relevant because the browser has to jump to the anchor but Slim will receive the complete url including the hash-character for processing.

Assume you click on a link like this http://en.wikipedia.org/wiki/Francis_Drake#Award_of_knighthood the server has to deliver the page http://en.wikipedia.org/wiki/Francis_Drake and the browser would scroll to the anchor #Award_of_knighthood.

But currently this does not work because the regex in the router uses hashes as begin/end marker.

tuupola:
AFAIK hash is not even sent to server so PHP cannot access it anyway. So there is no real need to include code for removing it.

stephenmittag:
Sry but that is not right. Try it.

tuupola:
I do not know a browser which shows string `foo` when doing `print_r($_SERVER)` at address below. Hashes are not sent in the http request. They are used in the browser only.

http://www.appelsiini.net/demo/hash.php#foo

If you have a browser which sends the hash / fragment in the http request let me know which one.

stephenmittag:
I've seen the requests in the access log and I'm quite sure that I'm not hallucinating. When im back at the office on Monday I will examine the logs again and see what the user agent was.

codeguy:
It's also my understanding that fragments are not passed to the server... they are handled only by the HTTP client. You can ready this for more info:

http://www.w3.org/TR/webarch/#media-type-fragid

tuupola:
I can see fragments in referrer logs but not in the actual requests.

```
85.253.45.82 - - [26/Oct/2013:00:09:03 +0300] "GET /xxx/css/skin.css HTTP/
1.1" 200 2387 "http://www.example.com/xxx#content" "Mozilla/4.0 (compatible; M
SIE 8.0; Windows NT 5.1; Trident/4.0; .NET CLR 2.0.50727; .NET CLR 3.0.4506.2152
; .NET CLR 3.5.30729; InfoPath.1; .NET CLR 1.1.4322; OfficeLiveConnector.1.3; Of
ficeLivePatch.0.0)"
```

However those are actually against [specs](http://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-07.txt).

```
14.37 Referer
... 
If the field value is a partial URI, it SHOULD be interpreted relative
to the Request-URI. The URI MUST NOT include a fragment.
```

stephenmittag:
Well, today I tried this: 
telnet localhost 80
GET /m.php#foo HTTP/1.1
Host: localhost

Array
(
    [SCRIPT_FILENAME] => /opt/www/m.php
    [REQUEST_METHOD] => GET
    [SCRIPT_NAME] => /m.php
    [REQUEST_URI] => /m.php#foo <----
    [DOCUMENT_URI] => /m.php
    [DOCUMENT_ROOT] => /opt/www
    [SERVER_PROTOCOL] => HTTP/1.1
    [GATEWAY_INTERFACE] => CGI/1.1
    [SERVER_SOFTWARE] => nginx/1.4.3
    [REMOTE_ADDR] => 127.0.0.1
    [REMOTE_PORT] => 52162
    [SERVER_ADDR] => 127.0.0.1
    [SERVER_PORT] => 80
    [SERVER_NAME] => localhost
    [REDIRECT_STATUS] => 200
    [HTTP_HOST] => localhost
    [PHP_SELF] => /m.php
)

The http stack doesn't remove the fragment if the client sends it. We would have to rely on the clients correnct behavior.

It is up to you to decide whether that's a good idea.

codeguy:
@n3m0 When you telnet in, it looks like you are manually setting the raw request URL... whereas an HTTP client will strip that before it sends the actual request. For example, try this instead:

```
curl -X GET http://localhost/m.php#foo
```

Since curl is an actual HTTP client (as opposed to telnet issuing raw HTTP commands), it will properly strip the fragment.

stephenmittag:
@codeguy Thats the point. The client will strip it if its implemented properly. If its not I am the one with the exceptions.

Anyway ... I forked the project and we are using the patched version of it. For me the problem is solved.

PR #651, opened 2013-10-18
Added Missing Http Statuses From IANA Standard

dangerousdan:
Added missing Http Statuses from IANA (The Internet Assigned Numbers Authority).
http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml


codeguy:
Please resubmit to the develop branch and I'll pull it in.

PR #652, opened 2013-10-18
Define contoller callback as \class::method

renatko:
I propose a way to specified controllers

```
$slim->get('/', '\Renatko\MyAmazingController::action');
```

```
namespace Renatko;
class MyAmazingController
{
    public function action()
    {
    }
}
```

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -92,8 +92,8 @@ class Route
 
     /**
      * Constructor
-     * @param string $pattern  The URL pattern (e.g. "/books/:id")
-     * @param mixed  $callable Anything that returns TRUE for is_callable()
+     * @param string $pattern The URL pattern (e.g. "/books/:id")
+     * @param mixed $callable Anything that returns TRUE for is_callable()
      */
     public function __construct($pattern, $callable)
     {
@@ -154,8 +154,8 @@ class Route
      */
     public function setCallable($callable)
     {
-        $matches = [];
-        if(is_string($callable) && preg_match('!^([^\:]+)\:\:(.+)$!', $callable, $matches)) {
+        $matches = array();
+        if (is_string($callable) && preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
             $callable = array(new $matches[1], $matches[2]);
         }
 
@@ -199,7 +199,7 @@ class Route
      */
     public function setName($name)
     {
-        $this->name = (string) $name;
+        $this->name = (string)$name;
     }
 
     /**
@@ -222,7 +222,7 @@ class Route
 
     /**
      * Get route parameter value
-     * @param  string                    $index     Name of URL parameter
+     * @param  string $index Name of URL parameter
      * @return string
      * @throws \InvalidArgumentException If route parameter does not exist at index
      */
@@ -237,8 +237,8 @@ class Route
 
     /**
      * Set route parameter value
-     * @param  string                    $index     Name of URL parameter
-     * @param  mixed                     $value     The new parameter value
+     * @param  string $index Name of URL parameter
+     * @param  mixed $value The new parameter value
      * @throws \InvalidArgumentException If route parameter does not exist at index
      */
     public function setParam($index, $value)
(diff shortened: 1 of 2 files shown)


Slim/Route.php
```diff
@@ -154,6 +154,11 @@ public function getCallable()
      */
     public function setCallable($callable)
     {
+        $matches = [];
+        if(is_string($callable) && preg_match('!^([^\:]+)\:\:(.+)$!', $callable, $matches)) {
+            $callable = [new $matches[1], $matches[2]];
```
erizocosmico: The shortened array syntax was introduced in php 5.4 (http://docs.php.net/manual/en/language.types.array.php). Since Slim is compatible with php 5.3 or higher I would replace that line with array(new $matches[1], $matches[2]);

renatko: Fixed

designermonkey:
<del>There is already a PR for controller like functionality.</del>

I'm closing it out in favour of this one.

codeguy:
Do you all prefer to use the `::` separator even though that notation is used elsewhere in PHP? Should we instead use `@` to separate controller name from controller method name?

renatko:
Maybe then `->`?

```
$slim->get('/', '\Renatko\MyAmazingController->action');
```

codeguy:
I'd prefer to keep it limited to a one-character separator, or the same character typed twice. Some options are:
- `Foo::bar` (potentially confused with static method call)
- `Foo:bar` (my favorite)
- `Foo@bar` (like Laravel)
- `Foo>bar` (not as visible on quick glance as other options)

lplume:
:: as well -> could be confused (although it will be exacted what will be executed), in the end it is about taste. One character like : or @ is visible enough, two characters repeated are better than two different (thinking of typing).

I second for : or @

designermonkey:
When I see classes and methods documented, I see it as `\Namespace\Classname::methodName`, to me this how one would write it as a string, so seems most fitting in this case.

lrlopez:
Hmmm... does it have any sense to call a static method as a controller? If that's the case, we could use `:` for non-static methods (an instance should be created first before the call) and `::` for static ones (no need to create a instance).

codeguy:
@lrlopez I'm not sure whether it makes sense or not... that's really for the end developer to decide. But I'd prefer not to rule out the possibility. I think using `:` for non-static methods and `::` for static methods is a good idea. Is anyone adamantly opposed?

gmanricks:
I think that's a really elegant solution +1

codeguy:
Ok let's move forward with using a single `:` for instance methods and `::` for static methods. @renatbilalov Can you go ahead and make this update, and also change `[]` on line 157 to `array()`? Will merge when these two changes are complete.

erizocosmico:
I think using `:` for non-static methods and `::`for the static ones is a great idea too. Maybe a little confusing for the developer to distinguish between the two of them at quick glance since it's the same symbol.

Anyway, like @gmanricks said, that's a really elegant solution. +1 to that.

designermonkey:
This isn't about static vs non static methods, if you all look at the code clearly, it always creates an instance of the object before instantiation anyway, so there are no static methods in use here.

As a prime example of what I was talking about, here: http://us2.php.net/method_exists#65405 is an example of how many people write comments or docs pointing to any method on a class, and that's why in my PR and I suspect this one, there is the use of `::`.

I'm not trying to be difficult, I just want to make sure the decision is made for the correct reason, on the right information.

codeguy:
@designermonkey I understand that the code will always use object instances. What I'm saying is that the syntax it uses duplicates existing syntax used for static method calls. If we implement this, it will effectively override the ability to use static method calls for Slim route callables, and it may cause confusion. I'm not even sure it's good to use that syntax in documentation examples :) I personally find it confusing, purely because of PHP's `::` static method call syntax.

renatko:
I changed the separator for a method to `:`.
`::` - it is native separator for static method in php. Maybe it's someone already uses, so we must leave it for compatibility.

```
is_callable('\Slim\Slim::getInstance'); // true
```

codeguy:
Thanks! Merged into https://github.com/codeguy/Slim/commits/release-2.4.0

PR #654, opened 2013-10-19
Removes trailing slash from $_SERVER['DOCUMENT_ROOT'].

### LINKED ISSUE #653: Path parsing returns bad script name if Apache DocumentRoot includes trailing slash
`\Slim\Environment::__construct()` generates a bad `$env['SCRIPT_NAME']` if Apache `DocumentRoot` directive  includes a trailing slash.  Per the Apache [`DocumentRoot` documentation](http://httpd.apache.org/docs/2.2/mod/core.html#documentroot), the directive should not include a trailing slash.

jeremykendall:
Fixes #653.

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -152,8 +152,12 @@ class Environment extends \Slim\Collection
             // Request method
             $settings['REQUEST_METHOD'] = $_SERVER['REQUEST_METHOD'];
 
-	        // Root URI (physical path) and resource URI (virtual path)
-            $scriptName = str_replace($_SERVER['DOCUMENT_ROOT'], '', $_SERVER['SCRIPT_FILENAME']); // <-- "/physical/index.php"
+            // Ensure no trailing slash in $_SERVER['DOCUMENT_ROOT']
+            // Per Apache docs, "DocumentRoot should be specified without a trailing slash"
+            // http://httpd.apache.org/docs/2.2/mod/core.html#documentroot
+            $documentRoot = rtrim($_SERVER['DOCUMENT_ROOT'], '/');
+            // Root URI (physical path) and resource URI (virtual path)
+            $scriptName = str_replace($documentRoot, '', $_SERVER['SCRIPT_FILENAME']); // <-- "/physical/index.php"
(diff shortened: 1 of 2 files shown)


jeremykendall:
@petarov: I noticed [your comment](https://github.com/codeguy/Slim/commit/a35767746009ff818e6375b72f23a2cf7de5f1a4#commitcomment-4371264) on the `\Slim\Environment` HVVM commit.  Would you be willing to apply this fix to your local Slim install and see if it fixes your sandbox problem?  I'm interested to know if this PR fixes both my issue and yours.

petarov:
@jeremykendall: I'll check it out asap  and I'll write back. thx!

petarov:
I applied the #653 fix but sadly it did not solve my problem. I still get  `/sandbox/resource` instead of just `/resource`  in  `PATH_INFO` 

Like I wrote in the comment, I'm using an Apache Alias which points to a completely different server directory path than what comes in `DOCUMENT_ROOT`. I think the problem lays somewhere there but I've got little time to thoroughly trace it right now.

jeremykendall:
@petarov I'm sorry to hear it didn't help, although it was kind of a shot in the dark.  If you're willing and able to share a small application that will recreate the issue, I'd be interested in looking at it.  Maybe we can put our heads together and get it sorted out.

tuupola:
Just FYI. Slim routing is also broken when using [mod_vhost_alias](http://httpd.apache.org/docs/2.2/mod/mod_vhost_alias.html) I have a workaround for it but it breaks when upgrading from 2.3.1 to 2.3.3. Will do a proper bug report soon(ish).

4g0st1n0:
I have the same problem of @petarov. I'm using an Apache alias to point to a directory outside of virtual host document root.
After the upgrade from 2.3.1 t o 2.3.3 my routes does not works.
My alias is /backend/services. 
With request uri "/backend/services/search" before upgrade PATH_INFO = "/search", after upgrade PATH_INFO = "/backend/services/search"

jeremykendall:
@4g0st1n0, @petarov, and @tuupola: Let's move the Apache alias discussion over to #655.  That's a separate issue than #653, which this PR fixes.

codeguy:
Closing. This is fixed in the current release.

PR #659, opened 2013-10-23
Removed redundant code from sendFile

lrlopez:
Removed some dead code.

`$contentType` is checked for trueness on a parent `if`. The `else` block will be run only if it evaluates to `false`, so it's useless to check again for it inside.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1045,26 +1045,25 @@ class App extends \Slim\Pimple
      *
      * @param  string $file        The URI of the file, can be local or remote
      * @param  string $contentType Optional content type of the stream, if not specified Slim will attempt to get this
+     * @throws \RuntimeException   If provided $file cannot be opened
      * @api
      */
     public function sendFile($file, $contentType = false) {
         $fp = fopen($file, "r");
+        if ($fp === false) {
+            throw new \RuntimeException("Cannot open the stream at {$file}");
+        }
         $this->response->stream($fp);
         if ($contentType) {
             $this->response->headers->set("Content-Type", $contentType);
         } else {
             if (file_exists($file)) {
-                //Set Content-Type
-                if ($contentType) {
-                    $this->response->headers->set("Content-Type", $contentType);
+                if (extension_loaded('fileinfo')) {
+                    $finfo = new \finfo(FILEINFO_MIME_TYPE);
+                    $type = $finfo->file($file);
+                    $this->response->headers->set("Content-Type", $type);
                 } else {
-                    if (extension_loaded('fileinfo')) {
-                        $finfo = new \finfo(FILEINFO_MIME_TYPE);
-                        $type = $finfo->file($file);
-                        $this->response->headers->set("Content-Type", $type);
-                    } else {
-                        $this->response->headers->set("Content-Type", "application/octet-stream");
-                    }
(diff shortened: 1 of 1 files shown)


erizocosmico:
It doesn't have anything to do with the redundant code but shouldn't be checked at the top of the function that the remote or local resource exists and throw an exception or something if it doesn't?

I mean something like:

``` php
/**
     * Send a File
     *
     * This method streams a local or remote file to the client
     *
     * @param  string $file        The URI of the file, can be local or remote
     * @param  string $contentType Optional content type of the stream, if not specified Slim will attempt to get this
     * @api
     */
    public function sendFile($file, $contentType = false) {
        // We don't want that warning if the file does not exist
        @$fp = fopen($file, "r");
        if ($fp === false)
            throw new \RuntimeException("Requested file at '{$file}' does not exist.");
        $this->response->stream($fp);
        if ($contentType) {
            $this->response->headers->set("Content-Type", $contentType);
        } else {
            if (file_exists($file)) {
                //Set Content-Type
                if ($contentType) {
                    $this->response->headers->set("Content-Type", $contentType);
                } else {
                    if (extension_loaded('fileinfo')) {
                        $finfo = new \finfo(FILEINFO_MIME_TYPE);
                        $type = $finfo->file($file);
                        $this->response->headers->set("Content-Type", $type);
                    } else {
                        $this->response->headers->set("Content-Type", "application/octet-stream");
                    }
                }

                //Set Content-Length
                $stat = fstat($fp);
                $this->response->headers->set("Content-Length", $stat['size']);
            } else {
                //Set Content-Type and Content-Length
                $data = stream_get_meta_data($fp);

                foreach ($data['wrapper_data'] as $header) {
                    list($k, $v) = explode(": ", $header, 2);

                    if ($k === "Content-Type") {
                        if ($contentType) {
                            $this->response->headers->set("Content-Type", $contentType);
                        } else {
                            $this->response->headers->set("Content-Type", $v);
                        }
                    } else if ($k === "Content-Length") {
                        $this->response->headers->set("Content-Length", $v);
                    }
                }
            }
        }
        $this->finalize();
    }
```

It's pointless to stream `false` to the browser.

lrlopez:
I'm not sure, but I think the `file_exists()` call is to check if the `$file` param is an actual file or a remote URI...

erizocosmico:
Yes, but even if it is a remote URI `$fp` value will be false if the remote URI doesn't exist and currently this is not checked anywhere in the code.

lrlopez:
@mvader,
José Miguel, you're completely right. I've amended the commit to introduce the check. ¡Gracias!

codeguy:
Can you rebase your local branch against the latest develop branch and update your PR? As is I can't automatically merge.

lrlopez:
Sure!

lrlopez:
Josh, the PR is rather outdated. I think the only thing we could savage from it is throwing an exception if the file/stream cannot be opened. Do you want me to create a new PR with the change?

codeguy:
Sorry for the egregious delay :D Sorting PRs now. Yes, there's not much to salvage here. I actually killed this method in the App class, and I intend to migrate it into the Response class and named something like `withDownload($path, $type, $disposition = "attachment")`. Do you want to work on that?

PR #660, opened 2013-10-25
Refer to $app as $this inside a router

IlanFrumer:
### Instead of repeating the keyword use everywhere like that....

``` php
$app = new \Slim\Slim();

$app->get('/',function ()  use ($app) {

});

$app->group('/api',function ()  use ($app) {

    $app->get('/',function ()  use ($app) {

    });

});
```
### We can just use $this:

``` php
$app = new \Slim\Slim();

$app->get('/',function () {

});

$app->group('/api',function ()  {

    $this->get('/',function ()  {
          $this->halt(404);
    });

});
```

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -426,7 +426,7 @@ class Slim
     {
         $pattern = array_shift($args);
         $callable = array_pop($args);
-        $callable = $callable->bindTo($this);
+        if(PHP_MINOR_VERSION >= 4) $callable = $callable->bindTo($this);
         $route = new \Slim\Route($pattern, $callable);
         $this->router->map($route);
         if (count($args) > 0) {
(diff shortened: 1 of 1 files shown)


tuupola:
:+1:

erizocosmico:
It would be great to be able to do that but according to travis the method bindTo does not exist https://travis-ci.org/codeguy/Slim/jobs/13025874 (it was added in php 5.4). The tests also fail for 5.4.

IlanFrumer:
bindTo would only work if you have php > 5.4

tuupola:
I think Slim follows [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#5-control-structures) so you should include braces around control structures.

petarov:
This is a nice feature! But I would pull against the `develop` branch because tests obviously need be adapted too. There are also a lot of changes in `develop` and maybe Slim system reqs would also change to >= 5.4 in next major release

IlanFrumer:
[a pull against the develop branch](https://github.com/codeguy/Slim/pull/661)

Thanks a lot

PR #661, opened 2013-10-25
refer to $app as $this inside a router

IlanFrumer:
### Instead of repeating the keyword use everywhere like that....

``` php
$app = new \Slim\Slim();

$app->get('/',function ()  use ($app) {

});

$app->group('/api',function ()  use ($app) {

    $app->get('/',function ()  use ($app) {

    });

});
```
### We can just use $this:

``` php
$app = new \Slim\Slim();

$app->get('/',function () {

});

$app->group('/api',function ()  {

    $this->get('/',function ()  {
          $this->halt(404);
    });

});
```

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -394,6 +394,9 @@ class App extends \Slim\Pimple
     {
         $pattern = array_shift($args);
         $callable = array_pop($args);
+        if (PHP_MINOR_VERSION >= 4 && is_callable($callable)) {
+            $callable = $callable->bindTo($this);
+        }
         $route = new \Slim\Route($pattern, $callable);
         $this->router->map($route);
         if (count($args) > 0) {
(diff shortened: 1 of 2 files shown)


IlanFrumer:
this line is written in the test suite 7 times

``` php
$this->assertSame($callable, $route->getCallable());
```

The problem is that [bindTo](http://php.net/manual/en/closure.bindto.php) returns a new created Closure object....

codeguy:
Not ready to go 5.4 only just yet.

IlanFrumer:
What i ended up doing is extending the slim class like so:

``` php
namespace Slim;

class SlimX extends Slim
{

    protected function mapRoute($args)
    {
        $pattern = array_shift($args);
        $callable = array_pop($args);
        if (is_callable($callable)) {
            $callable = $callable->bindTo($this);
        }
        $route = new \Slim\Route($pattern, $callable);
        $this->router->map($route);
        if (count($args) > 0) {
            $route->setMiddleware($args);
        }

        return $route;
    }
    public function group()
    {
        $args = func_get_args();
        $pattern = array_shift($args);
        $callable = array_pop($args);
        $this->router->pushGroup($pattern, $args);
        if (is_callable($callable)) {
            $callable = $callable->bindTo($this);
            call_user_func($callable);
        }
        $this->router->popGroup();
    }
}
```

Than using the extended class

``` php
$app = new \Slim\SlimX();
```

Thanks a lot.

PR #665, opened 2013-10-28
Update Autoloader.php - to work properly with include_path

entenbraten:
If you want to include Slim via include_path, better use "stream_resolve_include_path" here. Be aware that the function might cache the results (see http://us1.php.net/manual/en/function.stream-resolve-include-path.php#111790).


codeguy:
Can you resubmit to the develop branch? Will get this into the 3.0 release since it will require us to bump the minimum supported PHP version.

Issue #663, opened 2013-10-27
2.3.4 version adds escape for slashes


DavidePastore:
I'm working with Twig. In the base page I have something like this:

```
{%
 set links = {
  'Home': urlFor('HomePage'),
  'Link1': urlFor('Link1'),
  'Link2': urlFor('Link2'),
  'Link3': urlFor('Link3')
 }
%}
{% for key, link in links %}
 <li>{{ link }}</li>
{% endfor %}
```

This code works fine in 2.3.3, but in 2.3.4 it adds escape backslashes.
HomePage is defined in this way:

```
$app->get('/', function () use($app){
  $app->render('index.twig');
})->name("HomePage");
```

The other routes are similar to this.

What I get:

```
/\/
/\/link1
/\/link2
/\/link3
```

erizocosmico:
I tried to reproduce your error but I don't get your results.

``` php
$app->get('/', function () use($app){
  $app->render('index.html');
})->name("HomePage");

$app->get('/hi/testing', function () use($app){
  $app->render('index.html');
})->name("Link1");
```

``` html
{%
 set links = {
  'Home': urlFor('HomePage'),
  'Link1': urlFor('Link1'),
  'Link2': urlFor('HomePage'),
  'Link3': urlFor('HomePage')
 }
%}
{% for key, link in links %}
 <li>{{ link }}</li>
{% endfor %}
```

And I get:

```
/
/hi/testing
/
/
```

No escape backslash is added for me.

Versions:

```
installed:
  slim/slim  2.3.4   Slim Framework, a PHP micro framework
  slim/views 0.1.0   Twig View Parser package for the Slim Framework
  twig/twig  v1.14.1 Twig, the flexible, fast, and secure template language f...
```

DavidePastore:
I tried with your versions but the problem is always here. It's so strange.

codeguy:
I'll touch base with @silentworks and see if the Slim-Views repo needs any tweaks for the 2.3.4 update.

silentworks:
Hi @DavidePastore can you state your OS and PHP version please?

DavidePastore:
Hi @silentworks.

OS: Microsoft Windows XP Home Edition Service Pack 3.
PHP: 5.3.8 on xampp.

tuupola:
Hardcoded "/" and DIRECTORY_SEPARATOR clashing somewhere?

DavidePastore:
I have not touched any of this.

[...]

PR #667, opened 2013-10-30
Fix urlFor adding extra slashes on Windows OS #663

silentworks:


codeguy:
Does this work when the physical directory path has more than one "\" in it?

codeguy:
Picked into 2.3.5.

tuupola:
Did you notice that some tests were failing?

silentworks:
Re-testing now, I didn't realize that travisci is only running on develop branch.

tuupola:
Actually the failed test is not caused by your commit. Instead it was some slow running test. Second changed before the test could finish.

```
1) SlimTest::testExpiresAsString
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Tue, 05 Nov 2013 00:37:01 GMT'
+'Tue, 05 Nov 2013 00:37:00 GMT'
```

silentworks:
I seem to be getting the ViewTest failing and not the one you mentioned. Are you testing master branch?

tuupola:
I am not testing myself. I saw it from the [details](https://travis-ci.org/codeguy/Slim/builds/13293406) link in the Travis CI build failed notification in this pull request.

silentworks:
ah yes what you said would make sense.

codeguy:
Yes, there's one test that will intermittently fail since it's coded poorly. It expects two time measurements to be the same, but they may differ by a second or so every once in a while. I need to fix that.

PR #669, opened 2013-10-31
Add (string) case in encodeSecureCookie call to hash_hmac

elgenie:
Per http://php.net/manual/en/function.hash-hmac.php, the second argument to hash_hmac is supposed to be a string. HHVM (http://www.hhvm.com) is more strict about checking the arguments.

(note: this functionality is no longer present on develop)

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -224,7 +224,7 @@ class Util
      */
     public static function encodeSecureCookie($value, $expires, $secret, $algorithm, $mode)
     {
-        $key = hash_hmac('sha1', $expires, $secret);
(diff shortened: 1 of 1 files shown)


elgenie:
This fixes SlimHttpUtilTest::testEncodeAndDecodeSecureCookie\* when run with HHVM.

JoelMarcey:
Any update on if we can get this pull request merged? Thanks.

@cc ptarjan

codeguy:
Will get it pulled into the forthcoming 2.4.1 release this weekend.

codeguy:
Merged into release-2.4.1 branch.

JoelMarcey:
@codeguy Thank you very much for this merge!

cc: ptarjan

PR #671, opened 2013-11-06
Add property typehints for proper IDE support.

dongilbert:
With the introduction of the Resource Locator in 2.3.0, we are able to directly access the `response`, `request`, `environment` and `router` (and others?) properties on a `Slim\Slim` app object. This is really great for chaining, as it removes the unnecessary `()`.

The downside of this, however, is that in most IDE's you lose the code completion that come with specifying the return type. You can get around this by using the old style syntax `->request()`, `->response()`, but then we're right back where we're started.

This PR proposes to add support specifically for IDE's to restore code completion abilities.

`@property` is supported by Netbeans, PHPStorm, PDT, ZendStudio, Eclipse and possibly others. (I think anything based on IntelliJ)

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -40,9 +40,14 @@ if (!extension_loaded('mcrypt')) {
 
 /**
  * Slim
- * @package Slim
- * @author  Josh Lockhart
- * @since   1.0.0
+ * @package  Slim
+ * @author   Josh Lockhart
+ * @since    1.0.0
+ *
+ * @property $environment \Slim\Environment
+ * @property $response    \Slim\Http\Response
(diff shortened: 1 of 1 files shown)


codeguy:
This looks great. Will pull in soon.

dongilbert:
So, there is a potential downside, and that is if someone has overwritten the request with their own implementation, `@property` wouldn't necessarily match anymore.

dongilbert:
Sorry - had to fix the ordering. It goes 

`@property \Slim\Environment $environment`
not
`@property $environment \Slim\Environment`

Oops

codeguy:
Merged into https://github.com/codeguy/Slim/commits/release-2.3.6

PR #674, opened 2013-11-11
Fixed virtual path with public directory

ghost:
PATH_INFO was being set to "/" breaking routes.

/
    public/
        .htaccess
        index.php
    .htaccess
- /.htaccess -
  RewriteEngine on
  RewriteCond %{REQUEST_URI} !public/
  RewriteRule (.*) /public/$1 [L]
- /public/.htaccess -
  RewriteEngine On
  RewriteRule ^ index.php [QSA,L]

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -140,7 +140,10 @@ class Environment implements \ArrayAccess, \IteratorAggregate
             $env['SCRIPT_NAME'] = rtrim($physicalPath, '/'); // <-- Remove trailing slashes
 
             // Virtual path
-            $env['PATH_INFO'] = substr_replace($requestUri, '', 0, strlen($physicalPath)); // <-- Remove physical path
+            $env['PATH_INFO'] = $requestUri;
+            if (substr($requestUri, 0, strlen($physicalPath)) == $physicalPath) {
+                $env['PATH_INFO'] = substr($requestUri, strlen($physicalPath)); // <-- Remove physical path
+            }
             $env['PATH_INFO'] = str_replace('?' . $queryString, '', $env['PATH_INFO']); // <-- Remove query string
             $env['PATH_INFO'] = '/' . ltrim($env['PATH_INFO'], '/'); // <-- Ensure leading slash
 


codeguy:
Can you clarify 1) what the issue is, and 2) how this solves it? I'm not saying you are wrong, I'm just not entirely clear on what your description above says.

ghost:
My Slim app is running with the file index.php in a subdirectory called public.
I have two .htaccess files, one in the root folder and one inside the public folder, as described in the pull request.
My issue was that with this configuration, all request are served by the default "/" route, and if the url is longer than the public folder name (6 chars) then app returns an 404 error.
Debugging the file Environment.php i found that url passed to it was already without the "public" prefix, but app try to remove it.
My fix remove the subdirectory prefix only if is present.

codeguy:
Ok. All Slim application routes should assume they are relative to the physical directory in which the Slim `index.php` file exists. So if I create my Slim app in `/foo/index.php`, my application routes should not begin with `/foo`. Can you confirm this issue exists in version 2.3.5?

ghost:
I have tried with 2.3.5 on apache 2.4 and the issue still remains.

ghost:
I have updated to Slim 2.4.0 and Apache 2.4.6, same configuration ad problem remains.
After adding this fix all works.

shashankmehta:
@codeguy Hi. I am facing a similar issue. My directory structure is as follows:

```
|-- application
|   -- test.php
|   -- foo
|       -- index.php
```

`application` folder consists of the app that is served at `example.com`. Apache serves the parent folder of application. That is, there is an htaccess file that converts `example.com/test` to `application/test.php`.

Now I am trying to serve an app at `example.com/foo` using Slim.

Taking an example for a URI request as `example.com/foo/hello`, I expect Slim to match this request to `/hello`.
On [L143](https://github.com/codeguy/Slim/blob/master/Slim/Environment.php#L143) of `Environment.php`, in my case the values of variables are as follows:

```
$requestUri = '/foo/hello';
$physicalPath = '/application/foo';
```

However, Slim expects `$physicalPath` to be `/foo` which is not in this case. So the value of `$env['PATH_INFO']` becomes an empty string to which `/` is appended in L145.

Hence, Slim sees the path as `/` and not `/hello`

I tested the patch in this PR and it is _not_ the right fix. If such a case cannot be supported by Slim, the workaround that I'll be using is that right in the start, run this:

```
$_SERVER['REQUEST_URI'] = '/application' . $_SERVER['REQUEST_URI'];
```

If it is somehow possible to figure out the root folder, `application` in my case, through variables in `$_SERVER`, that would be the best possible solution. Otherwise, I think this should be allowed via an option when instantiating Slim.

I hope I have been clear enough. Let me know if you need any further clarification. Thanks!

ctsstc:
Glad to see this fixed up :) 
This should beat my hacky fix, and fix the problem I was having as well

PR #675, opened 2013-11-18
Pass through 'templates.path' to view object singleton.

ziadoz:
Fixes `Slim\View` object not getting the correct 'templates.path' setting.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -742,7 +742,6 @@ class Slim
         if (!is_null($status)) {
             $this->response->status($status);
         }
-        $this->view->setTemplatesDirectory($this->config('templates.path'));
         $this->view->appendData($data);
         $this->view->display($template);
     }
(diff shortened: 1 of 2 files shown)


ziadoz:
Should I remove the `$app->view->setTemplatesDirectory()` line in the `$app->render()` method? It seems like it's redundant now, as either the DI container or `$app->view()` will set the correct templates path on any custom view objects.

codeguy:
@ziadoz Yes, you can remove the redundant call in the `render` method.

ziadoz:
Okay, I've done that, along with a horrible rebase that appears to have added more commits…

codeguy:
haha. That's fine I can cherry pick the relevant ones. Will do that today.

codeguy:
Merged into https://github.com/codeguy/Slim/commits/release-2.3.6

PR #679, opened 2013-11-20
Don't rely on system setting for error_reporting level

ptarjan:
This test shouldn't fail if the system's php.ini setting differs on the error level (This was failing in HHVM's test environment)

### DIFF
--- a/tests/Middleware/SessionCookieTest.php
+++ b/tests/Middleware/SessionCookieTest.php
@@ -135,6 +135,7 @@ class SessionCookieTest extends PHPUnit_Framework_TestCase
 
         $logWriter->expects($this->once())
             ->method('write');
+        $oldLevel = error_reporting(E_ALL);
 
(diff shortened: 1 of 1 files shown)


codeguy:
Is this issue limited to a particular unit test? And if so, should this solution be implemented in the Slim source code rather than in a unit test? I guess I'm just looking for more details about this issue.

ptarjan:
This is a problem with Slim's unit test in `tests/Middleware/SessionCookieTest.php`, so indeed the fix should be in this repository. It is assuming that I haven't touched my error reporting level in my system's `php.ini` or else it fails.

I ran into this issue when testing slim on HHVM since we have different default error levels than PHP (we emit notices by default).

codeguy:
Thanks for the clarification. Will get this pulled in soon.

codeguy:
Merged into https://github.com/codeguy/Slim/commits/release-2.4.0

PR #688, opened 2013-12-03
Array access

designermonkey:
As discussed, converted accessors to ArrayAccess.

I also moved some final properties into the container for consistency. Only one left is the hooks property.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -57,23 +57,6 @@ class App extends \Slim\Pimple
     const VERSION = '2.3.5';
 
     /**
-     * @var array
-     */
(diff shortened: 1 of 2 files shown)


designermonkey:
Did a load of fixes for `AppTest` as implementing `ArrayAccess` throughout `App` broke many many tests.

codeguy:
With these changes, do we need to retain these magic methods?

https://github.com/codeguy/Slim/blob/develop/Slim/App.php#L280

designermonkey:
I wouldn't have thought so, but I didn't want to start deleting things willy-nilly.

PR #691, opened 2013-12-05
Configuration

designermonkey:
First pass at Configuration handler.

Provides app configuration values stored as nested arrays, which can be accessed and stored using dot separated keys.

Works with already present `$app->config($name, $value);` method, but allows `ArrayAccess` to `$app->settings['dot.separated.key'];`

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -104,7 +104,9 @@ class App extends \Slim\Pimple
     public function __construct(array $userSettings = array())
     {
         // Settings
-        $this->settings = array_merge(static::getDefaultSettings(), $userSettings);
+        $this->settings = $this->share(function ($c) use ($userSettings) {
+            return new \Slim\Configuration($userSettings);
+        });
 
         // Environment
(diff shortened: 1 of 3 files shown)


designermonkey:
Balls. No unit test. Must sleep, will do it tomorrow.

codeguy:
Cool. Get some unit tests added and make sure Travis CI is happy then I'll get this merged in.

designermonkey:
I have begun work on a 'part 2' for this, to extend the functionality out to have a `ConfigurationHandler` loaded into the `Configuration` class.

This will allow scenarios like loading config values from a file for example.

This can be pulled in for now, as 'part 2' is built on top of this, and it will follow a very similar setup.

designermonkey:
There's one test that fails: `testDerivedClassCanOverrideStaticFunction`, but I'm not sure if it's even needed anymore?

codeguy:
Cool feel free to ignore and I'll have a look on my end.

PR #695, opened 2013-12-07
Fixed Route::setCallable not matching some valid controller names

itsgoingd:
Regexp used for matching valid controller names doesn't match "_" and some other valid characters in the method name. Replaced with a regexp specified in the PHP manual (http://www.php.net/manual/en/functions.user-defined.php).

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -155,7 +155,7 @@ class Route
     public function setCallable($callable)
     {
         $matches = array();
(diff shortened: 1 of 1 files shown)


codeguy:
Can you include a unit test for this?

poison:
Also posted by someone else: https://github.com/codeguy/Slim/pull/718 I think there's a need for this! +1

itsgoingd:
Ah, I completely forgot about this. Closing in favor of #718 which has the same fix + unit test.

PR #704, opened 2013-12-16
Configuration Handler

designermonkey:
This PR adds support for separate `ConfigurationHandler` classes to be used to manage the settings for Slim.

There is a default `ConfigurationHandler` which takes the previous work of nested keys and allows the handler to manage the keys and values.

Any `ConfigurationHandler`s will need to provide all settings as they currently are (which is `dot.nested.keys` as has always been with Slim), and also provide a `setArray()` method to allow values to be set from a PHP array. How the handler gets the values and does it's magic is completely up to the developer. Fancy using XML files to store configuration? Go ahead.

I have an interface prepared for each class here for when the interfaces branch is ready.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -89,7 +89,6 @@ class App extends \Pimple
         // Settings
         $this['settings'] = $this->share(function ($c) use ($userSettings) {
             $config = new \Slim\Configuration(new \Slim\ConfigurationHandler);
-            $config->setDefaults();
             $config->setArray($userSettings);
 
             return $config;
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -88,7 +88,11 @@ class App extends \Pimple
     {
         // Settings
         $this['settings'] = $this->share(function ($c) use ($userSettings) {
-            return new \Slim\Configuration($userSettings);
+            $config = new \Slim\Configuration(new \Slim\ConfigurationHandler);
+            $config->setDefaults();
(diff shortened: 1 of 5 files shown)


designermonkey:
As an example, I am building an app that uses XML, so I will create an `XMLConfigurationHandler` in my app's namespace. As I will also want nested keys in my app, I will just extend the `ConfigurationHandler` class, and add my XML support. Then all I will do is swap out the classes, like so.

``` php
$app = new \Slim\App();

$app['settings'] = $app->share(function ($app) {
    $config = new \Slim\Configuration(new \MyApp\XMLConfigurationHandler);

    $config->callHandlerMethod('xmlFileLoadMethod', '/some/file/path/to/config.xml'); // Loads from a file

    return $config;
});
```

codeguy:
There are merge conflicts against the latest develop branch. Can you take a look at those?

designermonkey:
Certainly can.

It's 1 hour before 'Santa' arrives here, so I will get back on this after the kids are in bed tomorrow night.

Happy Christmas!

codeguy:
Merry Christmas! I've a few hours left to prepare the milk and cookies :)

designermonkey:
Didn't see any conflicts there, I merged directly with the remote develop branch and everything went in OK.

codeguy:
Hmm ok. Weird that happened locally.

PR #705, opened 2013-12-16
Feature interfaces

designermonkey:
This is incomplete as I add the PR, but I'm losing track of where I am with it.

@csdco has expressed an interest in this, and a fresh pair of eyes would help.

Currently, there are interfaces for _most_ of the framework. While doing this (twice), I've started to question the need for some interfaces, the Collection class doesn't really need one, as it's a core component IMO. Many classes using it have interfaces, and if these were replaced, technically the developer could just use their own collection class.

You may also notice that I have blank interfaces for Cookies and Headers. I want to have a play with how this is currently implemented, and when done, there will need to be interfaces.

I think I have missed some parts, more than likely around the Request and Response area, but I can't tell now.

I need some others to look over this and let me know thoughts and changes, and I can then get them done.

cc @codeguy

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -149,7 +149,14 @@ class App extends \Pimple
 
         // Flash
         $this['flash'] = $this->share(function ($c) {
-            return new \Slim\Flash($c['session'], $c['settings']['session.flash_key']);
+            $flash = new \Slim\Flash($c['session'], $c['settings']['session.flash_key']);
+            // TODO: Build array-access to current request messages for easy view integration
+            //
+            // if ($c['settings']['view'] instanceof \Slim\View) {
+            //     $c['view']->set('flash', $flash->getMessages());
+            // }
+
+            return $flash;
         });
 
         // Mode
@@ -1054,23 +1061,26 @@ class App extends \Pimple
     }
 
     /**
-     * Call
+     * Dispatch request and build response
      *
-     * This method finds and iterates all route objects that match the current request URI.
+     * This method will route the provided Request object against all available
+     * application routes. The provided response will reflect the status, header, and body
+     * set by the invoked matching route.
+     *
+     * The provided Request and Response objects are updated by reference. There is no
+     * value returned by this method.
+     *
+     * @param  \Slim\Http\Request  The request instance
+     * @param  \Slim\Http\Response The response instance
      */
-    public function call()
+    protected function dispatchRequest(\Slim\Http\Request $request, \Slim\Http\Response $response)
     {
         try {
-            if ($this['settings']['view'] instanceof \Slim\View) {
-                $this['view']->set('flash', $this['flash']->getMessages());
-            }
-
             $this->applyHook('slim.before');
             ob_start();
             $this->applyHook('slim.before.router');
             $dispatched = false;
-            $matchedRoutes = $this['router']->getMatchedRoutes($this['request']->getMethod(), $this['request']->getPathInfo());
-
+            $matchedRoutes = $this['router']->getMatchedRoutes($request->getMethod(), $request->getPathInfo());
             foreach ($matchedRoutes as $route) {
                 try {
                     $this->applyHook('slim.before.dispatch');
@@ -1088,9 +1098,66 @@ class App extends \Pimple
             }
             $this->applyHook('slim.after.router');
         } catch (\Slim\Exception\Stop $e) {}
-
-        $this['response']->write(ob_get_clean());
+        $response->write(ob_get_clean());
         $this->applyHook('slim.after');
+    }
+
+    /**
+     * Perform a sub-request from within an application route
+     *
+     * This method allows you to prepare and initiate a sub-request, run within
+     * the context of the current request. This WILL NOT issue a remote HTTP
+     * request. Instead, it will route the provided URL, method, headers,
+     * cookies, body, and server variables against the set of registered
+     * application routes. The result response object is returned.
+     *
+     * @param  string $url             The request URL
+     * @param  string $method          The request method
+     * @param  array  $headers         Associative array of request headers
+     * @param  array  $cookies         Associative array of request cookies
+     * @param  string $body            The request body
+     * @param  array  $serverVariables Custom $_SERVER variables
+     * @return \Slim\Http\Response
+     */
+    public function subRequest($url, $method = 'GET', array $headers = array(), array $cookies = array(), $body = '', array $serverVariables = array())
+    {
+        // Build sub-request and sub-response
+        $environment = \Slim\Environment::mock(array_merge(array(
+            'REQUEST_METHOD' => $method,
+            'REQUEST_URI' => $url,
+            'SCRIPT_NAME' => '/index.php'
+        ), $serverVariables));
+        $headers = new \Slim\Http\Headers($headers);
+        $cookies = new \Slim\Collection($cookies);
+        $subRequest = new \Slim\Http\Request($environment, $headers, $cookies, $body);
+        $subResponse = new \Slim\Http\Response();
+
+        // Cache original request and response
+        $oldRequest = $this['request'];
+        $oldResponse = $this['response'];
+
+        // Set sub-request and sub-response
+        $this['request'] = $subRequest;
+        $this['response'] = $subResponse;
+
+        // Dispatch sub-request through application router
+        $this->dispatchRequest($subRequest, $subResponse);
+
+        // Restore original request and response
+        $this['request'] = $oldRequest;
+        $this['response'] = $oldResponse;
+
+        return $subResponse;
+    }
+
+    /**
+     * Call
+     *
+     * This method finds and iterates all route objects that match the current request URI.
+     */
+    public function call()
+    {
+        $this->dispatchRequest($this['request'], $this['response']);
     }
 
     /**
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -89,7 +89,6 @@ class App extends \Pimple
         // Settings
         $this['settings'] = $this->share(function ($c) use ($userSettings) {
             $config = new \Slim\Configuration(new \Slim\ConfigurationHandler);
-            $config->setDefaults();
             $config->setArray($userSettings);
 
             return $config;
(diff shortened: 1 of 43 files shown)


designermonkey:
I will rebase this again before a pull, as there are too many commits here.

charlie-s:
I'm digging into the code and commits now to get the hang of things. Looking forward to it.

designermonkey:
I knew I'd missed things ;) Thanks, I'll add that now.

designermonkey:
It makes sense that as Slim still fully supports 5.3.x, the interface should be fully part of Slim. I'll do that.

charlie-s:
Nice :) I'm looking at Request and Response next. Anywhere else you'd recommend I start? You can also email me, it's listed on my profile page.

Last question for the moment – all of this is being done against Slim 3.5. Should we port this stuff to the 4.0 codebase?

charlie-s:
My thoughts on the `Collection` class – even though it's a core component and an interface for it might seem like unnecessary cruft, I always like that interfaces provide a summary of the contract between one class and other classes that extend or use it.

designermonkey:
> Last question for the moment – all of this is being done against Slim 3.5. Should we port this stuff to the 4.0 codebase?

Did you mean `2.5` and `3.0`? We are doing this work for a breaking release of `3.0`, which is shaping up to be mega awesome.

The Request and Response classes are where I lost my way a bit. Last I looked was a very tired me at 1am, so I closed the laptop and slept. There may be bits missing, so looking here and checking if all methods required by the App class are present makes sense.

> My thoughts on the Collection class...

These are actually my thoughts too, but I didn't want to just overload Slim with interfaces. I am happy to make more, as there are many classes in the setup that extend the Collection.

I will also stick an Environment interface in there as well, even though it will be empty for now. A question has arisen about CLI use of Slim, and yes, making interfaces for every class would help if we start making changes along those lines too.

@csdco thanks for your time so far looking through this, all help is appreciated.

designermonkey:
Plans have suddenly changed this evening as I opened up Sublime to work on this. I will crack on tomorrow now.

charlie-s:
Sorry, I meant 2.3.5 and 2.4.0. Not enough coffee apparently. 

I was just comparing a local pull of 2.4.0 where I have class `Slim` (Slim/Slim.php) instead of `App` (Slim/App.php), it no longer extends `\Pimple`, etc. But after browsing through the commit history on the master branch, I'm a bit confused because your branch shows 2.3.5 but has App.php, my pull of 2.4.0 has Slim.php, but browsing the repo's latest commits shows it's currently App.php.

I'm thinking this is all because the latest work is being done on the `develop` branch which hasn't changed any version tags since 2.3.5 – is that correct or am I missing something else? 

Enjoy your evening :)

designermonkey:
That's correct, yep. All intermediary work, @codeguy is doing in release branches.

John Porter.

charlie-s:
Regarding `\Slim\Http\Request` – it looks to me like the interface you defined works with all interactions within Slim.

charlie-s:
`\Slim\App.php` uses `\Slim\Http\Response::getMessageForCode($status)`. Should this then be accounted for in `\Slim\Interfaces\Http\ResponseInterface`? I realize that historically putting a static method in an interface doesn't make a lot of sense, but as of PHP 5.3 with late static bindings you can do:

```
$foo = new Foo();
$foo::someStaticMethod();
```

so this is potentially helpful to have in the interface.

Otherwise I also saw this interface matching up with `\Slim\Http\Response`'s usage.

designermonkey:
Noted, I can add that too.

charlie-s:
Can I commit to this branch or would I have to fork and send PRs to you? I'm happy to pass commits along for your review. Also started on the Container interface and could share that.

designermonkey:
I've given you collaborator access so you can push to my repository. Feel free to make any additions and changes, I can always alter anything if need be.

charlie-s:
Thanks! Just pushed initial work on the `CollectionInterface`. 

The `encrypt()` / `decrypt()` methods are used in App, and both require that the class implements `\IteratorAggregate`. Is there a way to define an interface that must also implement other interfaces, or do we need to consider an abstract class for `Collection`?

Something like:

```
interface CollectionInterface
{
    // ...
}

abstract class AbstractCollection implements \Countable, \IteratorAggregate
{
    public function count()
    {
        return count($this->data);
    }

    public function getIterator()
    {
        return new \ArrayIterator($this->data);
    }
}

class Collection extends AbstractCollection implements \Slim\Interfaces\CollectionInterface {
    // ...
}
```

I could be way off, just thinking.

designermonkey:
I don't know about the Abstract idea, what does @codeguy think?

charlie-s:
Just fixed the issue that I introduced between `Slim\Interfaces\CollectionInterface` and `Slim\Collection`'s `replace()` method. I should have ran phpunit before I pushed but I wanted to get the code up for you to see. 

The abstract thing could well be overkill, but after looking at how the Collection class is iterated over in the `encrypt()` / `decrypt()` and such, it seems like you may have to add an abstract class that implements `IteratorAggregate`, `Countable`, and potentially handles `encrypt()` and `decrypt()`, or just ignore them all from an interface perspective. That maybe defeats the purpose since they are called from App and elsewhere... which all goes back to your first point about Collection not really requiring an interface.

designermonkey:
I wanted to refactor the Collection to extend Pimple itself and adjust the behaviour slightly. I had a chat with @codeguy about it. I'm too tired right now to figure if that would have an impact on what we're doing here or not.

designermonkey:
I've had a good old think about this one, and I think I've come up with a clean solution (I think being the operative statement there).

For class interfaces where the class extends another, the interface should extend the the inherited one

```
class Headers extends Collection
{

}

interface HeadersInterface extends CollectionInterface
{

}
```

For classes implementing multiple interfaces, like Collection, then we should do something like the following

```
interface CollectionInterface extends IteratorAggregate
{
    // Include defined methods from the Countable interface
}
```

What do you think?

I want to have `Collection` extend `Pimple` as well, and in doing so, I will make a `PimpleInterface` doing similar.

codeguy:
Sorry, just catching up on this thread. I'd like to avoid using abstract classes and stick with interfaces only. If that means we have to adjust some of Slim's implementation to make interfaces easier, that's fine.

charlie-s:
+1 to avoiding abstracts; keep it slimmer :)

@designermonkey that sounds great. I can play with the `CollectionInterface` and push some code for you to review.

designermonkey:
Again, I very much doubt this is going to merge cleanly. I've rebased the develop branch into this one today, and it caused a real mess. I'm in progress cleaning it up, and also re-thinking some of the interfaces...

An example is the Request interface. We have only listed public methods that are required by the App class, but I think it should have public methods that we would expect the implementing class to be able to handle. IMO, it's better to say "Slim expects a class to be able to do the following.", rather than "This is the bare minimum that Slim requires to function.".

charlie-s:
Regarding `Request`, do you think the interface should require the `->is{Method}()` methods? (e.g. `isPost()`)? Also stuff like `getQueryString()`, `getProtocol()`?

designermonkey:
I've gone through and made my personal judgement on that.

> do you think the interface should require the ->is{Method}() methods?

Yes, and I've also added the `->{Method}` functions as well, like `post` and `get`.

When it comes to the others, I've only added a few that, personally, I would deem required for an HTTP framework. Some of them are just accessors for the headers object, and so aren't so needed, but others are IMO.

I'm nearly ready to push again, so take a look later on.

designermonkey:
Ok. All done as far as I can see.

@codeguy I've just test merged this using squash-commits also (as there are way too many for a normal history), and everything was ok with the world.

@csdco, if you want to add anything, please take a look through and let me know.

Also, I've gone with the idea of extending interfaces, but we still need to have a think about the multiple implementing classes, as just adding the interface methods is not quite right. It will ensure the methods are present, but I don't know if there's some PHP magic to how interfaces like `ArrayAccess`, `IteratorAggregate` and `Countable` work, as these are the only three we may see issues with.

If anyone else reading this thread has ideas, please feel free to let us know.

charlie-s:
Can't we extend multiple interfaces in an interface? e.g.

```
interface CollectionInterface extends \ArrayAccess, \Countable, \IteratorAggregate
```

I honestly haven't used Countable and IteratorAggregate before, and have only touched ArrayAccess on occasion, but the methods they call don't seem to be very complicated:

```
public function count()
{
    return count($this->values);
}
...
public function getIterator()
{
    return new \ArrayIterator($this->values);
}
```

etc. The only way to tackle this that I'm aware of would be an abstract class.

Also, is it odd that `protected $values = array();` is declared in `Pimple`, but `CollectionInterface` doesn't extend Pimple and `$this->values` is always assumed to at least be an empty array?

designermonkey:
I haven't read anywhere that interfaces can extend multiple others. Classes certainly can't do that.

Also, interfaces can't extend classes, so `Collection` could never extend `Pimple`. I will ask @fabpot whether he would provide a `PimpleInterface` and see what the answer is.

We could certainly add the protected property to `Collection` to ensure it is set, but `Collection` does extend `Pimple`, so I saw this as pointless.

charlie-s:
Apparently interfaces can extend multiple interfaces, but classes certainly cannot. Check out example 3 @ php.net: http://php.net/language.oop5.interfaces#language.oop5.interfaces.examples.ex3

You're absolutely correct, it can't extend Pimple. I think I was just trying to illustrate that the interface doesn't require that property, so in the event that someone uses the `CollectionInterface` without `Pimple` they've got an object that begins throwing lots of errors because `$this->values` isn't set.

designermonkey:
> Apparently interfaces can extend multiple interfaces

Well, I never spotted that! Damn! (goes to do some more work).

> so in the event that someone uses the CollectionInterface without Pimple they've got an object that begins throwing lots of errors because $this->values isn't set.

How so? the interface doesn't reference `$this->values` anywhere, it's the class that does, and the class would be replaced when someone uses the interface.

charlie-s:
Fair point! I guess I'm looking at `count()`, `getIterator()`, etc., and was stuck in the mindset that those would be defined elsewhere and therefore the class would depend on the `values` property. I think you're spot on and it's a moot point.

designermonkey:
@codeguy, I think this is ready to pull :D

designermonkey:
This gives me a nice warm feeling inside :)

codeguy:
:+1:

PR #706, opened 2013-12-23
Detect circular middleware stacks

marcioAlmada:
This pull request aims to enhance Slim\App behavior when encountering circular middleware stacks. Instead of a PHP Fatal error, now an exception is thrown:

``` php
<?php
$middleware = new CustomMiddleware();
$app = new Slim\App;
$app->add($middleware);
$app->add($middleware);
$app->run(); // fatal error
```

Before:

> Fatal error: Maximum function nesting level of '100' reached. aborting! [...]

After:

> RuntimeException: Circular Middleware setup detected. Tried to queue the same Middleware instance (CustomMiddleware) twice.

The fatal error can be hard to debug mostly because of it's opaque error message. This usually happens by some accident in userland code and deserves a more comprehensible message IMMO.

Cheers!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1021,9 +1021,13 @@ class App extends \Pimple
      */
     public function add(\Slim\Middleware $newMiddleware)
     {
+        $middleware = $this['middleware'];
+        if(in_array($newMiddleware, $middleware)) {
+            $middleware_class = get_class($newMiddleware);
+            throw new \RuntimeException("Circular Middleware setup detected. Tried to queue the same Middleware instance ({$middleware_class}) twice.");
+        }
         $newMiddleware->setApplication($this);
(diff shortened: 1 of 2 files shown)


codeguy:
Can't argue with better error reporting :+1: Thanks!

codeguy:
Actually, can you re-submit this change against the release-2.4.1 branch? That way I can get it pulled in this week rather than wait for the 3.0 release.

marcioAlmada:
OK :smile:

marcioAlmada:
Oops, just noticed `develop` is now related to a future version. Branch `2.4.1` looks quite different from develop, but no big deal.

I'll port this from App to Slim class ~~after I recover from Christmas dinner~~ ASAP.

PR #707, opened 2013-12-23
Non static Http classes, and other works

designermonkey:
@codeguy I know the rule is to do single feature or implementation changes in separate branches, sorry for having three separate changes in one here.

I wanted to get the Http classes to be instantiated objects rather than rely on static methods as I tried to explain in my email. They seem to lend themselves more to that pattern IMO. Anyway, I got it working ok, so let me know if you think we could go this route here.

Also, as we discussed, I have the Collection extending Pimple now, more work could be done in the future to use more array syntax throughout (maybe?), currently works fine.

Also also, I've protected the Headers and Cookies instances in the Request and Response classes.

All unit tests pass, and after xmas, I will take a thorough look through to see if there is full code coverage with these changes.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -93,14 +93,14 @@ class App extends \Pimple
 
         // Environment
         $this['environment'] = $this->share(function ($c) {
-            return \Slim\Environment::createFromGlobals();
+            return \Slim\Environment($_SERVER);
         });
 
         // Request
         $this['request'] = $this->share(function ($c) {
             $environment = $c['environment'];
-            $headers = \Slim\Http\Headers::createFromEnvironment($environment);
-            $cookies = new \Slim\Collection(\Slim\Http\Cookies::extractFromHeaders($headers));
+            $headers = new \Slim\Http\Headers($environment);
+            $cookies = new \Slim\Http\Cookies($headers);
             if ($c['settings']['cookies.encrypt'] ===  true) {
                 $cookies->decrypt($c['crypt']);
             }
@@ -537,7 +537,7 @@ class App extends \Pimple
     public function lastModified($time)
     {
         if (is_integer($time)) {
-            $this['response']->headers->set('Last-Modified', gmdate('D, d M Y H:i:s T', $time));
+            $this['response']->getHeaders()->set('Last-Modified', gmdate('D, d M Y H:i:s T', $time));
             if ($time === strtotime($this['request']->getHeaders()->get('IF_MODIFIED_SINCE'))) {
                 $this->halt(304);
             }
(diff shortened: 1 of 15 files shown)


designermonkey:
Just to note this will affect how the interfaces branch works, so if you decide you like this, and it is to be pulled, it should go in first.

codeguy:
Just fyi, the request and response objects have always been instantiated objects... the static methods were simply factories that return instances based on different types of input arguments.

Anyway, will have a look through this soon and get it merged in. Thanks!

designermonkey:
I've not edited those methods, as I can see they are simple helpers, just the Cookies and Headers classes, and how these are used within Request and Response. Should have been clearer there :)

codeguy:
:+1:  Changes look great. Gotta head out to a family lunch, but will work on this some more tonight. v3 is going to be great!

designermonkey:
You're as bad as me. It's xmas and we're working on code.

Hope you have a great xmas though man!

prisis:
hey, if i try that pull i get a error like that: "Fatal error: Uncaught exception 'InvalidArgumentException' with message 'Passed variable is not an array or object, using empty array instead' in /Slim/Collection.php:167 Stack trace: #0 /Slim/Collection.php(167): ArrayIterator->__construct(NULL) #1 Slim/App.php(1224): Slim\Collection->getIterator() #2 Slim/App.php(1081): Slim\App->finalize() #3 /index.php(116): Slim\App->run() #4 {main} thrown in Slim/Collection.php on line 167

designermonkey:
When do you see this error? Is it when you run the bundled `index.php` file?

designermonkey:
Aha! Have you run `composer update`, as the Pimple class is now loaded by composer instead of being included in the project.

prisis:
I didn't use composer, i included all files manually "pimple and slim" in my index.php, it worked before, but if included your code i get that error

designermonkey:
The reason the error is happening (at least the only way I can re-create it) is that Pimple isn't being auto-loaded or found, and therefore not giving access to it's `values` property.

Can we see your `index.php` file, and how you've auto-loaded the files?

prisis:
Of course ;), 
autoloadManager is something like composer.
And now after i added the new changes i get a white page.

$autoloadManager = new autoloadManager();
$autoloadManager->setScanOptions(autoloadManager::SCAN_ALWAYS);
$autoloadManager->setSaveFile('System/_Storage/_autoload/_cache_autoload.php');
$autoloadManager->addFolder(realpath(dirname(__FILE__)) . '/System/Vendor');
$autoloadManager->register();

/*
|--------------------------------------------------------------------------
| Turn On The Lights
|--------------------------------------------------------------------------
|
| We need to illuminate PHP development, so let's turn on the lights.
| This bootstraps the framework and gets it ready for use, then it
| will load up this application so that we can run it and send
| the responses back to the browser and delight these users.
|
*/
$brain = new Pimple();
$brain['app'] = new App();

$brain['app']->run();

designermonkey:
I don't think there's anything I can do to help here, sorry.

codeguy:
@prisis I recommend you post a question to http://help.slimframework.com. The issue is almost certainly related to your custom autoloader.

PR #717, opened 2013-12-27
Allow changing the middleware which is responsible for printing pretty exceptions

mcuelenaere:


codeguy:
This is already resolved in the develop branch and is scheduled to be included in the 3.0 release.

PR #718, opened 2013-12-28
Allow all valid characters in function names of getCallableAsClass

mcuelenaere:
This uses the [a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]\* regex as specified by https://php.net/manual/en/functions.user-defined.php

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -194,7 +194,7 @@ class Route implements RouteInterface
     public function setCallable($callable)
     {
         $matches = array();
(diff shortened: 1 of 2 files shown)


mcuelenaere:
@codeguy Any comments?

poison:
+1 I need the underscore as well :)

codeguy:
Merged into develop and release-2.4.3 branches.

PR #721, opened 2013-12-30
Lazy-initialize class in \Slim\Route::setCallable()

mcuelenaere:
You don't want to construct all kinds of objects when setting up your routes, so don't initialize those objects until
the route is actually dispatched. Add some basic caching so multiple route dispatches don't result in multiple class
instantiations.

Note: I had to comment out RouteTest::testGetCallableAsClass(), as the callable is now a lambda function instead of an array. WDYT is the best way to solve this? (I think that the test was depending on an implementation detail and should just test whether the set callable is invoked when the route gets dispatched)

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -195,7 +195,15 @@ class Route implements RouteInterface
     {
         $matches = array();
         if (is_string($callable) && preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
-            $callable = array(new $matches[1], $matches[2]);
+            $class = $matches[1];
+            $method = $matches[2];
+            $callable = function() use ($class, $method) {
+                static $obj = null;
+                if ($obj === null) {
+                    $obj = new $class;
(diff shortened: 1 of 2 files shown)


anlutro:
Definitely this. I guess the "fix" for the broken test is to make `foo()` return something and do an `assertEquals` on `$route->dispatch()`.

designermonkey:
I definitely like this. +1 from me.

mcuelenaere:
@codeguy Any comments?

poison:
I like this too! +1!

haroldkyle:
This works much better, and helps the memory footprint a lot. +1

scottux:
+1

codeguy:
Scheduled for 2.4.3

codeguy:
This is merged into the develop branch. Do you want to send a similar PR for the `release-2.4.3` branch?

mcuelenaere:
Sure, see #779

PR #734, opened 2014-01-28
Add require-dev: phpunit (with proper .gitignores)

zergin:


designermonkey:
@zergin You should add this PR to the composer.json file in the develop branch.

PR #735, opened 2014-01-28
Add require-dev: phpunit (with proper .gitignores)

zergin:


composer.json
```diff
@@ -19,6 +19,11 @@
     "require-dev": {
         "mockery/mockery": "dev-master@dev"
     },
+
+    "require-dev": {
+        "phpunit/phpunit": "*"
```
nubs, line 21: There's a require-dev right above here - shouldn't it go in there?

zergin, line 21: It should and it's already fixed - it was an artefact after git cherry-picking when switching base branches.

PR #736, opened 2014-01-28
Fix tests when mcrypt extension is not loaded

zergin:
Mcrypt is only marked as suggested so tests should pass with extension not loaded.

### DIFF
--- a/tests/CollectionTest.php
+++ b/tests/CollectionTest.php
@@ -135,6 +135,11 @@ class CollectionTest extends PHPUnit_Framework_TestCase
 
     public function testEncryptAndDecrypt()
     {
+        if (!extension_loaded('mcrypt')) {
+            $this->markTestSkipped('The mcrypt extension is not loaded');
+            return;
+        }
+
         // Prepare bag
         $data = array(
             'abc' => '123',
--- a/tests/ConfigurationTest.php
+++ b/tests/ConfigurationTest.php
@@ -41,28 +41,34 @@ class HandlerTest extends ConfigurationHandler
  */
 class ConfigurationTest extends \PHPUnit_Framework_TestCase
 {
-    public $defaults = array(
-        // Application
-        'mode' => 'development',
-        'view' => null,
-        // Cookies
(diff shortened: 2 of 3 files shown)


zergin:
I see that MCRYPT consts should be defined by App.php ( https://github.com/codeguy/Slim/blob/develop/Slim/App.php#L35 ) but it is not the case for tests. So may be this patch should be changed to reflect that. What is your recommended way?

codeguy:
I would probably just update composer dev requirements to require the mcrypt lib.

zergin:
Wouldn't that defeat the purpose of checking code when there is no mcrypt extension? If tests always work with the extension present then there is a risk of not caching some dependency.

lrlopez:
@zergin: Yes, you're right, but (ideally) tests should also check the code that runs when there is no mcrypt extension.

Also think the other way around: Tests should be also able to detect when a code change affects code that depends on mcrypt and if you skip them when is not installed, you risk to unintentionally break code that worked before.

I agree with Josh in that adding mcrypt as a dev dependency should be the way to go. I didn't peer at the tests, but the failback code should also be checked there for completeness.

lrlopez:
Just to clarify: I'd add mcrypt as a dev dependency but also add tests for the expected behaviour when encrypted cookies are used but no mcrypt extension is available.

codeguy:
@lrlopez Nails it. We need to require mcrypt so we can have tests that cover cases with and without mcrypt.

codeguy:
@zergin Do you want to take the lead on this? We'll need a test or two that confirms behavior if mcrypt is not loaded. Let me know if you're busy and I can write some up this weekend.

sarciszewski:
If mcrypt isn't loaded, this falls back to null. Would it be better to also support openssl? Or is this irrelevant with `Zend\Crypt`?

codeguy:
@sarciszewski This is now irrelevant. I need to clean up obsolete tests, which I'll try to do tonight or tomorrow.

codeguy:
Leaving this thread open until that is done.

PR #742, opened 2014-02-03
Update Route.php

entenbraten:
Method "array_values()" on $this->getParams() removed.
Added an index based array with one index to match call_user_func_array requirements.
So you get an associate array as the parameter in your callback function.
IHMO easier to deal with.

=> documentaion gotta be changed


silentworks:
This is now outdated and cant see the immediate value of the change, closing off for now. Reopen with updated code and the value it adds over using array_values and will reconsider.

PR #747, opened 2014-02-11
Allow underscores in method names for controllers

hischeapmoves:
This adds an underscore to the regex used to compose the $callable variable to match cases where the method name contains an underscore.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -155,7 +155,7 @@ class Route
     public function setCallable($callable)
     {
         $matches = array();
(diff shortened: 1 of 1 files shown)


calmdev:
I can also reproduce this on 2.4.0 and can confirm that the PR resolves it.

davestewart:
This would be useful, esp for post routes, i.e. function add_post()

mcuelenaere:
Similar to #718

codeguy:
Closing in favor of https://github.com/codeguy/Slim/pull/718

PR #750, opened 2014-02-16
Add additional "reRun" method to Slim

ryan-mahoney:
No I haven't created any unit tests -- yet!

Just wanted to see if you could take a quick look at this change.  Essentially, this is to "reRun" Slim on a different HTTP method and URI so that you could call additional routes.  This code assumes that Slim has already been "run" and now it's being run again in the same request.

I had mentioned this a couple of months ago, the most common use case I can think of (mine) is when you want to call your own RESTful API from you own Slim based application... without creating a new and unnecessary HTTP request to your own server.  This has a lot of benefits.  The first, is speed, because some overhead is eliminated.  The second, is that exceptions from the reRuns will now bubble up.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1293,6 +1293,64 @@ class Slim
     }
 
     /**
+     * ReRun
+     *
+     * This method is used to call a secondary URL after Slim has been executed in
+     * the normal way.
+     */
+
+    public function reRun($method, $uri)
+    {
+        set_error_handler(array('\Slim\Slim', 'handleErrors'));
+
+        //Apply final outer middleware layers
+        if ($this->config('debug')) {
+            //Apply pretty exceptions only in debug to avoid accidental information leakage in production
+            $this->add(new \Slim\Middleware\PrettyExceptions());
+        }
+
+        //Invoke middleware and application stack
+        try {
+            ob_start();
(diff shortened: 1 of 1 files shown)


designermonkey:
I think Josh added a sub request method to Slim for the soon to be released version 3, let me have a look...

designermonkey:
Aha!

here it is: https://github.com/codeguy/Slim/blob/develop/Slim/App.php#L1128 in the develop branch.

Is this the sort of thing you want? If so, then it's been tackled, if not, maybe we could work towards a common goal.

codeguy:
Yep this is already available in the develop branch scheduled for the 3.0 release.

ryan-mahoney:
LOL, I can believe I missed that!  Thanks

ryan-mahoney:
So, I obtained the dev version.  Right off the bat, the code was generating a few errors.  Those errors could be fixed w/ two changes:

//parseHeaders expects Environment object
$headers->parseHeaders($environment);

//replace expects cookies to be and array, not an object
$cookies->replace((Array)$cookies);

At this point, I wasn't getting run-time errors, but it's returning an empty string for simple predefined routes.  I am using it with ReactPHP... so that could be a factor... will keep testing.

Issue #741, opened 2014-02-01
notFound & error handlers with 2.4.0 class controllers


calmdev:
Should the notFound and error handlers work with the new [2.4.0](https://github.com/codeguy/Slim/releases/tag/2.4.0) style controllers?

My controller(s):

``` php
// Load controllers.
$home = new \App\Controller\Home($app);
```

This works:

``` php
// Define routes.
$app->get('/', array($home, 'index'));
$app->notFound(array($home, 'notFound'));
$app->error(array($home, 'error'));
```

This works:

``` php
// Define routes.
$app->get('/', '\App\Controller\Home:index');
$app->notFound(array($home, 'notFound'));
$app->error(array($home, 'error'));
```

This does not work:

``` php
// Define routes.
$app->get('/', '\App\Controller\Home:index');
$app->notFound('\App\Controller\Home:notFound');
$app->error('\App\Controller\Home:error');
```

Here's the error I encounter in the last example:

![screen shot 2014-02-01 at 3 41 33 pm](https://f.cloud.github.com/assets/461249/2058734/7d26ce0a-8b81-11e3-8fa6-d8b9e7ef9af0.png)

erizocosmico:
Nope, `notFound` and `error` currently don't support the new style controllers.

calmdev:
Ok, I was wondering if it was intentionally left out for some reason or just overlooked.

I'd like to use your PR #743, so hope it's merged :+1:

codeguy:
Closing. Addressed in PR #752.

PR #752, opened 2014-02-16
Allow passing controllers to error and notFound

erizocosmico:
This PR adds functionality to allow the user to pass a controller instead of a callable to notFound and error methods.

Adresses #741.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -144,6 +144,23 @@ class Route
     {
         $this->pattern = $pattern;
     }
+    
+    /**
+     * Parses controller string to an almost callable array
+     * @param string $callable
+     * @return bool|array
+     */
+    public static function stringToCallable($callable)
+    {
+        $matches = array();
+        if (preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
+            $callable = array($matches[1], $matches[2]);
+        } else {
+            $callable = false;
+        }
+        
+        return $callable;
+    }
 
     /**
      * Get route callable
@@ -151,6 +168,9 @@ class Route
      */
     public function getCallable()
     {
+        if (is_array($this->callable)) {
+            $this->callable = array(new $this->callable[0], $this->callable[1]);
+        }
         return $this->callable;
     }
 
@@ -161,12 +181,11 @@ class Route
      */
     public function setCallable($callable)
     {
-        $matches = array();
-        if (is_string($callable) && preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
-            $callable = array(new $matches[1], $matches[2]);
+        if (is_string($callable) && !is_callable($callable)) {
+            $callable = self::stringToCallable($callable);
         }
 
-        if (!is_callable($callable)) {
+        if (!is_callable($callable) && !is_array($callable) || $callable === false) {
             throw new \InvalidArgumentException('Route callable must be callable');
         }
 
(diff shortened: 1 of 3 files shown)


calmdev:
FWI, I've cherry-picked commit into a fork and it's working good for me.

It's a small change, but it feels so much cleaner now, ha.

calmdev:
Hrm...this only appears to be working with static methods for me?

erizocosmico:
@calmdev According to the tests it should work with instance methods too. Can you give me an example where it doesn't work? I'll take a look at it later.

calmdev:
@mvader here's [my branch](https://github.com/calmdev/Slim/compare/codeguy:master...master) with the cherry picked commit and below is what I'm running into when visiting a bogus URL to test the notFound handler.

Routes:

``` php
// Define routes.
$app->get('/', '\App\Controller\Home:index');
$app->notFound('\App\Controller\Home:notFound');
$app->error('\App\Controller\Home:error');
```

Controller:

``` php
namespace App\Controller;

class Home {

    public function __construct() {
        $this->app = \Slim\Slim::getInstance();
    }

    public function index() {
        $this->app->redirect('/users');
    }

    public function notFound() {
        $this->app->render('404.tpl');
    }

    public function error(\Exception $error) {
        $this->app->render('500.tpl', array(
            'message' => $error->getMessage()
        ));
    }

}
```

Error:
screen shot 2014-02-21 at 11 42 32 am

If I alter the controller like this it works:

``` php
namespace App\Controller;

class Home {

    public function __construct() {
        $this->app = \Slim\Slim::getInstance();
    }

    public function index() {
        $this->app->redirect('/users');
    }

    public static function notFound() {
        $app = \Slim\Slim::getInstance();
        $app->render('404.tpl');
    }

    public static function error(\Exception $error) {
        $app = \Slim\Slim::getInstance();
        $app->render('500.tpl', array(
            'message' => $error->getMessage()
        ));
    }

    public function logout() {
        $this->app->user->logout();
    }

}
```

Let me know if you want more information. Maybe I'm making a mistake, but hope it's not just me making noise in here for no reason. This place is always so damn quiet. I appear to get stuck on these [two](https://github.com/calmdev/Slim/blob/7f71657f868f66cb665ce7faf40faf500ec28ea1/Slim/Slim.php#L602) [lines](https://github.com/calmdev/Slim/blob/7f71657f868f66cb665ce7faf40faf500ec28ea1/Slim/Slim.php#L670) - not making it to the next `else if` which is what we want I think?

erizocosmico:
@calmdev You were right. Somehow php was treating the arrays in `error` and `notFound` as callables. It's weird because it doesn't with `\Slim\Slim:defaultNotFound`. Anyway, I fixed it and rewrote the tests.

calmdev:
Nice work @mvader.

Both the error and notFound handlers are working perfectly for me now.

I just noticed something with Slim which completely caught me off guard while testing the custom error handler. Check out [these lines](https://github.com/codeguy/Slim/blob/master/Slim/Slim.php#L1334-L1342). If you are running in debug mode then the custom error handler never gets called. It will invoke Slim's default error handler instead. I'm not sure if this desirable for folks who have defined their own - not really the behavior I was expecting. In my application if debug mode is enabled it does more verbose logging, but as it is now I lose the custom error handler. @codeguy is it intentional behavior?

IMO, if it was to change this PR would be a good spot for it since it's related. Thoughts?

codeguy:
@calmdev That is intentional (although not ideal, in hindsight) behavior. It has been fixed in the develop branch (what will become 3.0 soon).

Issue #748, opened 2014-02-12
Slim instance as parameter in controller constructor


danielgsims:
We would like to pass Slim as a dependency in controllers without having to use the Slim singleton instance. This would help us to write more thorough tests.

scottux:
+1

lectrotext:
This would make for better unit testing.

ciaranmaher:
I came across a way pass Slim into your controllers without using the Slim singleton. https://github.com/codeguy/Slim/pull/755#issuecomment-35495095

silentworks:
Closing this off in favour of #822

PR #755, opened 2014-02-17
Slim instance as parameter in controller constructor

ciaranmaher:
This PR addresses issue #748 by passing the \Slim\Slim application object to the controller constructor.

Just add something like the following to your controller class and \Slim\Slim object will be available within your controllers.

``` php
/**
 * Constructor
 * @param \Slim\Slim $application The Slim application
 */
public function __construct(\Slim\Slim $application)
{
    $this->app = $application;
}
```

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -99,9 +99,9 @@ class Route
      * Constructor
      * @param string $pattern The URL pattern (e.g. "/books/:id")
      * @param mixed $callable Anything that returns TRUE for is_callable()
-     * @param \Slim\Slim $application
+     * @param \Slim\Slim|null $application The primary Slim application instance
      */
-    public function __construct($pattern, $callable, $application)
+    public function __construct($pattern, $callable, $application = null)
     {
         $this->setApplication($application);
         $this->setPattern($pattern);
@@ -140,13 +140,26 @@ class Route
      * Set application
      *
      * This method injects the primary Slim application instance into
-     * this middleware.
+     * this class.
      *
      * @param  \Slim\Slim $application
      */
     public function setApplication($application)
     {
         $this->app = $application;
+    }
+
+    /**
+     * Get application
+     *
+     * This method retrieves the application previously injected
+     * into this class.
+     *
+     * @return \Slim\Slim
+     */
+    public function getApplication()
+    {
+        return $this->app;
     }
 
     /**
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -51,6 +51,11 @@ class Route
     protected $callable;
 
     /**
+     * @var \Slim\Slim Reference to the primary application instance
+     */
+    protected $app;
+
+    /**
      * @var array Conditions for this route's URL parameters
      */
     protected $conditions = array();
@@ -94,9 +99,11 @@ class Route
      * Constructor
      * @param string $pattern The URL pattern (e.g. "/books/:id")
      * @param mixed $callable Anything that returns TRUE for is_callable()
+     * @param \Slim\Slim $application
      */
-    public function __construct($pattern, $callable)
+    public function __construct($pattern, $callable, $application)
     {
+        $this->setApplication($application);
         $this->setPattern($pattern);
         $this->setCallable($callable);
         $this->setConditions(self::getDefaultConditions());
@@ -130,6 +137,19 @@ class Route
     }
 
     /**
+     * Set application
+     *
+     * This method injects the primary Slim application instance into
+     * this middleware.
+     *
+     * @param  \Slim\Slim $application
+     */
+    public function setApplication($application)
+    {
+        $this->app = $application;
+    }
+
+    /**
      * Set route pattern
      * @param  string $pattern
      */
@@ -156,7 +176,7 @@ class Route
     {
         $matches = array();
         if (is_string($callable) && preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
-            $callable = array(new $matches[1], $matches[2]);
+            $callable = array(new $matches[1]($this->app), $matches[2]);
         }
 
         if (!is_callable($callable)) {
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -434,7 +434,7 @@ class Slim
     {
         $pattern = array_shift($args);
         $callable = array_pop($args);
-        $route = new \Slim\Route($pattern, $callable);
+        $route = new \Slim\Route($pattern, $callable, $this);
         $this->router->map($route);
         if (count($args) > 0) {
             $route->setMiddleware($args);
--- a/tests/RouteTest.php
+++ b/tests/RouteTest.php
@@ -76,6 +76,18 @@ class RouteTest extends PHPUnit_Framework_TestCase
         $this->assertEquals('getCurrentRoute', $callable[1]);
     }
 
+    public function testGetCallableAsClassPassingApp()
+    {   
+        $app = new \Slim\Slim();
+        $route = new \Slim\Route('/foo', '\Slim\Router:getCurrentRoute', $app);
+
+        $callable = $route->getCallable();
+        $application = $route->getApplication();
+        $this->assertInstanceOf('\Slim\Router', $callable[0]);
+        $this->assertEquals('getCurrentRoute', $callable[1]);
+        $this->assertInstanceOf('\Slim\Slim', $application);
+    }
+
     public function testGetCallableAsStaticMethod()
     {
         $route = new \Slim\Route('/bar', '\Slim\Slim::getInstance');
--- a/tests/RouteTest.php
+++ b/tests/RouteTest.php
@@ -78,8 +78,8 @@ class RouteTest extends PHPUnit_Framework_TestCase
 
     public function testGetCallableAsClassPassingApp()
     {   
-        $app = new \Slim\Slim();
-        $route = new \Slim\Route('/foo', '\Slim\Router:getCurrentRoute', $app);
+        $s = new \Slim\Slim();
+        $route = new \Slim\Route('/foo', '\Slim\Router:getCurrentRoute', $s);
 
         $callable = $route->getCallable();
         $application = $route->getApplication();
--- a/tests/RouteTest.php
+++ b/tests/RouteTest.php
@@ -77,7 +77,7 @@ class RouteTest extends PHPUnit_Framework_TestCase
     }
 
     public function testGetCallableAsClassPassingApp()
-    {   
+    {
         $s = new \Slim\Slim();
         $route = new \Slim\Route('/foo', '\Slim\Router:getCurrentRoute', $s);
 


erizocosmico:
How about a base controller? Copy and paste that snippet on every controller it's not a very elegant solution to this problem. 

This issue, however, needs to be addressed. Currently there is no way to pass the instance of the app to the controller. Well, there is...

``` php
public function myMethod()
{
    $app = \Slim\Slim::getInstance();
}
```

This approach has the same problem I see in your approach. You have to copy and paste that line in every method (or class in your approach).

I know a base controller goes against the Slim concept but, IMHO, it's really necessary. I think it's better extending a base controller than repeating over and over the same lines of code.

Something like:

``` php
class BaseController
{
    protected $app = null;
    public function setApplication($app)
    {
        $this->app = $app;
    }
}
```

It would be checked that the controller class has the method `setApplication` with reflection (or whatever you can think of to check that) so it wouldn't be compulsory to extend the base controller.

ciaranmaher:
Slim doesn't dictate how you structure your own controller classes, that's the beauty of it.

The above constructor was just to highlight how you would integrate it, I wasn't suggesting that you'd have to copy it into all your controllers. It's really up to you how you use the above constructor, it can sit in your own base controller or individual controller. 

Myself personally, I have the above code sitting in my abstract base controller class and all my other controllers extend from it.

erizocosmico:
According to your code the `$app` parameter is always passed to the controller class. Your controllers will always have to have a constructor that accepts that parameter. That does dictate how you structure your own controller classes. Then if Slim doesn't have a base controller the user will have to write one. Or a constructor. Or whatever. The thing is that the user will have to write it.
Having the choice to use a base controller doesn't force you to use it.

ciaranmaher:
@mvader I get your point about how this would dictate how you structure the constructor within your controller class, although declaring a class constructor is entirely optional. On the flip side of that you could also argue that Slim is currently dictating how your controllers constructor is defined, i.e. it must be declared with no parameters (or else with default parameters set)

If I choose to have a constructor in my controller class that accepts any parameters, Slim forces me to set a parameter default otherwise a PHP warning will be thrown. `Missing argument 1 for Controller::__construct(), called in Slim/Route.php on line 166`

I approached this PR based on what people were looking for in issue #748 and it's something that I need in all my apps. This approach would allow for better unit testing of controllers avoiding the Singleton approach `\Slim\Slim::getInstance()` (which is to be [removed from Slim 3.0](http://help.slimframework.com/discussions/questions/1211-what-is-the-best-way-to-pass-app-to-class-controllers#comment_30845786))

An other option of passing `$app` to the controllers was mentioned in the help forum by way of a [go between class](http://help.slimframework.com/discussions/questions/1211-what-is-the-best-way-to-pass-app-to-class-controllers#comment_31164386) but it doesn't seem that elegant.

In saying all that having some sort of BaseController that's part of Slims core would also work very nicely :)

erizocosmico:
@ciaranmaher Well, if `\Slim\Slim::getInstance()` is going to be removed soon it makes sense to force controllers to have a constructor that accepts the `$app` parameter.

designermonkey:
I went through all this to some degree before, and got argued with.

https://github.com/codeguy/Slim/pull/639

IMO, if you are allowing the ability to pass a route through a 'Controller' style class and method, then of course that class needs to receive the main app, otherwise it can't get access to the container, or any of the other functions.

lalop:
I'm not sure but maybe that can be done in a middleware ?
And the code put in slim extra.

designermonkey:
Slim Extras is being deprecated when we release v3.

Also, I just realised that this has been tackled to some degree in v3 (the develop branch). The Route class' `setCallable` now accepts a string representation of a `Class:method`...

``` php
    /**
     * Set route callable
     * @param  mixed                     $callable
     * @throws \InvalidArgumentException If argument is not callable
     * @api
     */
    public function setCallable($callable)
    {
        $matches = array();
        if (is_string($callable) && preg_match('!^([^\:]+)\:([[:alnum:]]+)$!', $callable, $matches)) {
            $callable = array(new $matches[1], $matches[2]);
        }

        if (!is_callable($callable)) {
            throw new \InvalidArgumentException('Route callable must be callable');
        }

        $this->callable = $callable;
    }
```

Also, parameters are passed on to the callable by using the `setParams` method of a Route. One could therefore write a Route as follows:

``` php
$app->get('/hello/:name', 'Class:method')->setParams(array(
    $app,
    $someVariable
));
```

While this is not ideal as each method would receive the `$app` instance, it is a way of doing it.

ciaranmaher:
@designermonkey I wasn't aware "Route::setParams()" was available in 2.4.2

That's great news because now I can do something like this:

In my routes file:

``` php
$app->get('/hello/:first/:second', '\Hello:sayHello')->setParams(
    array($app)
);

```

In the \Hello class:

``` php

public function sayHello($app, $first, $last)
{        
    echo 'Hello ' . $first . ' ' . $last;
    echo get_class($app);
}

```

The route `/hello/ciaran/maher` will output:

``` php
Hello ciaran maher
\Slim\Slim
```

It still would be nice to pass $app to the controllers constructor but this is a nice workaround that will allow me access to $app without using the Singleton method or some kind of go between class!

danielgsims:
The set params option is a better way to go about it than using the Singleton instance. The approach at least decouples the dependency. I would prefer to have it available for constructor dependency injection instead of during the method call. The constructor DI approach would cut down on ceremony. Maybe the approach of a follow up function would be good if the controller class was lazy loaded and we could use a setConstructorParams method in addition to setParams. This would still involve extra ceremony, which I am ok with, but it would allow constructor DI which is ideal. Maybe this approach would be good for the Slim 3 with the lazy loading efforts.

designermonkey:
`setConstructorParams` I really like that idea. I'll get a PR together for that!

designermonkey:
A PR is in for @ciaranmaher's repository as there is already work affecting this area of the Route class. It makes sense to merge the two together before merging into the main repository. @codeguy has said that the work by @ciaranmaher will be pulled for the next point release. Lets hope this gets the go-ahead!

danielgsims:
@designermonkey is this the lazy loading controller work? This would make it a lot better.

designermonkey:
It allows you to add parameters to pass to the constructor of any class you use for the route.

``` php
$app->get('/hello/:first/:second', '\Hello:sayHello')->setConstructorParams(
    array($app, $someParam)
);
```

``` php
class Hello
{
    public function __construct($app, $someParam) {}
}
```

It will only try to pass along parameters if they have been set using the `setConstructorParams` method.

ciaranmaher:
I think that's a great way to do it, just spotted a couple of things in the PR https://github.com/ciaranmaher/Slim/pull/1

silentworks:
@designermonkey how would this work if we had more than one method and route for that controller?

``` php
$app->get('/hello/:first/:second', '\Hello:sayHello')->setConstructorParams(
    array($app, $someParam)
);

$app->get('/yo/:world', '\Hello:yoWorld');
```

Would I have to pass setConstructorParams on the second route too? I can see this becoming a long array list of dependencies on my routing code if I had lets say 6 dependencies and if I have to repeat that, it would start to become a bit tedious.

lalop:
One other way could be using the first part of the route callback `Hello` in `Hello\sayHello` not as a class name but as a key in a DI container. Or using an other format to that.

erizocosmico:
How about `setDefaultConstructorParams()`?

``` php
$app = new \Slim\Slim();
...
$app->setDefaultConstructorParams($app, $db, $logger, $foo);
...
$app->get('/', '\App\Home:index');
...
```

silentworks:
@mvader but that means you just tied me into all the same objects for all my controllers which is not ideal.

@lalop I like that idea, the problem I am finding with this is how would one find the key in the di container without injecting the di container into the Route?

erizocosmico:
@silentworks then something like `setConstructorParamsFor()`?

``` php
$app = new \Slim\Slim();
...
$app->setConstructorParamsFor('\App\Home', $app, $db, $logger, $foo);
$app->setConstructorParamsFor('\App\Users', $app, $db);
...
$app->get('/', '\App\Home:index');
$app->get('/users', '\App\Users:index');
...
```

lalop:
@silentworks controllers key could be prefixed by `controllers` in the container `controllers.hello` for exemple.

silentworks:
@mvader that would be a better solution. Maybe change the method name, I so dislike that name, something like inject sound better and makes more sense.

``` php
$app->inject('\App\Home', $app);
```

silentworks:
@lalop I understand that bit, but how do you anticipate getting this into the Route.php file without injecting the di container?

lalop:
@silentworks good point, I'm not aware of all the internal mecanisms into Slim but Route need to use the app container

designermonkey:
How about doing it with the Configuration object?

``` php
$app['settings']['route.foo'] = array(
    $app, $foo, $bar
);

$app['settings']['route.bar'] = array(
    $foo, $app, $bar
);

$app->get('/', '\App\Home:index')->setConstructorParams('route.foo');

$app->get('/bar', '\App\Home:barRoute')->setConstructorParams('route.bar');

$app->get('/baz', '\App\Home:bazRoute')->setConstructorParams(array(
    $bar, $foo, $app
));

$app->get('/foobar', '\App\Home:fooBarRoute');
```

We could have the `setConstructorParams` method identify a string and look up from the config object.

The `Route` object would need the current `$app` passing into it's constructor, but I don't see this as too big a problem personally.

After all, we have the `Configuration` object available for storing things, so why not use it?

designermonkey:
Expanding on the above, it wouldn't call the `settings` object until `getCallable` is called, so as to not store the array of params in more than one location.

silentworks:
@designermonkey This still feels like I am repeating code for the sake of repeating code, I will have to call setConstructorParams on each route of this exact same Controller. I am more for @mvader suggestion or automatic resolution like they have in Laravel.

danielgsims:
@silentworks I like the idea of being able to designate dependencies for multiple routes that use the same controller. I would like to use the DI sugestion from @lalop however, the route doesn't have scope of the pimple container. I think the Route needs to have it's own setConstructorParams method so that one would still be able to directly add DI to the controller without having to use the Slim instance. For example, someone might want to use Slim routing without the Slim application ala Symfony HTTP Foundation. 

What would happen if we used setConstructorParams and an $app->inject method?

danielgsims:
Also, what about adding a setConstructorParams on a Route Group? It stands to reason that shared routes within an API that use controllers might inherit from a base controller. Shared dependencies would be injected via the constructor. Methods that have unique dependencies could inject those dependencies through the setParams method.

silentworks:
@danielgsims I am not sure if Slim Route is being used anywhere outside of Slim itself, but I can see your point. I think adding the basic setConstructorParams as @designermonkey for now is ok, but the Route class need to be added to the di container in Slim\App to make it easily interchangable. Currently I have to extend the Slim\App and changing the Slim\Route class in the mapRoute method.

danielgsims:
@silentworks I don't understand. You want to add your Route the container the same time you add it to the Router so you can modify it after the fact?

silentworks:
@danielgsims I have submitted a PR but for Slim 3 rather than 2, You can check it out #767 ti get a better understanding of what I am saying.

With this PR, I can change \Slim\Route after instantiating \Slim\App like this

``` php
$app = new \Slim\App();
$app['route'] = $app->share(function () {
    return new MyOwnRouteClass();
});
```

With this I can now create my own way of Route and implement my own way of dealing with the constructor params or even change out Slim\Route for another Route library all-together.

danielgsims:
@silentworks To keep the Route unaware of Application, you could just instantiate your Routes directly and add them to $app->router manually. You could keep them in pimple.

codeguy:
So if my understanding is correct, the entire gist of this thread is how to define what arguments are injected into the controller object referenced by a string callable like "\MyController:methodName". Correct? Would it not be simpler to instantiate the controller first, and then associate it to a route like this?

```
$app->get('/foo', [$fooController, "methodName"]);
```

This allows you to instantiate the controller however you want to, and maintain a single instance for however many routes use it.

silentworks:
@codeguy The problem with doing that is I would have to instantiate all my controllers even if they aren't called.

ciaranmaher:
@codeguy The main reason that I opened this PR was to have the ability to pass $app into the controllers constructor.

Yes, you can instantiate the controller first but then, exactly as @silentworks put it, you'd have to instantiate all your controllers even if they aren't called.

codeguy:
Understood. Will probably leave this open and roll into a 3.x point release so we don't rush the implementation.

krzksz:
In my opinion this improvement is to important to be left for consideration until 3.x. 

@ciaranmaher's initial idea is just Good Enough™ since it allows us to easily access application object inside every controller without making code more complicated and involving all of the singleton and globals stuff. Every other's solution can be accomplished by already existing methods e.x. dependency injection.

4nd:
+1 for @ciaranmaher 's solution.  The 'Slim doesn't dictate how you structure your own controller classes' doesn't hold true, when it already dictates how you structure the action method. Although you can use setParams, it always passes the route matched vars, why is this any more acceptable. 

Does anyone have a problem with always passing the $app in the constructor?  You don't have to do anything with it?

danielgsims:
It would preferable to dictate which dependencies you want. If you want your controllers to framework agnostic, you won't want to have Slim as a dependency. It's so easy to make it customizable, why would you want to lock yourself into the default.

4nd:
Most, if not all, other frameworks I've experienced, pass parameters to the Controller constructor, how about an assumption that if you're using the shorthand '\Some\Controller:action' that you're integrating with Slim, otherwise you can use a closure to construct the controller and execute your method?

danielgsims:
You're using \Some\Controller:action' within the router. Your controller class is not aware of that. Say for example that you want to customize the Slim framework. Say you decide not to extend Slim, but to create an Adapter around it. If the controller expects you to use Slim as a dependency, you would not be able to pass your new Slim Adapter as a dependency. Those assumptions locked yourself into a corner. It's much more developer friendly to give us some way to inject dependencies into controllers without relying on Slim as a service locator to do so.

danielgsims:
Check out my solution in #822

silentworks:
Closing this off in favour of #822 which seem like a more robust implementation.

PR #758, opened 2014-02-18
always merge settings if 1st parameter is an array

pine3ree:
- avoids accidental assignments when $name in an array and num parameters > 1 
- avoids usage of magic methods

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -335,16 +335,16 @@ class Slim
      */
     public function config($name, $value = null)
     {
-        if (func_num_args() === 1) {
-            if (is_array($name)) {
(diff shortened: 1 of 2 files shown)


codeguy:
Can you provide unit tests for the new recursive merge feature?

pine3ree:
Hello,

i added a test. There seems to be an issue with travis build with php 5.5 in this test:
SlimTest::testExpiresAsString

kind regards

tuupola:
Looks like timing issue in tests.

pine3ree:
yeah, 
that test is so trivial.

codeguy:
Merged into release-2.4.3. Thanks!

PR #739, opened 2014-01-29
Only instantiate the controller class if the route is requested

ciaranmaher:
Not sure if this is a document error or a bug but according to the [2.40 release notes](https://github.com/codeguy/Slim/releases/tag/2.4.0) the app will instantiate the controller class when the route is requested.

This is not the case, presently the class is instantiated even if the route is not requested. 
e.g. 
$app->get('/hello/:name', '\Hello:sayHello');
$app->get('/bye/:name', '\GoodBye:sayBye');

If you request `/hello/Josh`, the constructor of both `\Hello` and `\GoodBye` will be triggered. 

To get around this issue I've just moved the instantiation of the class from `setCallable` to `getCallable`

PR #761, opened 2014-02-20
Improved functionality and performance when adding class:method routes

davestewart:
The current routing setup, in order to check for is_callable(), instantiates and caches class instances in Route::setCallable() when adding class:method routes. This has the negative side-effect of firing class (controller) constructors each and every time a route is added, as well as caching redundant instances.

See: http://help.slimframework.com/discussions/problems/6304-slims-classmethod-routing-syntax-instantiates-a-new-class-for-each-and-every-route-added

This PR checks for callable methods using reflection, and only instantiates a class instance on the first getCallable() call, usually from dispatch();

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ php:
   - 5.3
   - 5.4
   - 5.5
+  - hhvm
 
 script: phpunit --coverage-text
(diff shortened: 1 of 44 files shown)


erizocosmico:
I think that problem has already been adressed in #739

davestewart:
Ah... so it has, and much more elegantly... !

silentworks:
Thanks for the effort, you mind closing out this PR please.

PR #764, opened 2014-02-21
Constructor params

danielgsims:
Allows setting construction parameters for Controller Classes discussed in issue #748 and PR #755. Expands on the work of @ciaranmaher in #739

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -182,6 +182,11 @@ class Route implements RouteInterface
      */
     public function getCallable()
     {
+        //Instantiate class constructor
+        if(is_array($this->callable)) {
+            $this->callable = array(new $this->callable[0], $this->callable[1]);
+        }
+
         return $this->callable;
     }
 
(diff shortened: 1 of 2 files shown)


danielgsims:
@codeguy Any feedback Josh?

danielgsims:
#739 was closed in favor of #721. I will look at redeveloping this with #721.

4nd:
Is this still under development? I know its been discussed before, but some way of getting the App instance into the controller, aside from as a route parameter, is pretty important.

I'm happy to take a look at it if nobody else is.

danielgsims:
I was hoping for more feedback!

4nd:
How do we push this forward, am I missing something?  Is there some way in my controller to access the Slim instance, or its view without this functionality?

danielgsims:
I canned this because there were two different solutions to Lazy Loading of Controllers: #721 and #739. #721 fixes this by creating a callable closure at the time of Route construction. #739 fixes this by only creating the controller when it was called. The way #739 went about this allowed me to create a setParams method to create a list of dependencies that could be used when the controller was constructed. However, since #721 created the callable immediately, there would be no time to handle a setParams method. I'm reviewing the code now, and there doesn't seem to be an immediately identifiable solution to me with this code review. I have some ideas brewing, however they require more heavy lifting than this solution which is no longer applicable :(

danielgsims:
I created an alternative solution in #822

PR #767, opened 2014-02-23
Removed direct instantiation of Route class and added it to the di conta...

silentworks:
...iner

Added new methods to Route class in order to setCaseSensitive
Modified the setCallable to check if callable is not null before doing anything else

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -124,8 +124,12 @@ class App extends \Pimple
         });
 
         // Route
-        $this['route'] = function ($c) {
-            return new \Slim\Route();
+        $this['route_factory'] = function($app) {
+            return function($pattern, $callable) use ($app) {
+                $case = $app['settings']['routes.case_sensitive'];
+
+                return new \Slim\Route($pattern, $callable, $case);
+            };
         };
 
         // View
@@ -278,11 +282,7 @@ class App extends \Pimple
     {
         $pattern = array_shift($args);
         $callable = array_pop($args);
-        $route = $this['route'];
-
-        $route->setPattern($pattern);
-        $route->setCallable($callable);
-        $route->setCaseSensitive($this['settings']['routes.case_sensitive']);
+        $route = $this['route_factory']($pattern, $callable);
         $this['router']->map($route);
         if (count($args) > 0) {
             $route->setMiddleware($args);
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -123,6 +123,11 @@ class App extends \Pimple
             return new \Slim\Router();
         });
 
+        // Route
+        $this['route'] = $this->share(function ($c) {
+            return new \Slim\Route();
+        });
+
         // View
         $this['view'] = $this->share(function ($c) {
             $view = $c['settings']['view'];
@@ -273,13 +278,15 @@ class App extends \Pimple
     {
         $pattern = array_shift($args);
         $callable = array_pop($args);
-        $route = new \Slim\Route($pattern, $callable, $this['settings']['routes.case_sensitive']);
-        $this['router']->map($route);
+        $this['route']->setPattern($pattern);
+        $this['route']->setCallable($callable);
+        $this['route']->setCaseSensitive($this['settings']['routes.case_sensitive']);
+        $this['router']->map($this['route']);
         if (count($args) > 0) {
-            $route->setMiddleware($args);
+            $this['route']->setMiddleware($args);
         }
 
-        return $route;
+        return $this['route'];
     }
 
     /**
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -124,9 +124,9 @@ class App extends \Pimple
         });
 
         // Route
-        $this['route'] = $this->share(function ($c) {
+        $this['route'] = function ($c) {
             return new \Slim\Route();
-        });
+        };
 
         // View
         $this['view'] = $this->share(function ($c) {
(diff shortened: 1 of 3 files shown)


silentworks:
@designermonkey all tests that require a route failed when changed to your suggestion.

designermonkey:
@silentworks I'll have a peek later at your branch, as that is how the Pimple container works. Something must be wrong somewhere else.

designermonkey:
I made a PR on this branch for you @silentworks

silentworks:
@designermonkey thanks for that, my oversight. Forgot that I was calling a new instance each time when I changed the code to your example. Great stuff, learning more each day.

andrewmman:
guys, why would you need to move the Route to the container? Route it's not really a service class and by calling $this['route'] you're already instantiating the route, so it's basically the same.

The Route $pattern and $callable constructor parameters shouldn't be optional parameters.

silentworks:
@andrewmman How else would you change out the Route object if you extended it? The only other way would be that you would have to extend Slim\App itself then change the route mapping at that point.

silentworks:
The other option would be to move the mapRoute method into the router instead then we wouldn't need this as the Router itself would be interchangable. But as this is not the case this is the only option I can think of.

andrewmman:
@silentworks , you could declare a factory function in Application:

```
$this['route_factory'] = function($app) {
    return function($pattern, $callable) use ($app) {
        $case = $app['settings']['routes.case_sensitive'];

        return new Slim\Route($pattern, $callable, $case);
    };
};
```

and in mapRoute do:

```
    protected function mapRoute($args)
    {
        $pattern = array_shift($args);
        $callable = array_pop($args);
        $route = $this['route_factory']($pattern, $callable);
        $this['router']->map($route);
        if (count($args) > 0) {
            $route->setMiddleware($args);
        }

        return $route;
    }
```

codeguy:
+1 for using a route factory method suggested by @andrewmman.

silentworks:
@codeguy will change the PR to use @andrewmman suggested factory method.

andrewmman:
@silentworks mind checking this [branch](https://github.com/andrewmman/Slim/tree/patch/routing)?

Basically if the callable is a string it creates a callback that will resolve an instance from a defined service in the container or a controller class when the route is dispatched, but it does not support automatic resolution of dependencies for controllers.

silentworks:
@andrewmman this looks good, is there a reason why you didn't push the new service back into the container stack? I am just thinking if I have the same dependency on more than one route callable then it would create a new one each time if I don't manually add it to the container stack.

andrewmman:
it just instantiates a controller class, one without dependencies, so didn't see the need to. The Closure that resolves the service or controller is called once per request.

Every route that is defined using a string as the callable, the string will be converted to that function. Once the route is dispatched the instance will be resolved.

andrewmman:
@codeguy, @silentworks would you guys prefer to have the RouteFactory as a dependency of Router?

App::mapRoute would then be:

```
    protected function mapRoute($args)
    {
        $pattern = array_shift($args);
        $callable = array_pop($args);

        return $this['router']->makeRoute($pattern, $callable, $args);
    }
```

and Router::makeRoute:

```
    public function makeRoute($pattern, $callable, $middlewares = array())
    {
        $route = $this->factory->make($pattern, $callable);

        $this->map($route);

        if (count($middlewares) > 0) {
            $route->setMiddleware($middlewares);
        }

        return $route;
    }
```

designermonkey:
Why does there need to be _another_ class to achieve all this?

silentworks:
@andrewmman it needs to remain separate, this was done sometime ago to not couple the router and route class #452 , hence why its now in the Slim\App class.

andrewmman:
@designermonkey mainly because it has the App as a dependency.

@silentworks I made a [change](https://github.com/andrewmman/Slim/commit/cdb9a07ea3282830c16a449f1078c7c87f60fc5b), so the factory just calls the Route resolver.

If you want to use a custom route, you just change an option and the resolver closure in the container.

andrewmman:
An alternative could be to use the 'slim.before.dispatch' hook to modify the matched routes callable to the controller instance, but one would need the route to be passed as an argument to the hook.

andrewmman:
Made a new [branch](https://github.com/andrewmman/Slim/tree/patch/routing-hook) to show the alternative.

codeguy:
My thoughts so far:
- I agree that we need to use a factory (closure) to generate Route objects. This will let a developer easily swap the Route class with another that implements the same Route interface.
- I do not think it is a big deal if the route factory exists in \Slim\App or in \Slim\Router. The result is the same, just a different place to override the route factory implementation.
- String to callable functionality has already been merged recently into the develop branch, so be sure you accommodate that existing code with any of these changes.

codeguy:
After some thought I'm leaning toward moving the route factory into the \Route class and have a `setFactory(Closure $factory)` method. We can then use a static `\Route::create()` method that will return the result of the factory.

andrewmman:
@codeguy, @silentworks I believe the real issue is allowing the developer to declare how they instantiate the controller, some people want to pass the App or be able to resolve their controllers.

The merged P.R that resolves the callable still presents this issue, how a controller is instantiated shouldn't be the reason to modify the Route class.

I think it would be best if Slim presented a flexible solution, so controllers could be developed in a separate package, declaring an interface, allowing middlewares, etc. It doesn't need to be in the core, a lot of people use Slim without them.

silentworks:
@andrewmman after some thought, I came to the same conclusion.  Although I think you are typing this comment in the wrong issue. But yeah, this should be a separate project and not included in the core of Slim.

lalop:
Sorry for the basic question but can a middleware do that ?

andrewmman:
First of all, @silentworks sorry for hijacking the P.R. I made the previous comment taking into consideration some P.R and open issues related to controllers. Maybe it's best to open a new issue to discuss it.

@lalop I mentioned it just as an example of what could be done.

PR #774, opened 2014-03-03
File Slim/Http/Util.php

rndm2:
Cookie with '=' in value no more ignored by getCookie

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -402,9 +402,10 @@ class Util
         $headerPieces = preg_split('@\s*[;,]\s*@', $header);
         foreach ($headerPieces as $c) {
             $cParts = explode('=', $c);
-            if (count($cParts) === 2) {
+            if (count($cParts) >= 2) {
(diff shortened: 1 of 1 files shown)


Slim/Http/Util.php
```diff
@@ -402,9 +402,10 @@ public static function parseCookieHeader($header)
         $headerPieces = preg_split('@\s*[;,]\s*@', $header);
         foreach ($headerPieces as $c) {
             $cParts = explode('=', $c);
```
mcuelenaere, line 404: Why not just change this to `$cParts = explode('=', $c, 2);` ?

codeguy, line 404: :+1: This is a cleaner solution. @rndm2 Do you want to make this change?

rndm2, line 404: Sorry, I have no time now.

codeguy, line 404: No worries. I'll do this and link to your PR for historical reference.

PR #784, opened 2014-03-08
Load Groups on Route Match

pwhelan:
This PR essentially loads the group once a match has been made on the base of the route. This allows loading separate views for a group as well as optimizing page load by loading only the callbacks necessary to fulfill the route.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -384,11 +384,7 @@ class App extends \Pimple
         $args = func_get_args();
         $pattern = array_shift($args);
         $callable = array_pop($args);
-        $this['router']->pushGroup($pattern, $args);
-        if (is_callable($callable)) {
-            call_user_func($callable);
-        }
-        $this['router']->popGroup();
+        $this['router']->addGroup($pattern, $callable, $args);
     }
 
     /**
--- a/Slim/Interfaces/RouterInterface.php
+++ b/Slim/Interfaces/RouterInterface.php
@@ -49,9 +49,7 @@ interface RouterInterface
 
     public function map(RouteInterface $route);
 
-    public function pushGroup($group, $middleware = array());
-
-    public function popGroup();
+    public function addGroup($group, $callable, $middleware = array());
 
     public function urlFor($name, $params = array());
 
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -212,10 +212,10 @@ class Router implements RouterInterface
     {
         return array_push(
             $this->routeGroups, 
-            [$group => (object)[
+            array($group => (object)array(
                 'routes'        => $routes,
                 'middleware'    => $middleware
-            ]]
+            ))
         );
     }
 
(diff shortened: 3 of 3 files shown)


pwhelan:
All the tests passed but I'm still wondering what functionality this change will affect, such as named routes, etc...

codeguy:
Cool. I'd like to test this thoroughly to avoid any inadvertent breaking changes. This will probably find its way into a 3.x point release.

gmanricks:
It won't pickup named routes defined inside groups like @pwhelan mentioned, (which may be a test case we should add in-general). I do like the idea though, but to get around this we can maybe convert this into a lazily loaded property (both the routes name, and path).

Add we can allow for groups to be named, and that can be the base name, so for example if the group has a name of 'foo' and the route was named 'bar' you can access it with `$app->urlFor('foo_bar');`

We should probably break Group into it's own class, and have a `loaded` boolean property, and we can then cache the results on the first load.

This would allow us to move some of these concerns out of the Router object into the groups object. From the router object we should only have to call: `$group->matchesUri` and `$group->matchesPath` in which the group would perform these checks using @pwhelan algorithm for matching bases. Once loaded, (either for path or name) it would cache all the properties for the routes in the group object

codeguy:
I'd prefer to avoid another class.

pwhelan:
The algorithm I used for groups in itself could stand to be improved. I tried to rock the boat as little as possible there.

One of the bigger problems with fully resolving groups on demand will be inadvertently executing code that is meant solely for that group. This was actually the behaviour which I had expected when adding a group which I sought to support, ie:

```
$app->group('/rest', function() use ($app) {

    $app->view(new \JsonApiView);
    $app->add(new \JsonApiMiddleware());

    $app->get('/people', function() use ($app) {
        $app->render(200, [ /* a bunch of people .... */ ]);
    });
});

$app->group('/web', function() use ($app) {

    $app->view(new Slim\Views\Twig);

        $app->render('template.twig', [ /* some data... */ ]);
    });
});
```

Upon executing a group to fully resolve it's routes and get their names it would also unintentionally set the view. This is what happens currently. 

The complicated solution to this is to use a separate instance of $app to execute the groups. This would not avoid all possible side effects but it would avoid most side effects to Slim itself. I know how to set $this for a closure and how to pass arguments but I have not seen how to manipulate it's scope set with the use keyword before executing or invoking a Closure.

Adding another group for Router groups to me seems excessive. Waiting for a cache to be filled to use named routes on groups also seems dysfunctional to me (using a cache is a good idea but should be optional).

pwhelan:
Another solution would be to add a function we can call to check if the group is being routed or not. That way initialization can be executed conditionally, ie:

``` php
    $app->group('/rest', function() use ($app) {

        $app->routed(function() {
                $app->view(new \JsonApiView);
                $app->add(new \JsonApiMiddleware());
        });
    });
```

I'm not married to the actual syntax or even the use of the callback. Any thoughts?

smmoosavi:
@gmanricks `$app->urlFor('foo:bar');` will be better than `$app->urlFor('foo_bar');`

codeguy:
I think this is getting too complicated for the sake of performance. Closing.

PR #792, opened 2014-03-14
Set the flash variable in the view

StJudeWasHere:
ArrayAccess was already built into the Flash object so it made sense to just add it to the view.


codeguy:
Works for me.

PR #795, opened 2014-03-17
Automatic json_encode when header is set

HelloChrisBrown:
I mainly use Slim for backend api calls and find that I have to constantly json_encode my data. This would allow for automatic encoding if the header is set. I'm open for better ways for comparison. I figured checking for a string was the best performance wise.

Example use (very basic):

```
$app->response->headers->set('Content-Type', 'application/json');
$result = array("key" => "value");
$app->response->setBody($result);
```


HelloChrisBrown:
Sorry. Realized I have the wrong branch

PR #796, opened 2014-03-18
Fix race condition in testExpiresAsString

fredemmott:
This race condition is common when running the tests with HHVM + JIT,
due to JIT overhead when not warmed up.


codeguy:
Can you send to the release-2.4.3 branch?

PR #798, opened 2014-03-19
Implement addHeader and addHeaders

StJudeWasHere:
Implement the addHeader and AddHeaders in the Http\Request and Http\Response classes.

According to the proposed FIG http message interface, the Http\Header class now uses arrays to store the headers and returns a string on getHeader() if the $asArray parameter is false or not set.


StJudeWasHere:
Added a second commit as Travis wasn't passing.

codeguy:
Thanks!

PR #801, opened 2014-03-30
Correct the config handler

designermonkey:
Sorry for taking so long to redo this PR.


codeguy:
Looks like this breaks unit tests. Can you have a look?

designermonkey:
Will do.

designermonkey:
Sorry, still getting used to remembering about tests.

codeguy:
Looking

PR #808, opened 2014-04-09
update guzzle streams

StJudeWasHere:
Update guzzle streams to guzzlehttp/streams 1.*


asppsa:
The new Guzzle requires PHP 5.4.  Is this break with 5.3 compatibility intentional?

PR #816, opened 2014-04-22
use php://input if body is null

StJudeWasHere:
Use php://input if the body is null as [noted](https://github.com/codeguy/Slim/commit/1c4a990dc834728e1ad5f2d597071d9700504e7d#commitcomment-6083981) by @svemir. This was removed by mistake in the guzzle update pull request.

Some tests where added to make sure the read-only stream is used when the body is null.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -127,10 +127,12 @@ class Request implements RequestInterface
         $this->env = $env;
         $this->headers = $headers;
         $this->cookies = $cookies;
-        $this->bodyRaw = new \GuzzleHttp\Stream\Stream(fopen('php://temp', 'r+'));
 
         if (is_string($body) === true) {
+            $this->bodyRaw = new \GuzzleHttp\Stream\Stream(fopen('php://temp', 'r+'));
             $this->bodyRaw->write($body);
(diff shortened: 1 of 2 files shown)


svemir:
Thanks! Travis seems to fail due to unrelated issues right now. I pulled this locally and all the tests pass.

silentworks:
@codeguy can we look at getting this back in please, as currently without this PR develop branch is broken when trying to get data from Slim\Http\Request.

And can you also rerun this build as it seem to have failed for no reason to do with this PR, but more to do with composer issues.

ffflabs:
If you commit on top of an existing pull request the travis test will be ran again. 

It seems that Guzzle requires PHP 5.4+, so Slim whould have to amend its travis.yml to reflect the same. (or add some kind of shenanigans to use a former Guzzle version for 5.3 environments).

The 5.4 and 5.5 builds failed due to a 403 error in pimple's repo. Should have been just a lapsus.

silentworks:
@mnlg can you update your branch with latest changes from develop please as PHP 5.3 has been removed from the travis.yml file. I want to merge this in today, but can only do so once you have a made this change.

PR #821, opened 2014-05-03
Update Slim.php

Concentum:
Why not give the opportunity to do so:
$app->render(array('header.php',  '???.php',  'footer.php'), ... etc


silentworks:
This is something your would do in a template parser or create a layout engine to manage this kind of setup, I don't think this belongs inside the core of Slim.

PR #822, opened 2014-05-05
Callable Resolver

danielgsims:
I moved the responsibility of creating the callable to a new dependency called a Callable Resolver. In the App constructor, I created a default Resolver that creates route callable methods the exact way they are created now. In addition, I created two more Resolvers. One by default will inject the Slim Pimple container into the constructor. The other will look for services inside of Pimple. For example, if you say "mycontroller:index", it will look for $app['mycontroller'] and call the index method on that controller. There is an interface to create your own Resolver.

To implement a resolver, you just change the $app['resolver'] class prior to adding routes.

I have some questions in this which I would like to discuss. For one, I made Pimple a dependency, but maybe it should be the Slim application itself. Also, I'm curious on naming. Third, I haven't written tests/doc for this but all current tests pass. I wanted to chat about this rough draft and nail it out

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -24,7 +24,6 @@ use Pimple\ServiceProviderInterface;
  */
 class App extends \Pimple\Container
 {
-    use ResolveCallable;
     use MiddlewareAware;
 
     /**
@@ -162,6 +161,14 @@ class App extends \Pimple\Container
         $this['notAllowedHandler'] = function ($c) {
             return new Handlers\NotAllowed;
         };
+
+        /**
+         * This Pimple service MUST return a SHARED instance
+         * of \Slim\Interfaces\RouterInterface.
+         */
+        $this['resolver'] = function($c) {
+            return new Resolvers\DependencyResolver($c);
+        };
     }
 
     /********************************************************************************
@@ -261,7 +268,8 @@ class App extends \Pimple\Container
             throw new \InvalidArgumentException('Route pattern must be a string');
         }
 
-        $callable = $this->resolveCallable($callable);
+        $callable = $this['resolver']->build($callable);
+
         if ($callable instanceof \Closure) {
             $callable = $callable->bindTo($this);
         }
--- a/Slim/Interfaces/CallableResolverInterface.php
+++ b/Slim/Interfaces/CallableResolverInterface.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Slim\Interfaces;
+
+interface CallableResolverInterface
+{
+    public function build($callable); 
+}
--- a/Slim/Resolvers/DependencyResolver.php
+++ b/Slim/Resolvers/DependencyResolver.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Slim\Resolvers;
+
+use Slim\Interfaces\CallableResolverInterface;
+
+class DependencyResolver implements CallableResolverInterface
+{
+    private $container;
+
+    public function __construct(\Pimple\Container $container)
+    {
+        $this->container = $container;
+    }
+
+    public function build($callable)
+    {
+        if (is_string($callable) && preg_match('!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!', $callable, $matches)) {
+
+            $class = $matches[1];
+            $method = $matches[2];
+
+            $container = $this->container;
+
+            $callable = function() use ($class, $method, $container) {
+                if (isset($container[$class])) {
+                    $obj = $container[$class];
+                } else {
+                    if (!class_exists($class)) {
+                        throw new \RuntimeException('Route callable class does not exist');
+                    }
+                    $obj = new $class;
+                }
+
+                if (!method_exists($obj, $method)) {
+                    throw new \RuntimeException('Route callable method does not exist');
+                }
+
+                return call_user_func_array(array($obj, $method), func_get_args());
+            };
+        }
+
+        return $callable;
+    }
+}


danielgsims:
Test failure was due to testing an invalid callable and throwing an InvalidArgumentException. That was something I would like to ask about. That responsibility would now be a part of the Resolver. With the Route, do we want to type hint callable for setCallable.

public function setCallable(callable $callable)...

Or, do we want to keep the InvalidArgumentCheck in place when !is_callable($callable).

danielgsims:
Here is a gist using the Container Resolver:

https://gist.github.com/danielgsims/b9e5ad85cf30a5c5ece7

danielgsims:
And the Dependency Resolver!

https://gist.github.com/danielgsims/84efe4c10ac97c56b18c

designermonkey:
+1 from me on how to solve the dependencies for controller based routing. It's not too complicated, and even though it defines a restriction for developers to some degree, it is one that works efficiently.

danielgsims:
@designermonkey Do you have any input on naming this thing? Do you like "Resolver"?

designermonkey:
Yeah resolver seems rine to me, as that's what it's doing.

silentworks:
This is a great solution I think, I have currently implemented this as a separate dependency for a project I am working on and its working nicely so far.

danielgsims:
@silentworks That's great to hear. I will implement the pattern for the current version of Slim too when we get feedback from @codeguy

silentworks:
@danielgsims Can you fix the current issue that is causing this PR to not be ready for merge, @codeguy will probably take a look at it soon and if its ready for PR and he is happy with it, he can merge it in easily.

danielgsims:
Yeah absolutely. I'll get it in over the next couple of days

danielgsims:
@silentworks @codeguy there we go guys

ecoreng:
I know I am butting in here, but I did something to solve this problem without knowing there was already a solution waiting for feedback. I created two traits to help me inject dependencies from the container (Pimple) into my controllers without having to inject the whole container/app:

[removed link]

Wouldn't something like this work? It would certainly allow the Controller Author to declare it's own dependencies from the container Inside the controller itself. It works by doing setter injection, which is a little bit different from constructor injection.. so it's not as "pretty".

The problem I see with my own solution is when someone requires to inject the App itself instead of the router, response or any other service defined inside the container.

silentworks:
Sorry folks, I am still waiting on feedback from @codeguy on this. He has been real busy lately but should be back on PR's at the end of this month. 

@ecoreng your solution looks like a good one but I'm not too sure if it keeps the simplicity of Slim in its execution.

danielgsims:
@ecoreng It's really the difference between Dependency Injection and Service Location. Your solution doesn't actually allow you to inject dependencies, it just gives you access to a service locator through Pimple. Unfortunately, even with a service locator your Controller still has the responsibility of getting it's own dependencies. It seems like it might be splitting hairs, but by actually injecting dependencies, you get the benefits of typehinting and everything that comes with it, polymorphism etc. Also, it's easier to unit test!

In this PR, I made it so that you can customize how you inject things. You can inject Pimple and use a service locator, or you can use Pimple as a dependency injection container.

scottux:
:+1:

ecoreng:
@danielgsims, you are right. Thanks for the explanation, makes sense.

danielgsims:
@silentworks - Pulled in latest changes so tests pass. Have you heard any feed back yet. :)

dac514:
Hi, I'm coming in from Issue #977 

The literature I have read hints that passing the entire Pimple class into your constructor is an anti-pattern. I am talking about the example in Container Resolver posted on May 5, 2014

```
    protected $app;
    public function __construct(\Pimple $app)
    {
        $this->app = $app;
    }
```

Some sources:: 
- https://github.com/silexphp/Pimple/issues/71 _"Never inject the container."_
- http://www.mendoweb.be/blog/dependency-injection-service-locator-factory-constructor-injection-dic/
- http://www.devtrends.co.uk/blog/how-not-to-do-dependency-injection-the-static-or-singleton-container

So not sure about that one... It's easy and solves the problem but it's not the best.

I am into "Dependency Resolver" though. That makes sense. +1 for that part of the feature.

Hope this helps.

danielgsims:
Thanks @connerbw. It's better to not inject the container, but I think many users are going to want to do it. I would absolutely suggest that everyone uses the Dependency Resolver. It helps your application be tested more easily.

silentworks:
@danielgsims sorry for the cat and mouse chase, can you update this again, we will be taking a look at this very soon as we now have the psr7 branch merged into develop.

danielgsims:
Hey guys, I've been swamped but I'm going to pick this back up. Give me a day or two to get this resolved.

codeguy:
Assigning @akrabat to work on integrating this with his existing callable resolver stuff.

designermonkey:
@akrabat I took the concept from here and embellished it into one class here http://github.com/usebeagle/callable-resolver

Please feel free to pull it apart as it was based on this code, and it won't be used as it is there.

danielgsims:
Okay guys, here's the update. I think it would be best to ditch the Resolve trait and stick with a class that we can overwrite if we want to change how route resolving works. I ditched two of the Resolvers that I wrote because that functionality was deprecated and replaced in trait code. I edited the final Resolver to use the majority of the logic from the Resolve trait. I was able to pull the trait successfully out of App and it's using the Resolver class fine. The trait is still being used in Route however when adding middleware.

danielgsims:
My main complaint, and something I tried to avoid in this initial proposal, is that it's shouldn't be the Route's responsibility to figure out how to interpret something into a callable. It should just take a callable.  Something else inside of App or the Router should know how to make it happen.

designermonkey:
I agree. A callable should be passed through, something like

``` php
$app->get('/', $app->resolve('MyClass:method'));
```

akrabat:
This is looking good so far @danielgsims.

The concern about which objects become aware of the DIC has has come up before. See the discussion on https://github.com/slimphp/Slim/pull/1112

From my point of view, the main things that the solution needs to achieve are:
1. Lazy retrieval of the class from Pimple
2. Consistent API between setting the dispatched callable and the route middleware callable.

As long as we satisfy these, then we'd like the cleanest solution possible :)

danielgsims:
@akrabat 

I created a new branch that makes the Route unaware of the container. I took the Resolver functionality out of App but now, the Router needs an instance of the ResolverInterface. The good news is the Router isn't aware of the container either because it's encapsulated behind the Resolver, and if someone really wanted to get away from Pimple, they could create their own Resolver. And again, if someone wants to change the functionality at all, they would just up something with the ResolverInterface.

My only concern with this method is changing the constructor signature for Router

Here's the diff:

https://github.com/slimphp/Slim/compare/slimphp:develop...danielgsims:RouterResolver?expand=1

lalop:
Hy @danielgsims ,
Maybe We can add the resolver to the router after its initialization.

```
$router = new Router();
$router->addResolver($c['resolver']);
```

The simple router only handle callable, then anyone ( libraries, middleware ...) can add its own resolver.
When the Router has to resolve a callable it send the param to each resolver until it get a callable.
For sure we can add the default resolver to handle `Class:method`.

What do you think ?

danielgsims:
Hey @lalop.

I'm not too crazy about making the resolver an aggregate by default. I think it might be confusing. What if two resolvers could resolve the same string differently? It could be a mess. 

We could inject the resolver with a setter though. It would make the code in the Router slightly more complex though, but nothing too crazy.

```
public function map($methods, $pattern, $handler)
    {
        if (!is_callable($handler)) {

            if (!isset($this->resolver)) {
                throw new \Exception("Cannot process Route handler");
            }

            $handler = $this->resolver->build($handler);
        }

        //rest of map
    }
```

danielgsims:
Hey @akrabat - Did you get a chance to check out the diff here yet? https://github.com/slimphp/Slim/compare/slimphp:develop...danielgsims:RouterResolver?expand=1

akrabat:
@danielgsims 

I much prefer that the resolver is separated to its own class. However, your current diff has an inconsistent API which I dislike:

```
$app->get('/foo', 'foo:bar');
```

but

```
$route->add($app['resolver']->build('foo:bar'));
```

i.e. I'd like to see a way for the route to to know about the resolver so that:

```
$route->add('foo:bar');
```

works. 

i.e adding the callable to `get` or to a route should be as simple as possible for users of Slim.

geggleto:
I know this is a bit nit-picky, but if we are going to remove the trait; can you please remove it from the PR repo so that it shows up as removed on the diff. Helps with reading the diff and understanding what is changing without needing multiple tabs.

danielgsims:
@akrabat I see what you're saying about adding a route directly via the add method. In order to toss a string just at the route, the route has to be aware of the container. I feel like we should be straight forward. If the route needs a callable, give it a callable. I suppose we could switch gears and make the route take the resolver itself instead of the Router, but it seems like a creational responsibility, not a behavioral one, but it could be reasonable. What do you think?

akrabat:
I think that ease of use and consistency are the most important, so yes, I like the idea of the Route taking the resolver.

danielgsims:
I'll whip it up shortly.

PR #827, opened 2014-05-10
Route contexts

gmanricks:
Added the ability to bind contexts to routes. By default if no context is set the slim app object will be used allowing you to run things like `$this->render();` in your routes without pulling the `$app` variable into scope.

This PR also adds a `Slim` setting 'routes.context' which can be used to set a default context object for all routes (instead of the Slim app object).

On the route object you have two new methods `getContext()` and `setContext(object)` which can be used to get and set the routes personal context.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -51,6 +51,11 @@ class Route
     protected $callable;
 
     /**
+     * @var object The routes context
+     */
+    protected $context;
+
+    /**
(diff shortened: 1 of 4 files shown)


codeguy:
Just realized you submitted your PR to the master branch. Can you re-submit to the develop branch instead? I'll get this pulled into the 3.0 release branch.

PR #829, opened 2014-05-12
Ability to override error handler when extending \Slim\App

amcsi:
I do not like how Slim forces you to use ErrorExceptions, so I made this change to use late static binding when referencing the error handler so you can use your own if you extend \Slim\App.

Alternatively I could make a configuration option for the main \Slim\App class to just not set the error handler and back in `run()`.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1042,7 +1042,7 @@ class App extends \Pimple
      */
     public function run()
     {
-        set_error_handler(array('\Slim\App', 'handleErrors'));
+        set_error_handler(array(get_called_class(), 'handleErrors'));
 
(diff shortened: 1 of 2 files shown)


mattzuba:
Slim doesn't force you to use it's built in error handler - the beauty of middleware is that you can set your own if you want.  Create a middleware class called 'ErrorHandler' and in it's call method, call set_error_handler to set your own.

amcsi:
Oh OK!

akrabat:
In addition to creating your own error handler, you can now completely disable Slim's error handling by removing the error handler from the container:

```
unset($app->getContainer()['errorHandler']);
```

Issue #846, opened 2014-06-22
logger messages miss array typehint


hakre:
as per https://github.com/codeguy/Slim/commit/61518e81#commitcomment-6756678 the **Log** implentation can follow psr-3 better with an array typehint for `$context`.

PR #847, opened 2014-06-22
Modified Slim/Log to be more PSR-3 Compliant

hakre:
Seems like in 61518e8 it was forgotten to typehint the context parameter
which is always an array.

This issue has been removed by adding the typehint.

References:
- 61518e8
- #846
- https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#13-context
- https://github.com/codeguy/Slim/issues/846
- https://github.com/codeguy/Slim/commit/61518e81#commitcomment-6756678


hakre:
Nother never made it. This once neither.

Issue #840, opened 2014-06-07
log.writer non-string(able) values.


guice:
Per the documentation:
http://docs.slimframework.com/#Logging

> Each log object method accepts one mixed argument. The argument is usually a string, but the argument **can be anything**. The log object will pass the argument to its log writer. It is the log writer’s responsibility to write arbitrary input to the appropriate destination.

However, per the code, "mixed" parameter has to be a string or string-able object, preventing the ability to send arrays (even though a log writer, such as php-console, can support arrays). 

https://github.com/codeguy/Slim/blob/master/Slim/Log.php#L309

``` php
            $message = (string)$object;
```

Any attempt to send an array or non-stringable object results in an error.

guice:
Interesting. I didn't even realize there was a PSR specific to logger interfaces.

I agree with your assessment. The dirty fix doesn't work very well--key issue being when context has data, the interpolate replace blows up with (again) non-string data.

I would add one more patch in checking "is_array()" on $context before assuming array. Php-Console--the logger I was using--and old ZF1 (ZF2? Maybe has changed now), supplied their messages as: ([mixed] $message, [string] $label = null). Supplying a string into $context will cause interpolate call to fail.

bensquire:
This is also causing me some issues as I'm trying to pass an array into my custom LogWriter...

hakre:
That shouldn't be an issue, array is converted to string in PHP as "`Array`". Perhaps not what you want or expect, but that's how that code works. I remove my other comments as this Slim seems dead and I don't wanna put any hurdles.

bensquire:
This line means I cannot pass an array into my custom log writer or am I missing something?

```
https://github.com/codeguy/Slim/blob/master/Slim/Log.php#L309
```

Also (and this may be a stupid question) but what do you mean by "this Slim seems dead"

Thanks :)

tuupola:
Not directly related, but I have been using [Monolog](https://github.com/Seldaek/monolog) which is PSR-3 compliant with Slim for a while. Something like:

``` php
/* Setup Slim */
$app = new \Slim\Slim();

/* Setup Monolog */
use Monolog\Logger;
use Monolog\Handler\RotatingFileHandler;
use Monolog\Handler\PushoverHandler;
use Monolog\Formatter\LineFormatter;

$log = new Logger("slim");
$formatter = new LineFormatter("[%datetime%] [%level_name%]: %message%\n");
[...]

PR #848, opened 2014-06-22
Patch 840

hakre:
see #840

this fixes additional  minor issues like additional docblock comments and an unused variable in `LogWriter::write()`.


lrlopez:
I think Josh usually prefers PRs that merge into develop...

hakre:
yeah, and I prefer traction so here goes the delete branch button.

PR #849, opened 2014-06-22
Switch to composer autoloader for Phpunit

hakre:
> _Note:_ This is the leftover from #846 which turned out to be in the wrong branch

Autoloading can be fully delegated to composer and the bootstrap file
for testing can be replaced by it.

This only needs to remove one test that already looks superfluous as it
tests the Slim autoloader  but composers autoloader is used and we can
expect that the composer autoloader does work. And if not, it's not an
issue in Slim.

Thanks to that test removal, the class Foo can be removed as well.

Additionally the php ini setting from bootstrap.php to disable session
cookies as otherwise headers would be sent has been moved into the
phpunit configuration.


hakre:
this has lived its time.

PR #851, opened 2014-06-25
fix function log from Slim/Log.php

alexandrubagu:
Now you can print mixed variables: $app->log->debug(mixed )

$response = array(  "data" => "value" );
$app->log->debug($response);

and also:

$my_obj = new stdClass;
$my_obj->fix = 1;
$my_obj->email = "contact@alexandrubagu.info";
$app->log->debug($my_obj);

### DIFF
--- a/Slim/Log.php
+++ b/Slim/Log.php
@@ -307,7 +307,7 @@ class Log
             throw new \InvalidArgumentException('Invalid log level supplied to function');
         } else if ($this->enabled && $this->writer && $level <= $this->level) {
             if (is_array($object) || (is_object($object) && !method_exists($object, "__toString"))) {
-                $message = print_r($object,1);
+                $message = print_r($object, true);
             } else {
                 $message = (string) $object;
             }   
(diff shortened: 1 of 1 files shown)


danielgsims:
I see one problem with this. If you have an object, it will always return the print_r of that object. If the object can be cast a string using the __toString magic method, it won't be triggered by the print_r function.

http://us3.php.net/manual/en/function.print-r.php#111738

Of course you could always go with something like this:

``` php
if (is_array($object) || (is_object($object) && !method_exists($object, "__toString"))) {
    $message = print_r($object,1);
} else {
    $message = (string) $object;
}
```

There are a lot of other ways to cast an object to a string out there. This kind of code is an antipattern because you can't cover all of your cases. For example, what if I implement the Json Serializable interface and want to log all of my classes as JSON? I think it's preferable to go one of two paths: handle all objects the same way, or create a user definable way to handle these logging methods.

alexandrubagu:
I think your solution is the best.  I modified the Log.php file to reflect your sulution.

For other behavior I think a solution is to define a  CustomLogWriter
<?php
class CustomLogWriter {
    public function write(mixed $message) {

   }
}
$app = new \Slim\Slim(array(
    'log.writer' => new CustomLogWriter()
));

maldoinc:
I'd also recommend passing `true` to the `print_r` function as per it's signature, rather than relying on php's implicit type conversion

alexandrubagu:
I modified the print_r statement to use true instead of 1(int)

bensquire:
Is there any chance this fix can make it into a tagged release? That would be really useful :)

PR #859, opened 2014-07-17
gitignore improved

ghost:


silentworks:
This should be added to your own personal .gitignore, not the project's .gitignore.

silentworks:
I was waiting on this @tim-bezhashvyly you will note in that other issue I stated this "This issue will remain a project and team preference, some folks don't like it while some do, its hard to please everyone unfortunately." this still stands.

silentworks:
@tim-bezhashvyly maybe we should have a chat one evening on Skype or something, I think these comments wont adhere enough and it might eventually start sounding like a flame war even though we both have no intention of starting one.

silentworks:
@tim-bezhashvyly Just to clarify on the project with framework ambitions part, if you take a look at zend framework 2 they are taking the same approach. And one of the reason why others are doing this is so that each time you change dev machine you don't need to add .idea back into your `.git/info/exclude` as this doesn't move with the project, its stays locally at all times.

https://github.com/zendframework/zf2

tuupola:
What happened to #slim? 

https://github.com/codeguy/Slim/wiki/IRC-Chat

codeguy:
Use two "#"s

silentworks:
Ah even better @tuupola @tim-bezhashvyly  catch me over at ##slim on irc. Its on freenode.

PR #860, opened 2014-07-17
Escape special chars on route marching

ghost:
Trying out the project I noticed that a route declared for example as /v2.1/ matches /v2.1/, /v2a1/, /v2b1/ and so on... the dot is translated as 'any char'.
This patch fixes this kida issue.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -69,6 +69,12 @@ class Route implements RouteInterface
     protected $pattern;
 
     /**
+     * Determine is the route pattern should be escaped or not
+     * @var bool
+     */
+    protected $escapePattern;
+
+    /**
      * The route callable
      * @var mixed
      */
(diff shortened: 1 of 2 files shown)


lrlopez:
Although I find the PR useful (especially for newcomers), I find two problems with:
1. Sometimes is useful to specify a RegExp on the route (the example you provided as incorrect con be the desired behavior).
2. It's a breaking change. It could potentially break old apps relying on it.

I suggest adding a switch somewhere. One possibility is introducing a new configuration option to enable/disable the escaping or add a third parameter to `\Slim\Route` constructor (i.e. a boolean).

What do you think about it?

ghost:
I agree with you @lrlopez 
Talking in this way I prefer to add this feature to `\Slim\Route`. I don't want to force all the routes with a single/global configuration.
Could it be somethig similar to the last commit?
The next easy (I think) step is to implement all this on `Slim\Slim` so it can be applied using
`$app->get('...')`

codeguy:
Can you rebase against the latest develop branch code and resolve any conflicts?

harikt:
@codeguy I assume the user himself have deleted. Not sure whether he himself have written the same though. https://github.com/ghost

codeguy:
@harikt Ah thanks for the heads up. I'll bring in this code manually then.

harikt:
@codeguy when you see `from unknown repository` the repo will be deleted by user ;) .

codeguy:
indeed

PR #862, opened 2014-07-21
no need for is_array test

pine3ree:
set/parse always set values as array
get($key, true) always get an (maybe empty) array;

I still have a doubt about the (Collection) get method override. It passes E_STRICT but in this case the second parameter has a different meaning than the original $default value.


pine3ree:
what was i looking at?

PR #864, opened 2014-07-22
Explicit flow control for better static analysis

eridal:
## Why?

The motivation for this is to explicitly tell static analysis tools that `App::stop` (and friends) won't return flow control.
## What changed?

_No code change was required_. Only signature (aka docblocks) changes were added. All we need is those functions to explicit declare that they `@return \Exception`

> to be fair a `throw` was required to satisfy `App::halt`'s signature
> although it's guaranteed it will not be used as `App::stop` won't return flow control
## What's the benefit of this?

``` php
$app->get('/some/route', function () use ($app) {
  if (someCondition()) {
      throw $app->halt(404); // <-- explicit flow control
  }
  $app->render('some/page');
});
```

Although it's completely possible to write client code like the example shown here, static analysis tools will complain about `App::halt` not returning `Exception`.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -736,6 +736,7 @@ class App extends \Pimple
      * The thrown exception will be caught in application's `call()` method
      * and the response will be sent as is to the HTTP client.
      *
+     * @return \Exception
      * @throws \Slim\Exception\Stop
      * @api
      */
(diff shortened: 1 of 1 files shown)


eridal:
Well.. I have created a branch `explicit-flow-control` but the commit went somehow into `develop`
Please let me know if that's an issue, and I will look into fix/resend this

silentworks:
Since this is not an actual issue in client code, I am closing this one out.

PR #866, opened 2014-07-28
Each route gets a canonical name by default, i.e. a number.

disc5:
By this, it is possible to programmatically list all available routes.
Example:
$routes = $app->router()->getNamedRoutes();

For course, routes can still labeled with custom names, i.e. the canonical names can be overridden.

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -98,6 +98,11 @@ class Slim
         'slim.after' => array(array())
     );
 
+    /**
+     * @var integer Counts the number of available routes. 
+     */
+    private $routeCount = 0;
+
     /********************************************************************************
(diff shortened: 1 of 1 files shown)


silentworks:
This would be solving a problem, but I wonder if the problem should be rather solved by adding a public method to get all routes instead. Also note you can extend the router and use your own router if you wish like the example below:

``` php
$app = new \Slim\Slim();
$app->container->singleton('router', function() {
    return new MyOwnRouter();
});
```

MyOwnRouter class

``` php
class MyOwnRouter extends \Slim\Router
{
    public function getAllRoutes() 
    {
        return $this->routes;
    }
}
```

PR #870, opened 2014-08-07
Modrewrite environment php bug fix

blackmamba:
Incase of rewrite the $scriptName is null, which is causing the 'Without rewriting' case to break, as the strpos endue with an error 

 Uncaught exception 'ErrorException' with message 'strpos(): Empty needle' in  Environment.php


silentworks:
Too many changes in here to know what was actually changed. Can you send a PR with only the files you have changed, this PR seem to be including other changes too.

PR #876, opened 2014-08-13
extra params to query string params

pine3ree:
- cached urls

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -224,12 +224,6 @@ class Router implements RouterInterface
      */
     public function urlFor($name, $params = array())
     {
-        $key = md5($name . serialize($params));
-
-        if (isset($this->urls[$key])) {
-            return $this->urls[$key];
-        }
-
         if (!$this->hasNamedRoute($name)) {
             throw new \RuntimeException('Named route not found for name: ' . $name);
         }
@@ -252,8 +246,6 @@ class Router implements RouterInterface
             $url .= '?' . http_build_query($params);
         }
         
-        $this->urls[$key] = $url;
-
         return $url;
     }
 
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -210,13 +210,14 @@ class Router implements RouterInterface
 
     /**
      * Get URL for named route
-     * @param  string            $name   The name of the route
-     * @param  array             $params Associative array of URL parameter names and replacement values. Unmatched parameters will be used to build the query string.
-     * @return string                    The URL for the given route populated with provided replacement values
-     * @throws \RuntimeException         If named route not found
+     * @param  string            $name    The name of the route
+     * @param  array             $params  Associative array of URL parameter names and replacement values.
+     * @param  array             $options Associative array of query string parameters.
+     * @return string                     The URL for the given route populated with provided replacement values
+     * @throws \RuntimeException          If named route not found
      * @api
      */
-    public function urlFor($name, $params = array())
+    public function urlFor($name, $params = array(), $options = array())
     {
         if (!$this->hasNamedRoute($name)) {
             throw new \RuntimeException('Named route not found for name: ' . $name);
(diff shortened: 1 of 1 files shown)


tuupola:
What does this PR do? Commit message is not very helpful.

pine3ree:
@tuupola 

Hello. It does 2 things.

1.
it caches urls based on route name and params so if urlFor is called twice with the same arguments the result is taken from cache (memory cache for the current request cycle)

2.
it allows you to append a query string to the url generated from all the parameters that aren't matched as route parameters. es

```
$app->get('user/index(:/page)', function ($page=1) {
// implementation here
})
->name('user-list'')
->conditions(array('page' => '\d+'));

$url = $app->urlFor('user-list', array('page' => 12, 'sort' => 'firstname.desc'));
```

will give you smt like:
/user/index/12?sort=firstname.desc

kind regards

silentworks:
Can you send this as two separate PRs please, route caching as one and route parameter to query string as another.

lrlopez:
I like the idea!

tuupola:
Any tests how much faster this make `$app->urlFor()`?

pine3ree:
@tuupola 
The generated url cache is useful when you have a rich web application including portlets, areas etc.
When links are generated in different parts of the application chances are that there will be duplicate urls...(imagine a blog post page and portlets displaying latest/featured posts or an ecommerce website with product/category list displayed everywhere).
Caching them will ensure that generation for the same url will occur only once. But speed tests are meaningful only for that application as they depends on many factor like the complexity of route patterns, the number of duplicated urls, the number of total urls...etc etc..

tuupola:
I understand. Caching is only useful when when it speeds up things. Otherwise it just adds complexity. This is why I asked if there are any tests about possible speed improvements.

pine3ree:
@tuupola 
anyway, i am going to remove caching implementation from this PR as per @silentworks request.

silentworks:
Please change the phpdoc comments to match the changes you made.

silentworks:
@pine3ree Can you resend this please but add the extra params as an additional param for the urlFor method.

``` php
$app->urlFor('user-list', ['page' => 12], ['sort' => 'firstname.desc']);
```

This way we are taking in the third param as a query string, this will allow the user to set the desired expectation rather than one being imposed on them.

pine3ree:
@silentworks 
Hello,

in my opinion a third optional parameters looks a bit uglier and the purpose of the PR was also to add some sort of /auto(detection|mation)/
If a parameter they is not a route parameter then it's automatically added to the url as a query string parameter. If you change,say remove, any routing parametersin the route definition, you will get all leftover params as qs, so you will always get this (probably useful) extra information with the request, without the need to move it to the 3rd method arg.

kindly

codeguy:
We discussed this at length in IRC, and the consensus is to move query params to an optional additional argument. This way you don't get any unintentional/unexpected side-effects. The Zend Framework does it this way, as do many other frameworks.

pine3ree:
Yes, i have used zf2 url view helper a lot (that's why i tried to avoid the 3rd method argument)

questions:
1) i used the name $options as in zf2. Any better but still short name? naming things always takes me a lot of time.

2) should we type hint as array $params and $options? (http_build_query also supports objects but i don't think it makes any sense here)

3) should we add support for url fragments as in zf2 inside the $options arguments?

kindly

silentworks:
@pine3ree 

1) maybe queryParams would be more descriptive
2) yes please
3) not sure as I have no experience with zf2 and don't know what this does

Also jump on irc channel #slimphp on freenode if you want to chat more in realtime.

pine3ree:
queryParams was my 1st choice even if a little longer. 
readability and semantic are always more important  and i have just remembered i wondered what was the meaning of the $options argument in zf2.
so => I'll change it instantly

silentworks:
@pine3ree thanks for updating, but this still looks like its not aligned with the latest dev branch updates.

pine3ree:
yeah, i notice now,....it was added for the 2 dev branch and it was also splitted. I think it's better to move the code to a new PR and close this.

pine3ree:
in the dev branch there will be no need to update the urlFor proxy method in the main app class

pine3ree:
#1007

there was a travis erro report but the error is found in:
PHP Parse error:  syntax error, unexpected '.', expecting ',' or ';' in /home/travis/build/codeguy/Slim/tests/Http/BodyTest.php on line 37

the same error reported in the next PR

PR #882, opened 2014-08-26
Derace test expires as string

fredemmott:
This race condition is common when running the tests with HHVM + JIT,
due to JIT overhead when not warmed up.

Replacing codeguy/Slim#796

Github only gives me a choice between develop and master. It will not let me request a PR against the release.


tuupola:
You probably want to include only 949768bbf3773466051120c069789972e18121d9 in the PR.

fredemmott:
Heh, oops, I got my remotes muddled up. replacement PR incoming.

PR #884, opened 2014-08-26
Fix race condition in testExpiresAsString

fredemmott:
Calling time functions twice in a row isn't guaranteed to get you the
same results.


fredemmott:
Was this fixed? This does cause intermittent failures when continually testing, and means that HHVM tests my fork of Slim instead of the main repository.

PR #887, opened 2014-08-30
Delete .gitignore

tim-bezhashvyly:
The `.idea` directory stores the project settings of PHPStorm IDE. The IDE selection is a matter of personal preference of a developer and doesn't belong to the project scope. The `.idea` directory should be ignored in local `.git/info/exclude` file instead.


silentworks:
This should remain as it currently is, the user should define their own .gitignore for their project. But as for the Slim codebase and contribution we will keep this to make sure no one will commit this by mistake.

tim-bezhashvyly:
Then it will "make" sense to add `/.project` for eclipse, `/.vagrant/` and `/Vagrantfile` for Vagrant, `.DS_Store` for MacOS and who know what else people can commit by mistake.

silentworks:
Personally I think its better to try and prevent people from committing such files, I have seen many projects take this approach and I'm not saying its good because they take this approach, but its a lot easier to remove a .gitignore file from Slim from your project than it is to accidentally commit such files. I am all for adding .DS_Store and Thumb.db in the .gitignore file.

This issue will remain a project and team preference, some folks don't like it while some do, its hard to please everyone unfortunately.

Here are some other projects doing the same:

https://github.com/zendframework/ZendSkeletonApplication
https://github.com/yiisoft/yii
https://github.com/laravel/laravel
https://github.com/magento/magento2

tim-bezhashvyly:
Thanks for taking time for replying. Even though I think this approach to be wrong way to go it is still viable. Issue can be considered closed.

PR #889, opened 2014-09-02
json can also just contain null or false.

Langmans:
used json_last_error to check if json was decoded successfully.

Not sure if exceptions should be thrown however, I mean you should be able to use a HTTP DELETE for null values...


DavidePastore:
:+1:

PR #892, opened 2014-09-03
Add basic reverse proxy support.

DavidePastore:


tuupola:
Does this work for setup like following?

```
<Proxy *>
  Order deny,allow
  Allow from all
</Proxy>
ProxyPass          /foo  http://192.168.0.12/subdirectory/
ProxyPassReverse   /foo   http://192.168.0.12/subdirectory/
```

DavidePastore:
I don't know. In my case the configuration is a little different (and simpler):

```
<Proxy *>
  Order deny,allow
  Allow from all
</Proxy>
ProxyPass          / http://192.168.0.12/subdirectory/
ProxyPassReverse   / http://192.168.0.12/subdirectory/
```

Slim/Environment.php
```diff
@@ -156,6 +156,13 @@ private function __construct($settings = null)
             //HTTP request headers (retains HTTP_ prefix to match $_SERVER)
             $headers = \Slim\Http\Headers::extract($_SERVER);
             foreach ($headers as $key => $value) {
+                if($key == 'HTTP_X_FORWARDED_FOR'){
+                    //This request is from a reverse proxy, so overwrite SCRIPT_NAME
+                    $env['SCRIPT_NAME'] = '';
+                    
+                    $env['HTTP_HOST'] = $headers['HTTP_X_FORWARDED_HOST'];
+                }
+
```
msjyoo, line 165: Wouldn't there be a chance that the SCRIPT_NAME and HTTP_HOST be overwritten to the incorrect values if it comes after the HTTP_X_FORWARDED_HOST?

DavidePastore, line 165: Yes, you're right. Do you have ideas on how to fix it? :+1:

DavidePastore:
News about this? Why did you close this?

msjyoo:
What's the use case for this?

codeguy:
Yep, closed in favor of #1029 which should have a more secure implementation.

DavidePastore:
@sekjun9878 The use case of this is the one showed [here](https://github.com/slimphp/Slim/pull/892#issuecomment-54281432).

PR #893, opened 2014-09-03
Slim View extract fix; removed possibility to include arbitrary file

strngr:
Here is method `render` of class `\Slim\View`:

```
    protected function render($template, array $data = array())
    {
        // Resolve and verify template file
        $templatePathname = $this->templateDirectory . DIRECTORY_SEPARATOR . ltrim($template, DIRECTORY_SEPARATOR);
        if (!is_file($templatePathname)) {
            throw new \RuntimeException("Cannot render template `$templatePathname` because the template does not exist. Make sure your view's template directory is correct.");
        }

        // Render template with view variables into a temporary output buffer
        $this->replace($data);
        extract($this->all());
        ob_start();
        require $templatePathname;

        // Return temporary output buffer content, destroy output buffer
        return ob_get_clean();
    }

```

Pay attention to `extract`. If you pass variable `templatePathname` to template, it will overwrite path to template and arbitrary file may be included. Such behaviour if default for `extract`.

How to reproduce:

```
public function testAction() {
    $this->getApp()->render('test.php', array('templatePathname' => '/etc/passwd'));
}
```

### DIFF
--- a/Slim/View.php
+++ b/Slim/View.php
@@ -134,7 +134,7 @@ class View extends Collection implements ViewInterface
 
         // Render template with view variables into a temporary output buffer
         $this->replace($data);
-        extract($this->all());
+        extract($this->all(), EXTR_SKIP);
         ob_start();
         require $templatePathname;
 


StJudeWasHere:
I can confirm this is a bug and we should fix it. However, I'm not sure using EXTR_SKIP is the best approach here.

If I set a templatePathname variable I would expect to be able to use it in my views, instead of getting some internal value that I don't need to know about. The same goes for the $data variable, which can also be overwritten.

Maybe we could just unset this variables before using extract, that way we would avoid conflicts. What do you think?

strngr:
Your idea seems better.
I'm not sure about unsetting `$data`. It's not needed for avoiding conflicts. Btw, `$this` would be accessible in view anyway.

StJudeWasHere:
In my opinion the `$data` variable should also be unset, otherwise it'd be always set in the view. Doing something like `isset($data)` in the view would always be `true`.

Regarding `$this` I'm not sure there's an easy way to unset it without complicating the view too much. Anyway, we could take advantage of it. We could extend the view to add new functions and do something like `$this->urlFor()' or`$this->escape()`from within the template, so it may be good to leave it there.

strngr:
I'd like to say that unsetting `$data` is senseless if you want to protect data. That data anyway may be accessed in this way: `$this->all()`.
I have nothing against unsetting `$data`. It is OK for clearing scope. But this won't protect data.

codeguy:
Closed in favor of #895

PR #898, opened 2014-09-15
Include redirected items when parsing headers.

tuupola:
Undocumented Apache [annoyance](http://goo.gl/tRKHba). In some cases when using mod_rewrite _REDIRECT__ prefix is added to the headers.  This happens for example when using workaround to make HTTP Basic Authentication work together with FastCGI.

`RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization}]`

Currently Slim does not include these items when parsing the headers. Only way to access them is through _$_SERVER_ superglobal.

### DIFF
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -93,7 +93,7 @@ class Headers extends Collection implements HeadersInterface
         foreach ($environment as $key => $value) {
             $key = strtoupper($key);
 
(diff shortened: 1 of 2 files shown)


silentworks:
@tuupola this is gonna have to wait a little until Josh is back on dealing with PRs.

codeguy:
This needs to be refactored against the latest code. You can find the relevant header parsing code in `\Slim\Http\Headers::createFromEnvironment()`.

PR #902, opened 2014-09-25
Update PrettyExceptions.php

hardlynoticeable:
The str_replace function was not terminating lines in the trace because the newline was enclosed in single quotes and not double quotes.  I also formatted the returned HTML slightly and added DOCTYPE.

### DIFF
--- a/Slim/Middleware/PrettyExceptions.php
+++ b/Slim/Middleware/PrettyExceptions.php
@@ -89,7 +89,7 @@ class PrettyExceptions extends \Slim\Middleware
         $message = $exception->getMessage();
         $file = $exception->getFile();
         $line = $exception->getLine();
-        $trace = str_replace(array('#', '\n'), array('<div>#', '</div>'), $exception->getTraceAsString());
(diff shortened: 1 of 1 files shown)


tuupola:
Non patch related question. What git client are you using? I have been trying to find out which client makes the default "Update filename.txt" commit messages.

StJudeWasHere:
This was fixed in PR #879. I didn't change the doctype though.

hardlynoticeable:
I didn't.  I used Github's web based interface.  Actually I don't use
github often and I found their interface a little confusing.

Stephen J Akins
https://google.com/+StephenAkins/about

hardlynoticeable:
I had another change immediately afterwards that I may not have pushed
properly...  I added an extra newline so the last div in the trace closed
properly and I made a bunch of little formatting changes including taking
out the PRE tags as they weren't needed any longer....  Did you see those
changes also?  Just wondering.

Stephen J Akins
https://google.com/+StephenAkins/about

silentworks:
Fixed and merged fe1c215.

PR #904, opened 2014-10-09
Prevent fatal error if callable does not exist in Route->setCallable

wpmvangroesen:
Add class and method checks to setCallable to avoid fatal errors when callable does not exist.


timovanderzanden:
+1

PR #907, opened 2014-10-15
Allow multiple arguments for applyHook

wadewomersley:
Not using variadic due to PHP5.6+ only.


wadewomersley:
Here's an idea...copy and paste the right code changes and not your test code!

PR #908, opened 2014-10-15
Allow multiple arguments for applyHook, pass to callable in order given

wadewomersley:


wadewomersley:
Fixed my silly var_dump and added a test.

PR #910, opened 2014-10-16
add getParam for router, now we can get Route Param anywhere

louis-she:
I can't using Route Param IN the Route Middleware Since now I can only(?) access Route Param via the Route Callback Parameters.
This patch fixed this problem.
e.g.

``` php
$app->get('/:param1/:param2', function (/*now it's optional*/) use ($app) {
    $param1 = $app->router->getParam('param1');
    $param2 = $app->router->getParam('param2');
    $allparam = $app->router->getParam();
});
```

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -80,6 +80,12 @@ class Router implements RouterInterface
     protected $routeGroups;
 
     /**
+     * all params of the matched route
+     * @var array
+     */
+    protected $routeParams;
+
+    /**
      * Constructor
      * @api
      */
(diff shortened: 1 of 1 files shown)


silentworks:
Sorry for taking so long to get back to this but you dont need the additional method on router to do this, you can already do this like the example below:

``` php
$app->get('/:param1/:param2', function (/*now it's optional*/) use ($app) {
    $route = $app->router->getCurrentRoute();
    $param1 = $route->getParam('param1');
    $param2 = $route->getParam('param2');
    $allparam = $route->getParams();
});
```

PR #914, opened 2014-10-19
batch config

elibyy:
added batch config option using php/ini/json/xml file

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -67,6 +67,17 @@ class Slim
     protected static $apps = array();
 
     /**
+     * @var array
+     * @since 1.0
+     */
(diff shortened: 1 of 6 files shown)


silentworks:
This should not be in the core of Slim, you can already use third party libraries to do this as they all convert back to an PHP array in the end.

elibyy:
yes but it's still make it much easier to load system configurations or make an installation config result file

StJudeWasHere:
I'm with @silentworks too, I don't think this should go into Slim. I'd rather use a third party library when needed.

prisis:
if you like you can you my config loader, https://github.com/n-brainwave/config. I will make some small changes today so it should work with slim

PR #922, opened 2014-10-30
Error with virtual path in case a physical path is not a part of it.

andrew-forest:
Hi everyone!
I have problem with routing on my hosting where it is not possible to set "DocumentRoot" to 'public' subfolder (in my project 'index.php' placed in 'public' directory). I was created a '.htaccess' file with content:

```
DirectoryIndex /public/index.php
RewriteEngine On
RewriteCond %{REQUEST_FILENAME} -f
RewriteRule ^(.+) $1 [L]
RewriteCond %{DOCUMENT_ROOT}/public%{REQUEST_URI} -f
RewriteRule ^(.+) /public/$1 [L]
RewriteCond %{DOCUMENT_ROOT}/public/index.php -f
RewriteRule ^(.*) /public/index.php
```

And I see that all my routes was broken in $env['PATH_INFO'].
For example: 'ts' instead '/contacts'.
It happened because on line 143 in 'Slim/Environment.php':

```
$env['PATH_INFO'] = substr_replace($requestUri, '', 0, strlen($physicalPath)); // <-- Remove physical path
```

first seven symbols (strlen('/public')) was removed from $env['PATH_INFO'] but it does not contain '/public'.
And I was added a logic to check if physical path is a really part of $env['PATH_INFO']:

```
if (strpos($requestUri, $physicalPath) === 0) {
    $env['PATH_INFO'] = substr($requestUri, strlen($physicalPath)); // <-- Remove physical path
} else {
    $env['PATH_INFO'] = $requestUri; // <-- Physical path is not a part of request uri
}
```

P.S. Sorry for my bad English. :)

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -140,7 +140,11 @@ class Environment implements \ArrayAccess, \IteratorAggregate
             $env['SCRIPT_NAME'] = rtrim($physicalPath, '/'); // <-- Remove trailing slashes
 
             // Virtual path
-            $env['PATH_INFO'] = substr_replace($requestUri, '', 0, strlen($physicalPath)); // <-- Remove physical path
+            if (strpos($requestUri, $physicalPath) === 0) {
+                $env['PATH_INFO'] = substr($requestUri, strlen($physicalPath)); // <-- Remove physical path
+            } else {
+                $env['PATH_INFO'] = $requestUri; // <-- Physical path is not a part of request uri
+            }
             $env['PATH_INFO'] = str_replace('?' . $queryString, '', $env['PATH_INFO']); // <-- Remove query string
             $env['PATH_INFO'] = '/' . ltrim($env['PATH_INFO'], '/'); // <-- Ensure leading slash
 


ctsstc:
I have this exact some problem and tracked it down the the same file and location

For example I'm using Nginx with:

```
location /staging/insomnia/ {
    try_files /index.php?$args /staging/insomnia/web/index.php?$args;
}

location ~ ^/staging/insomnia/(css|images|img|js|svg|fonts)/(.*) {
    try_files /staging/insomnia/web/assets/$1/$2 =404;
}
```

I also have a public folder which is

> /var/www/phpliteframework.tk/web/staging/insomnia/web/

and a private folder with slim, includes, etc

> /var/www/phpliteframework.tk/web/staging/insomnia/private/

So the site loads up via:
/staging/insomnia/

Runtime Hitting Route: /staging/insomnia/memoryUse/
Produces the following variables in the Environment //Virtual Path

```
$requestUri =>    "/staging/insomnia/memoryUse/"
$physicalPath => "/staging/insomnia/web"
$env['PATH_INFO'] => "/oryUse/"
```

The patch andrew-forest provides does fix this issue, but I'd still like an option to set the base path during Slim initialization which would override the $physicalPath, that way I could pass in "/staging/insomnia/" for my case, which would return "memoryUse" for the path_info/resourceUri.

With andrew-forest's patch all my routes then require: "/staging/insomnia/" before all the routes

**Edit**
I'm sure the whole environment variables can be overridden it looks like, but all I want to do is change that single variable, rather than write the whole environment myself.  Maybe a middleware could be combined with andrew's fix, but I'd still rather not if there's a way around this, and I'd rather not make a global variable like $prefixRoot='/staging/insomnia/' that would be prefixed to every route.

**Edit2**
What about rewriting the constructor to act like the mock function how it array_merge's the passed settings into the pre-built, this way only environment variables that want to be overwritten can. Toss out the if check in the constructor, default the parameter to $settings=array() then toss in the array merge at the end.

codeguy:
Pulled in #674 which fixes this issue.

PR #930, opened 2014-11-10
How to run Slim on Google App Engine/Cloud Platform

formigone:
Not only do you need a default script handler that points to your index.php file, but you will also need to manually populate the PATH_INFO attribute of Slim's Environment object. Additionally, App Engine doesn't set $_SERVER['SERVER_PORT'], which Slim doesn't use directly, but attempts to set to its Environment object. If you don't manually set SERVER_PORT on the $_SERVER object, GAE will fire a warning every time Slim loads up.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -132,6 +132,31 @@ Ensure the `Web.config` and `index.php` files are in the same public-accessible 
         </system.webServer>
     </configuration>
 
+#### Google App Engine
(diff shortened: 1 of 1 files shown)


prisis:
:+1:

silentworks:
Thanks for this, will get this merged in now.

PR #931, opened 2014-11-11
Take the global Closure of the type hinting View::keep method.

### LINKED ISSUE #891: View->keep() is unnecessary and doesn't work
- The type hint is not properly namespaced (as `\Closure` or via `use`)
- The method tries to delegate to `Helper\Set->keep()`, which is not defined
- It's not needed, since the data is passed to the template via `Helper\Set->all()` which doesn't unwrap closures

So I'd suggest removing it altogether.

m3y:
fixed #891


prisis:
can you rebase you changes?

m3y:
Sorry, I would be wrong pull-request destination branch.

PR #933, opened 2014-11-13
Route file

josephscott:
Another very easy method for routing would be to allow for routes to map directly to an individual file.  This diff modifies Slim\Route to support that.

Changes were minimal, just adding to the `is_callable()` logic.

The route file will have access to the `$route_params` variable.  It can also access `$this`, which is the Slim\Route object ( so `$this->params` also works ).  No other variables are in scope for the required file.

An example of this in action would be: `$app->get( '/hello(/:name)', __DIR__ . '/routes/hello.php' );`


prisis:
Can you please rebase your changes an make a new pull request?

silentworks:
I think we have multiple cleaner ways of doing routes already, what you have proposed would only confuse users. I am going to reject it on that basis.

Issue #924, opened 2014-11-02
Deprecated methods used in framework's source code


federicojasson:
The framework's source code uses methods that are marked as deprecated.

---

How did I stumble into this? (Just an example of why this can become a problem)

In Slim.php, in the halt method (https://github.com/codeguy/Slim/blob/master/Slim/Slim.php#L1036), the response body is set using response->body instead of response->setBody.

Both functions do the same thing (when response->body is called with a not-null parameter), as you can see here:
https://github.com/codeguy/Slim/blob/master/Slim/Http/Response.php#L202
https://github.com/codeguy/Slim/blob/master/Slim/Http/Response.php#L214

In my app, I coded a class that inherits from the \Slim\Http\Response class, in order to redefine the setBody method.

The problem: when I call app->halt, the response->body method is used, so the redefined method is never invoked.
The (ugly) solution: redefine response->body in order to achieve the goal. This is ugly because it is supposed to be a deprecated method.

---

Reviewing the source code, I see that this issues repeats itself with other methods (e.g. using response->status, instead of response->setStatus).

I'd suggest updating the code so that the deprecated functions aren't used in the framework's source code (leave it available only for backward-compatibility purposes).

dgmike:
:+1:

silentworks:
Thanks for pointing this out, we will get this updated soon.

kiswa:
:+1:

codeguy:
Can someone send a PR to the develop branch?

PR #935, opened 2014-11-15
Remove framework dependencies on deprecated methods

ecoreng:
Addressing this issue:
https://github.com/codeguy/Slim/issues/924

I didn't know what to do with the tests, but I guessed that we wanted to update them where they still applied, and keep the ones that test specific behavior for the deprecated function until it gets removed.

https://github.com/ecoreng/Slim/blob/fw-remove-dep-depr/tests/ViewTest.php#L144
https://github.com/ecoreng/Slim/blob/fw-remove-dep-depr/tests/ViewTest.php#L100

### DIFF
--- a/Slim/Helper/Set.php
+++ b/Slim/Helper/Set.php
@@ -140,6 +140,16 @@ class Set implements \ArrayAccess, \Countable, \IteratorAggregate
     }
 
     /**
+     * Add value as Closure with key
+     * @param string $key
+     * @param mixed $value
+     */
+    public function keep($key, \Closure $value)
+    {
+        $this->set($key, $this->protect($value));
+    }
+    
+    /**
      * Property Overloading
      */
 
--- a/Slim/Middleware/PrettyExceptions.php
+++ b/Slim/Middleware/PrettyExceptions.php
@@ -71,8 +71,8 @@ class PrettyExceptions extends \Slim\Middleware
             $env['slim.log'] = $log;
             $env['slim.log']->error($e);
             $this->app->contentType('text/html');
-            $this->app->response()->status(500);
-            $this->app->response()->body($this->renderBody($env, $e));
+            $this->app->response()->setStatus(500);
+            $this->app->response()->setBody($this->renderBody($env, $e));
         }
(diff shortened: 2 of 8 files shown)


ecoreng:
I know, but internally .. appendData uses a "replace" method directly in the data set:
https://github.com/codeguy/Slim/blob/master/Slim/View.php#L188

that same function is available as proxy here:
https://github.com/codeguy/Slim/blob/master/Slim/View.php#L129

ecoreng:
... And I had to add the method "keep" to the data set because that function is being used without existing (no tests for this?, should I add one?)

define data as \Slim\Helper\Set:
https://github.com/codeguy/Slim/blob/8760ac091c284f4e93d80d157a32d12c517084d0/Slim/View.php#L69

call data->keep:
https://github.com/codeguy/Slim/blob/8760ac091c284f4e93d80d157a32d12c517084d0/Slim/View.php#L113

but it's not available nowhere in the class:
https://github.com/codeguy/Slim/blob/8760ac091c284f4e93d80d157a32d12c517084d0/Slim/Helper/Set.php

so I added that function, copying this proxy call signature:
https://github.com/codeguy/Slim/blob/8760ac091c284f4e93d80d157a32d12c517084d0/Slim/View.php#L111

and using this built in method:
https://github.com/codeguy/Slim/blob/8760ac091c284f4e93d80d157a32d12c517084d0/Slim/Helper/Set.php#L240

ecoreng:
This last part complements:
https://github.com/codeguy/Slim/pull/934

but:
https://github.com/codeguy/Slim/issues/891

says it's not even used, however this test fails if we use $view->set('fooClosure', ...:
https://github.com/codeguy/Slim/blob/master/tests/ViewTest.php#L81

[Failed asserting that 'foo' is an instance of class "Closure".]

so we have to add that function or drop support for closures as view data

silentworks:
Can this PR be pointed at develop branch instead as per @codeguy request on #924 please.

ecoreng:
I think this code only applies to this branch.

PR #936, opened 2014-11-15
Dont need to check if some variable or functions are true

prisis:
We dont need to check if is_null === true or function_exists or $this->env->has($key)

or?

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -567,10 +567,10 @@ class Request implements RequestInterface
     public function get($key = null, $default = null)
     {
         // Parse and cache query parameters
-        if (is_null($this->queryParameters) === true) {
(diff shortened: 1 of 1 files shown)


strngr:
php.net docs:
`is_null`

> Returns TRUE if var is null, FALSE otherwise.

`function_exists`

> Returns TRUE if function_name exists and is a function, FALSE otherwise.

So what's the point to compare strictly results of `is_null` and `function_exists`?

prisis:
i don't see any point to do it, because we can only get true or false.

strngr:
I agree

prisis:
I did use slim for my framework, ok is not more really slim and i did remove all strict results, and it works fine :)

codeguy:
Agreed. Can you rebase against the latest code?

PR #939, opened 2014-11-18
update pimple to 3.*

ecoreng:
referencing this: https://github.com/codeguy/Slim/issues/938


silentworks:
This is currently out of sync and I think we will be including Pimple as part of the repo rather then adding it to the composer require.

PR #941, opened 2014-11-20
Allow Slim to be initialized with a route base

ctsstc:
First pull request in my life ever, so please bear with me... - constructive feedback is great.

Slim can now be initialized with a route base set
which works great for sub folders when the
physical uri base doesn't match the request uri base
Often the case when using using htaccess or vhost on a sub folder
pretending to run the uri from root, which often happens when
trying to create public and private folders in the Server Root;
some hosts prevent access to above the root directory.

```
new \Slim\Slim(array(
 'environment.settings'=> array(
  'route.base'=>'/staging/insomnia/')
));
```

Builds off of: https://github.com/codeguy/Slim/pull/922

Changed documentation for version - not sure if this is the file's
last changed release version or if it should match the latest release
version.

Slim.php
- Checks settings for 'environment.settings' and passes them over
  to the environment if the key exists.

Environment.php
- allow getInstance to pass userSettings through to initializing the
  environment.
- in the constructor changed $settings to default to an empty array
- removed the check if settings was set
- added Andrew's change:
  https://github.com/andrew-forest/Slim/commit/da29004adc0d6011e189fc715b3a5fdac0b55882
  This works great if the $requestUri uses the root for its base
- added check for $settings['route.base'], if it's set it will
  remove that for the PATH_INFO
- the server variables passed in through $settings now get
  merged over the default global server variables

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -70,10 +70,10 @@ class Environment implements \ArrayAccess, \IteratorAggregate
      * @param  bool             $refresh Refresh properties using global server variables?
      * @return \Slim\Environment
      */
-    public static function getInstance($refresh = false)
+    public static function getInstance($refresh = false, $userSettings = array())
     {
         if (is_null(self::$environment) || $refresh) {
-            self::$environment = new self();
+            self::$environment = new self($userSettings);
         }
 
         return self::$environment;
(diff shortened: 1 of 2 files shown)


codeguy:
Closing. This is not a commentary on the quality of your PR. I just don't think Slim needs this feature, and a lot has changed in the develop branch (future 3.x codebase). The master branch is accepting only security fixes at this time.

PR #953, opened 2014-12-05
Add settings option to skip passing routes for more speed

alanpich:
Adds a new settings default `route.no_pass` which if set to true will force the router to return the first matched route it finds without iterating through the entire stack. Should improve performance of the router for all but specialist use cases. Is set to false by default to maintain current behavior.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1070,14 +1070,14 @@ class App extends \Pimple
      * @param  \Slim\Http\Request  The request instance
      * @param  \Slim\Http\Response The response instance
      */
-    protected function dispatchRequest(\Slim\Http\Request $request, \Slim\Http\Response $response)
+    protected function dispatchRequest(\Slim\Http\Request $request, \Slim\Http\Response $response, $noPass)
     {
         try {
             $this->applyHook('slim.before');
             ob_start();
             $this->applyHook('slim.before.router');
             $dispatched = false;
-            $matchedRoutes = $this['router']->getMatchedRoutes($request->getMethod(), $request->getPathInfo(), false);
+            $matchedRoutes = $this['router']->getMatchedRoutes($request->getMethod(), $request->getPathInfo(), false, $noPass);
             foreach ($matchedRoutes as $route) {
                 try {
                     $this->applyHook('slim.before.dispatch');
@@ -1156,7 +1156,10 @@ class App extends \Pimple
      */
     public function call()
     {
-        $this->dispatchRequest($this['request'], $this['response']);
+        /** @var \Slim\Interfaces\ConfigurationInterface $settings */
+        $settings = $this['settings'];
+        $noPass = $settings['routes.no_pass'];
+        $this->dispatchRequest($this['request'], $this['response'], $noPass);
     }
 
     /**
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1067,17 +1067,18 @@ class App extends \Pimple
      * The provided Request and Response objects are updated by reference. There is no
      * value returned by this method.
      *
-     * @param  \Slim\Http\Request  The request instance
-     * @param  \Slim\Http\Response The response instance
+     * @param  \Slim\Http\Request  $request   The request instance
+     * @param  \Slim\Http\Response $response  The response instance
+     * @param  bool                $passable  Allow multiple route matches and Pass() exceptions
      */
-    protected function dispatchRequest(\Slim\Http\Request $request, \Slim\Http\Response $response, $noPass)
+    protected function dispatchRequest(\Slim\Http\Request $request, \Slim\Http\Response $response, $passable = true)
     {
         try {
             $this->applyHook('slim.before');
             ob_start();
             $this->applyHook('slim.before.router');
             $dispatched = false;
-            $matchedRoutes = $this['router']->getMatchedRoutes($request->getMethod(), $request->getPathInfo(), false, $noPass);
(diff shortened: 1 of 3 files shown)


tuupola:
I like the idea but double negatives makes brain hurt `routes.no_pass => false`.  Why not just  `routes.pass => true`?

silentworks:
This gives such great speed boost on large projects. I have tested this myself and it works great, I am yet to see it breaking anything but maybe @codeguy could say if it would break anything existing.

alanpich:
So have swapped around the negatives in the variable names so that:
- Behaviour remains the same as before as default (as in passable routes)
- Settings `routes.passable` in configuration to `false` will enable fast matching (returning the first matched route only, disabling Pass-ing

alanpich:
It probably won't cause any issues for newer projects. I can see a small possibility of doing some crazy stuff with fallback urls, but its definitely an edge case.

As an example, this would work with `passable=TRUE`:
- `/foo/:id`    ID not recognised, so throw a PassException
- `/foo/:unknownId`    Falls through to here, where the invalid ID is logged or something...

... but wouldn't work with `passable=FALSE`, so the `/foo/:id` callable would need to handle both known and unknown foo IDs ...

I don't think using Exceptions to switch between Routes is the right approach. It works as a nice way to bail out,  but it smells a bit iffy to me...

tuupola:
It will of course break applications which use `pass()` method in any of the routes. But it is not a problem since fast matching is not enabled by default.

StJudeWasHere:
Is `pass()` really that useful? I've never used it myself and I 'd rather not have it in Slim at all.

I guess most projects could do fine without `pass()`, and removing it would simplify the code and give us that speed boost by default. I know some of you guys use this feature but I'm wondering if we could get rid of it.

silentworks:
I think this is better as a setting than removing `pass()` there is probably someone somewhere using it and not sure if this is worth the break over this feature which can be enabled anyway.

alanpich:
Personally I think the use of Exceptions as a part of the planned process flow is a grubby misuse of the construct. Which is a shame, 'cos I can see the ability to bail out of a route as something that _could_ be useful at some point in the future (I also keep plastic bags and elastic bands in a drawer just in case...)

My vote would be to switch FastRouting on by default to give the speed boost out of the box, but leave the setting in for people to enable who still really want to `pass()`, as its more of an advanced feature anyway

tuupola:
Removing `pass()` breaks BC and requires major version number upgrade. Same goes for enabling this feature by default.

alanpich:
This is a PR for the develop branch.
The entire API is different (`\Slim\App` vs `\Slim\Slim`), so its already a breaking change and a major bump.

Also, I'm not advocating the complete removal of the `pass()` method, but changing the underlying implementation to not (mis)use Exceptions

tuupola:
Oh, did not notice this was against development branch. Yes I agree. I too was rather confused when reading the code and found out exceptions were used as flow control.

alanpich:
Have created a new PR (#958) for exception removal discussions

codeguy:
I'm honestly for removing the `pass()` method. I have never used it myself. I'm just worried others may depend on it. I'll see what others think in IRC.

codeguy:
Eh, let's just remove the pass method altogether. The performance gains not having it outweigh its benefits, imo. @alanpich how do you want to handle this? Do you want to submit a new PR that removes the pass functionality? If so I'll close this one. Let me know.

alanpich:
Yeah i'll try and carve out some time tomorrow and strip out all the redundant code. I'd say leave this issue open as a reference until I get the PR submitted.

Glad to see some activity on here again :)

codeguy:
Sounds like a plan. Thanks!

alanpich:
OK so after pulling down the latest `develop` commits to update my branch locally it appears that the unit tests are borked and full of errors... There doesn't seem much point submitting a PR to action this discussion until it can be validated properly.

Am i missing a setup step, or have the unit tests just been neglected recently?

lalop:
Hi,
I'm wondering that if the first issue that we will solved with that is performance maybe we can start by testing the matching routes at a group level.
If we have a lot of routes the router tests all of them but if they are grouped we could start by filtring routes on matching group, that could be a much more performance gain.

silentworks:
@alanpich yes develop is broken the moment, but you can still go ahead and submit your PR. Also can we split this into two PRs, once for the skip route and another for removing `pass()`.

alanpich:
Submit a PR without tests?? :scream: 

And also, aren't they the same thing?

silentworks:
Ah sorry my bad, I read through the thread again and notice its all the same thing. We will try and fix the tests soon, or you can jump on it if you want @alanpich

codeguy:
The \Slim\App class unit tests should be passing now (assuming you run _only_ the tests for this class). You can stick your tests in the tests/AppTest.php file.

codeguy:
Any update on this?

alanpich:
This change also applies to `Router::getMatchedRoutes` so would need to run `RouterTest` also. Haven't had a chance to sync my branch so will try to find some time this week to check out if the tests can run. Do we know why the tests are borked? I've a few other things I want to try but am loath to make any core changes unless I can test fully as I go.

Also @silentworks where u been bro?

codeguy:
@alanpich Hold on this. Exploring alternative router implementations and that could affect your work on this. Will post more details soon.

codeguy:
This PR is moot. @alanpich As we discussed, send a PR against the develop branch with your router mounting code.

PR #956, opened 2014-12-09
app_engine_support

karpachev:
On Google App engine the $_SERVER['SERVER_PORT'] is not defined.. and as a result Slim does not work. Workaraund is to define it in your application. 

This patch will check if this variable is defined(Slim/Environment.php). If not - will use default of 80. Works beautifully on App engine.


prisis:
Pls change your pull request to develop!

PR #958, opened 2014-12-10
remove `Slim\Exception\Pass` exception and replace with `return false`

alanpich:
Changed router flow control to bail out of a route by returning `false` from the route's callable rather than throwing a `\Slim\Exception\Pass`. 

Have also added in a `\Slim\Exception\Depreciated` that is thrown if a user tries to use the now redundant $app->pass() method and wonders why nothing happens.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -778,7 +778,7 @@ class App extends \Pimple
     public function pass()
     {
         $this->cleanBuffer();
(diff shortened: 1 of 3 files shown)


codeguy:
Closing. Can you send a new PR against the latest develop branch code? Sorry for the back and forth. Lots of changes being made lately.

Issue #972, opened 2014-12-30
Mixed return types from getCookie()


nochso:
The doc comment for getCookie() says
`* @return string|null`

If encryption is disabled, this will return null thanks to `$value = $this->request->cookies->get($name);`

However with encryption enabled `$value = \Slim\Http\Util::decodeSecureCookie` will make getCookie() return false when the decryption of the cookie has failed.

I suggest that in this case getCookie() should also return null.

krydos:
I changed return value to null. Tests are passed. 

It seems like cosmetic change but anyway, I hope it will allow me to start to contribute more into this framework :)

nochso:
Thank you. It may feel cosmetic, but something not working properly because of the implementation differing from documentation is either a bug or improper documentation :)

PR #975, opened 2015-01-03
fixed issue #972 when getCookie is returning mixed instead of string|null

krydos:
Hi :)

The issue #972 is fixed. I hope you will like it.


krydos:
Ahh, first read then do :(
I had to create pull request to the develop branch. Sorry

krydos:
So, I found that I can push to master :)

PR #982, opened 2015-01-22
Update .htaccess to support existing directories

cpl-repo:
With this option, not just existing files are skipped but existing directories as well.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -6,5 +6,6 @@ RewriteEngine On
 #
 # RewriteBase /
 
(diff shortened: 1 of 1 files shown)


tuupola:
:+1: This is what I am using myself too.

StJudeWasHere:
:+1: I like this too.

ffflabs:
Just to be on the safe side, I suggest skipping symlinks as well

```
RewriteCond %{REQUEST_FILENAME} !-l
```

brandisher:
:+1:

PR #983, opened 2015-01-29
Support for HTTP_X_FORWARDED_PROTO

dopesong:
$_SERVER[‘HTTPS’] not set when using proxy. I saw that its fixed in develop, but I'm using Slim in production, so I can't use dev version of Slim...

It's just simple workaround.


msjyoo:
A note that this will allow the method to be spoofable by users, which is probably not desired. For XF headers a trusted proxies system needs to be implemented,

PR #993, opened 2015-02-09
Add new standarized HTTP response codes

msjyoo:
Source: http://en.wikipedia.org/wiki/List_of_HTTP_status_codes

A particularly needed one is the 429 Too Many Requests, as that would be very useful for APIs with rate limiting.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -116,6 +116,10 @@ class Response implements \ArrayAccess, \Countable, \IteratorAggregate
         418 => '418 I\'m a teapot',
         422 => '422 Unprocessable Entity',
         423 => '423 Locked',
+        426 => '426 Upgrade Required',
+        428 => '428 Precondition Required',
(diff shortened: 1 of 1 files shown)


msjyoo:
The problem with this is that until this is implemented, halt()'ing with an undefined status code will return "HTTP/1.1" in the HTTP response status line (empty status code). Until this is implemented however, you can get around this issue by using Reflection.

```
//Fix for https://github.com/codeguy/Slim/pull/993 until new version is released
$obj = new Slim\Http\Response();
$refObject = new ReflectionObject($obj);
$refProperty = $refObject->getProperty("messages");
$refProperty->setAccessible(true);
$array = $refProperty->getValue(null);
$array[429] = '429 Too Many Requests';
$refProperty->setValue(null, $array);
unset($obj, $refObject, $refProperty, $array);
```

Edit: Or forking a custom version of Slim. Either way, same result.

PR #995, opened 2015-02-10
Fixed wrong PATH_INFO when REQUEST_URI doesn't start with PHYSICAL_PATH.

olejnjak:
My server doesn't put physical path on start of request uri so the path was malformed.


codeguy:
This is obsolete. Can you test against the latest code in the develop branch and report back? The relevant code is in `\Slim\Http\Uri::createFromEnvironment()`.

PR #996, opened 2015-02-10
Fix: Remove useless assignment

localheinz:
This PR
- [x] removes a variable which is assigned a value before a condition, but never used when the condition evaluates to false and instantly overwritten when the condition evaluates to true

Because, you know, cycles!

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -119,7 +119,6 @@ class Slim
 
         $className = ltrim($className, '\\');
         $fileName  = $baseDir;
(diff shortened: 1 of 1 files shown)


localheinz:
Duplicate of #994.

Hey, @codeguy, since you just applied for membership to PHP-FIG, how about sorting out the PRs stacking up here? :wink:

codeguy:
On my todo list.

localheinz:
@codeguy 

:+1:

PR #1000, opened 2015-02-12
Update Set.php

ruchengtang:


tuupola:
What does this PR do? Commit message does not reveal anything.

lalop:
There is an other pr for that https://github.com/codeguy/Slim/pull/998

PR #1001, opened 2015-02-14
Update index.php to fix broken "Hello World" link

joshbuchea:


silentworks:
Good catch, thanks for that.

PR #1002, opened 2015-02-14
.editorconfig file

bjarneo:
Adding editorconfig file. Makes it easier for developers to get their editor to follow your rules if they use some kind of editorconfig plugin.


bjarneo:
Wow, this was all wrong

PR #1004, opened 2015-02-14
editorconfig file added

bjarneo:
Use editorconfig file to easily follow your contribution rules.


silentworks:
I am going to reject this PR as I don't think it should be included in the repo, since not everyone uses editorconfig plugin with their IDE/editor and we have already stated that contributions should follow PSR-2 style guide.

PR #1009, opened 2015-02-18
Use the same middleware mecanism for App and Route

lalop:
This is a work in progress.
The target is to build a common entity that can manage app middleware's stack as well as the route middleware's stack

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -210,7 +210,8 @@ class App extends \Pimple
         $route = new \Slim\Route($pattern, $callable, $this['settings']['routes.case_sensitive']);
         $this['router']->map($route);
         if (count($args) > 0) {
-            $route->setMiddleware($args);
+            //add Middleware
+            $route->add($args);
         }
 
         return $route;
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -50,6 +50,8 @@ if (!extension_loaded('mcrypt')) {
  */
 class App extends \Pimple
 {
+    use Middlewared;
+
     /**
      * @const string
      */
@@ -71,12 +73,6 @@ class App extends \Pimple
         'slim.after.dispatch' => array(array()),
         'slim.after' => array(array())
     );
-
-    /**
-     * Middleware stack
-     * @var array
-     */
-    protected $middleware;
 
     /********************************************************************************
     * Instantiation and Configuration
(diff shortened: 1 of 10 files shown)


lalop:
That 's is ready for review

lalop:
@codeguy since the Middleware class is exists only for internal business I think we can transform it into a more abstract class to wrap the app to, to do it we just have to make the next param optinal.
Since it shouldn't be used outside of the library that seems ok for me. Maybe we should change its name.

codeguy:
Ok, this is next on my todo list. Will be posting feedback soon.

codeguy:
Here's some basic code I put together really quick based on what you've done so far. Let me know what you think. Basically, all we'd need to do is change the `Middleware` class in this code to a trait, and drop that into the `Slim` and `Route` classes.

``` php
<?php
class Middleware
{
    protected $stack = [];

    public function initializeStack()
    {
        $this->stack[] = $this;
    }

    public function add(callable $callable)
    {
        $next = $this->stack[0];

        array_unshift($this->stack, function ($req, $res) use ($callable, $next) {
            return $callable($req, $res, $next);
        });
    }

    public function call($req, $res)
    {
        return $this->stack[0]($req, $res);
    }

    public function __invoke($req, $res)
    {
        $res .= '-CENTER-';
        return $res;
    }
}

$mw = new Middleware();
$mw->initializeStack();
$mw->add(function ($req, $res, $next) {
    $res .= 'Foo';
    $res = $next($req, $res);
    $res .= 'Bar';

    return $res;
});
$mw->add(function ($req, $res, $next) {
    $res .= 'Xyz';
    $res = $next($req, $res);
    $res .= '123';

    return $res;
});
$out = $mw->call('', '');

echo $out;
```

codeguy:
So much has changed I went ahead and implemented a rudimentary version in https://github.com/slimphp/Slim/commit/f45fe83c6f976210b5b334229bd406cc932db00e. Can you look over and make sure it's ok? Also, can you integrate into the Route class and make sure Route middleware is working correctly?

lalop:
Hello.
Afk for the all week.
I will take a look next week.
Just to be sure you want me to add your implementation in the route class ?
Do we leave my pr ?

codeguy:
I think your existing PR can be closed, but open a new PR for new changes and we can move discussions there.

PR #1011, opened 2015-02-23
Updated interface that Route matches responses against

wdelfuego:
Updated interface that Route matches responses against from legacy \Slim\Interfaces\Http\ResponseInterface to \Psr\Http\Message\ResponseInterface.

I hope this is an actual contribution. I'm using Slim 3 to develop a PSR-7 compatible library, and my Slim Routes would ignore the responses I generated because they checked for the legacy ResponseInterface.
This fix solves that for me. I am guessing and hoping this is how Slim 3 is intended to work, would love to hear why if I'm mistaken.


codeguy:
Good catch! Merged :)

PR #1013, opened 2015-02-24
Fix bug that caused no valid headers to be sent

wdelfuego:
No header values would be sent without this fix, I think this is how it was supposed to work.


codeguy:
Thanks!

PR #1016, opened 2015-02-25
fix ConfigurationHandler::flattenArray()

ghost:
Fix flatten array

```
return array(
    '123' => array(
        '456' => array(
            '789' => 1
        ),
    ),
);
```

should be
`'123.456.789' => 1`
but it become
`'456123.789' => 1`


codeguy:
Can you provide a unit test for this?

ghost:
Check this [here](https://github.com/slimphp/Slim/pull/1021)

PR #1017, opened 2015-02-25
query string parameters support in urlFor

pine3ree:


silentworks:
Thanks for this.

nickolasgregory:
Does the signature for `Slim\Interfaces\RouterInterface::urlFor()` need updating too?

```
public function urlFor($name, array $params, array $queryParams);
```

I can make a PR if it helps.

pine3ree:
you are totally right! :-) a fixing PR would be good

PR #1019, opened 2015-02-25
separate create route to override Route easier

ghost:


codeguy:
Closing. Route creation should happen either in a container service or inside the Router object. This is something that needs refactored for the 3.x release.

PR #1039, opened 2015-03-01
Fixed some unit tests by changing the via from a string to an array

silentworks:


codeguy:
Already outdated. Sorry :) I need to rewrite many of these tests.

PR #1042, opened 2015-03-02
Update the phpdocs and constant to match the current release

tw2113:
Covers both the actual framework as well as the tests folder.


codeguy:
Merged and tagged. Thanks!

PR #1048, opened 2015-03-04
Fix ConfigurationHandler::mergeArray

ghost:


codeguy:
Assigning this to @silentworks for review. He'll merge in when appropriate.

codeguy:
@silentworks Something in one of these pulls broke settings... looks like you have to separate setting values with "_" instead of ".". Can you restore this behavior asap?

ghost:
I don't think so @codeguy, this method like array_replace_recursive, but merge when element's key is integer. Not separate settings with `_` instead of `.` ??

gmanricks:
Before the keys were added with setValue and in this pull request they are just inserted directly, setValue handles keys with the separator string '.' line 221 in the pull request should be replaced with setValue

Issue #1052, opened 2015-03-04
[develop] Suggest changing order of route params


silentworks:
With the current setup it is expecting route name first then pattern and then callable. I reckon we should move name to be the last param.

current implementation

``` php
$app->get('home', '/', 'Class:method');
```

proposed implementation

``` php
$app->get('/', 'Class:method', 'home');
```

This way existing Slim 2 apps can port routing easier and also it leaves name optional at this point if its at the end as some users don't actually use route names.

codeguy:
:+1: Send PR please

harikt:
nice silentworks :) . 

Also this is much readable .

PR #1054, opened 2015-03-05
Changed the order of params and added in automatic naming #1052

silentworks:
Added in automatic route naming from #866 because it looks like its not optional to have a route without a name in FastRoute.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -37,6 +37,8 @@ class App extends \Pimple\Container
      */
     protected $responded = false;
 
+    private $routeCount = 0;
+
     /********************************************************************************
     * Instantiation and Configuration
     *******************************************************************************/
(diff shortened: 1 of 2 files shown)


akrabat:
Oh good. I have just come across this while looking at `devlop`.

I don't see that FastRoute cares about the name though, as the only thing that uses the name is Router::urlFor() as far as I can tell.

silentworks:
@akrabat you are correct on this, I have now found the issue in the tests which led me to thinking this was necessary.

@codeguy it would seem that `$this->addRoute` method is looking for an string as the method and not an array, If I move this into a foreach loop it fixes the issues. I can see this only being necessary to do because we are still supporting the `$app->map` method. Let me know your thoughts.

codeguy:
@silentworks Pull the latest commits. I asked Nikita to push out a new 0.4.0 FastRoute release that accepts an array of HTTP methods.

silentworks:
@codeguy yes that worked.

PR #1069, opened 2015-03-09
route use middlewareAware

lalop:
I've updated Route to use MiddlewareAware trait.
It copies the same integration then the one made in the Application class.

### DIFF
--- a/Slim/Interfaces/RouteInterface.php
+++ b/Slim/Interfaces/RouteInterface.php
@@ -33,5 +33,5 @@ interface RouteInterface
      *
      * @return ResponseInterface
      */
-    public function __invoke(RequestInterface $request, ResponseInterface $response, array $args);
+    public function __invoke(RequestInterface $request, ResponseInterface $response);
 }
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -17,6 +17,9 @@ use Slim\Interfaces\RouteInterface;
  */
 class Route implements RouteInterface
 {
+
+    use MiddlewareAware;
+
     /**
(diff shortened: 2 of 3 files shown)


codeguy:
This looks awesome so far. The only thing you missed is `Slim\App` on line 267. Need to change the `setMiddleware()` method call with `add()`.

lalop:
I commited that.
In my first PR I've added a lock to prevent middleware addition once the stack is called. Like that we can't add a middleware on the stack that will not be call.
An other point, the seedMiddlewareStack method has to be call by the class that uses MiddlewareAware but the trait could do that itself, in that way the usage could be painless.

designermonkey:
MiddlewareAware was just updated to seed the stack on the first add operation.

codeguy:
Does this still work if no middleware is added?

lalop:
We should add the same things in callMiddlewareStack

``` php
    public function callMiddlewareStack(RequestInterface $req, ResponseInterface $res)
    {
        if (is_null($this->stack)) {
            $this->seedMiddlewareStack();
        }
        $start = $this->stack->top();
        return $start($req, $res);
    }
```

codeguy:
Sounds gd. @lalop do you want to add that to this pr?

lalop:
I'm making a new one

2015-03-11 1:16 GMT+01:00 Josh Lockhart notifications@github.com:

lalop:
https://github.com/slimphp/Slim/pull/1078

Based on this new pr we can specify the kernel of the middleware stack to be an other method of the class or anything callable.
Like that we can create groupStack and other cool stuf.

PR #1074, opened 2015-03-10
[develop] Fix PSR-2 coding style

zoontek:
I try to do a massive cleanup of the codebase.
Tell me what you think about it !


JoeBengalen:
Note that the array short syntax requires PHP 5.4. Slim seems to also support version 5.3

zoontek:
Slim 3 seems to need at least PHP 5.4, so I guess it's OK
https://github.com/slimphp/Slim/blob/develop/composer.json

JoeBengalen:
Ah my mistake. I looked at the master requirements ;)

codeguy:
I think this is all fine except for the EditorConfig thing. Can you remove that?

zoontek:
No problem, not everyone use this indeed.

prisis:
You can try to use https://github.com/FriendsOfPHP/PHP-CS-Fixer is much faster to get the right coding style :)

zoontek:
Yep, it's already what I use, with the corrects fixers :)

prisis:
Maybe you can create a ".php_cs" so all of us can use it, if @codeguy like the idea

codeguy:
Just the formatting changes, please. I don't want anything in the repo that doesn't absolutely need to be there.

zoontek:
I think it's all good. I don't see another optimisations.
Maybe PSR-4 autoloading, but in a different PR?

JoeBengalen:
For psr-4 see #1075

codeguy:
Crap. Can you rebase against the latest develop branch?

zoontek:
No problem, It should be OK.

PR #1075, opened 2015-03-10
Update to psr-4 autoloader

JoeBengalen:
Moved all source files in the `src/` directory.
Updated the `Slim\Autoloader` to support the psr-4 loading and register other namespaces

I purposely did not add the Slim namespace as default to the loader, as I can image the use case where Slim is already loaded via composer and the Slim\Autoloader is only used to include custom, application specific, classes.

Do you want to update to psr-4 at all? Seems usefull to me as psr-0 is marked deprecated.


JoeBengalen:
Also, would you like to add a sub namespace? Most packages start with `{vendor}\{package}\` as namespace. Would `SlimPHP\Slim\` fit in this case? Other repos from the Slim Framework could then use the same vendor name (eg `SlimPHP\Session\`)

codeguy:
We will convert to PSR-4 and move source code into a `src/` directory closer to the 3.0 release. I don't want to do this now so I don't cause too many conflicts with outstanding PRs in progress.

harikt:
@codeguy nice. I have been thinking to tell you about src folder. And good to know . Great +1 :) . and thanks!

JoeBengalen:
@codeguy, good reason to wait. Shall I close this pull for now? Or do you wanna keep it open.

codeguy:
We can keep this one open if you don't mind rebasing in the future. Otherwise feel free to close and re-open a new one closer to the 3.0 release.

akrabat:
Reading https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md, I think that we already have a vendor namespace: **`Slim`**

I also don't see why we need a `src` directory. Could someone explain why?

silentworks:
I don't think we need `src`, I think that is just a convention that has been used by most projects.

tuupola:
As @akrabat said there already is vendor namespace `Slim`. I do not see any reason to change it. Current namespacing works just fine.

harikt:
@akrabat `src` folder is not for psr-4 . Just a convenience of saying source resides here, tests resides here.

akrabat:
No longer applicable as we've removed the autoloader. See #1256

PR #1076, opened 2015-03-10
Rename App::mapRoute() to map() and remove map()

akrabat:
We don't need map() as a separate method as the first parameter must always be an array which is what mapRoute() expects.

Also, I fixed the docblock's list of array elements to match the new reality of the name being last.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -228,7 +228,7 @@ class App extends \Pimple\Container
      *
      * @return Route
      */
-    protected function map(array $methods, $args)
+    public function map(array $methods, $args)
     {
         static $routeCount = 0;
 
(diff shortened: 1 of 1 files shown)


JoeBengalen:
Maybe I am wrong, but I thought the `map()` was a public method that could be used to set a route to multiple methods: `$app->map(['PUT', 'GET'], {route}, {handler});`.

The `mapRoute()` was for internal handling of both `map()` and all other _method_ methods (put, get, delete, etc)

akrabat:
@JoeBengalen Yes map() should be public. Fixed.

All the other helper methods (get, post, put, patch, delete & options) fill in the method array for you. map() doesn't as you specify the set of methods you want. Hence mapRoute() and map() have exactly the same function signature & so map() is redundant and slows things down.

akrabat:
Actually, ignore me - this makes more sense for 5.6 :)

PR #1077, opened 2015-03-10
Strip leading slash from use statements

designermonkey:
They're not required, my bad for putting them all in in the first place :D


designermonkey:
Nothing like starting small to get back into it :D

codeguy:
:+1:

PR #1078, opened 2015-03-11
auto seed middleware + add seed parameter

lalop:


lalop:
Maybe we should throw an exception when the seed method is called twice.

lalop:
I'm wondering if we shouldn't prevent seeding multiple times the stack by throwing an exception if that happens ?
But I'm not sure if the MiddlewareAware should give the hability to update the stack.

codeguy:
Seeding is a protected method and only accessible to the framework itself. I imagine we'll notice that if it happens without using an exception :)

lalop:
Yes but the trait can be used outside of the framework.
I imagine creating a stack of stack of middleware, with my common middleware, then authentifications one... etc.
In that case using this trait will be nice, so it should be write to an external use too.

codeguy:
:+1:

PR #1090, opened 2015-03-13
docblock typo's

Allineer:


JoeBengalen:
It also says `Resposne` instead of `Response`

Allineer:
@JoeBengalen Thanks :)

Issue #960, opened 2014-12-12
Request::getIP() may not return the correct result when the user uses multiple proxies


brian978:
If the user uses more than 1 proxy to connect to an application then the getIP() method might not return the correct result. Now, I haven't found much information on the HTTP_CLIENT_IP header, but, the HTTP_X_FORWARDED_FOR / X_FORWARDED_FOR header may not contain a single IP address but a list IPs separated by commas. 

Considering this, the getIP() method should return only 1 IP from that list (theoretically the first one, in practice it can be either one, since that header can be spoofed).

I can send a pull request with the changes, if we agree that the behavior should be changed.

tassoevan:
:+1:

GeeH:
I think that @weierophinney has a library that fixes this

codeguy:
Link?

PR #1095, opened 2015-03-15
Update Request.php

r3wt:
I think this can address #960 

I'm using it in my own Slim App.


codeguy:
The master branch is only accepting bug/security fixes for the current code. We already have a PR open for improved IP and X-Forwarded-\* header parsing. This PR will be back-ported to the 2.x branch when ready.

PR #1098, opened 2015-03-15
Also perform method_exists() on objects from Pimple

JoeBengalen:


codeguy:
Can you describe what this PR is fixing?

JoeBengalen:
In the method `\Slim\App::resolveCallable()` you perform a `method_exists()` check on objects created outside of the Pimple, but not on objects that came from the Pimple, which should be done as you cannot  be sure the method you are going to call does actually exist.

This PR makes sure that the method being called is checked for existance in both objects coming from Pimple as well as other objects.

JoeBengalen:
In case it is not clear what I mean, an example:

``` php
$app = new \Slim\App();

class Foo
{
    public function bar($request, $response, $args)
    {
        return $response->write('Foo::bar() called!');
    }
}

$app['foo'] = function() {
  return new Foo();  
};

$app->get('/test', 'foo:bar');     // This calls Foo::bar()
$app->get('/test', 'foo:unknown'); // This first resulted in a PHP error, with the PR an InvalidArgumentException is thrown.
```

akrabat:
Good catch. I'll look at it this week.

JoeBengalen:
Also as I created the example, this came to mind: Since the `InvalidArgumentException` is thrown from within the closure, which will not be called immediately, it should maybe be a `RuntimeException` instead?

codeguy:
Be sure any PRs are accompanied with appropriate unit tests.

JoeBengalen:
Im not too experienced with PHPUnit, but gave it a shot. Please let me know if the tests are good enough.

PR #1104, opened 2015-03-17
Reopend Callback middleware

kanellov:
Hello,
I'm opening a new PR (previous #1103) against master branch, instead develop, since it's use is  for Slim 3.
I have created a middleware class for **Slim 2** that uses a callable in its call method. I have been using it for a while in my projects with Slim and I think it's very handy, especially in prototypes. Here's an example

``` php
<?php

$app = new Slim\Slim();
$app->add(new Slim\Middleware\Callback(function ($app, $next) {
    // do some work ...
    // call next middleware
   $next->call(); 
   // do some extra work ...
});
```

### DIFF
--- a/Slim/Middleware/Callback.php
+++ b/Slim/Middleware/Callback.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Slim - a micro PHP 5 framework.
+ *
+ * @author      Josh Lockhart <info@slimframework.com>
+ * @copyright   2011 Josh Lockhart
+ *
+ * @link        http://www.slimframework.com
+ *
+ * @license     http://www.slimframework.com/license
+ *
+ * @version     2.6.2
+ */
+
+namespace Slim\Middleware;
+
+/**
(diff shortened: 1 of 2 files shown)


JoeBengalen:
Maybe Im missing somehting, but Slim 3 middleware MUST be a callable. What does this `Middleware\Callback` improve?

http://docs-new.slimframework.com/concepts/middleware/

codeguy:
Master branch is not accepting new features.

kanellov:
@JoeBengalen this class is intended for Slim 2 version. @codeguy is the development of ver 2 frozen? If you are interested in this feature, could you tell from where to start my branch in order to merge right?
Thanks.

JoeBengalen:
Ah I see. There's a type in the OP then. It says its intended for Slim **3**

codeguy:
v2 is frozen except for security fixes.

kanellov:
ok thanks

PR #1108, opened 2015-03-18
Return Http\Response from App::redirect

### LINKED ISSUE #1107: Return Http\Response from App::redirect
The redirect helper is nice but wouldn't it be better to do this:

```
return $app->redirect('/page');
```

instead of the current:

```
$app->redirect('/page');
```

The current implementation relies on `App::stop` and exceptions but this proposal makes the control flow much clearer. For example [Silex](http://silex.sensiolabs.org/doc/usage.html#redirects) already does this.

siiptuo:
Resolves #1107 but raises questions about similiar behavior in `App::halt`. Will `App::stop` be needed?

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -367,17 +367,16 @@ class App extends \Pimple\Container
     /**
      * Redirect
      *
-     * This method immediately redirects to a new URL by preparing
-     * and sending a new 3XX HTTP response object.
+     * This method returns a new 3XX HTTP response object to specific URL.
      *
      * @param string $url    The destination URL
      * @param int    $status The HTTP redirect status code (optional)
      *
-     * @throws \Slim\Exception
+     * @throws \Slim\Http\Response
      */
     public function redirect($url, $status = 302)
     {
-        $this->stop($this['response']->withStatus($status)->withHeader('Location', $url));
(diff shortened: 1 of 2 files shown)


JoeBengalen:
On line 375 of App.php there is a typo. It returns the Response, its not throwing.

siiptuo:
@JoeBengalen oh no... thanks for noticing it!

JoeBengalen:
The difference in behaviour of returning and throwing an Exception is if the middleware stack gets called or not. On returning if will travel all the way down and up the stack. On throwing the Exception it will not and go straight to `App::run()`.

I think returning will be fine, but the difference should be considered. My vote will be returning, as I never liked the way an Exception was used to stop the application heren.

codeguy:
I do want to maintain a way to short-circuit the middleware stack and exit _immediately_, which is what stop currently does.

siiptuo:
So should the short-circuit also apply to `redirect`? Or only `halt` and `stop`? In point of view of the user hiding the throw exception can be misleading. Is there any way to combine these two approaches?

codeguy:
For now let's leave `halt()` and `stop()` as they are. Let's modify only the `redirect()` method.

PR #1112, opened 2015-03-18
DIC-aware route middleware

### LINKED ISSUE #1014: Using 'Class:method' as a route middleware
Hello!

If I understand correctly, currently Slim does not allow to use 'Class:method' string as a route or group middleware. It would be very nice to have this possibility.

And, optionally, maybe some method called "filter" or "before" would be handy—in addition to "group" mehod—where route pattern will not be necessary.

This will allow to transfer all logic, including filtering methods, to Controller classes, leaving routes.php file simple and slim.

Thank you.

akrabat:
This PR resolves #1014: "Using 'Class:method' as a route middleware" which means that it supports:

``` php
$app->get(
    '/',
    function($request, $response) {
        echo "<p>Hello world</p>";
    }
)
->add('ClassRouteMiddleware:run');
```

This is more invasive than I expected as  `$app->map()` now returns a `Route` with which we use `add()` in order to add middleware to the route.

However, the Route object doesn't know about the DI container and its  `add()` method comes from the `MiddlewareAware` trait which has a signature that is type hinted to `callable`. Also, the `Route` object is  directly instantiated by the `Router` object in `Router::map`.

As a result, I've created a `ResolveCallable` trait that is used by both `App` and `Route` and modified `Route` to use it. I also decided to make the `Route` object `Container`-aware, but injected the container via a setter so that I didn't need to change the Router class. The `resolveCallable()` method checks that the container has been set if you try to use the 'class:method' string and throws a runtime exception if it doesn't have a valid container.

I'm mostly concerned about the `testAddMiddlewareAsString` test as it seems more of an integration than unit test, but I couldn't work out how else to test. If there's a better way to do it, then it should be changed!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -20,6 +20,7 @@ use Psr\Http\Message\ResponseInterface;
  */
 class App extends \Pimple\Container
 {
+    use ResolveCallable;
     use MiddlewareAware;
 
     /**
@@ -288,42 +289,6 @@ class App extends \Pimple\Container
             call_user_func($callable);
         }
         $this['router']->popGroup();
-    }
-
-    /**
-     * Resolve a string of the format 'class:method' into a closure that the
-     * router can dispatch.
-     *
-     * @param  string $callable
-     *
-     * @return Closure
-     */
-    protected function resolveCallable($callable)
-    {
-        if (is_string($callable) && preg_match('!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!', $callable, $matches)) {
-            // $callable is a class:method string, so wrap it into a closure, retriving the class from Pimple if registered there
-            $class = $matches[1];
-            $method = $matches[2];
-            $callable = function() use ($class, $method) {
-                static $obj = null;
-                if ($obj === null) {
-                    if (isset($this[$class])) {
-                        $obj = $this[$class];
-                    } else {
-                        if (!class_exists($class)) {
-                            throw new \RuntimeException('Route callable class does not exist');
-                        }
-                        $obj = new $class;
-                    }
-                    if (!method_exists($obj, $method)) {
-                        throw new \RuntimeException('Route callable method does not exist');
-                    }
-                }
-                return call_user_func_array(array($obj, $method), func_get_args());
-            };
-        }
-
-        return $callable;
     }
 
     /********************************************************************************
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -10,6 +10,7 @@ namespace Slim;
 
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Pimple\ServiceProviderInterface;
 
 /**
  * App
@@ -266,7 +267,11 @@ class App extends \Pimple\Container
             $callable = $callable->bindTo($this);
         }
 
-        return $this['router']->map($methods, $pattern, $callable);
+        $route = $this['router']->map($methods, $pattern, $callable);
+        if ($route instanceof ServiceProviderInterface) {
+            $route->register($this);
+        }
+        return $route;
     }
 
     /**
--- a/Slim/MiddlewareAware.php
+++ b/Slim/MiddlewareAware.php
@@ -46,9 +46,13 @@ trait MiddlewareAware
      *                                3. A "next" middleware callable
      * @return self
      */
-    public function add(callable $callable)
+    public function add($callable)
     {
-        if($this->middlewareLock) {
+        if (!is_callable($callable)) {
+            throw new \RuntimeException('Expected a callable to be added');
+        }
+
+        if ($this->middlewareLock) {
             throw new \RuntimeException('Middleware can’t be added once the stack is dequeuing');
         }
 
--- a/Slim/ResolveCallable.php
+++ b/Slim/ResolveCallable.php
@@ -8,6 +8,8 @@
  */
 namespace Slim;
 
+use Pimple\Container;
+
 /**
  * ResolveCallable
  *
@@ -17,6 +19,13 @@ namespace Slim;
  */
 trait ResolveCallable
 {
+    /**
+     * Container
+     *
+     * @var Container
+     */
+    protected $container;
+
     /**
      * Resolve a string of the format 'class:method' into a closure that the
      * router can dispatch.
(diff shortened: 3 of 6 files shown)


codeguy:
Are there any architectural changes that would make this easier for you? I didn't want the `Route` class implementation exposed outside the `Router` class, because I see the Route as more an implementation detail of the `Router`. However, if exposing the `Route` class as a first-class service makes your life and this implementation easier, I think that's fine.

As is, I think your solution is good given the current architectural limitations. But I'm wondering if it can be better. Thoughts?

akrabat:
Having slept on it & looked at the diff again this morning, I don't see any obvious architectural changes that would simplify things without making other stuff more complex. I'm generally happy with this solution.

I agree that `Route` is an integral part of `Router` and as the user can replace the `Router` implementation with their own, it's not a problem. I've deliberately made it so that we don't try to register the container into the `Route` instance unless its an instance of `ServiceProviderInterface`, so I'm confident nothing will break if another `Route` implementation is used.

If someone wants to override the `resolveCallable` strategy, then they can extend `App` to change the routable callable and override `Router` & `Route` and inject their new implementation into `$app`, so it's not too hard to do.

JoeBengalen:
It's maybe not what the `Pimple\ServiceProviderInterface` is ment for in the first place*, but it works quite nice I think.

\* [ServiceProviderInterface](http://pimple.sensiolabs.org/#extending-a-container) can be used to register some services and parameters on $pimple.

lalop:
Instead of expecting Route and App to be a Container maybe we can inject the container in this class to use the same everywhere and even outside of Slim.
No ?

akrabat:
`Route` isn't a `Container`; `Route` has a `Container`.
However, `App`, is a `Container` as it extends it.

JoeBengalen:
@lalop That is exactly what the `Pimple\ServiceProviderInterface` is (mis)used for now. To register the container into the route. So you do only have a single instance of `Pimple\Container` which is `Slim\App` itself.

akrabat:
@JoeBengalen I considered creating a "`ContainerInterface`" containing just `setContainer(Container $container);`, but the `ServiceProviderInterface` is essentially exactly that, so I just used it.

JoeBengalen:
@akrabat That would be a `ContainerAwareInterace` then, which can be implemented by a `ContainerAware` trait. But then you are encouraging injecting the container into an object, which is discouraged by pimple itself.

Bottomline: With keeping Slim simple I understand the choice of using the `Pimple\ServiceProviderInterface` in this case, just wanted to point it out to keep in mind if the same situation is encoutered again. Then it could be wise to create the `ContainerAwareInterface`.

akrabat:
@JoeBengalen Is there a better way to resolve a string that represents a key to Pimple when calling `$route->add()`  ?

JoeBengalen:
Not that I can think of. You will always need the `Pimple\Container` instance to resolve the 'callable' middleware. Injecting it to the route is indeed the most clean way to do this I think.

lalop:
@akrabat ok sorry, I didn't read the all diff, specially this https://github.com/akrabat/Slim/commit/232ee757ba2087b298a16cb7497a1ab242acec32#diff-78add50f4cab9c745966442f655ef57fL33

Why don't do this in MiddlewareAwre directly  ?

JoeBengalen:
@lalop Because callable middleware does only need the container if it is a closure.

akrabat:
Yeah, I figured that the MiddlewareAware trait didn't really care how the callable came to be, so I separated it.

jaapverloop:
Looks good, but I wonder if it's wise to inject the container into the route
objects. Aren't we taught not to do this?

The following example shows how to do it manually. Is the code really that
ugly/inconvenient?

``` php
$route = $app->get('/', function() {
    // do something here
});

$route->add($app->resolveCallback('foo:bar'));
```

Or with method chaining.

``` php
$app->get('/', function() {
    // do something here
})->add($app->resolveCallback('foo:bar'));
```

Or introduce an extra optional argument for further configuration of the route 
object. A closure expecting the route object and bound to the application.

``` php
$app->get('/', function() {
    // do something here
}, function($route) {
    $route->setName('home');
    $route->add($this->resolveCallback('foo:bar'));
});
```

akrabat:
I'm _not_ actually against @jaapverloop suggestions here.

There's a minor consistency issue as you'd get:

`$app->get('/', 'this:that')->add($app->resolveCallback('foo:bar'));`

compared to:

`$app->get('/', 'this:that')->add('foo:bar');`

But it does keep the DIC outside of `Route` and all we'd need to do is make `resolveCallable` public.

I think that adding an additional route-config argument to `get()` doesn't seem to gain us anything other than additional typing to set a name which is I think quite common.

What does everyone else think?

JoeBengalen:
I like the idea of keeping the container outside of the Route, but I certainly do not like the inconsistent API

Having to wrap the 'callable' in a extra method is inconvenient. Not because of the typing, but because of the inconsistant API. You are calling the `add()` method twice. One time you have to wrap it and the other time not. You cannot directly see that the `add()` method is called on a different object. 

Having to use the (as it looks) same method in two different ways seems not right to me.

``` php
$app->add('foo:bar');
$app->get('/', 'this:that')->add($app->resolveCallback('foo:other'));
```

The difference with the non middleware 'callable' (2nd arg of `get()`) is less of a concern to me.

jaapverloop:
The inconsistent API bothers me too. I'm still trying to come up with a better solution, but it's not so easy! It seems that adding app middleware using 'Class:method' does not work, on purpose or forgotten?

akrabat:
You can't at the moment as `add()` is type hinted with `callable`. If this one goes in, then I can do something about App's `add` to match Route's.

akrabat:
On the whole, I'd rather Route knew about the container, so we can have a nice consistent UI that's clean and easy to use.

I don't think it hurts that much to have Route know about Container as it doesn't do anything else with it and doesn't need to. Application developers won't be accessing it either as there's no way to get at the `$container` as it's protected and there's no accessor.

JoeBengalen:
Agreed!

codeguy:
Agreed. I'll merge in tonight.

JoeBengalen:
Do not forget to use the callableResolver for `App::add()` as @jaapverloop pointed out ;) Or should I add this in a separate pull request.

jaapverloop:
I guess you're right! The consistency is very welcome!

The following code is what I had in mind.

Slim\MiddlewareAware.php

``` php
trait MiddlewareAware
{
    protected $callableResolver;

    public function setCallableResolver(callable $callableResolver)
    {
        $this->callableResolver = $callableResolver;
        return $this;
    }

    public function add($callable)
    {
        $callable = $this->resolveCallable($callable);
        // ...
    }

    public function resolveCallable($callable)
    {
        if (null !== $this->callableResolver) {
            $callable = call_user_func($this->callableResolver, $callable);
        }

        if (! is_callable($callable)) {
             throw new \RuntimeException('Cannot resolve callable');
        }

        return $callback;
    }
}
```

Slim\App.php:

``` php
public function map(array $methods, $pattern, $callable)
{
    // ...
    $route = $this['router']->map($methods, $pattern, $callable);
    $route->setCallableResolver([$this, 'resolveCallable']);
    return $route;
}

public function resolveCallable($callable)
{
    // leave as it is
}
```

Thanks for the conversation!

PR #1113, opened 2015-03-19
Fix #1110

### LINKED ISSUE #1110: Error when creating Uri from string without a port
When creating a uri from string (in v3) without a port I get the following exception:

```
InvalidArgumentException: Uri port must be null or an integer between 1 and 65535 (inclusive)
```

jenssegers:
This quickly fixes the `createFromString` issue #1110. It basically lets the constructor take care of the default values.


jenssegers:
What is the main reason you prefer to pass empty strings and 80 to the constructor compared to passing null values?

Issue #1099, opened 2015-03-15
Implement `$_FILES` support into Http\Request class.


codeguy:
The `Http\Request` class needs to support the PSR-7 interface methods for dealing with the `$_FILES` request data. The `\Slim\Http\Request::getFileParams()` method should return the current `$_FILES` data. This data should be injected into the Request object at instantiation.

r3wt:
I'm down

codeguy:
Awesome! Go for it. Feel free to ask any questions here as you're working on it.

r3wt:
Well, looks like i will have to back out of this after all. I checked out the develop branch and i'm not happy with the direction the project is going. I love the current version of Slim php so i'll just stick with that. if you stop supporting it ( The `2.x` branch ), i will gladly fork and continue it.

r3wt:
I went ahead and pulled my changes.

r3wt:
@codeguy It's not immediately clear in the PSR if  `getFileParams()` is expected to take an optional key as argument. thoughts?

codeguy:
It does not. It should return a data structure that mimics `$_FILES`. Re: your current PR, we cannot rely on direct access to the `$_FILES` superglobal. Instead, we should inject an array that mimics this structure via the constructor. We can also leverage the Environment object to contain a key (e.g., `slim.files`) that contains this data. This makes it much easier to test and craft sub-requests.

geggleto:
A little updating is in order with the adoption of PSR-7.
PSR-7 defines an UploadedFileInterface, which for the most part is pretty straight forward. However it does also support returning the File as a stream. As such we need a more encapsulated Object that implements the StreamInterface (current supported in \Slim\Http\Body).

I would think that the implementation of the StreamInterface should be extracted from Body into a new class that both the "UploadedFile" Object and the Body Object can Utilize.

The Linkage is in the ServerRequestInterface where getUploadedFiles() is defined.

opengeek:
I'll take a stab at this shortly.

akrabat:
Fixed by https://github.com/slimphp/Slim/pull/1263

geggleto:
Hurray! :+1:

PR #1115, opened 2015-03-19
Implement $_FILES

r3wt:
Request feedback/merge addresses #1099


JoeBengalen:
I don't think $_SERVER should be used in this direct way. Why not use the same structure as cookies and headers?

PR #1117, opened 2015-03-19
Updated Slim\Http\Uri to latest psr-7 specs

JoeBengalen:


codeguy:
Thanks! This is a big help.

PR #1118, opened 2015-03-19
Fix getPort() following psr-7 spec

JoeBengalen:
Following the discussion in #1110 

``` php
// If a port is present, and it is non-standard for the current scheme, this method MUST return it as an integer.
$uri = new Slim\Http\Uri('https', 'www.example.com', 80);
var_dump($uri->getPort());
// result -> 80

// If the port is the standard port used with the current scheme, this method SHOULD return null.
$uri = new Slim\Http\Uri('http', 'www.example.com', 80);
var_dump($uri->getPort());
// result -> null

// If no port is present, and no scheme is present, this method MUST return a null value.
$uri = new Slim\Http\Uri('', 'www.example.com');
var_dump($uri->getPort());
// result -> null

// If no port is present, but a scheme is present, this method MAY return the standard port for that scheme, but SHOULD return null.
$uri = new Slim\Http\Uri('http', 'www.example.com');
var_dump($uri->getPort());
// result -> null
```

Let me know what you think!

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -94,7 +94,7 @@ class Uri implements \Psr\Http\Message\UriInterface
      * @param string $user     Uri user
      * @param string $password Uri password
      */
-    public function __construct($scheme, $host, $port = 80, $path = '/', $query = '', $user = '', $password = '')
+    public function __construct($scheme, $host, $port = null, $path = '/', $query = '', $user = '', $password = '')
     {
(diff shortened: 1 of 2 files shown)


codeguy:
Are these backed up with tests?

JoeBengalen:
If this is OK I will add tests for the cases shown above.

codeguy:
:+1: Go for it

JoeBengalen:
@codeguy Tests updated!

PR #1121, opened 2015-03-19
Replaced $name with $key for consistancy in Cookies

JoeBengalen:
In all function the term $key was used, except for getAsString().
Changes this.


JoeBengalen:
Bit me bein autistic, but it is more consistent :p

PR #1131, opened 2015-03-26
Fixes #1130

### LINKED ISSUE #1130: Issues with grouped routing containing '/' in v3
For an API application I have routes similar to this:

```
$app->group('/1.0', function () use ($app) {
    $app->group('/users/{userId:[0-9]+}/property', function () use ($app) {
        $app->get('/', '...');
        $app->get('/{propertyId:[0-9]+}', '...');
        $app->post('/', '...');
        $app->put('/{propertyId:[0-9]+}', '...');
        $app->delete('/{propertyId:[0-9]+}', '...');
    });
});
```

But using `'/'` as a route pattern inside a group doesn't seem to be working. If I replace it with just an empty string `''` it works.

jenssegers:
This change allows route patterns to be matched on both with and without trailing slashes.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -93,7 +93,7 @@ class Router extends \FastRoute\RouteCollector implements RouterInterface
 
         return $dispatcher->dispatch(
             $request->getMethod(),
-            $request->getUri()->getPath()
+            rtrim($request->getUri()->getPath(), '/')
         );
     }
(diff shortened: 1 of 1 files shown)


JoeBengalen:
I thought there has been an pull request of this before, but it was denied because the was the expected behaviour as is.

PR #1136, opened 2015-03-31
Adds Slim Request and Response Interfaces

jeremykendall:
Since the Slim Request and Response add methods beyond the PSR-7
interfaces, Slim should have its own Request and Response interfaces
that extend the appropriate http-message interfaces.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -8,8 +8,8 @@
  */
 namespace Slim;
 
-use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ResponseInterface;
+use Slim\Interfaces\Http\RequestInterface;
+use Slim\Interfaces\Http\ResponseInterface;
 use Pimple\ServiceProviderInterface;
 
 /**
@@ -71,7 +71,7 @@ class App extends \Pimple\Container
          * Request factory
          *
          * This factory method MUST return a NEW instance
-         * of \Psr\Http\Message\RequestInterface.
+         * of \Slim\Interfaces\Http\RequestInterface.
          */
         $this['request'] = $this->factory(function ($c) {
             $env = $c['environment'];
@@ -89,7 +89,7 @@ class App extends \Pimple\Container
          * Response factory
          *
          * This factory method MUST return a NEW instance
-         * of \Psr\Http\Message\ResponseInterface.
+         * of \Slim\Interfaces\Http\ResponseInterface.
          */
         $this['response'] = $this->factory(function ($c) {
             $headers = new Http\Headers(['Content-Type' => 'text/html']);
@@ -121,12 +121,12 @@ class App extends \Pimple\Container
          * This factory method MUST return a callable
          * that accepts three arguments:
          *
-         * 1. Instance of \Psr\Http\Message\RequestInterface
-         * 2. Instance of \Psr\Http\Message\ResponseInterface
+         * 1. Instance of \Slim\Interfaces\Http\RequestInterface
+         * 2. Instance of \Slim\Interfaces\Http\ResponseInterface
          * 3. Instance of \Exception
          *
          * The callable MUST return an instance of
-         * \Psr\Http\Message\ResponseInterface.
+         * \Slim\Interfaces\Http\ResponseInterface.
          */
         $this['errorHandler'] = function ($c) {
             return new Handlers\Error;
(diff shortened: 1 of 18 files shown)


JoeBengalen:
The whole point about psr7 is that you can swap them out. So the application should expect the psr7 interface, not some slim interface.

jeremykendall:
@JoeBengalen You're absolutely correct, and that's still possible using this PR, but both `Slim\Http\Response` and `Slim\Http\Request` expose more methods than are defined by the PSR-7 interfaces. It's still possible to type hint on the PSR-7 interfaces when necessary, but without providing a Slim specific interfaces, that leaves the rest of the Slim specific implementations unavailable to internal or external classes that need that additional functionality.

I may be missing something, and I'm open to being wrong here, but it seems like Slim's concrete implementations of the `http-message` interfaces should either implement _only_ what is defined in PSR-7 _or_ should extend those interfaces with Slim specific Request and Response interfaces. That is the intent of this PR.

codeguy:
@jeremykendall Which internal classes need access to the additional methods? I've tried to avoid this kind of tight coupling. If an external class needs the additional methods provided by this specific implementation, can they not typehint for `\Slim\Http\Request`?

JoeBengalen:
@jeremykendall You can add the interfaces, but you cannot use them to typehint in Slim objects and expect them to be returned from middleware, else you will still loose the option to swap the psr7 messages objects out for others.

lalop:
@codeguy I've found a call to one of this method while I was working on middlewares by mocking psr request and response.
https://github.com/slimphp/Slim/blob/develop/Slim/Route.php#L223
At this time I choose to mock the slim request and response instead

codeguy:
@lalop Good catch. That needs to be fixed then to rely on `$response->getBody()->write()` instead.

jeremykendall:
@codeguy Ah, I see where you're going. Yes, of course internal classes can typehint for `Slim\Http\*`. I just hate having implementations without interfaces, as a general rule of thumb. If I can throw an interface at it, then I do.

With that philosophy in mind, _every_ internal Slim class that would use methods on the Response and Request objects that are not provided by the http-message interfaces would need the Slim interfaces.

@JoeBengalen Agree, although if the general practice is to typehint against the concrete `Slim\Http\*` implementations when necessary, then the point (other than academic "always code to an interface ..." discussion or considerations) is moot in this case.

(Also, I know I'm late to the Slim 3 game, so forgive me if I'm throwing a wrench into the works.)

codeguy:
@jeremykendall "Always code to an interface" is a good mantra, but I think it's overkill in this particular scenario. We already accept interfaces for the PSR7 message objects. Our particular implementation provides some additional helper methods. These methods are _not_ (or should not) be relied upon by Slim's other internal classes. We _could_ make interfaces, but I don't think the additional weight is really all that helpful.

lalop:
Since the custom method isn't used in the Slim core, why did they exist ?
And more, we should not use them in middleware to keep frameworks interoperability.

codeguy:
@lalop Because they are helpful to the end user and provide a reason to use our own implementation?

jeremykendall:
@codeguy What @lalop said. Also, I've found those methods to be extremely useful in my Slim <3.0 projects.

lalop:
Ok, I understand

jeremykendall:
> Because they are helpful to the end user and provide a reason to use our own implementation?

Which to me says, "Add interfaces." (Full disclosure, I'm a pedant who frequently shoots for  textbook code, _sometimes_ to the detriment of my codebases.)

jeremykendall:
@codeguy Please feel free to close this if it's not in line with current Slim PHP dev philosophy and direction. There will be no hard feelings on my part :smile:

For the sake of discussion, how do interfaces introduce tight coupling? I've always understood them to be the very definition of loose coupling.

codeguy:
Oh they don't add tight coupling... that was in reference to something I said above about how I was trying to _not_ tightly couple the custom non-PSR methods to elsewhere in the framework.

jeremykendall:
Ah. Gotcha.

Again, for the sake of discussion, would it be better to remove those additional methods? Or move them to a Request helper/decorator? I've made heavy use of them in other projects (example: https://github.com/jeremykendall/flaming-archer/blob/develop/public/index.php#L283), and I really like them, but in this case they're causing the Slim Request and Response to be non-compliant with PSR-7 ([`Slim\Http\Response::write()`](https://github.com/slimphp/Slim/blob/develop/Slim/Http/Response.php#L562) violates the immutability of the Response object, for example).

codeguy:
I'm not sure how they are non-compliant with PSR-7... the Slim request and response objects are fully-compliant and implement the interfaces in full. They just happen to have additional methods that do not negate any of the PSR7 functionality.

The `write` method is indeed a mistake and needs to be corrected. Also, the Request and Response bodies _are_ mutable... but everything else in the HTTP message objects is immutable.

jeremykendall:
> The write method is indeed a mistake and needs to be corrected.

Gotcha.

> Also, the Request and Response bodies are mutable... 

Right. Absolutely.

> ... but everything else in the HTTP message objects is immutable.

Excepting the write method, I stand corrected :smile:

jeremykendall:
Closing as not in line with Slim PHP dev philosophy.

Many thanks for the excellent discussion.

PR #1152, opened 2015-04-04
Headers added in response not set

harikt:
Hi, 

I have came to notice that the headers set from the closure are not set on the headers.

```
Connection → close
Content-Length → 15
Content-Type → text/html
Host → localhost:8000
X-Powered-By → PHP/5.5.9-1ubuntu4.5
```


harikt:
My bad, I was returning a different response object.

I do recall some discussion where it mentions the injected response is different from that of returned. Not sure where the thread was, and whether it was in Slim itself or not.

Thanks

harikt:
Closing in favor of https://github.com/slimphp/Slim/pull/1153

PR #1153, opened 2015-04-04
ETag in response

harikt:
In favor of https://github.com/slimphp/Slim/pull/1152


index.php
```diff
@@ -26,10 +26,10 @@
 })->setName('testGet');
 
 $app->get('/etaghit', function ($request, $response, $args) {
-    $response->withEtag('abc123');
-    $response->write("Test body");
-
-    return $response;
+    return $response
+        ->withHeader('Content-Type', 'application/json')
+        ->withAddedHeader('ETag', 'abc123')
```
harikt, line 31: may be  good to use `withHeader` ? .

PR #1157, opened 2015-04-08
adding query string workaround if query string is empty when it shouldn'...

MarcusSchwarz:
...t be

Under some environments QUERY_STRING could be empty. This happens when using lighthttpd with php as fast-cgi module for example. 
With that configuration Request->get() always returns an empty value.

This patch tries to avoid this problem by building its own QUERY_STRING from REQUEST_URI

See for lighty bug report (open for 7 years now): http://redmine.lighttpd.net/issues/1339
See for xhgui-bug report: https://github.com/perftools/xhgui/issues/111


lalop:
Hello,

Is this affecting the development branch ?

MarcusSchwarz:
Hi @lalop 
honestly, I don't know. I found the bug in xhgui and did a quick fix. I never used the Slim Framework. I created the pull request against the dev-branch. I think I shouldn't have created the branch from master :/

PR #1159, opened 2015-04-08
Arguments passed to route callable, no more an array of args

rivetchip:
Allow using route argument(s) directly into its callable, no more an array of results.
In my opinion, it's more clean.

Example : 

``` php
$app->get('/category/{slug}', 'categoryHandler');

$app->get('/category/{slug}/{page}', 'categoryHandler');
```

``` php
function categoryHandler( $request, $response, $slug, $page = 1 )
{
    // Do something
}
```

instead of :

``` php
function categoryHandler( $request, $response, $args )
{
    $slug = $args['slug'];
    $page = isset($args['page']) ? $args['page'] : 1;

    // Do something
}
```

Test is done under : RouteTest::testRouteHandlerCallableParams()

What do guys thinking ?

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -206,7 +206,7 @@ class Route implements RouteInterface, ServiceProviderInterface
         // Invoke route callable
         try {
             ob_start();
-            $newResponse = call_user_func_array($this->callable, [$request, $response, $this->parsedArgs]);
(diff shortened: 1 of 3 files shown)


codeguy:
I prefer a more predictable argument list as opposed to a signature that can vary based on route segment parameters.

PR #1162, opened 2015-04-10
When specifying a Pimple callable, __call magic methods should be allowed

mariano:
Currently Slim3 allows you to define callables that are actually fetched via Pimple, when using the form `PimpleKey:method`. However this only works if `method` is actually a defined method in whatever object is returned by Pimple when the `PimpleKey` is fetched.

This PR takes into account the `__call` magic method, so objects that define such method and are returned as part of a Pimple fetch, will now work. The unit test attachted to this PR should describe this further.

### DIFF
--- a/Slim/ResolveCallable.php
+++ b/Slim/ResolveCallable.php
@@ -56,7 +56,7 @@ trait ResolveCallable
                         }
                         $obj = new $class;
                     }
-                    if ((is_object($obj) && !is_callable([$obj, $method])) || (!is_object($obj) && !method_exists($obj, $method))) {
+                    if (!is_callable([$obj, $method])) {
                         throw new \RuntimeException('Route callable method does not exist');
                     }
                 }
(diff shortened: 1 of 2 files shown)


codeguy:
CCing @akrabat

Slim/ResolveCallable.php
```diff
@@ -56,7 +56,7 @@ protected function resolveCallable($callable)
                         }
                         $obj = new $class;
                     }
-                    if (!method_exists($obj, $method)) {
+                    if ((is_object($obj) && !is_callable([$obj, $method])) || (!is_object($obj) && !method_exists($obj, $method))) {
```
akrabat: Line 59 can just be 

``` php
if (!is_callable([$obj, $method])] {
```

as it won't generate an error if $obj isn't an object and is_callable is effectively a super-set of method_exists functionality, so we don't need to test for that.

mariano:
@akrabat forgot to ping you back after that code change

akrabat:
Thanks @mariano

Issue #1142, opened 2015-04-01
Move \Slim\Collection beneath \Slim\Http namespace


codeguy:
See title. Preparation for separating `\Slim\Http` into its own component.

PR #1172, opened 2015-04-16
Move CollectionInterface beneath \Slim\Interfaces\Http namespace

zither:
see #1142 and #1146, my bad.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -73,7 +73,7 @@ class App extends \Pimple\Container
 
         /**
          * This Pimple service MUST return a shared instance
-         * of \Slim\Interfaces\EnvironmentInterface.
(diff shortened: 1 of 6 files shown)


lalop:
Why do you move CollectionInterace into Http ?

JoeBengalen:
@lalop He did not, codeguy did. If I remember correctly he did it to prepare the separation of the Http classes into its own repo.

Edit:
Ohw nvm we are talking about the Interface ... But the answer is actualy the same. Codeguy moved the Collection class into Http, so it is consitent to put the Interface into Http as well.

codeguy:
What @JoeBengalen said.

lalop:
ok, I see

Issue #1134, opened 2015-03-30
Input validation and sanitization middleware


funkatron:
One pain point for me in Slim 2.x was how to do input validation and sanitization. Retrieving values from the request object's `->get()` and `->post()` always return raw, unfiltered data. To turn it into safe(r) data, you need retrieve the items individually or as an array and pass them each through some kind of sanitization functionality, or bypass the request object entirely and do something to the superglobals directly (which may or may not be problematic depending on how Slim is internally handling that data).

I have a project I've worked on for a long time called [Inspekt](http://funkatron.github.io/inspekt/) that sorta tries to do this to the superglobals. It wraps them inside a "cage" object, and you have to use accessor methods to retrieve data. You have to use `getRaw()` to get completely raw, un-sanitized input. The two big wins are, I think:
1. You're not accessing superglobals directly -- harder to make mistakes
2. You have to _show intent_ when getting raw data -- harder to make mistakes

I think lots of people will have different approaches to how to handle input safely, and I don't think building an opinionated approach into Slim makes sense. However, I do think that it would be good if it was easier to override how request variables are accessed, both on a one-by-one basis, and as an array.

Were I designing a plugin or add-on for Slim that did this, I would want to be able to:
- eliminate the `request::get()` method entirely
- attach a number of input accessor methods that could either be applied to single values or Arrays (or ArrayObjects).

I have a few thoughts about how something like this might look API-wise, but I wanted to just get the idea out there first.

codeguy:
What if the `Request` object supported an object that implemented some form of filterable interface. We'd inject the filter object into Slim's request object in its Pimple factory method. We would then run all values currently returned from `getParam()` and `getParams()` through the filter object if present. This would let you register the filter object on the container, too. Obviously, this stops working if you swap in an alternative PSR7 request object. Thoughts?

codeguy:
For example, we'd register the custom request filter object like this:

``` php
$app = new \Slim\App();
$app['requestFilter'] = function ($c) {
    return new \MyRequestFilter();
};
```

We'd update the current Request Pimple service to look like this (note the last two lines):

``` php
$this['request'] = $this->factory(function ($c) {
    $env = $c['environment'];
    $method = $env['REQUEST_METHOD'];
    $uri = Http\Uri::createFromEnvironment($env);
    $headers = Http\Headers::createFromEnvironment($env);
    $cookies = new Collection(Http\Cookies::parseHeader($headers->get('Cookie')));
[...]

PR #1177, opened 2015-04-17
Add optional callable arguments to Request::getParam() and Request::getParams()

funkatron:
This adds optional callable args to `Request::getParam()` and `Request::getParams()`.

With `Request::getParam()`, the optional second argument is a callable that is passed the parameter value, and returns a modified version of the value. This modified version is returned by `Request::getParam()`.

The optional argument for `Request::getParams()` is similar, except the callable is passed the entire array of parameters, and should return a version of the array with some modifications.

The intent is to allowing a simple approach to applying sanitization and "test" style filters to input.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1142,10 +1142,11 @@ class Request implements RequestInterface
      * body or query string (in that order).
      *
      * @param  string $key The parameter key
+     * @param  callable $callable a callable executed on the param value, returning the parsed result
      *
      * @return mixed The parameter value
      */
-    public function getParam($key)
+    public function getParam($key, callable $callable = null)
     {
         $postParams = $this->getParsedBody();
         $getParams = $this->getQueryParams();
@@ -1158,20 +1159,32 @@ class Request implements RequestInterface
             $result = $getParams[$key];
         }
 
+        if ($callable) {
+            $callable->bindTo($this);
+            $parsed_result = $callable($result);
+            return $parsed_result;
+        }
+
         return $result;
     }
 
     /**
-     * Fetch assocative array of body and query string parameters
-     *
+     * Fetch associative array of body and query string parameters
+     * @param callable $callable a callable executed on array of parameters, returning the result. Should return an array
      * @return array
      */
-    public function getParams()
+    public function getParams(callable $callable = null)
     {
         $params = $this->getQueryParams();
         $postParams = $this->getParsedBody();
         if ($postParams) {
             $params = array_merge($params, (array)$postParams);
+        }
+
+        if ($callable) {
+            $callable->bindTo($this);
+            $parsed_params = $callable($params);
+            return $parsed_params;
         }
 
         return $params;
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -867,6 +867,22 @@ class RequestTest extends PHPUnit_Framework_TestCase
         $this->assertEquals('bar', $request->getParam('foo'));
     }
 
+    public function testGetParameterFromBodyWithCallable()
+    {
+        $body = new Body(fopen('php://temp', 'r+'));
+        $body->write('xss=<script language="javascript">alert("foo")</script>');
+        $body->rewind();
+        $request = $this->requestFactory()
+                   ->withBody($body)
+                   ->withHeader('Content-Type', 'application/x-www-form-urlencoded');
+
+        $callable = function ($raw) {
+            return filter_var($raw, FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES);
+        };
+
+        $this->assertEquals('alert("foo")', $request->getParam('xss', $callable));
+    }
+
     public function testGetParameterFromQuery()
     {
         $request = $this->requestFactory()->withHeader('Content-Type', 'application/x-www-form-urlencoded');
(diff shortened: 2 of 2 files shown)


JoeBengalen:
I do not see what this gives us ... Why not calling the callable on the returned value? Why does it have to happen within the Request object?

silentworks:
This should be done on the returned value as stated by @JoeBengalen. Otherwise better to extend the Request class in your own project.

funkatron:
Good points made, plus I didn't do a good job of explaining this, and I didn't refer to the issue (#1134) where this was discussed. I'll explain my thinking at least.

Ultimately I think it's a problem that the default action of `Request::getParam(s)` is to return raw values (of course, this is how almost all frameworks approach this, but I think it's a mistake there, too). I think it makes it more likely to make input handling mistakes. Were I doing what I felt like, I would rename the methods to `Request::getRawParam(s)`, to indicate to the developer exactly what they are getting: raw, potentially unsafe input, and then additional add a suite of accessor methods that apply common sanitizations.

_Note that the intent here isn't to handle data _validation_, but for de-fanging potentially dangerous input. I agree with the discussion in #1134 that you're better off handing validation with route middleware._

The simple changes here are really a compromise to try to make it a bit easier to implement input filtering, potentially as middleware, and/or on a piecemeal basis.

Here's my thinking for this approach:
- I didn't want to change the default behavior and potentially confuse folks, especially so close to beta.
- I didn't want to impose a set of input sanitization filters on everyone using Slim. Possibly down the road you could include a class that offers sanitization filters that can be applied. I think that would make it super handy to do something like 

``` php
$req->getParam('body_text', Input\stripHtml);
```
- For auditing purposes, this approach makes it a bit easier to check for raw data access -- if there's no second param (or no first param for `Request::getParams`), we are probably getting raw input, which should raise a red flag.
- I personally find this approach a fair bit cleaner/less messy than performing it on the return value in the route handler. The use case I imagine would be defining sanitization callables in a separate class or namespace, and reference those callables within the `getParam/getParams` calls.

``` php
$req->getParam('body_text', Input\stripHtml);
```

vs

``` php
Input\stripHtml($req->getParam('body_text'));
```

or

``` php
$body_text_unsafe = $req->getParam('body_text');
$body_text = Input\stripHtml($body_text_unsafe);
```

YMMV, of course. As mentioned above, I think auditing the first is easier.

Extending the Request class is another decent option, although my inexperience with Slim3 means I'm not sure exactly how easy it would be to use that extended Request class by default everywhere. Also, I think most devs won't do this out of the box, and I think this approach encourages more thoughtful consideration of handling input.

All that being said, this may be too small of a change/too little value to be worth including. This lays some groundwork, I think, but it could be better to put it off until a more comprehensive solution for safer, explicit input handling can be developed. I'll leave that up to you folks.

I'm pulling in @codeguy because he initially gave a thumbs-up on the approach 8)

codeguy:
I'm not against this. I do think it's a nice compromise, like Ed said. Another potential solution is to move this functionality into a service provider.

``` php
class FilterProvider
{
    protected $request;

    public function register(\Pimple\Container $c)
    {
        $c['filter'] = $this;
        $this->request = $c['request'];
    }

    public function get($name, FILTER_FLAGS)
    {
        $raw = $this->request->getParam($name);

        return $this->doFilter($raw);
    }

    protected function doFilter($raw)
    {
        // Return filtered value
    }
}
```

And register like so:

``` php
$app->register(new FilterProvider)
```

And use like this:

```
$app->get('/foo', function ($req, $res, $args) {
    $param = $this['filter']->get('foo');
});
```

This decouples it from the core and nicely encapsulates the functionality in a way that is easily portable .

GeeH:
Doesn't this break PSR-7?

(I love the concept BTW and is something I always do in my own projects)

codeguy:
No, it's specific to our own implementation. So long as we don't depend on this elsewhere in the core framework we're fine afaik

silentworks:
I think this should be provided as a separate component to the framework, I like the solution @codeguy provided above but it wouldn't be sufficient enough for what @funkatron has mentioned. As it doesn't actually override the provided Request class, I am also not for ServiceProviders implementing themselves, I think the ServiceProvider class should only provide an interface for an external Library, so this would fit will with the concept of Inspekt.

@funkatron if you need some pointers, catch me on irc next week and I will see where I can help out. (Still have the Session Library to work on, before @codeguy asks about it again :-))

akrabat:
I tend to agree with @silentworks. 

A service provider as suggested by @codeguy implies that seeing a $req->getParam() is automatically a signal that it's unsafe. Alternatively, route middleware allows for replacing the data in the request with filtered data so that it's outside the controller altogether, so I see that we already have two different non-framework-invasive ways to handle this.

funkatron:
I think the service provider idea may be sufficient, because it still
allows for good auditability -- if anyone is accessing input outside of the
service provider, that would be a red flag for detailed review. It would be
easier to add all sorts of custom functionality this was as well, without
extending a core class like Request. At some point I think a default (but
optional) provider could be included in the framework, but you could
replace it with your own, or extend it easily.

I prefer this approach over middleware, where how things are getting
sanitized is separated from where the input access takes place. I think
that makes it a bit harder to know what's happening ("Why are all my alpha
characters missing!" 10m later: "Oh, because this middleware defined
elsewhere is removing them"). There are advantages to using middleware in
some cases, of course, but I think it's important that the developer think
and show intent at the point of input access.

codeguy:
Cool. I think the best course of action is what @silentworks suggested above https://github.com/slimphp/Slim/pull/1177#issuecomment-93985985. Closing this out.

GeeH:
Can this be a middleware?

Issue #1155, opened 2015-04-07
Replace Pimple with a DI-Contract


rkrx:
Please consider replacing Pimple with something much more useful like PHP-DI or (even better) a DI-Contract to make me able to set a container by configuration. Please do NOT force me to use Pimple.

ecoreng:
+1

Consider:
https://github.com/container-interop/container-interop

mackenza:
perhaps some further feedback into _why_ Pimple is not a good choice would be in order? You say "able to set a container by configuration". Is that the extent of it or are there other reasons? I think @codeguy would probably appreciate some more detail here to distinguish this request from that of opinion or personal bias.

akrabat:
Personally, I like Pimple for simple DI needs. When I need something more powerful, then I personally use Zend\ServiceManager - it's also required if I'm using Zend\Form. With Slim 3, this means that I have two DI containers in play at the same time which is a small nuisance as you have to either make one aware of the other or keep track of what is where. In these cases, it would be nice to be able to replace Pimple with Zend\ServiceManager.

rkrx:
Just as @akrabat wrote: _May be I already use another DI Container_

This.

@mackenza: _Is that the extent of it or are there other reasons?_
[PHP-DI](http://php-di.org/) (for example) supports autowiring. That means (for me) that I don't have to ship a DI-Configuration for more then 95% of all classes involved. Autowiring is simply a technique that tries to automatically instantiate depending classes by looking at a method's (e.g. Constructor) signature and it's parameter-types (Interfaces, Abstract Classes, Classes). For interfaces and abstract classes as well as for scalar values you still have to specify by configuration what their counterpart is. That is in some way comparable to `include/require` vs. `autoloading`.

Some more?
- Lazy instance proxy
- Sophisticated configuration methods
- Annotation-based configuration (if you like annotations; i don't)
- Class factory

PHP-DI would support a call-Method, with could be interesting for slim, too. So in theory you could use something like:

``` PHP
<?php
$app = new \Slim\Slim();
$app->get('/hello/:name', function ($name, MyService $service) {
    $salutation = $service->getSalutationForName($name);
    echo "Hello, {$salutation} {$name}";
});
$app->run();
```

geggleto:
How will this work with older apps upgrading to 3.x ?

I really dislike in the example above the signature change, that for me personally would be an absolute reason why I would not upgrade my current projects.

codeguy:
[...]

PR #1179, opened 2015-04-17
DIC independence

akrabat:
Slim now composes a ContainerInterface which it expects to be configured correctly. 

`App`'s constructor will accept either an array of user settings so that there is no change for the standard use-case or an instance of a `Interop\Container\ContainerInterface`.

i.e you can either do:

```
$settings = [];
$app = new \Slim\App($settings);
```

as normal, or you can do:

```
$settings = [];
$app = new \Slim\App(new \My\InteropContainer($settings));
```

See #1155 for initial discussion.


codeguy:
Can we rename `\Slim\DefaultContainerFactory` to something not so Java-esque? :)

codeguy:
After some deliberation, I've decided to move forward with PHP-DI as our primary DI container. PHP-DI supports the container-interop interface, and the additional delegate container interface. I think this is the best out-of-the-box solution that also affords us additional flexibility if needed.

@akrabat Do you agree? If so do you want to get that process moving?

JoeBengalen:
Why? Maybe its a good choice, I dont know. Never really used it before. It just feelsa bit weird that the whole Slim framework is about the same size as only the Container being used. I like the small footprint of Pimple.

codeguy:
I'm going to do some more research this weekend on alternative lighter-weight solutions (League/Container, Aura/Di). Hold off on any implementations until Monday.

akrabat:
I will try & do the same.

Jusomeara:
PHP Dependency Injection Container Performance [Benchmarks](http://www.sitepoint.com/php-dependency-injection-container-performance-benchmarks/). Note these tests have been performed by the maintainer of Dice. Orno (now league) and Aura have had a few releases since these tests. The conclusion is that Orno is easier to configure than Aura may have changed.

codeguy:
@Jusomeara Thanks. I saw that a few days ago. We're working on a solution that will prevent us from favoring any one container over another. More info coming soon.

mnapoli:
@Jusomeara please be aware that these benchmarks are skewed, have a look at the comments. Some containers should have a cache, or should be compiled, but the author didn't do it. Same for ZF2 he didn't benchmark the correct package. And the biggest problem is he's using a simple loop to benchmark. The truth is that benchmarking containers is unfortunately very difficult because a lot of things come into play (e.g. caches, opcache, autoloading, what is configured, etc.) and the time involved is very small. Putting some code in a loop doesn't work because autoloading, cache reads, etc have an impact (like in a real application where every new request starts from scratch). For a realistic test you have to benchmark through a web request (so that opcache works), but then response time are so fast (because you shouldn't use loops to make it artificially slower) that the HTTP + PHP init overhead is greater than the code execution. That should be doable though, but it takes some work and thinking :) (if anyone wants to try it I would be willing to help)

In any case I'm really happy with the direction this is all taking. I'd love to see PHP-DI in Slim, but I also think that container-interop is a very good option and would make Slim stand out (PSR-7 + container-interop makes a hell of a flexible architecture). Pimple is great for starting easily but becomes limited when the application grows. A different container would allow Slim applications to scale much more easily and create very complex applications.

Of course Slim must come with a default container and work out of the box though ;) (whichever you decide to use)

Regarding PHP-DI the v5 is currently in development and will be finished in a few weeks. I have been working on a new site + new documentation, it's still WIP but it's probably a little bit better than the current documentation: http://phpdi5.herokuapp.com/ (this URL is temporary just to give you an idea)

The most important changes in v5 are:
- lighter package, lighter defaults (less composer dependencies, annotations disabled by default, etc.)
- simpler usage, syntactic sugar (i.e. if you are used to Pimple it will look familiar): http://phpdi5.herokuapp.com/news/14-php-di-5-syntaxic-sugar.html
- support for modular systems (we're working on that for [Piwik](https://github.com/piwik/piwik) which has a plugin system): overriding or extending previous entries, lists, etc. (I can detail more if you are interested)

I did a few performance tests yesterday with this branch using Pimple and PHP-DI, there's a small perf difference (15%) but it can be improved, I've already pushed some optimizations (v5 is not polished yet).

I have done a little Silex + PHP-DI integration, and I'd love to push it further with Slim (either through official support, or a PHP-DI integration). In any case I would love to help and get involved in Slim, and if you have suggestions for improving PHP-DI (new features…) the v5 it is still open.

Now regarding this pull request, it is really great. I also thought about a few things:
- I wouldn't put the request and response factories in the container: those are only helpers used by the `App` class, so I would move them to private `createRequest()` and `createResponse()` methods
- I would document somewhere the container entries (or "services") that Slim expect to be registered in the container: that way user it is easier to know what to define in containers for Slim to work
- a way to keep the simple API of Pimple would be to have `App` implement `ArrayAccess`, and redirect to the container - of course that wouldn't work with `offsetSet` in which case an exception would be thrown (or it could work only with Pimple…). That means the interface is not fully implemented, but the convenience is kept. It's just an idea, I'm not too sure about that myself…

I have another suggestion about dependency injection too but my comment is getting big, I'll open a separate issue ;)

_Edit:_ here is the other issue: #1186

akrabat:
> I wouldn't put the request and response factories in the container: those are only helpers used by the App class, so I would move them to private createRequest() and createResponse() methods

If they aren't in the container, how would a user swap them out for a different Request/Response implementation?

> I would document somewhere the container entries (or "services") that Slim expect to be registered in the container: that way user it is easier to know what to define in containers for Slim to work

Agreed. Along with which interfaces we expect any given service key to implement.

> a way to keep the simple API of Pimple would be to have App implement ArrayAccess, and redirect to the container - of course that wouldn't work with offsetSet in which case an exception would be thrown (or it could work only with Pimple…). That means the interface is not fully implemented, but the convenience is kept. It's just an idea, I'm not too sure about that myself…

I, personally, am not a fan of the array access interface of Pimple. However, `$app->getContainer->get()` maybe is a bit longwinded? So I can see the argument for implementing `ArrayAccess` and `offsetGet` with `offsetSet` throwing an exception. Alternatively, we could create our own SlimContainerInterface interface which extends Interop\Container\ContainerInterface and adds `set()` and `isset()`?

Thoughts anyone?

mnapoli:
> If they aren't in the container, how would a user swap them out for a different Request/Response implementation?

Good point, but if users wants to create a request or response they don't need to go through an abstraction, they can create it by knowing which implementation they want to use. Is there a use case I'm missing?

> I, personally, am not a fan of the array access interface of Pimple.

Me neither :) But I admit that I can see how it makes writing small apps really easy…

> However, $app->getContainer->get() maybe is a bit longwinded? So I can see the argument for implementing ArrayAccess and offsetGet with offsetSet throwing an exception. Alternatively, we could create our own SlimContainerInterface interface which extends Interop\Container\ContainerInterface and adds set() and isset()?

It would destroy the ability to use any container :(

The way I see it, ArrayAccess makes sense only with Pimple. So a Slim edition with Pimple makes perfect sense, it's super easy to learn and use (but limited when growing). In another "edition" (whatever the wording) of Slim with a more advanced container the ArrayAccess because useless, because then users will interact with the container directly.

I have a weird idea: Slim could have a `Application` class that is container agnostic (i.e. relies on ContainerInterface), and there would be a subclass that implements ArrayAccess. So by default you use Slim with Pimple and the "Slim ArrayAccess-Pimple" version (i.e. the subclass). And everything would be simple and well. But then you can also choose to use the agnostic `Application` class and provide your own container. I have not thought this through a lot, just throwing the idea…

Slim/Container.php
```diff
@@ -0,0 +1,171 @@
...
+        parent::__construct();
+
+        /**
+         * This service MUST return an array or an
+         * instance of \ArrayAccess.
+         */
+        $defaultSettings = $this->defaultSettings;
+        $this['settings'] = function ($c) use ($userSettings, $defaultSettings) {
+            return array_merge($defaultSettings, $userSettings);
+        };
```
lalop, line 46: I think we should separate the container implementation than the container provisioning to help provisioning custom container

akrabat, line 46: Why? This `Container` is specifically for Pimple. If I wanted to use Zend\ServiceManager I would need to write a new bridge class and the configuration would be different too.

lalop:
> I have a weird idea: Slim could have a Application class that is container agnostic (i.e. relies on ContainerInterface), and there would be a subclass that implements ArrayAccess. So by default you use Slim with Pimple and the "Slim ArrayAccess-Pimple" version (i.e. the subclass). And everything would be simple and well. But then you can also choose to use the agnostic Application class and provide your own container. I have not thought this through a lot, just throwing the idea…

Am I wrong or that what @akrabat say in its firs comment ? :)

Slim/Container.php
lalop, line 46: If App can work with any container implementation we could provision any container, isn't it ?

akrabat, line 46: Not really as `Interop\Container\ContainerInterface` does not define any way to set up a container. 

More importantly there are so many different features of containers that configuring them is very container-specific. An auto-wired one like Zend\Di is configured very differently to one that uses annotations for instance. For Zend\ServiceManager I would use separate factory classes, but for Pimple, I use closures.

lalop, line 46: Ok,
I didn't know.
That would have been too easy.

akrabat, line 46: I'm very comfortable that anyone who wants to integrate with a different container has to write their own bridge class (if required) and also write the configuration for the required Slim entries.

I expect that those who do this will create a composer package for it anyway.

lalop, line 46: At least, separating  provisioning and definition would permit to provisione a Pimple container that is not the slim one.

akrabat, line 46: Want to create a PR against https://github.com/akrabat/Slim/tree/dic-independence so we can see how it looks?

lalop, line 46: I don't know yet how that could look like but I will try.

mnapoli:
You mean the pull request description or his most recent comment? In the PR description there are not 2 different application classes (which leads to the problem of ArrayAccess not a good match with ContainerInterface). And in his latest comments he talks about "we could create our own SlimContainerInterface interface" which is not the same thing as having 2 application classes. Did I miss it ;)?

lalop:
@mnapoli you're right, I misunderstoud, for me the 2 classes was the container and the application but not 2 Application classes.
But the container is used in other files than Application so we should reimplement them too.

akrabat:
I see two separate issues:
1. Do we allow for a common way to configure the container? (e.g. `$app->getContainer()->set(…);`)

If we did this, then we'd need our own interface, which is why I brought it up earlier. Having thought about it, I think that we don't want to do this as configuration is so different between different DIC implementations.
1. How do we access the dependencies from `$app`?

With this change, the canonical way is `$foo = $app->getContainer->get('foo')`. Should we have a 'shortcut' way to get at the dependencies?

If w choices that I can see are:
1. Write shorter methods on App: e.g. `$foo = $app->getDependency('foo')` & `$app->hasDependency('foo')`
2. Implement `__get()` and `__issset()` on App which would match the way we do it in Slim 2. This would allow us to do `$foo = $app->foo`.
3. Implement ArrayAccess and implement `offsetGet()` and `offsetExists()`. We would also have to through exceptions for the implementations of `offsetSet` and `offsetUnset`. This would allow us to access via `$foo = $app['foo']`.

Personally, I don't see a DIC as an array, so am leaning towards option 2. Matching the way Slim 2 does it makes upgrading easier too which would be nice.

JoeBengalen:
I'd say do NOT implement a common way to configure the container, for the reason you gave; They are all so different.

For queston 2 I do not really care. I never use the container trough slim, but use the container directly.

``` php
class Controller
{
  public function __construct(Model $model)
  {
     $this->model = $model;
  }

  public function index($model)
  {
     $this->model->doSomething();
  }
}

// Example in Pimple
$container['database'] = function() {
  return new PDO(...);
};
$container['model'] = function ($c) {
    return new Model($c['database']));
};
$container['controller'] = function ($c) {
    return new Controller($c['model']));
};

$app = new Slim\App($container);
$app->get('/', 'controller:index');
```

akrabat:
@JoeBengalen I do the same as you.

Others don't though. This is especially true for `$app['router']->urlFor()` and for `$app['settings']`.

tuupola:
For me this feels like over engineering. Is DIC independence really needed? Just choose one and let it be.

geggleto:
:+1: I am with tuupola on this issue. I think its over complicating a framework that was meant to be simple to use. If you want a different DIC, fork slim and have at it.

JoeBengalen:
What does it complicate? using a different DIC is optional.  Default is still pimple here. Only difference here is how you get to an entry. That is different, but certainly not more complicated. Using a specified method instead of array access. 

The container being a separate object as the app makes it more clear. Different functionality - different class. More separation of concerns.

akrabat:
Rebased onto develop and have added __get and __isset.

JoeBengalen:
Why did you make the Container final? I hate final classes xD. You will always come up to a point where you want to extend it. Especially in such a framework I would not make a class final.

silentworks:
@JoeBengalen because this http://ocramius.github.io/blog/when-to-declare-classes-final/

JoeBengalen:
Al right I can get into that "Make your classes always final, if they implement an interface, and no other public methods are defined".

Another point I want to mention; would it be more elegant to prevent the try catch statement in the `get($id)` method? I see we have to throw a different exception then Pimple does, but wouldn't is be nicer to check for ourselves (using `offsetExists`) before calling the `offsetGet`?

``` php
public function get($id)
{
    if (!$this->offsetExists($id)) {
        throw new NotFoundException(sprintf('Identifier "%s" is not defined.', $id));
    }
    return $this->offsetGet($id);
}
```

Slim/Container.php
jaapverloop, line 46: What do you think about the idea of using service providers, the ServiceProviderInterface provided by Pimple, instead of creating the required components in the container? 

``` php
<?php

// Slim\Container.php
namespace Slim;

final class Container extends PimpleContainer implements ContainerInterface
{
    private $defaultSettings = [
        'cookieLifetime' => '20 minutes',
        'cookiePath' => '/',
        'cookieDomain' => null,
        'cookieSecure' => false,
        'cookieHttpOnly' => false,
        'httpVersion' => '1.1',
        'responseChunkSize' => 4096,
        'providers' => [
            'router' => 'Slim\Provider\RouterProvider',
            'request' => 'Slim\Provider\RequestProvider',
            'response' => 'Slim\Provider\ResponseProvider',
        ],
    ];

    public function __construct(array $userSettings = [])
    {
        parent::__construct();

        $defaultSettings = $this->defaultSettings;
        $this['settings'] = function ($c) use ($userSettings, $defaultSettings) {
            return array_merge($defaultSettings, $userSettings);
        };

        foreach ($this['settings']['providers'] as $class) {
            $provider = new $class;
            $provider->register($this);
        }
    }
}
```

``` php
<?php
// Slim\Provider\RouterProvider.php
namespace Slim\Provider;

use Slim\Router;
use Pimple\ServiceProviderInterface;

class RouterProvider implements ServiceProviderInterface
{
    public function register(Container $pimple)
    {
        $pimple['router'] = function ($c) {
            return new Router();
        };
    }
}
```

This makes it easy to swap each component with another implementation. The settings require another merge strategy to change a single component instead of all at once.

JoeBengalen, line 46: Talking about over complicating ... You want to create a separate provider for each class?

I do not really see the gain here. You can easily overwrite a registered service in Pimple.

Also the whole point of this PR is to be not Pimple dependant. So using their service provider interface would go against that.

jaapverloop, line 46: Over complicating... I don't know, the idea is very simple and a provider for each class is optional I guess.

What I really ment is something similar like the ServiceProvider of Pimple.

JoeBengalen, line 46: But how would you do that DIC independent? or do you want to make it pimple specific? If so, what would be the gain?

silentworks, line 46: For the core of Slim we will keep this as is at the moment, as @JoeBengalen said, your suggestion would add more classes which is unnecessary as you can override any dependency at the moment through the container or even using your own ServiceProvider class that implements the ServiceProviderInterface (If you are going to use default implementation). We are trying to make Slim 3 as flexible as possible to the Container someone will choose without too much added complexity.

jaapverloop, line 46: I do understand your concerns and my intention is not to add any unneeded complexity.

> We are trying to make Slim 3 as flexible as possible to the Container someone will choose without too much added complexity.

Awesome, but that's exactly the reason for my suggestion. IMHO it's not just plug and play if you want to use
another implementation of `Interop\Container\ContainerInterface`. You have to register all the required services by yourself. Even if these services remain unchanged! I'm aware this is not a lot of work, just copy and paste. But I hope we're able to address this particular issue in an easy manner.

silentworks, line 46: This would not make a difference as **Container Interop** does not define how you set dependencies in the container but only if it `has` them and how to `get` them. So your suggestion would only mean a bunch of redundant classes, while the current implementation makes sure there is only 1 redundant class if you should choose another implementation.

jaapverloop, line 46: Yes, I know that. The ability to reuse the default service definitions would be beneficial IMHO. I'm open to other suggestions. But it seems I'm the only one bothered by this. Thanks anyway.

mnapoli, line 46: @jaapverloop I am, for example, one interested in providing integrations of Slim with PHP-DI. I don't expect Slim to configure my container as 1) Container Interop doesn't cover this and 2) this is very complicated and not optimal because all containers are different (which is exactly why Container Interop doesn't cover this). So the thing that makes sense is that I distributed a package that configures Slim with a PHP-DI instance preconfigured with all the necessary services.

What's important though will be to document the services that Slim expect, so that developers like me can create integrations easily.

jaapverloop, line 46: > So the thing that makes sense is that I distributed a package that configures Slim with a PHP-DI instance preconfigured with all the necessary services.

@mnapoli I guess it's a better solution indeed!

PR #1178, opened 2015-04-17
[IN PROGRESS] Resolve middleware on demand

lalop:
Resolve middleware only if the route matches the requested path.

PR #1180, opened 2015-04-17
Wrap callable resolution

lalop:
Wrap the callable resolution to resolve it when it's called

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -13,16 +13,16 @@ class CallableResolver
     
     protected $resolved;
     
-    public function __construct($to_resolve, Container $container)
+    public function __construct($toResolve, Container $container)
     {
-        $this->to_resolve = $to_resolve;
+        $this->toResolve = $toResolve;
         $this->container = $container;
     }
     
     public function resolve()
     {
         static $obj = null;
-        preg_match('!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!', $this->to_resolve, $matches);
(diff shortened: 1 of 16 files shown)


JoeBengalen:
You are mixing `to_resolve` and `toResolve`

lalop:
Oh yes, 
thank you

lalop:
I will be happy to have some review on this 
\cc @codeguy @akrabat

codeguy:
Deferring to @akrabat on this.

akrabat:
What does the change give us and how does it affect what's happening in #822 ? 

Also you'll need to rebase and take into account the change in #1184

akrabat:
OK, I've read #1178 now, so know why this PR exists. Just need the rebase with the #1184 change taken into account.

lalop:
Fun fact, I have added a test that pass thanks to #1178 but as a side effect of this patch.
We never checked if the string past to the resolver is callable.
Can we list all the callable things we have : 
- `closure`
- `'function name'`
- `[$obj, $method]`
- `'Class:method'`
- `'container_key:method'`

Maybe we should add `container_key` that can be a function after all.

Other point, I'm not so happy with the name I gave to the class name `CallableResolver` in front of `ResolveCallable` for the trait, any inspiration is welcome.

Slim/CallableResolver.php
```diff
@@ -0,0 +1,49 @@
...
+    
+    protected $resolved;
+    
+    public function __construct($toResolve, Container $container)
+    {
+        $this->toResolve = $toResolve;
+        $this->container = $container;
+    }
+    
+    public function resolve()
```
mnapoli, line 22: @lalop why is this method public? I don't see it being called anywhere except in this class.

lalop, line 22: yes, good point.
I've fixed that in an other pr https://github.com/slimphp/Slim/pull/1188
Thank you

mnapoli, line 22: :+1:

PR #1182, opened 2015-04-18
Various minor fixes: phpdoc, type-hints, imports

mnapoli:
Hi there, first contribution here :) No a big change, but I opened the project in PhpStorm and there were a few warnings here and there:
- wrong on missing phpdoc
- missing type-hints that would help autocompletion and static analysis
- unused class imports
- unused variables

I have fixed some of them, I didn't add extensive phpdoc (which would help even more in autocompletion and static analysis) because I wasn't sure if it was OK regarding the style guide of the project.

### DIFF
--- a/Slim/Autoloader.php
+++ b/Slim/Autoloader.php
@@ -19,7 +19,6 @@ class Autoloader
     {
         $className = ltrim($className, '\\');
         $fileName  = '';
-        $namespace = '';
         if ($lastNsPos = strrpos($className, '\\')) {
             $namespace = substr($className, 0, $lastNsPos);
             $className = substr($className, $lastNsPos + 1);
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -12,7 +12,6 @@ use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\UriInterface;
 use Psr\Http\Message\StreamInterface;
 use Slim\Interfaces\Http\HeadersInterface;
-use Slim\Interfaces\Http\CollectionInterface;
 
 /**
  * Request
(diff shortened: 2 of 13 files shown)


Slim/Autoloader.php
```diff
@@ -19,7 +19,6 @@ public static function autoload($className)
     {
         $className = ltrim($className, '\\');
         $fileName  = '';
-        $namespace = '';
```
mnapoli, line 22: unused outside of the `if`, immediately overwritten inside the `if`

Slim/Http/Request.php
```diff
@@ -308,7 +307,7 @@ public function withMethod($method)
      *
      * @param  null|string $method
      * @return null|string
-     * @throws InvalidArgumentException on invalid HTTP method.
```
mnapoli, line 311: without the `\` it was resolved in the current namespace

Slim/Http/Uri.php
```diff
@@ -284,7 +284,6 @@ protected function filterScheme($scheme)
      */
     public function getAuthority()
     {
-        $scheme = $this->getScheme();
```
mnapoli, line 287: unused

Slim/Interfaces/Http/HeadersInterface.php
```diff
@@ -8,8 +8,6 @@
  */
 namespace Slim\Interfaces\Http;
 
-use \Slim\Interfaces\Http\CollectionInterface;
```
mnapoli, line 11: not necessary, same namespace

Slim/Router.php
```diff
@@ -23,13 +23,15 @@ class Router extends \FastRoute\RouteCollector implements RouterInterface
...
+     *
+     * @var Route[]
      */
     protected $routes = [];
 
     /**
      * Named routes
      *
-     * @var null|array
+     * @var null|Route[]
```
mnapoli, line 34: I wasn't sure if it was OK with the project's coding style (I can remove this change), but documenting like that offers autocompletion and static analysis when iterating the array, which is pretty useful to detect potential errors or to navigate the code (Ctrl+click on a method call for example).

codeguy:
Thank you!

PR #1183, opened 2015-04-18
Fix README in regards to sample app

mariano:
The sample app provided in README.md does not take into account the `Response` interface for writing a response

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,9 @@ Requires PHP 5.4.0 or newer.
 
 ```php
 $app = new \Slim\App();
-$app->get('/hello/{name}', function ($request, $response, $args) {
-    echo "Hello, " . $args['name'];
+$app->get('/hello/{name}', function (\Slim\Http\Request $req, \Slim\Http\Response $res, array $args) {
+    $res->write("Hi {$args['name']}");
+    return $res;
 });
 $app->run();
 ```


nhymxu:
I don't think anyone want write route very long like this

mariano:
@nhymxu i actually agree. I added this because the current form is broken, as callbacks are expected to return the response object now.

@codeguy should this echo sample work as-is then? If so I can fix App so it allows such cases

akrabat:
Type hinting the $request and $response is optional.

mariano:
@akrabat ok. What about echo vs Response?

akrabat:
Personally, I much prefer returning the response as it's very clear what's happening.

tuupola:
Please bring back the echoing. Makes life much easier. For example before it was possible to ghetto debug with something like.

``` php
print_r($foo);
```

Now you have to do hoops and loops.

``` php
$response->write(print_r($foo, true));
return $response;
```

akrabat:
I'm confused.

``` php

$app = new \Slim\App();
$app->get(
    '/',
    function($request, $response) {
        echo "HERE";
        $response->write("<p>Hello World</p>");
        return $response;
    }
);
$app->run();
```

works as you'd expect: 

screen shot 2015-04-18 at 19 38 06

alexantr:
Without `return $response;` we have fatal error:
`Call to a member function getStatusCode() on a non-object in .../vendor/slim/slim/Slim/App.php on line 354`
So sad.

codeguy:
haha there are vocal people on both sides. Some for `echo`, some for strict PSR-7 and returning the response object. I'll figure out if I want to restore `echo` before v3 release.

mariano:
@codeguy I'm also pro-bringing back echo.

I can get a PR ready if you like

akrabat:
FWiW, I'm not a fan of bringing back echo as it makes the controller callable different from middleware callables. 

I like the consistency of "always return a response" as it's simple to remember.

tuupola:
You can still return just keep echoing as an alternative. It is a tradeoff between usability and by the book engineering. Those who want to return can still return since it is easier to remember. Those who wish to echo still can echo.

mnapoli:
Why not keep support for `echo` by not erroring if the controller returns `null`?

The documentation could teach the right way (return a response), but the application wouldn't crash if you don't return one. I guess that could help a lot of people and keep the framework easy to use (debugging was one of the use case mentioned). It might also help for the migration to v3.

silentworks:
I also agree with @tuupola and @mnapoli about getting echo back in there. As the person who deals with most of the support forum requests, I know this is going to become something that will be asked more than 100 times and the cause of people's frustration with the framework.

I think from early days we have always kept Slim somewhat non opinionated and feel this would be making it a little more opinionated for no major gain.

PR #1195, opened 2015-04-25
handler return value in example

pine3ree:
Updating the hello example in readme to reflect latest slim-develop changes.
Route handlers must return a \Psr\Http\Message\ResponseInterface since they are part of the middleware stack.


mnapoli:
I think this was already done in #1187?

pine3ree:
@mnapoli 

oh right, thanks. i totally missed that. I pulled the latest dev about 30-40 minutes ago..

(updated) 
got it! 
Updating slim/slim dev-develop (d929342 => a18ed21)...a matter of minutes i think...

PR #1196, opened 2015-04-25
URL decode router parameters before setting into the request object

akrabat:
Sorry @codeguy, I put the url-decoding in the wrong place in PR #1194.

This PR also ensures that the `Request`'s `attributes` and the `$args` parameter that is passed to the route callable are the same at the point that we dispatch the callable. This means that route middleware can sanitise the `attributes` and be sure that `$args` also has the sanitised data.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -414,7 +414,12 @@ class App extends \Pimple\Container
     {
         $routeInfo = $this['router']->dispatch($request);
         if ($routeInfo[0] === \FastRoute\Dispatcher::FOUND) {
-            return $routeInfo[1]($request->withAttributes($routeInfo[2]), $response, $routeInfo[2]);
+            // URL decode the named arguments from the router
+            $attributes = $routeInfo[2];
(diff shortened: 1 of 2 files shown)


Slim/App.php
```diff
@@ -414,7 +414,12 @@ public function __invoke(RequestInterface $request, ResponseInterface $response)
     {
         $routeInfo = $this['router']->dispatch($request);
         if ($routeInfo[0] === \FastRoute\Dispatcher::FOUND) {
-            return $routeInfo[1]($request->withAttributes($routeInfo[2]), $response, $routeInfo[2]);
+            // URL decode the named arguments from the router
+            $attributes = $routeInfo[2];
+            array_walk($attributes, function(&$v, $k) {
+                $v = urldecode($v);
+            });
+            return $routeInfo[1]($request->withAttributes($attributes), $response);
```
mnapoli: Here this example wouldn't work:

```php
    ...
});
```

The `$args` parameter is not passed.

PR #1200, opened 2015-04-30
Allow custom errors in debug mode

chasepeeler:
A very small change that adds a new configuration option, "custom_errors," so that you can still use your custom errors even in debug mode. The option must be set to "true" to work. If the option is set to anything else (or not set at all) then it will continue to operate as before (custom error handling is ignored in debug mode)

### DIFF
--- a/Slim/Slim.php
+++ b/Slim/Slim.php
@@ -1371,7 +1371,7 @@ class Slim
         } catch (\Slim\Exception\Stop $e) {
             $this->response()->write(ob_get_clean());
         } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


akrabat:
We're only accepting security fixes to Slim 2 now. Is this relevant for what's in develop?

chasepeeler:
I don't know. I haven't had a chance to dive into v3 that much to determine that. It definitely would not be a straight-forward merge even if it was.

I understand if you're only doing security fixes, I thought that was the case, but figured it wouldn't hurt to create a pull request just in case.

Issue #1203, opened 2015-05-02
Middleware returns int(1)


realart:
Disclaimer: Most likely not an issue, just lack of insight on my part.

I am using v3 development version and my project relies on things like:
- Subrequests (killer feature, makes it a HMVC framework)
- Application-wide middleware for things like wrapping page routes inside layout views etc.

All was fine and dandy until I updated to latest commit of the develop branch, and middleware calls started to act strange:
- Return integer 1, instead of Response object
- Leak output

For example, a call like this "$newResponse = $next($req, $res)" inside application middleware would:
- Return integer 1 as the "$newResponse" variable
- Dump the output (the contents of what should have been response body) of the "$next" callable to the screen

akrabat:
Are you able to create a small example project that shows the problem?

realart:
If I invoke /test, it prints "some nice output" and crashes with exception:

```
UnexpectedValueException: Middleware must return instance of \Psr\Http\Message\ResponseInterface'

#0 C:\projects\flsite\vendor\slim\slim\Slim\MiddlewareAware.php(106): Slim\App->Slim\{closure}(Object(Slim\Http\Request), Object(Slim\Http\Response))
#1 C:\projects\flsite\vendor\slim\slim\Slim\App.php(346): Slim\App->callMiddlewareStack(Object(Slim\Http\Request), Object(Slim\Http\Response))
#2 C:\projects\flsite\index.php(16): Slim\App->run()
#3 {main}
```

Index.php

```
require 'vendor/autoload.php';

$app = new \Slim\App();

$app->get('/test', function($req, $res, $args) {
    echo 'some nice output';
});

$app->add(function($req, $res, $next) use($app) {
    $newResponse = $next($req, $res);
    return $newResponse;
});

$app->run();
```

akrabat:
You have to return a Response from the route's action callable:

```
$app->get('/test', function($req, $res, $args) {
   return $response->write('some nice output');
});
```

realart:
But this will force me to manually do output buffering (start my callable with ob_start and end with writing contents of ob_get_clean into response body) in each and every route callable. Previously this was handled automatically by framework.

My applications route definitions are structured like this:

```
$app->get('/', function($req, $res, $args) { return include('pages/home.php'); });
$app->get('/search', function($req, $res, $args) { return include('pages/search.php'); });
$app->get('/fragments/users', function($req, $res, $args) { return include('fragments/users.php'); });
etc.
```

[...]

Issue #1198, opened 2015-04-26
Thoughts


pine3ree:
@codeguy 
Hi Josh,
i've spent some time thinking about where we are moving towards with this new psr-7 / middleware interoperability revolution.

I believe we are putting too much emphasis on request/response paradigm. That is the basis of a web application, but shouldn't be the goal.

As an example, slim-3-dev now have url handlers behaving like middlewares and returning responses.
I'm doing that too....but...

back in the (not so) old times....

we returned a string, (maybe calling "The View" render method with a template-vars parameters.
That string could be then, based on how we setup the view and on what rendering engin we(built and)  used, be decorated with a layout.
Now we can (and we also have full responsibility) of manipulating the whole response. That's cool, but that should be only an option not a must do in my opinion.
If we need to render a partial view into a layout, we would need to do a lot of extra work in every controller action (route handler) and then fill the response body with the result of rendering + decoration.
I believe this pattern is going distract us developers from the real goal of the action, the business logic.

In short, the old way offered a bit more automation (most of the time implemented by us as well) in building an application and let us focus on the real  job.

what's your opinion about this?

best regars

akrabat:
Fixed in #1206.

Issue #1189, opened 2015-04-23
is output buffering disabled in version3?


omarkhd:
The example you have in the main page, does not work. You can echo something in a route, but then you got a Fatal error because the route is not returning the response. And to display something, you have to call the write method. Is this how it is supposed to work?

geggleto:
It is currently how it works, there are talks about re-enabling output buffering support.

omarkhd:
If you decide to re-enable it, I'd be glad to help.

akrabat:
Fixed in #1206.

PR #1206, opened 2015-05-04
Reinstate output buffering for the route callable

akrabat:
This issue has come up enough times since removal that I believe it should be reinstated.

Any `echo` calls in a route callable are now again captured in an output buffer and appended to the response object. If a response object is returned from the route callable, then this new object is used for appending to and is returned. This allows you to echo from a route callable, but also set the status code. We also allow for returning a string fro the route callable and appending that to the response.

See #1203, #1189, #1198

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -194,8 +194,32 @@ class Route implements RouteInterface, ServiceProviderInterface
      */
     public function __invoke(RequestInterface $request, ResponseInterface $response)
     {
-        $function = $this->callable;
+        // invoke route callable
+        try {
+            ob_start();
+            $function = $this->callable;
+            $newReponse = $function($request, $response, $request->getAttributes());
+            $output = ob_get_clean();
+        } catch (\Exception $e) {
+            ob_end_clean();
+            throw $e;
+        }
+        
+        // if route callback returns a ResponseInterface, then use it
+        if ($newReponse instanceof ResponseInterface) {
+            $response = $newReponse;
+        }
 
-        return $function($request, $response, $request->getAttributes());
+        // if route callback retuns a string, then append it to the response
+        if (is_string($newReponse)) {
+            $response->getBody()->write($newReponse);
+        }
+        
+        // append output buffer content if there is any
+        if ($output) {
+            $response->getBody()->write($output);
+        }
+
+        return $response;
     }
 }
--- a/tests/RouteTest.php
+++ b/tests/RouteTest.php
@@ -148,4 +148,80 @@ class RouteTest extends PHPUnit_Framework_TestCase
 
     // TODO: Test adding controller callables with "Foo:bar" syntax
 
+    /**
+     * Ensure that the response returned by a route callable is the response
+     * object that is returned by __invoke().
+     */
+    public function testInvokeWhenReturningAResponse()
+    {
+        $callable = function ($req, $res, $args) {
+            return $res->write('foo');
+        };
+        $route = new Route(['GET'], '/', $callable);
+
+        $env = \Slim\Http\Environment::mock();
+        $uri = \Slim\Http\Uri::createFromString('https://example.com:80');
+        $headers = new \Slim\Http\Headers();
+        $cookies = [];
+        $serverParams = $env->all();
+        $body = new \Slim\Http\Body(fopen('php://temp', 'r+'));
+        $request = new \Slim\Http\Request('GET', $uri, $headers, $cookies, $serverParams, $body);
+        $response = new \Slim\Http\Response;
+
+        $response = $route->__invoke($request, $response);
+
+        $this->assertEquals('foo', (string)$response->getBody());
+    }
+
+    /**
+     * Ensure that anything echo'd in a route callable is added to the response
+     * object that is returned by __invoke().
+     */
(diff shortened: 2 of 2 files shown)


tuupola:
Yeah! :+1:

akrabat:
Note, that I also looked at implementing this as Middleware that we ship as part of Slim 3, as shown here: https://github.com/slimphp/Slim/compare/develop...akrabat:output-buffering-middleware?expand=1

The main problem with this approach is that we'd be breaking the convention that the route callable always returns a `Response`, which is why I chose to implement this directly in `Route::__invoke()`. 

Also, the user would also have to ensure that the OutputBuffer middleware was the innermost one to avoid errors in any other middleware, which may increase the support burden, which is another reason why implementing it directly seemed better.

pine3ree:
Another way could be creating a Route descendant which has an default output buffering capturing  middleware (and why not also a string capturing) as its innermost middleware.

I had the same doubts and the same needs about ob and string capturing: we are now all about request-response(-next)=>response. But i guess this is what iteroperability needs: a common and easy pattern. @codeguy  Josh adopted this pattern and applied it to the Route cass too. Basically everything, even a route handler (and the handler wrapper in the route mw stack), is a middleware. That simplifies development design and code.

In any case allowing any route handler to return something else (even void) would sound like breaking the law a little.

In my case, because i also wanted route attributes as first paramaters in the route handler, i had to remove the middleware pattern in the route implementation. It works, it allows me to capture echoed string and returned string. But the resulting code, even if you can still call it clean, is a bit less elegant than the slim way. 

I believe that interoperability will force us to rethink the structure of http-application. We just need to move application business logic elsewhere and let the controllers do what they are really supposed to do: handle requests, return responses delegating what happens in between to other methods or objects.  

``` php
// pseudo(i-mean-it)-code
$app->get('/catalog/products/{page}', function ($request, $response, $args) use ($app) {
    $response->write($app['productReadController']->index((int) $args['page']));
    return $response;
});
```

Here a productReadController could be a standard mvc framework controller instance where actions are responsible of building the html page with a view renderer etc etc in a more traditional way. $app or any other service locator is responsible of injecting controller class dependencies.

In a restful application you would probably do the rendering work inside the route handlers, as you just need to send json encoded data and headers. In any case, little work inside the middleware, just give back a response.

kind regards

akrabat:
Yes, I came to the same conclusion that the result of `Route::__invoke()` must be a  `Response` for consistency.

A different `Route` implementation would be an reasonable alternative, though the docs for how to enable it would have to be easy to find as it's clear that echo'ing in the route callable is a very common use-case. This is why I went for putting it back into the default `Route` for this PR.

pine3ree:
@akrabat 
In my implementation (it started basically as a fork of slim 2.4 with added layers to allow a zf2-like mvc module structure) i have the router instantiated with a $routeClass parameter (st like the objectPrototype in zf2 db-result) which is used by the router when it acts as a route factory. It defaults to the basic namespace\routing\Route class. When I want to use the zf2-mvc-like-module layer, I start a namespace\mvc\App class which as a default not overwrit(e?)able configuration instantiates the router with a namespace\routing\mvc\Route. This class has one overriden method to intercept route mvc parameters (basically module, controller, action names).

maybe we should ask Josh if he is willing to implement something similar, if it's worth doing it.

Otherwise I believe it will only necessary to have a createRoute($methods, $pattern, $handler) method inside the router.
We could extend the Route class (to incude an always attached capturing middlware) and the Router (we would need to override just the createRoute method) and use pimple standard way to override/extend the router service.

pine3ree:
...but again, it's still like breaking the contract, as something in the chain (the route handler) will be allowed to return something that is not a response.
:-)

lalop:
I think it's a great enhancement, particularly for new users but maybe you can add it as an option by a settings or by adding the output buffering in the container.

With this pr I can have some strange behaviour, for exemple if I have a route like that :

```
$app->get('/', function($req, $resp) {
$resp->write('hel');
echo 'lo';
$resp->write(' world');
return $resp;
});
```

The final response will contain `hel worldlo`.

Knowing how work the response object, I prefer to see an exception when somehing is echoed.

tuupola:
Above behaviour is expectable and exactly the same as how Slim 2 works now. If you mix echoing and writing  via response object you cannot expect Slim to know the order of those calls.

``` php
$app->get("/hello", function() use ($app) {
    $app->response->write("hel");
    echo "lo";
    $app->response->write(" world");
});
```

Above yields `hel worldlo` as expected.

silentworks:
As @tuupola stated above, this is expected behaviour, I think we shouldn't be constraining the user as to how to return their response, we should only guide them. So in all documentation we should strictly use `return $response->write();`. A new user is most likely to follow what the docs say than going their own way, if we make this a recurring pattern in tutorials/open source projects we produce also it will make it easier for everyone to follow.

lalop:
I understand your point, and the fact it's the default is a good thing.
I notice that we can make it unpluggable too for more experienced user.

tuupola:
It is not about experience. I consider myself a non-beginner and I hate the hoops and loops I now have to make to get some output from Slim. Draconian measures such as throwing an exception if someone echoes can be included but it must be optional and turned off by default.

PR #1207, opened 2015-05-04
ordered request (route) attributes (params)

pine3ree:
Since we can easily use getAttributes() on the 1st argument, i believe it would be better to have any route attributes as orderer optional handler arguments (ala old style handlers).
example:

``` php
$handler = function (RequestInterface $request, ResponseInterface $response, $sort = 'firstname', $dir='asc') {
    //....
   return $response;
}
```

PS
That said, i believe we will seldom use extra handler params since we are now all about the req, res, res bermuda triangle...the mistery of what happens inside being your responsibility to implement.


akrabat:
We have no current plans to change from `$args`.

PR #1208, opened 2015-05-04
containerize callable resolver

lalop:
Add the callable resolver to the container.

### DIFF
--- a/Slim/CallableResolverAware.php
+++ b/Slim/CallableResolverAware.php
@@ -35,7 +35,7 @@ trait CallableResolverAware
             } else {
                 throw new \RuntimeException('Cannot resolve callable string');
             }
-            $resolver = $container['callableResolver'];
+            $resolver = $container->get('callableResolver');
             $resolver->setToResolve($callable);
             $callable = $resolver;
         }
--- a/Slim/CallableResolverAware.php
+++ b/Slim/CallableResolverAware.php
@@ -17,7 +17,7 @@ use Interop\Container\ContainerInterface;
  * into a closure. This class is an implementation detail and is used only inside
  * of the Slim application.
  */
-trait ResolveCallable
+trait CallableResolverAware
 {
     /**
      * Resolve a string of the format 'class:method' into a closure that the
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -36,7 +36,6 @@ class App
         add as addMiddleware;
     }
 
-
     /**
      * Current version
      *
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -30,7 +30,6 @@ use Interop\Container\ContainerInterface;
  */
 class App
 {
-
     use CallableResolverAware;
     use MiddlewareAware {
         add as addMiddleware;
@@ -61,7 +60,6 @@ class App
      */
     public function __construct($container = [])
     {
-
         if (is_array($container)) {
             $container = new Container($container);
         }
@@ -112,7 +110,6 @@ class App
     public function __isset($name)
     {
         return $this->container->has($name);
-
     }
 
     /********************************************************************************
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -113,6 +113,7 @@ class App
     public function __isset($name)
     {
         return $this->container->has($name);
+
     }
 
     /********************************************************************************
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -36,6 +36,7 @@ class App
         add as addMiddleware;
     }
 
+
     /**
      * Current version
      *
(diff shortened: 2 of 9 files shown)


lalop:
I think this should come with an interface.

lalop:
Adding the `CallableResolver` in the container makes more difficult to build and maintain test that uses the resolver but that doesn't use `App`.
Since all the default services are added in the container in the App class if you don't use this class in a test that use some service we have to add each service manually.
Much of the test on the `CallableResolver` was done twice, in `Route` and  in `App`, now all this tests are made directly on the `CallableResolver` so we don't need them any more.

lalop:
Next step can be injecting a kind of callableResolver Builder in the Route instead of the container.

JoeBengalen:
Is this going to be implemented in v3.0.0?

lalop:
@JoeBengalen it seems yes

lalop:
I don’t see a good way to inject the callable resolver service ( the closure in the container ) into the resolver trait and keeping it as straightforward as possible.
help is welcome :)

rivetchip:
It may be a detail, but isn't it better to not use the `final` keyword for classes, and use `protected` over `private` for props/function ? So that we can more easily extend those classes and access props.

lalop:
I think you are asking that for the Container class, this patch is merged on develop today, it appears here because I misuse `git rebase`.
Anyway you will found an answer here https://github.com/slimphp/Slim/pull/1179#issuecomment-102471067

Slim/CallableResolverAware.php
```diff
@@ -37,16 +30,16 @@
...
-                $container = $this;
-            } elseif($this->container instanceof Container) {
+            if ($this->container instanceof ContainerInterface) {
                 $container = $this->container;
             } else {
                 throw new \RuntimeException('Cannot resolve callable string');
-            }            
-            return new CallableResolver($callable, $container);
+            }
+            $resolver = $container['callableResolver'];
```
akrabat: You need to use the container interface's get() method.

lalop: done

lalop:
Any news here ?

akrabat:
Intend to get to it this week.

akrabat:
@lalop would be handy if you could rebase it though…

lalop:
@akrabat done

mnapoli:
@lalop How are you rebasing? It is weird there are "rebase" commits in the pull request, whereas rebasing shouldn't introduce new commits. Also the PR is still in conflict with develop.

lalop:
hello @mnapoli,

Yes I use rebase, but I have some conflict while rebasing and maybe I have a bad usage of rebase while resolving them.

It seems that my local develop branch differs from the official, I think this is the problem.

lalop:
@mnapoli How and where do you see conflict with develop ?
I merge container-callable into develop whithout conflict

lalop:
I make this test whithout conflict 

```
$ git checkout -b origincc origin/container-callable
$ git checkout -b updev upstream/develop
$ git merge origincc
Updating 45cc95c..98fe66c
Fast-forward
 Slim/App.php                                            |   4 +++-
 Slim/CallableResolver.php                               |   9 +++++++--
 Slim/{ResolveCallable.php => CallableResolverAware.php} |   6 ++++--
 Slim/Container.php                                      |  10 +++++++++-
 Slim/Interfaces/CallableResolverInterface.php           |  10 ++++++++++
 Slim/Route.php                                          |   2 +-
 tests/CallableResolverTest.php                          |  14 +++++++-------
 tests/RouteTest.php                                     | 104 +--------------------------------------------------------------------------------------------------
 8 files changed, 42 insertions(+), 117 deletions(-)
 rename Slim/{ResolveCallable.php => CallableResolverAware.php} (86%)
 create mode 100644 Slim/Interfaces/CallableResolverInterface.php
```

Am I missing something ?

mnapoli:
This is what i see on the pull request (here on this page on GitHub):

> This pull request contains merge conflicts that must be resolved.

---

> Yes I use rebase, but I have some conflict while rebasing and maybe I have a bad usage of rebase while resolving them.

Usually when you rebase git goes through commit by commit and stops when there is a conflict. Then you can solve the conflict and continue rebasing. This results in no extra commit: when you solved the conflict you actually edited an old commit.

> It seems that my local develop branch differs from the official, I think this is the problem.

Yes that doesn't sound good. Usually when you have rebased, you want to "force push" because you need to overwrite the old commits with your new ones (when you edit a commit while rebasing it becomes a new commit with a different SHA1). Then the local branch and remote branch are the same.

lalop:
Ok, I think I will make a new PR all clean

lalop:
@mnapoli, thank you for this information.
Anyway I don't understand why I can't reproduce this conflicts.

lalop:
I finally found the issue, my upstream was linked on the @codeguy original repository wich is not anymore uptodate.

PR #1209, opened 2015-05-04
rename callable trait

lalop:


lalop:
I've accidentaly add this to #1208

lalop:
reopen since #1208 will be closed

PR #1213, opened 2015-05-05
PROPOSAL: allow string http-methods definitions

pine3ree:
Just a proposal, not sure if it would be much overhead.
It basically allows the following route defintion alternatives

``` php
$app->map('GET,POST', '/books', function ($request, $response, $args) {
    //...
    return $response;
});
$app->map('GET', '/books', function ($request, $response, $args) {
    //...
    return $response;
});
```

If accepted, maybe a better place for this change would be the router map method itself.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -243,14 +243,21 @@ class App extends \Pimple\Container
     /**
      * Add route with multiple methods
      *
-     * @param  string[] $methods  Numeric array of HTTP method names
-     * @param  string   $pattern  The route URI pattern
-     * @param  mixed    $callable The route callback routine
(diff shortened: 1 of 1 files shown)


silentworks:
In Slim 2 we had this and then we added array mapping, I can't see any great benefits in adding this back in, rather it could cause confusion if in a team some folks were using strings while others using array. I would say lets keep it simple and keep what we currently have only.

pine3ree:
hi @silentworks 
since nobody else subscribed or commented this proposal i believe you can safely close it.
or i will wait another 10 days and then i'll close it myself.
kind regards

PR #1215, opened 2015-05-05
respond method

kminek:


Slim/App.php
```diff
@@ -393,11 +380,35 @@ public function run()
...
 
     /**
+     * Run application
+     *
+     * This method traverses the application middleware stack,
+     * and it returns the resultant Response object to the HTTP client.
+     */
+    public function run()
+    {
+        static $responded = false;
```
JoeBengalen, line 394: Shouldn't this line be inside the `respond` method instead of being passed along as a param?

kminek, line 394: maybe you are right i am open to corrections, i came up with this method because i needed a way to return slim's error page inside register_shutdown_function, i think this method would be helpful in other situations also

JoeBengalen, line 394: I do agree separating running the application and responding into two separate methods is a good idea. I just do not see why '$respond' is still declared in `run`. Maybe you did this op purpose as you saw a use case for this, but if not you can move it.

akrabat:
What's the problem that is solved by this PR?

JoeBengalen:
Separation of concerns. Separating running the application and sending the response. Makes it easier to override one of those.

silentworks:
@JoeBengalen in that case, shouldn't `respond()` be a protected method?

kminek:
there is often a need to render user-friendly page in case of fatal-error. i thought it would be nice to render response from $app['errorHandler'] like this:

``` php
ini_set('display_errors', 0);

$app = new Slim\App();

register_shutdown_function(function () use ($app) {
    $error = error_get_last();
    $exception = new Exception($error['message']);
    $response = $app['errorHandler']($app['request'], $app['response'], $exception);
    $responded = false;
    $app->respond($response, $responded);
});

$app->get('/', function ($req, $res) {
    fatalerror(); // triggers fatal error
    $res->write('Something');
    return $res;
});

$app->run();
```

i created this PR just as a example, it can be canceled but i think
such separation of concerns is needed

JoeBengalen:
Another solution would be to put it into a separate Responder class:

https://github.com/slimphp/Slim/compare/develop...JoeBengalen:feature-responder?expand=1

kminek:
@JoeBengalen nice, this looks good :clap:

akrabat:
Why does the `static $responded` exist at all?

We don't call `run()` twice and if someone does that, that's their problem, surely?

codeguy:
@akrabat To avoid errors caused by sending headers multiple times in unit tests.

akrabat:
Original branch is gone, so I have reimplemented in 48a16a51a54ba57eec6f4543923f75cbfb5f9566

PR #1217, opened 2015-05-06
avoid creating stack for every route

pine3ree:
Please verify:
- add() will create the stack with this route as the kernel when adding a middleware
- if the route is a match, the run() method (=the FastRoute route handler) will call callMiddlewareStack() that will seed a new stack if it wasn't created before.

if no route middleware is added, the stack will be created only for the matching route with this route as the kernel and only middleware. (We could also skip stack creation in this case, and invoke the route directly...in my implementation with a separate class stack, the stack is invokable and when invoked it start calling the middlewares. So for a MiddlewareAware - StackKernelTrait like object i set a $runner var to the stack if present or the object itself and invoke the $runner)

``` php
//pseudo-code
$runner = $this->getStack() ?: $this;
$response = $runner($req, $res);
```

regards

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -64,7 +64,6 @@ class Route implements RouteInterface, ServiceProviderInterface
         $this->methods = $methods;
         $this->pattern = $pattern;
         $this->callable = $callable;
-        $this->seedMiddlewareStack();
     }
 
     /**


lalop:
You are right, seedMiddleware is useless, indeed App does'nt call it.

pine3ree:
Hello @lalop 
Actually the application run() method calls:

callMiddlewareStack()

which in turn creates a new stack for the application middleware using seedMiddleware() sets itself (the application) as the kernel.

For the application stack is not that much of a problem since it's done once per request.
But I believe it would be better not to create a middleware stack (the process is the same as for the application) for every route if no route middleware are used, i.e. a middleware stack with only the kernel.

This would benefit cases in which the numer of routes is significant.

Anyway the common design (and the consequential use of a common trait) is very clean and simple and (node.js-like) elegant.

lalop:
Hello @pine3ree ,
I see your point, but if this pr is merged, route whithout middleware will have a stack only if it's called since the stack will be created in `Route::run`, isn't that the target ?

pine3ree:
Hello @lalop ,

yes that's right. As for the application, the stack will be created only for the matching route when the run method is called or whenever a middleware is added tot he route. Maybe i misunderstood your comment. What i meant is that seedMiddleware is still called by the app (and by the matching route), but not invoked directly. It's indeed "lazily" invoked. So I believe there is no need to call it in the route constructor and it would be better to let only the matching route or routes having middlewares build their own stacks when needed (within the run() method or the add() method which calls the aliased method addMiddleware == trait MiddlewareAware::add, which in turn calls its own method seedMiddleware)

kind regards

lalop:
hum... either we are agree, the call to `seedMiddleware` is useless and your pr is welcome imho or I don't understand :)

pine3ree:
@lalop 
I believe we basically agree. :-)

At least we agree on the fact that seedMiddlewareStack() shouldn't be called directly in the route constructor. If we avoid that, we would have just 2 middleware stacks in most of the cases. 
1 belonging to the application, 1 to the matching route. 

We will have 1 more route stack per subRequest in case of subrequest
We will have 1 more rote stack creation for every route that has middlewares

If we call it DIRECTLY in the route constructor, we will create a stack for every route, which is not needed.

In your comment you said: "indeed App does'nt call it".

What i meant is that the App doesn't call it directly, but it still calls it indirectly, as it's invoked inside the MiddlewareAware trait methods callMiddlewareStack(...) and add(...).

When you call add(...)/addMiddleware(...) or run(...) from the app or from a route, you call those 2 trait methods and indirectly you will be also calling seedMiddlewareStack to create the stack.

kind regards

lalop:
So we are completely agree

PR #1219, opened 2015-05-07
Stack kernels MUST return a Response

pine3ree:
Please verify:
I believe that in the current implementation even route handler and the application, both kernel of their own middleware stack, are expected to return a psr-response-interface.

### DIFF
--- a/Slim/MiddlewareAware.php
+++ b/Slim/MiddlewareAware.php
@@ -59,15 +59,8 @@ trait MiddlewareAware
         if (is_null($this->stack)) {
             $this->seedMiddlewareStack();
         }
-        $next = $this->stack->top();
-        $this->stack[] = function (RequestInterface $req, ResponseInterface $res) use ($callable, $next) {
-            $result = $callable($req, $res, $next);
-            if ($result instanceof ResponseInterface === false) {
-                throw new \UnexpectedValueException('Middleware must return instance of \Psr\Http\Message\ResponseInterface');
-            }
 
-            return $result;
-        };
+        $this->addMiddlewareLayer($callable);
 
         return $this;
     }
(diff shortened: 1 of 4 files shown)


lalop:
I don't understand what you fix here. 
Route and App kernel have their own wrapper that returns a response object.

pine3ree:
Hi @lalop 
try this:

``` php
$app->get('/', function ($req, $res, $args) {
    echo 'Hello';
});
$app->add(function ($req, $res, $next) {
    $res->write('<p>APP-PRE</p>');
    if ($next) {
        $res = $next($req, $res);
    }
    $res->write('<p>APP-POST</p>');
    return $res;
});
```

and then this:

``` php
$app->get('/', function ($req, $res, $args) {
    return $res->write("<p>HELLO</p>");
});
$app->add(function ($req, $res, $next) {
    $res->write('<p>APP-PRE</p>');
    if ($next) {
        $res = $next($req, $res);
    }
    $res->write('<p>APP-POST</p>');
});
```

the second example will catch an exception,
the first will result in an unhandled error.

The application and the route handler use the MiddlewareAware trait and they are added directly into the stack without a callable wrapper.
I believe they were not wrapped because they don't really need to acces the $next layer, as they are the last (inner) layer in their own stack. But they are supposed to return a ResponseInterface anyway.

lalop:
I think this issue is solved by @akrabat https://github.com/slimphp/Slim/commit/2f02b4ba7ae3b72e781872d0d41bd4e19b74e711
Anyway You are right any middleware layer have to return a psr-response-interface and this could help some developer who want to extend Slim. 
Maybe you can change the error message to be more generic.

pine3ree:
@lalop 
Yes @akrabat  PR forces "the" route to be a "good" kernel.
As you stated, formally even the kernel should be restricted (and checked) using the same rules of any other middleware in case someone needs to extend slim router/route classes. The only difference with stack kernels is that there is no next (inner) layer.

I'll leave the PR open as a mere reference to this formali(sm|ty) for the repo managers.

akrabat:
I like the correctness, but this PR causes the unit tests to fail.

pine3ree:
Hello @akrabat ,
I updated the 4 failing tests, i had to extract the $callable by name from the closure "use ()" parts as there is no other way (as far as i know) using reflection. 
I rename $stack to $stackable (i could have used $mwAware as well) in the MiddlewareAwareTest as it is less confusing and more precise.
I also renamed the RouteTest and AppTest methods replacing "Is" with "Invokes" as app, route and middleware callables are now all wrapped inside middleware closures (the ones which are truly added to the stack) .
(...sorry for the many commits..with renaming mistakes, i did it online as i had my patch repo is cloned in other pc and it's kind of late here now)

best regards ...and good night

akrabat:

pine3ree:
Hello @akrabat ,

at the time i added the tests all was working fine. I'll have to check for what was changed afterwards.

PR #1223, opened 2015-05-11
unneeded string casting?

pine3ree:
str_replace should always return a string for a string subject/haystack. $scheme is a string because it is casted in the previous line.


geggleto:
It returns mixed string or array needs to be casted

pine3ree:
@geggleto 
I believe it returns an array only when an array subject is provided as argument, the search and replace being computed on every element in that case, but if i am missing a case, please give me an example.
kind regards

PR #1224, opened 2015-05-11
same test called twice?

pine3ree:
Hielo Josh, please verify:

the getPort() method internally calls ::hasStandardPort(). It returns null for standard port/scheme combination. I think we just need to check for non null port values here.

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -302,9 +302,8 @@ class Uri implements \Psr\Http\Message\UriInterface
         $userInfo = $this->getUserInfo();
         $host = $this->getHost();
         $port = $this->getPort();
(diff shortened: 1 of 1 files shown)


JoeBengalen:
Correct, the check is left from an earlier version of the psr-7 I guess.

nunocodex:
I think not want add the port if the port is standard, in your case you add in any case the port.

The mix of all 2 code is:

```
return ($userInfo ? $userInfo . '@' : '') . $host . (false === $this->hasStandardPort()) ? ':' . $port : '');
```

JoeBengalen:
@namaless I agree the port should not be included if default, as also stated in the psr-7 specs, but the default port check happens withing the `getPort()` method now.

PR #1227, opened 2015-05-13
little optimization on body parsing

lalop:


geggleto:
:+1:

Slim/Http/Request.php
```diff
@@ -1062,22 +1062,28 @@ public function withBody(StreamInterface $body)
...
-        if ($this->bodyParsed) {
+        if (isset($this->bodyParsed)) {
             return $this->bodyParsed;
         }
 
-        if (!$this->body) {
-            return;
-        }
-
+        $this->bodyParsed = null;
```
akrabat, line 1199: Surely if we've got to this line, then bodyParsed is already null?

akrabat:
Also, can you rebase?

Slim/Http/Request.php
lalop, line 1199: Yes but now it's set so the next time `getParsedBody` is called we will not try to parse the body.

lalop:
@akrabat yes I can and it's done :)

Slim/Http/Request.php
JoeBengalen, line 1199: What do you mean? Null value will still return false in the isset check (line 1195).

lalop, line 1199: Oh yes you are right, I didn't know that.
Damn ! Is there a way to do that whithout setting a second property to know if the body is parsed ?

Slim/Http/Request.php
```diff
@@ -1192,24 +1192,33 @@ public function withBody(StreamInterface $body)
...
             $parsed = $this->bodyParsers[$mediaType]($body);
 
             if (!is_null($parsed) && !is_object($parsed) && !is_array($parsed)) {
                 throw new RuntimeException('Request body media type parser return value must be an array, an object, or null');
+            
+            $body = (string)$this->getBody();
+            
+            if ($body) {
+            
+                $parsed = $this->bodyParsers[$mediaType]($body);
```
akrabat, line 1212: I'm unclear what's going on here as this is a repeat of line 1203.

akrabat:
Tests are failing due to a syntax error. I think it's a missing closing brace.

akrabat:

PR #1125, opened 2015-03-23
Implement Request::$fileParams

JoeBengalen:
Its is only a simple collection of $_FILES. No escaping or anything.

Is this enough @codeguy?

Working on #1099

PR #1228, opened 2015-05-13
Update to the latest PSR-7 changes

JoeBengalen:
Work in progress to update to the latest psr-7 changes (tag 0.11.0)

Apart from that I also updated to the ServerRequestInterface instead of the RequestInterface, because the latter is ment for client side implementations.

TODO
- ~~Update `Request` class~~
- ~~Update `Response` class~~
- ~~Update `Uri` class~~
- ~~Update `Body` class~~
- ~~implement `Request::withUri()` $preserveHost option~~
- Implement UploadedFiles in `Request`
- ~~Implement the $reasonPhrase in `Response` class~~
- ~~Fix Uri path handling.~~
- ~~Add warnings to non psr-7 methods~~


JoeBengalen:
There are several methods in the `Request` class that are not in the psr-7 spec. Be aware those cannot be used in the application if you want to keep interchangeability of other psr-7 compliant implementations. Maybe these methods should be listed in the docs? Or make a note in those docblocks?
- getOriginalMethod
- isMethod
- isGet
- isPost
- isPut
- isPatch
- isDelete
- isHead
- isOptions
- isAjax
- isXhr
- getContentType
- getMediaType
- getMediaTypeParams
- getContentCharset
- getContentLength
- withAttributes
- getIp

Same goes for the `Response`, `Uri` and `Body` classes.

JoeBengalen:
Need some input from @codeguy for the other points

Handling the path in the `Uri` object. How do you want to implement this? What is exactly the use case of the basePath?

```
      (from withPath())
     * The path can either be empty or absolute (starting with a slash) or
     * rootless (not starting with a slash). Implementations MUST support all
     * three syntaxes.
     *
     * If the path is intended to be domain-relative rather than path relative then
     * it must begin with a slash ("/"). Paths not starting with a slash ("/")
     * are assumed to be relative to some base path known to the application or
     * consumer.

      (from getPath())
     * Normally, the empty path "" and absolute path "/" are considered equal as
     * defined in RFC 7230 Section 2.7.3. But this method MUST NOT automatically
     * do this normalization because in contexts with a trimmed base path, e.g.
     * the front controller, this difference becomes significant. It's the task
     * of the user to handle both "" and "/".

      (from __tooString()
     * - The path can be concatenated without delimiters. But there are two
     *   cases where the path has to be adjusted to make the URI reference
     *   valid as PHP does not allow to throw an exception in __toString():
     *     - If the path is rootless and an authority is present, the path MUST
     *       be prefixed by "/".
     *     - If the path is starting with more than one "/" and no authority is
     *       present, the starting slashes MUST be reduced to one.
```

How do you want to implement the UploadedFile handling? I remember you had some ideas about that previously. You also started a [repo](https://github.com/thephpleague/uploads) I noticed. Do you want to use that? Or was that before the `UploadedFileInterface` was introduced in psr-7.

pine3ree:
Hello @JoeBengalen,
I haven't had time to read all your pr, the purpose being interesting. My eyes just hovered on the 
"There are several methods in the Request class that are not in the psr-7 spec...." comment.
Just to share my thoughts
At first I had the same feeling about it, but i believe that interoperability means that you can pass a request/response to an external middleware and it will utilize only interface methods to use instances or to return their "modified" clones. Any clone will be of the same original (internal) class so any framework method can still be used on the clone. Interoperable components will use the common interface, they will not change the nature of the response/request objects. A component intended to work as a middleware should never use new to create a response of a different class implementing the psr-7 interface, but only use the interface.

Problems will arise if we try to use a framework application as a middleware. In that case even internally it would need to use only contract methods.

But i believe that Slim/App is meant to start the process by itself since it creates the initial request/response instances in the run() method via factories stored in the container.

Please share more details on this subject and your point of view.

kind regards

JoeBengalen:
@pine3ree Internally slim only uses method from the psr spec. At least that was/is the the goal. For example `$response->write()` was replaced with `$response->getBody()->write()` internally. The extra "helper" methods are only ment for the developers building their app on Slim.

That aside, I do not think implementing and using methods that are not in the interface is a bad thing. Some of them may come in quite handy in some cases.

My point is that you cannot see, without going through the interfaces them selves, those methods are not part of the psr spec. So a developer may not know that he cannot use another response object whenever he wants to. That's why I said it may be a good idea to point that out somewhere. Either in the Slim documentation or within the phpdoc itself (or both).

Another option could be to make tweo separate objects. One implementing only the psr spec, and another, extending the first one, adding some useful "helper" methods. That way people can choose whether to use the extra methods or just the basic implementation of the psr spec.

I agree external libraries should only rely on the official interface.

All that said, that was not the purpose of this PR haha. I just made some updates regarding the updated version of the psr-7 spec, now that it is almost finalized (currently being in voting stage).

akrabat:
Thanks for your work on this @JoeBengalen. I would like to see the non-PSR-7 methods documented somewhere. Just in the docblocks would be fine for now.

As I understand it, `basePath` is the part of the URI that maps to where index.php is. e.g. for `http://example.com/my-app/`, the basePath is `/my-app` and '/' is the path used in `$app->get('/', …`. The basePath can then be used for generating the correct URLs in the router's urlFor() and also would be used for creating the absolute URLs for static assets such as JS/CSS links within view scripts.

JoeBengalen:
@akrabat Ah ok that makes sense about the absolutePath. Also I will add a note in the docblocks for non psr methods.

I will continue this PR after this weekend ;)

pine3ree:
@JoeBengalen 

I - very respectfully - don't agree with that either. 
A slim application is the kernel of its own middleware stack. It instantiates the initial request/response using slim internal implementations. 

``` php
$request = $this['request'];
$response = $this['response'];
//...
$response = $this->callMiddlewareStack($request, $response);
```

So even the application can safely use extra methods. I've seen many diagrams of middleware stacks in which the request-response process initially touch the external layer. In slim it's not the outer layer responsible of request/response instantiation. Nor are those objects instantiated elsewhere. They are first created by the application itself and then passed as arguments to the top-outer middleware. The application could receive request/responses instances (or clones of them) it created itself, so it can safely assume they are implementation of slim request response classes.

Even so, i also would prefer sticking to interface methods inside the framework implementation, just pointing out is not a must.

Problems begin if you replace the req/res factories ($['request'], $['response']) to return different implementations of the message interfaces. In that case neither the slim application nor user code will be able to use any extra slim methods. 

Thank You for sharing!!! 
Please reply if you see something wrong or missing in my argumentation.
kind regards.

lalop:
Since the request/response isn't hard coded we can use any implementation.
By hard coded I mean :

```
$response = $this->callMiddlewareStack(new Slim\Http\Request, new Slim\Http\Response);
```

The choice made was to open open the library to any psr7 complient request/response implementation by avoiding using this method in the kernel, they are here all for the end user.
That will certainly not be a big usage but it's a really good  choice I think in the psr7 philosophy.

pine3ree:
@lalop 
But you can't actually (more like practically) use any implemenation in a slim app. I believe the choice was made so that slim could use any 3rdy party psr-7 compliant middleware. The slim app creates the 1st req and res feeded into the stack. 3rd party components use the interface, but return either the same instances or clones of them.
If we want to start the stack allowing any other req/res implementation, we could never be safe using slim extra request/response methods as they would not probably be available/implemented.

lalop:
> If we want to start the stack allowing any other req/res implementation, we could never be safe using slim extra request/response methods as they would not probably be available/implemented.

Sorry but I don't understand what you mean here.
Are you saying that, if we start the stack with an other implementation we could not use the slim specific request/response 's methods ?

akrabat:
@pine3ree Actually Slim creates its Request and Response objects via the DIC. See https://github.com/slimphp/Slim/blob/develop/Slim/App.php#L85-L106.

As a result, you can do this:

``` php
$app = new \Slim\App();
$app['request'] = $app->factory(function ($c) {
    $request = \Phly\Http\ServerRequestFactory::fromGlobals(
        $_SERVER, $_GET, $_POST, $_COOKIES, $_FILES);
})
// set up routes, etc.
$app->run();
```

The app is now using Phly's Request object and Slim's Response.

pine3ree:
@akrabat 
Yes, it's using factories stored in the pimple container (which the app itself  extends) to build those instances. And you can extend (replace) services inside the container exactly like in your example.  By "you can't" i meant "you should not....,  otherwise...".  If we do that, none of the extra methods makes any sense because we are never certain that it will be available.

I believe though that the natural process of a slim application is to start from internal classes that implement the standard interfaces and utilize external middlewares. If we want complete portability we need to stick to a strict interface implementation + any number of only protected/private support methods.

as a side note:
There are a number of methods developers commonly use. For instance single parameter getters
like getQueryParam('name', $defaultValue) etc etc. fig-group is trying to keep a minimal interface but while we have single getter for request attributes, single getters for server, cookies, query.... are not included. And it feels kind of natural to have them implemented.

kind regards

pine3ree:
@lalop 
yes. If you try @akrabat example, extra methods of the Request instances and clones, will not be present because Matthew's implementation doesn't include them. Any middleware is meant to return the same instance or a new instance "altered" by with/without methods. So the nature of the object is not changed (with/without methods clone the original object) only its value, more precisely, being value objects we return clones representing different values not the same instances altered. 

One way to solve this could be implementing a factory method that takes an input request/response implementation and return an internal implementation instance with the same values, something like:

$req = Request::fromRequest(\Psr\Http\ServerRequestInterface $req);

this would need to be called inside the invoke method of the slim app, so that the route middleware stack receive a request (and response) that are instances of slim internal classes.

best regards

added comment
Another way would be, as suggested to me by Matthew elsewhere (sorry cannot find the reference right now), to have a simple and strict implementations of the interfaces and having internal classes (for the route middleware stack) that implement the decorator/adapter pattern on the stricter ones. the result would be the same: having more utility methods usable inside route handlers.

JoeBengalen:
Added notes to all non psr-7 public methods, apart from the static factories and constructors.

JoeBengalen:
Updated the path handling in Uri. @codeguy @weierophinney, can you please review if this is indeed the expected behaviour? I had some doubts about the base path implementation!

``` php
$uri = $this->uriFactory();
$this->assertEquals('https://josh:sekrit@example.com/foo/bar?abc=123#section3', (string) $uri);

// add / prefix to path
$uri = $uri->withPath('bar');
$this->assertEquals('https://josh:sekrit@example.com/bar?abc=123#section3', (string) $uri);

// uses the base path as $path is relative
$uri = $uri->withBasePath('foo/');
$this->assertEquals('https://josh:sekrit@example.com/foo/bar?abc=123#section3', (string) $uri);

// ignores the base path as $path as absolute
$uri = $uri->withPath('/bar');
$this->assertEquals('https://josh:sekrit@example.com/bar?abc=123#section3', (string) $uri);

```

JoeBengalen:
Psr-7 is final! https://github.com/php-fig/http-message/releases/tag/1.0

nhymxu:
congrat, hope slim beta version soon

akrabat:
@JoeBengalen Do you know when you'll get back to this?

JoeBengalen:
@akrabat, I'm still not sure about the Uri $path implementation. I just emailed Matthew, so I hope that helps him to confirm the current implementation.

About the fileParams, I made a [PR](https://github.com/slimphp/Slim/pull/1125) of it before on it ans Josh closed it and said he had his own ideas about implementing it. Not sure if any of this changed, but I want to hear from him before coding it. Also I am afraid I will not have much time in the upcoming few weeks to work on this.

Aside from that I think the PR is pretty much completed. @akrabat can you please review?

weierophinney:
@joebengalen Where did you email me? I haven't seen anything...

JoeBengalen:
@weierophinney the email from here: matthew@weierophinney.net

But since you are looking here anyway ...

Hellow Matthew,

While updating the Slim v3 framework to the final psr-7 interfaces I came to a point where I was not certain what to do. It involves the handling of the $path in the Uri implementation. I chatted with Rob Allen (@akrabat) and he adviced me to contact you about it, so here I am :)

Slim has a $basePath in the Uri to link to the index.php file: http://example.com/my-app/, the $basePath is /my-app and '/' is the path used in $app->get('/', …. I guessed "relative to some base path known to the application or consumer" was this case.

I was not too sure how to implement this. Where should the adding of the base path happen? The docblock mentions it in the 'withPath()' method, but I guess 'getPath()' should NOT include the base path. So it makes sense to only add the base path in the '__toString()' method. Am I right?

Currently I implemented in the following way. I'd like to get confirmation from you that this is indeed the right behaviour.

``` php
 $uri = $this->
uriFactory();
$this->assertEquals('https://josh:sekrit@example.com/foo/bar?abc=123#section3', (string) $uri);

// add / prefix to path
$uri = $uri->withPath('bar');
$this->assertEquals('https://josh:sekrit@example.com/bar?abc=123#section3', (string) $uri);

// uses the base path as $path is relative
$uri = $uri->withBasePath('foo/');
$this->assertEquals('https://josh:sekrit@example.com/foo/bar?abc=123#section3', (string) $uri);

// ignores the base path as $path as absolute
$uri = $uri->withPath('/bar');
$this->assertEquals('https://josh:sekrit@example.com/bar?abc=123#section3', (string) $uri);
```

FYI: the PR in which these changes are done: https://github.com/slimphp/Slim/pull/1228

Copied from the interface:

```
      (from withPath())
     * The path can either be empty or absolute (starting with a slash) or
     * rootless (not starting with a slash). Implementations MUST support all
     * three syntaxes.
     *
     * If the path is intended to be domain-relative rather than path relative then
     * it must begin with a slash ("/"). Paths not starting with a slash ("/")
     * are assumed to be relative to some base path known to the application or
     * consumer.

      (from getPath())
     * Normally, the empty path "" and absolute path "/" are considered equal as
     * defined in RFC 7230 Section 2.7.3. But this method MUST NOT automatically
     * do this normalization because in contexts with a trimmed base path, e.g.
     * the front controller, this difference becomes significant. It's the task
     * of the user to handle both "" and "/".
```

weierophinney:
It went to my spam. :-(

I've replied via email (didn't see you'd posted here as well).

JoeBengalen:
I don't know why it was marked as spam, but thnx for the reply.

For the interest of others: He answered the examples are 'correct' and 'reasonable', so thats good :)

Also he said this about the question of where to add the base path:
"The specification has no "base path" awareness. So any value you pass
to withPath() is expected to replace the current path.

We allow for both relative paths (those not prefixed with "/") and
absolute paths (those that are). In the case of a relative path,
however, if you cast the URI to a string, it MUST prefix with a "/" in
order to separate the authority and path components of the URI."

So again, if this PR is reviewed it will be ready for merge! The `$fileParams` will have to be added later on.

Slim/Http/Request.php
```diff
@@ -518,22 +545,49 @@ public function getUri()
...
+     * @param UriInterface $uri New request URI to use.
+     * @param bool $preserveHost Preserve the original state of the Host header.
      * @return self
      */
     public function withUri(UriInterface $uri, $preserveHost = false)
     {
-        //TODO Do something with $preserveHost
-
         $clone = clone $this;
         $clone->uri = $uri;
```
akrabat: Does this work? `uri` is a protected property.

Slim/Http/Request.php
```diff
@@ -518,22 +545,49 @@ public function getUri()
...
+        
+        if (!$preserveHost) {
+            if (!empty($uri->getHost())) {
+                $this->withHeader('Host', $uri->getHost());
```
akrabat, line 584: `withHeader` returns the changed request, so we need to assign to `$clone` or just update directly with:

```
$clone->headers->set('Host', $uri->getHost());
```

Slim/Http/Request.php
```diff
@@ -518,22 +545,49 @@ public function getUri()
...
+            }
+        } else {
+            if ((!$this->hasHeader('Host') || empty($this->getHeader('Host'))) && !empty($uri->getHost())) {
+                $this->withHeader('Host', $uri->getHost());
```
akrabat, line 588: `withHeader` returns the changed request, so we need to assign to `$clone` or just update directly with:

```
$clone->headers->set('Host', $uri->getHost());
```

JoeBengalen, line 588: Good point. Silly mistake.  Should be `$clone` ofcourse.

akrabat:
If updated the bits I commented on and have merged. As I had to rebase, GitHub didn't notice.

Merge commit: https://github.com/slimphp/Slim/commit/d3039eafc510f269985d9953d0accd0ea4c37a1b

PR #1231, opened 2015-05-15
Add helper function to retrieve accept headers if available

gabriel403:
Simplifies the ability to retrieve accept headers, helpful in those situation where `X-Requested-With` is missing and thus the `isAjax` function doesn't respond correctly as then we can check for `application/json` in the accept headers.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -763,6 +763,24 @@ class Request implements RequestInterface
         return $result ? (int)$result[0] : null;
     }
 
+    /**
+     * Get accept headers if available
+     *
+     * @return Array the accept headers
+     */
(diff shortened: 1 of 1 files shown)


akrabat:
Can you explain why we need this?

gabriel403:
@akrabat I added a description, hope that'll help.

akrabat:
I think I'd rather have a method that returns true if a given mime-type is in the list of accept headers.

What do others think?

(ping @codeguy, @silentworks, et al)

akrabat:

gabriel403:
@akrabat OK that sounds reasonable, `hasAcceptHeader`

PR #1237, opened 2015-05-19
move back app:responded

lalop:
Move back responded status as class property for testing purpose

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -52,6 +52,13 @@ class App
     const VERSION = '3.0.0';
 
     /**
+     * Has the app responsed to the HTTP client?
+     *
+     * @var bool
+     */
+    protected $responded = false;
+
+    /**
      * Container
      *
      * @var ContainerInterface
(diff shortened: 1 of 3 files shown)


akrabat:
I don't understand this PR. The `testRunSendBody` test works with the current code.

lalop:
It works because we have removed `testHaltInMiddleware`, but if we have two tests that does an `$app->run()` the second one will fail.

akrabat:
If we remove responded completely, then two tests will work…

lalop:
In both case the `responded` defined as static in the method isn't the good option

akrabat:
I agree. I'm just wondering if we can lose it completely.
@codeguy said that it was needed due to headers being sent, but I can't see why that's the case as we have a `headers_sent` check.

lalop:
It's what I see while rebasing.
I think we can removed it completely.
Do you want me to update this pr in this way ?

akrabat:
That's my gut feeling. We'll wait on @codeguy though.

akrabat:
As @codeguy never followed up, let's remove `$responded`.

codeguy:
If it can be removed without consequences go for it.

akrabat:

PR #1239, opened 2015-05-20
define properties proxied from container in app doc block

lalop:
add properties definitions to Slim\App.php header doc block

```
 * @property-read array $settings App settings
 * @property-read \Slim\Interfaces\Http\EnvironmentInterface environment 
 * @property-read \Psr\Http\Message\RequestInterface $request
 * @property-read \Psr\Http\Message\ResponseInterface $response
 * @property-read \Slim\Interfaces\RouterInterface $router
 * @property-read callable $errorHandler
 * @property-read callable function($request, $response) $notFoundHandler
 * @property-read callable function($request, $response, $allowedHttpMethods) $notAllowedHandler
```

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -23,7 +23,7 @@ use Interop\Container\ContainerInterface;
  * Slim Framework middleware.
  *
  * @property-read array $settings App settings
- * @property-read \Slim\Interfaces\Http\EnvironmentInterface environment 
+ * @property-read \Slim\Interfaces\Http\EnvironmentInterface $environment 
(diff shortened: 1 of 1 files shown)


Slim/App.php
```diff
@@ -21,6 +21,15 @@
  * register custom Pimple service providers on each
  * \Slim\App instance. The \Slim\App class also accepts
  * Slim Framework middleware.
+ *
+ * @property-read array $settings App settings
+ * @property-read \Slim\Interfaces\Http\EnvironmentInterface environment
```
JoeBengalen: Forgot dollar sign here

PR #1240, opened 2015-05-20
Convenient helper method

kminek:
express js-alike method :)

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -168,14 +168,14 @@ class App
     }
     
     /**
-     * Add route for every HTTP method
+     * Add route for any HTTP method
      *
      * @param  string $pattern  The route URI pattern
      * @param  mixed  $callable The route callback routine
      *
      * @return \Slim\Interfaces\RouteInterface
      */
-    public function all($pattern, $callable)
+    public function any($pattern, $callable)
     {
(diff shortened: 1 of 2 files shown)


nunocodex:
I like it +1

mnapoli:
:+1:

JoeBengalen:
Should `OPTIONS` be included?

mnapoli:
It is.

JoeBengalen:
It is indeed, but should it be? Has a whole different meaning compared to the other http methods.

mnapoli:
Ah sorry I thought you were asking to include it.

All means all to me ;) (i.e. I want to catch all methods) I don't see why it should be "all but options".

JoeBengalen:
True, just not too sure about what would be more useful. 

Also in Slim v2 there is a method called `any`. Isn't this the same functionality?
https://github.com/slimphp/Slim/blob/master/Slim/Slim.php#L563

silentworks:
Can you rename the method to `any` please? that way it will keep the same naming as it did in Slim 2.

kminek:
@silentworks done

silentworks:
@kminek sorry for the delay in getting back to you, but can we have a unit test for this method please?

silentworks:
Thank you very much for your contribution.

PR #1241, opened 2015-05-21
Updated the README with slightly clearer instructions on usage.

needcaffeine:
Not everyone realizes that they need to add the composer autoloader at the top of the page, so I added that to the usage. I also removed the testing section because there was no need for it. I added instructions to test the hello world application using PHP's inbuilt web server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,56 +1,81 @@
 # Slim Framework
 
 [![Build Status](https://travis-ci.org/slimphp/Slim.svg?branch=develop)](https://travis-ci.org/slimphp/Slim)
+[![Total Downloads](https://poser.pugx.org/slim/slim/downloads)](https://packagist.org/packages/slim/slim)
+[![License](https://poser.pugx.org/slim/slim/license)](https://packagist.org/packages/slim/slim)
 
-Slim is a PHP micro framework that helps you quickly write simple yet powerful web applications and APIs. Learn more at these links:
+Slim is a PHP micro-framework that helps you quickly write simple yet powerful web applications and APIs.
 
-- [Website](http://www.slimframework.com)
-- [Documentation](http://docs.slimframework.com)
-- [Support Forum](http://help.slimframework.com)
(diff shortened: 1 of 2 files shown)


mnapoli:
It seems you accidentally removed the `Testing` section.

needcaffeine:
That was not an accident. I intentionally removed it. There are no tests when you start off with Slim. Anyone who knows how to write unit tests knows how to use phpunit. It's just confusing for newcomers.

silentworks:
@needcaffeine thanks for your improvements to the README, but can you add the Testing section back please?

The Testing section does not confuse new users as new users would be pulling down Slim via composer or git clone the repo, the Testing section is necessary for contributors and there are already tests written for the framework itself.

nunocodex:
+1 for restore the testing section, I think who working with the frameworks is a developer so know phpunit :+1:

needcaffeine:
Okay, noted. I'll spend a little time making that section be more useful and restore it today. Thank you all for your thoughts on the subject.

needcaffeine:
Here we go.

README.md
```diff
@@ -1,56 +1,80 @@
...
-$app = new \Slim\App();
+<?php
+
+require 'vendor/autoload.php';
+
+$app = new Slim\App();
+
 $app->get('/hello/{name}', function ($request, $response, $args) {
     $response->write("Hello, " . $args['name']);
-    return $response;
```
akrabat, line 28: Best practice is to return the response. i.e. you are relying on an implementation detail of the current `Route` object if you don't return the `$response`.

needcaffeine, line 28: Hm, you're right.

needcaffeine, line 28: Fixed both in the README and in index.php.

needcaffeine:
Is there anything further you'd like me to do to get this merged in?

silentworks:
Thanks for this update @needcaffeine.

PR #1248, opened 2015-05-27
Updated subRequest to get serverParams correctly

silentworks:


lalop:
Reading your pr I see 2 questions/comments : 

1) I think we should use the same request type then the one in the container
That means the container should only give the Request class not a Request instance and the App should create this new Request for main and sub request.
Or you should clone the current request to make sub request, so `App::subRequest` could become an alias for `__invoke`.

```
$app->get('/bar', function($req, $res) use($app) {
    $uri = $req->getUri()->withPath('/foo');
    $sub_req = $req->withUri($uri);
    return $app->subRequest($sub_req,  $res);
});
```

2) Why do not run App middleware ? Maybe that make no sense for session but that can be usefull for a cachemiddleware.
Maybe We should add a request stack.
App stack once per HTTP request,
Request stack once per internal request,
Route stack once per route per request

silentworks:
@lalop not sure I understand point 2. In point 1 the code example seem to carry the wrong signature for the `subRequest` method.

lalop:
Hello @silentworks,

For point 1, `subRequest` has a wrong signature because it's what I propose, having the same signature for `App::subRequest` and `App::__invoke`.

For point 2, I wanted to notice that with this implementation app's middlewares will not see subrequests.
That can be okay in most of case but maybe not for all so I proposed to add a new middlewae stack between App and Route, App's stack will be run once, Request's stack will be run once per (sub)request, and Route's stack is defined for each route.
Is that more clear ?

akrabat:
Re point 1, `subRequest` exists solely to avoid you having to create the correct request object yourself.

i.e., you could do:

```
$app->get('/bar', function($req, $res) use($app) {
    $uri = $req->getUri()->withPath('/foo');
    $sub_req = $req->withUri($uri);
    return $app($sub_req,  $res);
});
```

without needing this method at all.

As such, `subRequest` is a convenience method, and that's okay.

Re point 2: I can see the argument, but disagree. a sub request runs in the context of the first request, not as a completely independent thing. If you have middleware that needs running on both the request and subrequest, then use route middleware.

 three places to put middleware would just be confusing for a very specialised edge-case.

akrabat:
Merged

PR #1249, opened 2015-05-27
more synthetic url  decode

lalop:


silentworks:
There was a bug in the original change which have been fixed in #1254, this PR is now outdated based on the new changes.

PR #1257, opened 2015-06-01
Added setter for request body

benjamij:
I have added a setter for request->body. There were instances in which my middleware had to modify the request body, and as such, a setter would be handy.

For example, having a security middleware for symmetric crypto (in addition running everything over HTTPS):      

// Decrypt
$request = $this->_slim->request();
$body = $request->getBody();
$plaintext = $this->decrypt($body, $key);

// Update body with plaintext, so that the services can process it
$request->setBody($plaintext);


akrabat:
The Request follows the [PSR-7](http://www.php-fig.org/psr/psr-7/) specification and so doesn't have a setter intentionally.
Instead, you would do this:

``` php
$request = $request->withBody($plaintext);
```

PR #1263, opened 2015-06-03
Provide Request::getUploadedFiles() implementation

opengeek:
This is being submitted for initial review and requires development of appropriate unit tests before being merged.


opengeek:
This should address #1099

akrabat:
When uploading a single file, the UploadedFile object should be directly against the keyname:

i.e given : `<input type="file" name="this_file">` and an route callable of:

```
$app->post('/fileupload', function ($request, $response) {
    var_dump($request->getUploadedFiles());
});
```

I should see:

```
array (size=1)
  'this_file' => 
    object(Slim\Http\UploadedFile)[70]
      protected 'name' => string '90by90.png' (length=10)
      protected 'type' => string 'image/png' (length=9)
      protected 'size' => int 3962
      protected 'error' => int 0
      protected 'sapi' => boolean true
      protected 'stream' => null
      protected 'moved' => boolean false
      public 'file' => string '/private/var/tmp/phplSKALT' (length=26)
```

rather than:

```
array (size=1)
  'this_file' => 
    array (size=1)
      0 => 
        object(Slim\Http\UploadedFile)[70]
          protected 'name' => string '90by90.png' (length=10)
          protected 'type' => string 'image/png' (length=9)
          protected 'size' => int 3962
          protected 'error' => int 0
          protected 'sapi' => boolean true
          protected 'stream' => null
          protected 'moved' => boolean false
          public 'file' => string '/private/var/tmp/phplSKALT' (length=26)
```

This is described by example in https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message.md#16-uploaded-files

Slim/Http/UploadedFile.php
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * Parse a non-normalized, i.e. $_FILES superglobal, tree of uploaded file data.
+     *
+     * @param array $uploadedFiles The non-normalized tree of uploaded file data.
+     *
+     * @return array A normalized tree of UploadedFile instances.
+     */
+    public static function parseUploadedFiles(array $uploadedFiles)
```
akrabat: Why public and not private?

opengeek: Good point. Just an oversight.

geggleto:
//todo @geggleto test pr

geggleto:
//cc @opengeek 

Additional Tests Required.
-- uploadedFileConstructTest
-- parseUploadedFilesTest
-- createFromEnvironmentTest
-- getStreamTest
-- moveToTest
-- moveToStreamTest
-- getErrorTest
-- getClientFilenameTest
-- getClientMediaTypeTest
-- getSizeTest

Additional Provider setup to handle slim.files

akrabat:
Do we feel this is ready to merge?

geggleto:
:+1: by me. I think it is solid enough to put up for beta-1

PR #1264, opened 2015-06-04
Proposed Middleware Helper

stratedge:
Hey there,

I have been working with the develop branch for a little while now on a new API and found it helpful to create a Middleware helper that exposes a before() and after() method. I thought it might be useful as part of the core library, so I'm submitting it for consideration.

For the developer, using the class is just a matter of extending Slim\Http\Middleware, and then overloading either the before() method, the after() method, or both. The helper class will handle calling the methods within the flow and also calling the next callable in the middleware stack in the middle.

**Example of before():**

``` php
use Psr\Http\Message\RequestInterface;
use Psr\Http\Message\ResponsetInterface;

class Authenticate extends \Slim\Http\Middleware {

    public function before(RequestInterface $request, ResponseInterface $response)
    {
        $request = AuthLayer::checkAuth($request);
        return [$request, $response];
    }
}
```

**Example of after():**

``` php
use Psr\Http\Message\RequestInterface;
use Psr\Http\Message\ResponsetInterface;

class SetJSONHeader extends \Slim\Http\Middleware {

    public function after(RequestInterface $request, ResponseInterface $response)
    {
        $response = $response->withHeader('Content-Type', 'application/json');
        return $response;
    }
}
```

For this implementation before() returns an array containing the Request and Response objects, while after() just returns the Response object. Honestly, it might be better if the objects are stored on the class itself, and the methods do not need to return anything. Happy to make that change if necessary.

Hope this proves to be of value. I find it cleans up my implementations a lot and helps me to better conceptualize just how exactly my middleware is working within the flow. On the other hand, if this is too specialized an implementation to go into the core framework, or is not of interest, I totally understand.

Take care,
JB


silentworks:
Hi @stratedge thanks for suggesting this, but I think it would move us back into the issue of making Middleware's Slim specific again. 

We are trying to move away from having Middleware being specific to 1 framework, this way we can share the same middleware between different frameworks. I know that most frameworks seem to be sticking to the trend of making everything framework specific, but we don't think this is the route we want to go down again, the developer should decide if they want to make their middleware specific to Slim rather than us directing them in that direction.

Yet again thanks for your suggestion but I don't think it sits well with the current goal of the project.

stratedge:
Totally understand - and in the long run I definitely agree that the best route is the one that fosters interoperability through some standardization.

I'll take a look at the list of issues and see if there are bugs or improvements to squash in my spare time, that'll probably be more helpful for everyone.

I'll close out the request so it doesn't linger.

PR #1291, opened 2015-06-15
Refactor/trait suffix

### LINKED ISSUE #1272: [v3] Traits
Just a quickie about traits.

Interfaces are defined in `Slim\Interfaces`. Should traits be in `Slim\Traits`?

Also, not sure if it is a standard or not in PSR, but interfaces should be defined as `MyAwesomeInterface`, abstracts as `MyNotSoCoolAbstract`, so should the traits be `WowSuchTrait`, all with the type at the end?

So, `CallableResolverAwareTrait` and `MiddlewareAwareTrait`

It's definitely a preference of mine to allow me to easily ascertain what is what in a codebase.

gabriel403:
Adds Trait suffix to the 2 traits we currently use. 

Resolves #1272

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -39,7 +39,8 @@ use Slim\Http\Request;
  */
 class App
 {
-    use CallableResolverAware;
+    use CallableResolverAwareTrait;
+
     use MiddlewareAware {
         add as addMiddleware;
     }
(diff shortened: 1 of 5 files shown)


geggleto:
:+1:

Issue #1290, opened 2015-06-14
Small mistake in exception message inside Slim\Http\Body class


wskorodecki:
There is a small mistake in exception message in getContents() method of Slim\Http\Body class in Slim 3.

```
/**
     * Returns the remaining contents in a string
     *
     * @return string
     * @throws \RuntimeException if unable to read or an error occurs while
     *     reading.
     */
    public function getContents()
    {
        if (!$this->isReadable() || ($contents = stream_get_contents($this->stream)) === false) {
            throw new RuntimeException('Could get contents of stream');
        }
        return $contents;
    }
```

Should be:

`Could not get contents of stream`

(missing "not")

zither:
Maybe you can send a PR.

akrabat:
Fixed.

PR #1292, opened 2015-06-15
fixes issue #1290

HardieBoeve:


akrabat:
Unfortunately, I've just merged a PR which means that this needs to be implemented against https://github.com/slimphp/Slim/blob/develop/Slim/Http/Stream.php.

HardieBoeve:
@akrabat, ok, i changed it.

PR #1293, opened 2015-06-15
Added missing "not" word in exception message

wskorodecki:
Added missing "not" word in exception message in Slim\Http\Stream@getContents() method.


akrabat:
Fixed in #1292

PR #1303, opened 2015-06-18
outputBuffer setting

kminek:
this PR introduces `outputBuffer` setting (with values: `false` or `'append'` or `'prepend'`) which let's user decide how output buffering is handled

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -205,20 +205,38 @@ class Route implements RouteInterface
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response)
     {
+        $outputBuffer = 'append';
+        if (($this->container !== null) && isset($this->container->get('settings')['outputBuffer'])) {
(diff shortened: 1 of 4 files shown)


Slim/Route.php
```diff
@@ -205,15 +205,25 @@ public function run(ServerRequestInterface $request, ResponseInterface $response
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response)
     {
+        $outputBuffer = 'append';
```
akrabat: Grabbing settings out of the container is wrong as the container is only in the route to resolve the callable and may go away if we find a better way to do it.

Therefore, please make `outputBuffer` a property of this class and read the setting for it within `App` and set the output buffer setting of the route instance when it's created [here](https://github.com/akrabat/Slim/blob/develop/Slim/App.php#L236).

akrabat:
Merged.

PR #1304, opened 2015-06-19
Fix urlFor() to work with optional parameters

### LINKED ISSUE #1289: Optional parameter in url
Slim doesn't support optional parameters as the underlying routing engine doesn't support them.

We need:
- Ability to specify optional parameters in the route path
- Ability to generate urls via `urlFor` for routes that have optional parameters.

This was originally reported as:

It would seem it currently isn't possible to add an optional paramater in the url if you are using urlFor to get the url via a named route.

``` php
$app->get('/test/{param1}[/{optional:[A-Za-z]}]')->setName('test');

$app->getContainer()['router']->urlFor('test', ['param1' => 'World']);
```

This will throw an error because urlFor method does not know how to handle optional regex. I saw another project using FastRoute and they handle it using this regex https://github.com/mrjgreen/phroute/blob/master/src/Phroute/RouteParser.php#L23-L29

akrabat:
Also, update composer.json to pick up the latest FastRoute until Nikita creates a new tag.

This PR fixes #1289.

### DIFF
--- a/tests/RouterTest.php
+++ b/tests/RouterTest.php
@@ -83,6 +83,30 @@ class RouterTest extends PHPUnit_Framework_TestCase
         );
     }
 
+    public function testUrlForWithOptionalParameters()
+    {
+        $methods = ['GET'];
+        $pattern = '/archive/{year}[/{month:[\d:{2}]}[/d/{day}]]';
+        $callable = function ($request, $response, $args) {
+            return $response;
+        };
+        $route = $this->router->map($methods, $pattern, $callable);
+        $route->setName('foo');
+
+        $this->assertEquals(
+            '/archive/2015',
+            $this->router->urlFor('foo', ['year' => '2015'])
+        );
+        $this->assertEquals(
+            '/archive/2015/07',
+            $this->router->urlFor('foo', ['year' => '2015', 'month' => '07'])
+        );
+        $this->assertEquals(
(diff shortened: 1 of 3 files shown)


JoeBengalen:
Not sure if it is smart to depend on version `dev-master`. Why not specify version `0.5`?

akrabat:
Version 0.5 doesn't have optional routes as it was tagged on 5th May 2015 and optional parameters was implemented on 18th June 2015.

JoeBengalen:
Oh OK, thought it was the update from 0.4 to 0.5. Makes sense now ;) But it would be nice if for the final version of Slim we could depend on a specific version of fast-route.

akrabat:
I agree. We'll ask @nikic nicely sometime after beta1 when it's had some more widespread testing :)

tests/RouterTest.php
```diff
@@ -83,6 +83,30 @@ public function testUrlFor()
         );
     }
 
+    public function testUrlForWithOptionalParameters()
+    {
+        $methods = ['GET'];
+        $pattern = '/archive/{year}[/{month:[\d:{2}]}[/d/{day}]]';
```
nikic, line 89: Is the `[\d:{2}]` here intentional?

akrabat, line 89: Yes - I wanted to ensure that it didn't break anything

PR #1312, opened 2015-06-23
Add getName(), setName() to RouteInterface

opengeek:
To have interoperability with a RouterInterface implementation, I believe we need at least these methods defined, and potentially getCallable() and setContainer() ?

### DIFF
--- a/Slim/Interfaces/RouteInterface.php
+++ b/Slim/Interfaces/RouteInterface.php
@@ -8,6 +8,7 @@
  */
 namespace Slim\Interfaces;
 
+use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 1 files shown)


Slim/Interfaces/RouteInterface.php
```diff
@@ -20,6 +21,20 @@
 interface RouteInterface
 {
     /**
+     * Get route methods
+     *
+     * @return string[]
+     */
+    public function getMethods();
```
akrabat: Where is this method used?

opengeek: I can't find any usages—perhaps I should remove this from the interface and it could be dependent on a specific implementation?

akrabat: I can't see any use for it in the interface

PR #1315, opened 2015-06-24
removed htaccess file

vutran:
Since the framework is supported on multiple servers, this should be removed since we're not tracking configs for other servers.

Perhaps a server setup page can be created in the wiki/docs for different setups.


tuupola:
There is too much confusion about this already. Removing .htaccess file would cause even more confusion. Most people are using Apache anyway. Those who use Nginx are generally more computer literate and understand that since they use Nginx they can safely remove this file.

mnapoli:
This repository will be installed as a dependency by Composer, it's not the project of users. The .htaccess will not be in the users project. So it's not really useful to have it here. Or maybe I'm missing a use case where users just clone this repository and don't use Composer?

codeguy:
@mnapoli Originally (read: with version 2) that was exactly the use case. With version 3, we are encouraging users to install with Composer. We include a sample .htaccess file in the documentation. I'm not sure why we would need to keep this file in the repo.

silentworks:
Yes this can be removed for v3, this was necessary in v2 but this is no longer the case.

DavidePastore:
If we remove it, it would be nice to have in the documentation (or in the site in general) a list of files to download with the recommended settings for each web server.

akrabat:
See http://www.slimframework.com/docs/start/web-servers.html
Improvements welcome, by PR of https://github.com/slimphp/Slim-Website/blob/gh-pages/docs/start/web-servers.md

PR #1317, opened 2015-06-26
Fix #1269

### LINKED ISSUE #1269: [Develop] Body Coupling
https://github.com/slimphp/Slim/blob/develop/Slim/App.php#L314

``` php
            if ($hasBody) {
                $body = $response->getBody(); //Returns the PSR Steam Interface
                if ($body->isAttached()) { //This method is NOT a part of the interface but is apart of the Slim\Http\Body
                    $body->rewind();
                    while (!$body->eof()) {
                        $settings = $this->container->get('settings');
                        echo $body->read($settings['responseChunkSize']);
                    }
                }
            }
```

geggleto:
- Changed visibility levels of NON-PSR-7 methods on stream object
- Updated test cases


tests/Http/BodyTest.php
```diff
@@ -151,14 +153,15 @@ public function testAttachStream()
...
-     */
+
+     * expectedException \InvalidArgumentException
+
     public function testAttachInvalidStream()
     {
         $body = new \Slim\Http\Body($this->resourceFactory());
         $body->attach('Foo');
     }
+    */
```
mnapoli, line 164: Commented code should rather be removed?

geggleto, line 164: I left it in just in case we wanted to move those methods elsewhere, but if required ill remove them.

PR #1321, opened 2015-06-28
Minor improvements

theodorejb:
- Fixed Stream exception message grammar
- Minor optimizations:
  -  Avoid retrieving container settings in a loop
  - If `$newResponse` is a `ResponseInterface`, it isn't necessary to also check if it is a string


akrabat:
Thanks.

PR #1323, opened 2015-06-28
Optimize Route::__invoke

theodorejb:
- Avoid extra body write call
- If output buffering is set to prepend, it isn't necessary to also check if it is set to append
- Improved comments


akrabat:
Thanks!

Issue #1326, opened 2015-06-28
Need a fully qualified urlFor() helper


akrabat:
In the current Slim 3, to create a fully qualified URL for a given route name, you need to do this inside a route callable:

``` php
$router = $this->router;
$uri = $request->getUri();

$url = $uri->getScheme() . '://' . $uri->getHost()
        . $uri->getPort() ? ':' . $uri->getPort() : ''
        . rtrim($uri->getBasePath(), '/')
        . $router->pathFor('news-archive', ['year'=>'2015'], ['foo' => 'bar']);
```

We need a helper method for this.

Melbournite:
Related question -- Are we going to rename url_for() in Twig-View to match what we named it here? Being I guess path_for() for consistency? Deprecate url_for() and I guess create a uri_for() or similar so that names are aligned with functionality and maintain a level of consistency.

akrabat:
See https://github.com/slimphp/Twig-View/pull/20

akrabat:
Based on IRC chat:

Another option would be to add a method to the Request that returns the combination of scheme, domain, port and base path. It could be used something like this:

``` php
$url = $app->request->siteUrl($app->router->pathFor('hello'));
```

or

``` php
$url = $app->request->siteUrl() . ltrim($app->router->pathFor('hello'), '/');
```

akrabat:
How about something like #1328 to provide an easy way to get the fully qualified URL including the base path?

calmdev:
This would be nice. I always run into the need for this when sending emails that need to provided a full URI so folks can click to access the application.

silentworks:
This have now been merged in.

PR #1327, opened 2015-06-29
Implement fullUrlFor() 

akrabat:
I've put this on both Router and App that builds a fully qualified URL for a named route.

Ideally, I'd have called the `urlFor` but we couldn't have the deprecated method if I did that.

See #1326.


akrabat:
Closing for now as probably need something on the Request object too.

PR #1328, opened 2015-06-30
Uri::getBaseUrl()

akrabat:
This addresses some of the issues in #1326.

It provides a new method on `Uri` called `getBaseUrl()` which provides a fully qualified URL with just the base path included. It never has a trailing slash and so is ideal for combing with `pathFor` like this:

``` php
$router = $this->router;
$uri = $request->getUri();

$fullUrl = $uri->getBaseUrl() . $router->pathFor('news', ['year'=>'2015'], ['foo' => 'bar']);
```

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -769,4 +769,35 @@ class Uri implements UriInterface
             . ($query ? '?' . $query : '')
             . ($fragment ? '#' . $fragment : '');
     }
+
+    /**
+     * Return the fully qualified base URL.
+     *
+     * Note that this method never includes a trailing /
+     *
+     * This method is not part of PSR-7.
+     *
+     * @return string
+     */
+    public function getBaseUrl()
+    {
+        $scheme = $this->getScheme();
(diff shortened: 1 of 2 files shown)


Slim/Http/Uri.php
```diff
@@ -769,4 +769,30 @@ public function __toString()
...
+    public function getBaseUrl()
+    {
+        $scheme = $this->getScheme();
+        $authority = $this->getAuthority();
+        $basePath = $this->getBasePath();
+        $query = $this->getQuery();
+        $fragment = $this->getFragment();
+
+        if ($authority && substr($basePath, 0, 1) !== '/') {
+            $basePath = $basePath . '/' . $basePath;
```
micheh, line 791: @akrabat Why do you add the base path to the base path? That doesn't seem right. Also, the method `withBasePath` always adds a root slash, so this statement can probably never be true?

akrabat, line 791: I would guess that there's an error there :)

PR #1329, opened 2015-06-30
Add helper function to retrieve accept headers if available

gabriel403:
Add 2 helper methods for checking what accept headers exist within the request.

Simplifies the ability to retrieve accept headers, helpful in those situation where X-Requested-With is missing and thus the isAjax function doesn't respond correctly as then we can check for application/json in the accept headers.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -874,6 +874,37 @@ class Request implements ServerRequestInterface
         return $result ? (int)$result[0] : null;
     }
 
+    /**
+     * Get accept headers if available
+     *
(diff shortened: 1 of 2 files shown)


gabriel403:
We already have plenty of other simple, many even simpler, header helper methods, look at `getContentLength` for example. :-1: just because it's possible to do else where doesn't make sense to me, considering the already implemented helper methods. 
The pseudo code you provided wouldn't work as you'd still need to split the return of `$req->getHeader('ACCEPT');`

gabriel403:
Giving up on getting this into slim

Issue #1265, opened 2015-06-04
[Develop] Adding Middleware to Route Groups


geggleto:
Come to find out that the process for adding middleware to route groups is different than adding them to routes.

What I expected:
`$app->group('/', function () { })->add('\Security:check');`

Current
`$app->group('/', \Security:check, function () { });`

Where \Security:check must return a callable which is the middleware wrapped in a function.

So I would suggest 2 things.

1) We change the middleware to be added like route middleware which is a ->add at the end
2) We also add the resolveCallable onto that so that things get resolved properly.

lalop:
:+1:

nhymxu:
:thumbsup:

designermonkey:
I was always a little uncomfortable with how routes and groups were done, so I wrote something to play with that made a new class of `RouteGroup` alongside `Route`.

I'll dig out the code and try to marge it back in to Slim3 (it was a while ago, and many things have changed).

What the code did was made routes and groups essentially the same thing (so things like middleware were handled exactly the same way), but one could have children. It followed the same principal for dispatching, but that part has to be updated due to the PSR-7 work in Slim since I wrote it.

dopesong:
@designermonkey any news about it?

designermonkey:
Sorry guys, responsibilities have taken over again.

silentworks:
@designermonkey is the code available anywhere on Github or can it be available for us to take a look at please?

designermonkey:
Yeah, by all means. This component is never going to get used, so feel free to pull it apart. It was based on Slim anyway.

https://github.com/usebeagle/routes

akrabat:
Implemented.

PR #1330, opened 2015-06-30
Add RouteGroup to allow registration of middleware consistently

opengeek:
Requires routes be collected and compiled before usage; see changes to App->run() and App->__invoke()

Attempts to address #1265

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -257,9 +257,6 @@ class App
     public function group($pattern, $callable/*, $middleware = []*/)
     {
         $group = new RouteGroup($pattern, $callable);
-//        foreach ($middleware as $callable) {
-//            $group->add($callable);
-//        }
         $this->container->get('router')->pushGroup($group);
         $group($this);
         $this->container->get('router')->popGroup();
@@ -331,10 +328,8 @@ class App
      */
     public function run()
     {
-        // Build routes here so we can bind group middleware late
-        /** @var Router $router */
-        $router = $this->container->get('router');
-        $router->compile();
+        // Finalize routes here for middleware stack
+        $this->container->get('router')->finalize();
 
         $request = $this->container->get('request');
         $response = $this->container->get('response');
@@ -360,8 +355,8 @@ class App
      *
      * This method implements the middleware interface. It receives
      * Request and Response objects, and it returns a Response object
-     * after dispatching the Request object to the appropriate Route
-     * callback routine.
+     * after compiling the routes registered in the Router and dispatching
+     * the Request object to the appropriate Route callback routine.
      *
      * @param  ServerRequestInterface $request  The most recent Request object
      * @param  ResponseInterface      $response The most recent Response object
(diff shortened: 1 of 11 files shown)


Slim/Interfaces/RouterInterface.php
```diff
@@ -42,12 +43,11 @@ public function dispatch(ServerRequestInterface $request);
...
      * Add a route group to the array
      *
-     * @param string     $group      The group pattern prefix
-     * @param array|null $middleware Optional middleware
+     * @param RouteGroup &$group The route group
      *
      * @return int The index of the new group
      */
-    public function pushGroup($group, $middleware = []);
+    public function pushGroup(&$group);
```
JoeBengalen: Why the reference sign? As it seems you are passing in an object, which will always be passed by reference.

Also maybe typehint for `RouteGroup`

Slim/App.php
```diff
@@ -233,7 +234,7 @@ public function map(array $methods, $pattern, $callable)
             $callable = $callable->bindTo($this);
         }
 
-        $route = $this->container->get('router')->map($methods, $pattern, $callable);
+        $route = $this->router->map($methods, $pattern, $callable);
```
akrabat: Why make this change?

Slim/RouteGroup.php
```diff
@@ -0,0 +1,61 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Invoke the group to register any Routable objects within it.
+     *
+     * @param App &$app The App to bind the callable to.
+     */
+    function __invoke(&$app)
```
JoeBengalen: Same here; No reference needed and typehint for `App`

Slim/Interfaces/RouterInterface.php
akrabat: Also why do we allow an array of middleware?

silentworks:
Can the creation of Group's happen inside of the router instead of in the App please?

Slim/Interfaces/RouterInterface.php
opengeek: Reference sign was leftover from some silly attempts to get add() to work on the already pushed groups (snicker), never mind that they are popped immediately after evaluation. :(

The array of middleware was an artifact of allowing any number of arguments in the call to `group()`, but this is gone now with `add()` working.

Slim/App.php
opengeek: Probably so I could see the methods available on the router which cannot be inferred from the `container->get()` call. I'll put it back...

Slim/RouteGroup.php
opengeek: Noted and reverting.

opengeek:
@silentworks I'm not sure how you envision the change to move the logic inside the Router. That would then require that I pass the App to a similar method inside Router. Is that really an improvement? It feels like I'm just moving the same dependency around, but if you think that is preferable I can make that change. Or if you have another suggestion?

opengeek:
See https://gist.github.com/opengeek/c8a2f8952560e226855b for example usage

Issue #1339, opened 2015-07-03
app middleware is not invoked when doing subRequest


kminek:
sample code:

``` php
$app = new Slim\App();

$app->add(function ($req, $res, $next) {
    $res->write('BEFORE');
    return $next($req, $res);
});

$app->get('/', function ($request, $response, $args) {
    return $this->subRequest('GET', '/hello/grzes');
});

$app->get('/hello/{name}', function ($request, $response, $args) {
    $response->write("Hello, " . $args['name']);
    return $response;
});

$app->run();
```

JoeBengalen:
That is the expected behaviour as far as I know.

Lewiscowles1986:
Surely the subRequest could be replaced with an API call or command pattern; if for example Auth and ACL, session data etc was needed, this would also be more descriptive. Using this command pattern in each route (the /hello/{name} request, and the '/') call a command and can share data from the app's di container. Is this the desired reason for using middleware on subRequest?

kminek:
to be honest i would like not to use subRequest at all. i wish something like following was possible:

``` php
$app->get('/hello/{name}', 'HelloController:hello');
$app->get('/', 'HelloController:hello')->setParam('name', 'grzes');
```

Lewiscowles1986:
it is possible, but not using just slim. The point of a micro-framework is to be able to bolt-in other libraries. I recommend https://github.com/jmathai/php-multi-curl. It has some pretty simple syntax, and if you need to add params just implode the values with the relevant separators.

Please also bear in mind this is not the best solution to calling other URL endpoints, but it will get you motoring, and if you target a specific domain, that has load balancing, it can be a poor man's load distribution.

kminek:
@Lewiscowles1986 man, i'm not talking about making curl requests. i'm talking about Route class feature that is missing in 3 but was in 2: https://github.com/slimphp/Slim/blob/2.x/Slim/Route.php#L259

Lewiscowles1986:
@kminek Sorry I don't understand why you would want to do this. From what maybe is my limited understanding of your goal, or problem, it seems easier to just send another request. Personally I would use command pattern for this rather than a route, then you can set data for the command.

akrabat:
One option would be to write as middleware:

``` php
$app->add(function($request, $response, $next) {
    if ($request->getMethod() == 'GET' &&
        $request->getUri()->getPath() == '/') {
        $request = $request->withUri($request->getUri()->withPath('/hello/rob'));
    }
[...]

PR #1343, opened 2015-07-08
Add default attributes to the route

AlexStansfield:
Added setDefaultAttributes method to the route class. Allows you to set default attributes when defining a route. 

For example:

```
$app->get('/hello/world', function ($request, $response, $args) {
    $response->write("Hello " . $args['extra']);
    return $response;
})->setDefaultAttributes(['extra' => 'world']);
```

### DIFF
--- a/Slim/Interfaces/RouteInterface.php
+++ b/Slim/Interfaces/RouteInterface.php
@@ -20,6 +20,24 @@ use Psr\Http\Message\ResponseInterface;
  */
 interface RouteInterface
 {
+
+    /**
+     * Retrieve a specific route argument
+     *
+     * @param string $name
+     * @param mixed $default
+     *
+     * @return mixed
+     */
+    public function getArgument($name, $default = null);
+
+    /**
+     * Get route arguments
+     *
+     * @return array
+     */
+    public function getArguments();
+
     /**
      * Get route name
      *
@@ -33,6 +51,25 @@ interface RouteInterface
      * @return string
      */
     public function getPattern();
+
+    /**
+     * Set a route argument
+     *
+     * @param string $name
+     * @param string $value
+     *
+     * @return static
+     */
+    public function setArgument($name, $value);
+
+    /**
+     * Replace route arguments
+     *
+     * @param array $arguments
+     *
+     * @return static
+     */
+    public function setArguments(array $arguments);
 
     /**
      * Set route name
--- a/index.php
+++ b/index.php
@@ -31,6 +31,11 @@ $app->get('/', function ($request, $response, $args) {
     return $response;
 });
 
+$app->get('/hello/world', function ($request, $response, $args) {
+    $response->write("Hello " . $args['extra']);
+    return $response;
+})->setArgument('extra', 'world');
+
 $app->get('/hello/{name}', function ($request, $response, $args) {
     $response->write("Hello, " . $args['name']);
     return $response;
--- a/index.php
+++ b/index.php
@@ -32,9 +32,9 @@ $app->get('/', function ($request, $response, $args) {
 });
 
 $app->get('/hello[/{name}]', function ($request, $response, $args) {
-    $response->write("Hello " . $args['name']);
+    $response->write("Hello, " . $args['name']);
     return $response;
-})->setArgument('name', 'world');
+})->setArgument('name', 'World!');
 
 /**
  * Step 4: Run the Slim application
--- a/index.php
+++ b/index.php
@@ -31,15 +31,10 @@ $app->get('/', function ($request, $response, $args) {
     return $response;
 });
 
-$app->get('/hello/world', function ($request, $response, $args) {
-    $response->write("Hello " . $args['extra']);
+$app->get('/hello[/{name}]', function ($request, $response, $args) {
+    $response->write("Hello " . $args['name']);
     return $response;
-})->setArgument('extra', 'world');
-
-$app->get('/hello/{name}', function ($request, $response, $args) {
-    $response->write("Hello, " . $args['name']);
-    return $response;
-});
+})->setArgument('name', 'world');
 
 /**
  * Step 4: Run the Slim application
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -233,7 +233,7 @@ class Route extends Routable implements RouteInterface
      *
      * @return $this
      */
-    public function setArguments($arguments)
+    public function setArguments(array $arguments)
     {
         $this->arguments = $arguments;
         return $this;
@@ -252,7 +252,10 @@ class Route extends Routable implements RouteInterface
     /**
      * Retrieve a specific route argument
      *
-     * @return array
+     * @param string $name
+     * @param mixed $default
+     *
+     * @return mixed
      */
     public function getArgument($name, $default = null)
     {
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -232,6 +232,61 @@ class AppTest extends PHPUnit_Framework_TestCase
         $this->assertEquals('Hello', (string)$res->getBody());
     }
 
+    public function testInvokeWithMatchingRouteWithSetArgument()
+    {
+        $app = new App();
+        $app->get('/foo/bar', function ($req, $res, $args) {
+            return $res->write("Hello {$args['attribute']}");
+        })->setArgument('attribute', 'world!');
+
+        // Prepare request and response objects
+        $env = Environment::mock([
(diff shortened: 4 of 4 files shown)


geggleto:
it is currently unclear in your PR what would happen if... 

``` php
//.... /hello/Alex
    $response->write("Hello " . $args['name']);
    return $response;
})->setDefaultAttributes(['name' => 'Bob']);
```

I think it should overwrite Bob...but I would like a test case for this exact situation please :)

Other than that it looks useful :+1:

AlexStansfield:
Yeah it would overwrite Bob. Basically it uses the default if nothing else over rules it. This thread gives a more info about my reasons behind it - http://help.slimframework.com/discussions/questions/7376-setting-attributes-in-a-route

It's just gone midnight here so will add a test to show default is overwritten by a parameter matched in the url tomorrow.

kminek:
this is nice 

exactly the kind of feature i was asking here: https://github.com/slimphp/Slim/issues/1339#issuecomment-118484579

AlexStansfield:
Who needs sleep. Added test case to show default attribute being used and also overridden by named parameter.

geggleto:
:+1:

silentworks:
Ok I think a solution might actually be:

``` php
    $response->write("Hello " . $request->getAttribute('name', 'Bob'));
    return $response;
});
```

Yes I know this is not happening from the outside in, but it is a solution I think.

AlexStansfield:
My use case sees me needing to access the attribute in Middleware. This middleware will be used across many routes, each route will be have different defaults. So it really needs to be set on the route. 

Example: http://help.slimframework.com/discussions/questions/7376-setting-attributes-in-a-route

Also if you have a reusable class method for the route (rather than a closure) the same theory applies.

silentworks:
We had some discussion on IRC yesterday about this and are still considering if this is the way to go, also could you make your implementation happen inside of the Route.php file rather than in App.php please.

AlexStansfield:
Makes perfect sense, sorry I didn't do that first time. Only thing was really wasn't sure the best name for the method that's called in the run to add the attributes. Now thinking addDefaultAttributesToRequest might have been better.

Crap just saw the php 5.4 error. Will fix.

akrabat:
Presumably, you mean that you want the default attribute available in the route middleware? I don't see how it's available in app-level middleware.

akrabat:
I would prefer two methods: `setAttributes(array $attributes)` and `setAttribute($name, $value)` as these are then analogous to the the Request's `withAttributes` and `withAttribute` methods.

AlexStansfield:
Hmm, that's a very good point. I'd only been looking at it from the point of a Route middleware. But should it also be accessible in App Middleware? I kinda feel it should even though my own use case doesn't need it. Would need to see if that's actually possible or not.

Method names make sense, will rename the original method and add the new one.

Thanks for being so quick to reply guys, I only picked up Slim 3 on Monday but really liking it a lot already.

AlexStansfield:
Renamed methods and added setAttribute.

As the framework executes the App middleware before resolving the route it seems these attributes will definitely not be able to be accessed in App middleware. That also goes for any attributes determined by the url matching (e.g /hello/{name}). I'm not sure if this was by design or a side effect of the order in which things are done.

JoeBengalen:
That is by design:
http://www.slimframework.com/docs/concepts/middleware.html

Routing happens in the app (most inner layer)

AlexStansfield:
Yeah that's fair enough. I was aware that the route isn't dispatched until after the application middleware is executed, just wasn't aware that the resolving of the route wasn't done until I dived into the code itself. 

It makes sense, but not sure if maybe the docs just need to make that a bit clearer so users know that any attributes determined by the route won't be available in the application middleware. But then maybe there are no real use cases for that anyway so it doesn't matter.

AlexStansfield:
So from the merge last night that introduced a Route Arguments attribute I have rebased my branch and squashed down my previous commits then fixed the tests (attributes now got from $request->getAttribute('blah') rather than $args['blah']).

However, I wasn't sure if you'd prefer these to still be going into the request attributes or if they should be treated as route arguments as before. 

Personally I prefer the latter as it means they will always be where I expect them. In my route middlewear I could check for a route argument 'name' whether it's value has been defined by /hello/{name} or explicitly on the route with setAttribute('name', 'blah'). 

The question then is, if we go with route arguments is Attribute really the right term for them now? should it be setRouteArgument (or setArgument) instead?

AlexStansfield:
As there is a fair timezone difference (GMT+7 here) I went ahead and changed this to setArguments. I can rollback this last commit if need be.

akrabat:
I agree with your thinking. `setArgument`/`setArguments` that store into the `routeArguments` sub-array is the right approach as it's the only way to ensure that `$args` has keys for the optional route parameters.

silentworks:
In this case, should this be renamed to `setParam/setParams` as we had in v2?

akrabat:
Fine with me - though I'd also call it `routeParams` rather than `routeArguments` in the rest of the code.

akrabat:
@AlexStansfield, I'm sorry, but I think that my work in #1348 to move the route argument assignment from the App to the strategy has inadvertently re-implemented a good proportion of your work. 

Can you check over and see if it does what you require and if not, update this PR to add what's missing?

AlexStansfield:
Haha, you guys are really making me work for this feature :stuck_out_tongue: 

Will take a look at you work and see what I need to do.

AlexStansfield:
@akrabat, looks like you've pretty much done the work. I need to checkout your branch and test but it looks like you've already added the setArgument public method so in theory you should be able to just add arguments in when defining the route.

I'm going to add the tests I added for this use case. Not sure how best to do this as your code isn't merged yet. Should I do a pull request onto your own branch?

silentworks:
@AlexStansfield it is now merged.

AlexStansfield:
Ok, so not much left for me in the end. I added the setter and getters to RouteInterface. Added the tests to AppTest for the use case of calling setArgument/setArguments in the route definition and I also fixed some type hinting and docblocks (hopefully correctly)

Slim/Interfaces/RouteInterface.php
```diff
@@ -35,6 +53,25 @@ public function getName();
...
 
     /**
+     * Set a route argument
+     *
+     * @param string $name
+     * @param string $value
+     *
+     * @return static
+     */
+    public function setArgument($name, $value);
```
JoeBengalen, line 63: Not sure if setters should be in the interface?

AlexStansfield, line 63: That's a good point and I guess it's a design question. 

Should Slim framework support setting arguments on a route definition (therefore something all route implementations should implement) or should it be a specific feature of the default router only?

Will happily remove them if the latter is preferred.

akrabat, line 63: That's a decision for @codeguy or @silentworks.

index.php
```diff
@@ -31,6 +31,11 @@
     return $response;
 });
 
+$app->get('/hello/world', function ($request, $response, $args) {
+    $response->write("Hello " . $args['extra']);
+    return $response;
+})->setArgument('extra', 'world');
+
```
akrabat: Would it be a better example to put a `->setArgument('name', 'world')` on the `/hello/{name}` route below and change that one's route path to `/hello[/{name}]`?

AlexStansfield: Good idea, will do it now.

AlexStansfield:
Just spotted @silentworks comment about renaming Arguments to Params. Do you want that done in this Pull Request?

silentworks:
@AlexStansfield no lets leave it as Arguments as it could get confused with the `Request` object `getParam/getParams` methods.

AlexStansfield:
Cool, then other than the question raised by @JoeBengalen about the setters in the interface I think this is pretty much final version :grinning:

silentworks:
Ok cool, will wait on @codeguy to discuss the setters in the interface, we already have `setName` in that particular interface so I don't really see it as a issue, but let me have a chat with @codeguy and hear what he says.

I am guessing you are getting ready for bed so, you might just see this all merged in when you wake up or further discussions on it.

Thanks again for all your help and your push in making this change possible.

AlexStansfield:
Only just gone 5pm here (Bangkok) so will be around for a few hours yet. 

Many thanks to you guys for being so helpful and quick to respond. Look forward to working more with Slim.

silentworks:
Will discuss with @codeguy about setters in interfaces later, going to merge this in for now.

AlexStansfield:
Many Thanks!!!

PR #1345, opened 2015-07-09
Add foundHandler that implements a new InvocationStrategyInterface

opengeek:
This will allow a foundHandler in the container to describe what parameters are injected into the callable defined for a found route.

### DIFF
--- a/Slim/Interfaces/InvocationStrategyInterface.php
+++ b/Slim/Interfaces/InvocationStrategyInterface.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Slim Framework (http://slimframework.com)
+ *
+ * @link      https://github.com/codeguy/Slim
+ * @copyright Copyright (c) 2011-2015 Josh Lockhart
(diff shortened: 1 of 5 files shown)


Slim/Interfaces/InvocationStrategyInterface.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    /**
+     * @param ContainerInterface     $container
+     * @param array                  $route
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface      $response
+     *
+     * @return mixed
+     */
+    public function invoke(ContainerInterface $container, array $route, ServerRequestInterface $request, ResponseInterface $response);
```
JoeBengalen, line 26: Why not the magic `__invoke()` method? This custom `invoke()` might be confusing.

And why is the container a parameter here? If it is needed for a strategy it can be inserted in the constructor.

opengeek, line 26: Both suggestions make good sense to me. I'll create another PR unless anyone else sees a problem.

PR #1346, opened 2015-07-09
Use __invoke for InvocationStrategyInterface and remove container

opengeek:
Refactored InvocationStrategyInterface and implementation; now with working code and tests to prove it for both included strategies!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -371,9 +371,12 @@ class App
     {
         $routeInfo = $this->container->get('router')->dispatch($request);
         if ($routeInfo[0] === Dispatcher::FOUND) {
-            /** @var InvocationStrategyInterface $handler */
-            $handler = $this->container->get('foundHandler');
-            return $handler->invoke($this->container, $routeInfo, $request, $response);
+            $routeArguments = [];
+            foreach ($routeInfo[2] as $k => $v) {
+                $routeArguments[$k] = urldecode($v);
+            }
+            $request = $request->withAttribute('routeArguments', $routeArguments);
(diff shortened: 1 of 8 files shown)


Slim/App.php
```diff
@@ -371,9 +371,12 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
         $routeInfo = $this->container->get('router')->dispatch($request);
         if ($routeInfo[0] === Dispatcher::FOUND) {
-            /** @var InvocationStrategyInterface $handler */
-            $handler = $this->container->get('foundHandler');
-            return $handler->invoke($this->container, $routeInfo, $request, $response);
+            $routeArguments = [];
+            foreach ($routeInfo[2] as $k => $v) {
+                $routeArguments[$k] = urldecode($v);
+            }
+            $request = $request->withAttribute('routeArguments', $routeArguments);
```
akrabat, line 378: Why have a sub-array for these arguments? It now means that you can no longer add an attribute in app middleware and have it available in the callback's `$args`.

akrabat, line 378: It also means that there's no way to do default arguments with a stock install of Slim3 without using `isset`.

opengeek, line 378: The isolation was at Josh's suggestion as attributes really aren't the right place for storing route arguments. Default arguments can be done—even in the default RequestResponse strategy—see latest commits on #1343.

silentworks, line 378: @opengeek take a look at #1348 as the currently merged changes broke the default setup, we think this will remedy it.

Issue #1335, opened 2015-07-01
PHPUnit >=4.2.0 warns "Test code or tested code did not (only) close its own output buffers" on PrettyExceptionsTest


grantcox:
PHPUnit 4.2.0 introduced a mandatory check that output buffering calls have not been orphaned (see https://github.com/sebastianbergmann/phpunit/blob/b6ca5dc1d4bc641809c348eea38dede5f722ab82/src/Framework/TestCase.php#L1923).

A route that throws an exception while Slim is in debug mode will not have the output buffer closed, see https://github.com/slimphp/Slim/blob/2.x/Slim/Slim.php#L1350 and the lack of `ob_get_clean()` calls in branches of the `} catch (\Exception $e) {`

This is demonstrated by running the existing PrettyExceptionsTest test with PHPUnit >= 4.2.0

```
grant$ phpunit tests/Middleware/PrettyExceptionsTest.php
PHPUnit 4.0.17 by Sebastian Bergmann.

.....

Time: 58 ms, Memory: 4.50Mb

OK (5 tests, 7 assertions)
```

```
grant$ phpunit tests/Middleware/PrettyExceptionsTest.php
PHPUnit 4.7.6 by Sebastian Bergmann and contributors.

.RRRR

Time: 174 ms, Memory: 14.25Mb

OK, but incomplete, skipped, or risky tests!
Tests: 5, Assertions: 7, Risky: 4.
```

PR #1351, opened 2015-07-12
Workaround for the failing unittests.

denshade:
Added an ob_get_clean in case of debugging. This fixes the annoying error of closed buffers mentioned in the unittests.
Extracted the rendering of Exceptions in a separate class. 
Also a few warnings were fixed.

### DIFF
--- a/Slim/Middleware/PrettyExceptionRenderer.php
+++ b/Slim/Middleware/PrettyExceptionRenderer.php
@@ -1,0 +1,47 @@
+<?php
+
+
+namespace Slim\Middleware;
+
+
+use Exception;
+
+class PrettyExceptionRenderer
+{
+
+    /**
+     * @param Exception $exception
+     * @return string
+     */
+    public static function renderException(Exception $exception)
+    {
+        $title = 'Slim Application Error';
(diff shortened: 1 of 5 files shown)


akrabat:
This is too big a change for Slim2. I'm also not seeing broken unit tests on 2.x.

denshade:
This fixes https://github.com/slimphp/Slim/issues/1335. But ok.

akrabat:
The main trouble is that you've done a fairly major refactor to fix this issue.

I was expecting to see a call `ob_get_clean()` just before https://github.com/slimphp/Slim/blob/2.x/Slim/Slim.php#L1375 or something similar.

denshade:
I actually cleaned up a lot of warnings as well. Extracting PrettyExceptionRenderer was probably not necessary. Stuff like  
@var array Lookup hash of all route objects 
 => 
-     \* @var Route[]
  Allows for decent auto-completion in most IDEs and is nothing more than changing a docblock. So risk free. Also missing return statements are on my grumpy list. On the other side I understand 2.x is no longer in development. You can also just add the one line with ob_get_clean which will fix the issue.

PR #1353, opened 2015-07-13
Code and comment cleanup

JoeBengalen:
Some code and comment cleanups.

Changed the `Exception` to `InvalidArgumentException` in the `Route::setOutputBuffering()` method.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -21,7 +21,6 @@ use Slim\Http\Headers;
 use Slim\Http\Body;
 use Slim\Http\Request;
 use Slim\Interfaces\Http\EnvironmentInterface;
-use Slim\Interfaces\InvocationStrategyInterface;
 use Slim\Interfaces\RouteGroupInterface;
 use Slim\Interfaces\RouteInterface;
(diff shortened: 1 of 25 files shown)


JoeBengalen:
Should this exception also be more specific?
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L82

geggleto:
Ideally @JoeBengalen it should probably throw ... \ContainerInterfaceException

geggleto:
Also if you would be so kind as to add a test case for that, as there currently is one to test for an exception should something bad be passed in for a container!

JoeBengalen:
@geggleto You mean an implementation of this?
https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/Exception/ContainerException.php

geggleto:
Yes :) :+1:

JoeBengalen:
Not sure if that would be right. Its not being thrown by the container, its an error from the application itself. It also requires slim to have its own implementation of the `ContainerException`.

I was more thinking of an `InvalidArgumentException`, as you should have passed an `array` or instance of `ContainerInterface`.

PR #1354, opened 2015-07-13
Added cast to string on Response::withHeader() call

AndrewCarterUK:
With the a zend-diactoros response object, the Response::withHeader() method will fail unless it's given a string.

https://github.com/zendframework/zend-diactoros/blob/master/src/MessageTrait.php (Line 191)

Not sure if you want to accept this change (or if there are other similar issues) - but I had to change this to get it to work with my project.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -287,7 +287,7 @@ class App
             if ($hasBody) {
                 $size = $response->getBody()->getSize();
                 if ($size !== null) {
-                    $response = $response->withHeader('Content-Length', $size);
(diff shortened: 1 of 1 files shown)


silentworks:
That line says if its a string convert it to an array, so I'm not sure how this relates to your change here.

AndrewCarterUK:
The next line (195) checks that it's an array of only strings.

akrabat:
Thanks.

The PSR-7 interface states that the `$value` parameter of `withHeader()` is a string or array of strings. Personally, I'd say that's a little too strict, but it is what it is :)

PR #1355, opened 2015-07-14
Remove unused method `Route::setCallable()`

JoeBengalen:
Couldn't find any use for this method.


silentworks:
@opengeek do you know if this is being used anywhere or what was the purpose of it?

PR #1356, opened 2015-07-14
When request error by ajax - show Prettyprint 

ghost:
Many times when i develop my APP i had problem with debug in on ajax.

I think that will be nice if is ( ajax ) request then Prettyprint res, otherwise i have to copy response to html file then read it

```
Details
Type: ErrorException
Code: 8
Message: Undefined index: ohno
File: [....]adm.php
Line: 23
Trace
[...]
```


DavidePastore:
You can see the rendered HTML in the **HTML** tab. Check it. :dango:

akrabat:
I think you may have opened this PR against the wrong branch? There's over 250 commits in this PR…

akrabat:
Please create a new PR by branching `3.x` then putting your change on your new branch and then creating the PR between your new branch and 3.x.

PR #1357, opened 2015-07-15
Route in App Middleware (proof of concept)

AlexStansfield:
So this is something that may not go down well or might be completely unwanted in Slim but I just wanted to show a proof of concept.

Example use case is an Auth App middleware. Say I have an api with 10 methods, 1 (auth - returns token) doesn't require an auth checking. The other 9 all want to check the Authorization header for a token they can verify before doing anything with the request. Currently I'd need to create the Auth middleware as a route middleware and add it to all 9 methods (could be less with grouping).

What would be nice is to have an App middleware that knows whether it needs to verify a token or not. This middleware would need to know the route.

code example:

```
$app = new Slim\App();

    $doAuth = true;

    if (($route = $request->getAttribute('route'))) {
        $doAuth = $route->getArgument('doAuth', $doAuth);
    }

    if ($doAuth) {
        $response->write('AUTH');
    } else {
        $response->write('NO AUTH');
    }

    return $response;
});

    $response->write("Welcome to Slim!");
    return $response;
})->setArgument('doAuth', false);

    $response->write("Hello, " . $args['name']);
    return $response;
});
```

Currently in slim the above code would prefix all the responses with "AUTH". With the code I've committed it would have the first route as "NO AUTH" and any others (where doAuth isn't explicitly set to false) would have "AUTH".

As I said this is a proof of concept. I don't know if you guys even want this and if you do then I'm sure there maybe a number of issues with the way I've implemented it. 

If this is a non-starter then cool, no worries. But if you're interested would appreciate feedback on how it can be done better. I also realise that adding this sort of change during beta is pretty late in the day, but I'm in no rush for this so happy to wait until a later release.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -341,7 +341,7 @@ class App
         $response = $this->container->get('response');
 
         // Dispatch the Router first if the setting for this is on
-        if ($this->container->get('settings')['dispatchBeforeAppMiddleware'] === true) {
+        if ($this->container->get('settings')['determineRouteBeforeAppMiddleware'] === true) {
             // Dispatch router (note: you won't be able to alter routes after this)
             $request = $this->dispatchRouterAndPrepareRoute($request);
         }
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -340,7 +340,11 @@ class App
         $request = $this->container->get('request');
         $response = $this->container->get('response');
 
-        $request = $this->dispatchRouterAndPrepareRoute($request);
+        // Dispatch the Router first if the setting for this is on
+        if ($this->container->get('settings')['dispatchBeforeAppMiddleware'] === true) {
+            // Dispatch router (note: you won't be able to alter routes after this)
+            $request = $this->dispatchRouterAndPrepareRoute($request);
+        }
 
         // Traverse middleware stack
         try {
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -340,6 +340,8 @@ class App
         $request = $this->container->get('request');
         $response = $this->container->get('response');
 
+        $request = $this->dispatchRouterAndPrepareRoute($request);
+
         // Traverse middleware stack
         try {
             $response = $this->callMiddlewareStack($request, $response);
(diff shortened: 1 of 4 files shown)


tuupola:
This can already be done without changes to Slim code. Middleware has access to `Psr\Http\Message\RequestInterface` which in turn gives you access to path of the request uri. It means the middleware can be passed a configurable list of routes it should passthrough without authenticating (or vice versa). This is how both [slim-basic-auth](https://github.com/tuupola/slim-basic-auth) and [slim-jwt-auth](https://github.com/tuupola/slim-jwt-auth) work.

AlexStansfield:
I appreciate that my example use case was quite simple and it appears there is already a work around. The real case I had for the project in mind is somewhat more complicated and would rely on both arguments set on the route definition and during url matching.

Whilst I could come up with a set of rules and regular expressions and match it against the URL it would duplicate much of the work the router is doing during dispatch which seems a bit counter productive.

AlexStansfield:
Have added URI checking to run dispatch again if the URI has changed before we execute the route. 

Not sure if this is the best solution, but as I said before this is a proof of concept and am happy for any feedback on how to do it better.

AlexStansfield:
Also worth considering, could it be the case that the method changes too?

akrabat:
Yes, the method may change as Slim2 supports the `_METHOD` POST parameter ( http://docs.slimframework.com/routing/put/#method-override). 

In Slim3 a neat way to handle that would be with App middleware that looks for `_METHOD` in the POST parameters and changes the Request's method.

AlexStansfield:
It now checks that the method has changed. Again feedback on just how I'm storing the method/path is welcome. I didn't really want to add another request attribute but wasn't sure if putting it in routeInfo was right either.

AlexStansfield:
Added a setting "dispatchBeforeAppMiddleware" (off by default) to enable this functionality.

Welcome suggestions of a better name for it and also any other improvements to the code.

geggleto:
:+1: This looks good :)

I am unsure of the name...but not a show stopper

AlexStansfield:
Have rebased onto latest 3.x commits to make sure there were no conflicts

akrabat:
Thanks! I intending to look at it today.

AlexStansfield:
Quick example of how to use it.

```
<?php

$app = new Slim\App(['dispatchBeforeAppMiddleware' => true]);

    if ($request->getAttribute('route')->getArgument('auth', true)) {
        $response->write('Authed: ');
    }
});

    $response->write("Welcome to Slim!");
    return $response;
})->setArgument('auth', false);

$app->get('/hello[/{name}]', function ($request, $response, $args) {
    $response->write("Hello, " . $args['name']);
    return $response;
})->setArgument('name', 'World!');

$app->run();
```

silentworks:
Is it me or the name doesn't have a meaning without context? I look at it without reading all the messages above and wonder what is it dispatching before the app middleware?

akrabat:
It's not just you. How about `determineRouteBeforeAppMiddleware` ?

AlexStansfield:
Yeah, you're right it does need context. Was a little worried about the name being too long. Happy to change it to whatever you guys agree on

akrabat:
Thanks for your work on this!

silentworks:
@AlexStansfield thanks for your time and hard work on this feature, we finally got it in.

AlexStansfield:
Yippee, thanks a lot guys!

PR #1359, opened 2015-07-15
make sure optional methods are callable

pine3ree:


akrabat:
Was there a specific use-case you came across that prompted this change?

silentworks:
I think this is to do with method visibility, I think `method_exists` will find private and protected methods even though they are unusable outside of their context.

silentworks:
Yeah tested this and my thinking above is correct.

lalop:
I think we should create a new interface that extends RouteInterface or some traits that define this methods but testing if the method exists doesn't seem to be a good partern since all is based on interface.

PR #1363, opened 2015-07-16
Move default register services

geggleto:
Move the default register of slim services to a trait for easy inclusion in other containers


codeguy:
Can we assume to know how other containers register services?

geggleto:
@codeguy I know this is jumping the gun a bit... and I figured we would need to wait to see where PSR-11 lands... but this seems to be a good bet imho.

Slim/DefaultServicesTrait.php
```diff
@@ -0,0 +1,209 @@
...
+ */
+trait DefaultServicesTrait
+{    
+
+    /**
+     * Default settings
+     *
+     * @var array
+     */
+    private $defaultSettings = [
```
nunocodex, line 55: You cannot access with private method, for trait you need to put public.

mnapoli:
> I know this is jumping the gun a bit... and I figured we would need to wait to see where PSR-11 lands... but this seems to be a good bet imho.

This is jumping the gun a lot ;) There is very little chance that a standard configuration format could exist, there is even less chances that it would be supported by Pimple in its current version and even less chance that it would look like this even if it existed.

I don't think this change brings any value today, and this could easily be done in the future if it turns out necessary. What do you think?

silentworks:
This is thinking way to far ahead, I am going to close this for now and if in the future PSR-11 should allow for this, then open PR again.

PR #1364, opened 2015-07-17
Allow use of FastRoute's cachedDispatcher

silentworks:
This would allow for changing out the dispatcher to use the cachedDispatcher, however the cachedDispatcher has some issues with the way we are handling routing. We also no longer need to extend `RouteCollector` as this all happens with the FastRoute function.

Example of how you would change the dispatcher.

``` php
$app = new Slim\App();
$app->get('/', function ($req, $res) {
    $res->write('Hello World');
});

// You can only change dispatcher after all your routes are defined 
// as you will need them for the cachedDispatcher

$router = $app->getContainer()['router'];
$router->setDispatcher(\FastRoute\cachedDispatcher(function ($r) use ($router) {
  foreach ($router->getRoutes() as $route) {
    $r->addRoute($route->getMethods(), $route->getPattern(),  $route->getIdentifier());
  }
}, [
    'cacheFile' => __DIR__ . '/route.cache'
]));

$app->run();
```

You will notice this throwing an error at the moment as FastRoute is using var_export to create the cache file and by default var_export attaches __set_state to all objects it exports, so on re-import when it imports the code it will look for that method on every object and closure it exported.

There is also a secondary issue with cachedDispatcher if you add more than 1 route where we have an issue of circular referencing.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -8,6 +8,7 @@
  */
 namespace Slim;
 
+use FastRoute\Dispatcher;
 use InvalidArgumentException;
 use RuntimeException;
 use Psr\Http\Message\ServerRequestInterface;
@@ -15,8 +16,6 @@ use FastRoute\RouteCollector;
 use FastRoute\RouteParser;
 use FastRoute\RouteParser\Std as StdParser;
 use FastRoute\DataGenerator;
-use FastRoute\DataGenerator\GroupCountBased as GroupCountBasedGenerator;
-use FastRoute\Dispatcher\GroupCountBased as GroupCountBasedDispatcher;
 use Slim\Interfaces\RouterInterface;
 use Slim\Interfaces\RouteInterface;
 
@@ -28,7 +27,7 @@ use Slim\Interfaces\RouteInterface;
  * finding routes that match the current HTTP request, and creating
  * URLs for a named route.
  */
-class Router extends RouteCollector implements RouterInterface
+class Router implements RouterInterface
 {
     /**
      * Parser
@@ -52,6 +51,12 @@ class Router extends RouteCollector implements RouterInterface
     protected $routes = [];
 
     /**
+     * Route counter incrementer
+     * @var int
+     */
+    protected $routeCounter = 0;
+
+    /**
      * Named routes
      *
      * @var null|Route[]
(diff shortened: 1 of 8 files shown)


codeguy:
Let's plan on pulling this in for 3.1.

silentworks:
Now we can use the Route caching above without any issues at all, thanks to @nikic for pointing out that you can't serialize Closures, so a lookup table was necessary.

https://github.com/nikic/FastRoute/issues/63

Slim/Router.php
```diff
@@ -45,6 +44,12 @@ class Router extends RouteCollector implements RouterInterface
     protected $routes = [];
 
     /**
+     * Route id incrementer
+     * @var int
+     */
+    protected $routeId = 0;
```
akrabat: Can we call this `routeCounter` please.

silentworks: Yes that is better naming, sorry I am bad with naming variables at times.

Slim/Router.php
```diff
@@ -191,6 +217,18 @@ public function popGroup()
     }
 
     /**
+     * @param $name
+     * @return \Slim\Route
+     */
+    public function lookupRoute($name)
```
akrabat: Change `$name` to `$identifier` to avoid it being confused with the route's name

yurevichcv:
:+1: Wanted to submit similar PR, but you were faster :) Ability to cache routes is crucial for apps with high load. Actually, it will be nice to have possibility to cache them in memory. @silentworks

Slim/Router.php
```diff
@@ -160,11 +166,34 @@ public function finalize()
...
+            $uri
+        );
+    }
+
+    /**
+     * @return \FastRoute\Dispatcher
+     */
+    protected function createDispatcher()
+    {
+        return $this->dispatcher ?: \FastRoute\simpleDispatcher(function (RouteCollector $r) {
```
akrabat: Can we have a `use` statement for this as isn't it a capital "S"?

JoeBengalen: Its a function, not a method so lowercase is OK. https://github.com/nikic/FastRoute/blob/master/src/functions.php#L12

Importing function within a namespace required PHP 5.6, so use cannot be used.

silentworks: Its a lowercase s, its an function.

akrabat:
Yeah - looks like a 3.1 issue as I can't get it work :)

This is my test case:

``` php
<?php
$app = new \Slim\App();

$app->get(
    '/',
    function ($request, $response) {
        return $response->write("Hello World!");
    }
);

$router = $app->getContainer()['router'];
$router->setDispatcher(\FastRoute\cachedDispatcher(function ($r) use ($router) {
    foreach ($router->getRoutes() as $route) {
        $r->addRoute($route->getMethods(), $route->getPattern(), [$route, 'run']);
    }
}, [
    'cacheFile' => __DIR__ . '/../route.cache'
]));


// Run app
$app->run();
```

I get this error: http://i.19ft.com/bbdea8a4.png

akrabat:
I'd like some of these changes for 3.0 though:
1. Updated RouteInterface and RouterInterface to match what we actually do
2. passing the router to `dispatchRouterAndPrepareRoute()`
3. Possibly the changes to `Route` & `Router` that affect the function signatures.

geggleto:
Since most of this stuff is internal there should be no user-land BC breaks right?

silentworks:
The interfaces would lead to BC breaks.

akrabat:
Okay. The problem was the way I was setting up the cachedDispatcher, not with the code itself. Once I fixed that, all was well.

I understand what the code does and why, so was happy to merge.

PR #1369, opened 2015-07-20
UploadedFile->moveTo() should expect targetPath to include filename

### LINKED ISSUE #1368: UploadedFile->moveTo() - specify filename?
Hi guys,

maybe I just overlooked how to do it, but I could not find a way to specify the file name when I want to move an uploaded file.

The current implementation of Slim 3 always uses the name provided by the client, one can only specify the target directory.
http://www.php-fig.org/psr/psr-7/ is not very clear on this subject.
On the one hand, the parameter for moveTo is called `$targetPath`, on the other there is an example, where the path AND filename are passed to the function.

Martin

opengeek:
As per PSR-7 example

Should resolve #1368

### DIFF
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -216,15 +216,14 @@ class UploadedFile implements UploadedFileInterface
             throw new RuntimeException('Uploaded file already moved');
         }
 
-        if (!is_writable($targetPath)) {
+        if (!is_writable(dirname($targetPath))) {
             throw new InvalidArgumentException('Upload target path is not writable');
         }
 
         $targetIsStream = strpos($targetPath, '://') > 0;
         if ($targetIsStream) {
-            $target = rtrim($targetPath, '/') . '/' . $this->name;
-            if (!copy($this->file, $target)) {
-                throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $target));
+            if (!copy($this->file, $targetPath)) {
+                throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $targetPath));
             }
             if (!unlink($this->file)) {
                 throw new RuntimeException(sprintf('Error removing uploaded file %1s', $this->name));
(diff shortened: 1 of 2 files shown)


Slim/Http/UploadedFile.php
```diff
@@ -238,9 +237,8 @@ public function moveTo($targetPath)
                 throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $targetPath));
             }
         } else {
-            $target = rtrim($targetPath, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $this->name;
-            if (!rename($this->file, $target)) {
-                throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $target));
+            if (!rename($this->file, $targetPath)) {
```
harikt, line 240: Did you missed the name ?

Slim/Http/UploadedFile.php
```diff
@@ -216,15 +216,14 @@ public function moveTo($targetPath)
...
+        if (!is_writable(dirname($targetPath))) {
             throw new InvalidArgumentException('Upload target path is not writable');
         }
 
         $targetIsStream = strpos($targetPath, '://') > 0;
         if ($targetIsStream) {
-            $target = rtrim($targetPath, '/') . '/' . $this->name;
-            if (!copy($this->file, $target)) {
-                throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $target));
+            if (!copy($this->file, $targetPath)) {
```
harikt, line 225: same here. I think this will just copy the file, but not to the name it needs to be.

JoeBengalen, line 225: Isnt the `$targetPath` including the new filename?

opengeek, line 225: Read the issue folks — it was using the client provided name previously (i.e. the original filename) appended to the targetPath, but targetPath should be the full path including the file name you want to move it to. This change fixes the incorrect implementation.

Slim/Http/UploadedFile.php
opengeek, line 240: No — this is the correction to not use the client-provided file name. The intention is that `moveTo()`'s targetPath includes the file name to move it to.

Slim/Http/UploadedFile.php
harikt, line 225: Thank you for clarifying @opengeek .

PR #1374, opened 2015-07-22
Router class public method "getNamedRoute" added

yurevichcv:
As far as I can see, you already have functionality of getting named routes inside `pathFor` method:

```
public function pathFor($name, array $data = [], array $queryParams = [])
{
        if (is_null($this->namedRoutes)) {
            $this->buildNameIndex();
        }
        if (!isset($this->namedRoutes[$name])) {
            throw new RuntimeException('Named route does not exist for name: ' . $name);
        }
        $route = $this->namedRoutes[$name];
        ...
        ... 
}
```

But method for getting route by name is missing. It was extremely useful in Slim 2.

Why not to add separate public method `getNamedRoute`:

```
public function getNamedRoute($name)
{
        if (is_null($this->namedRoutes)) {
            $this->buildNameIndex();
        }
        if (!isset($this->namedRoutes[$name])) {
            throw new RuntimeException('Named route does not exist for name: ' . $name);
        }
        return $this->namedRoutes[$name];
}
```

And refactor `pathFor` method to use it as well:

```
public function pathFor($name, array $data = [], array $queryParams = [])
{
        $route = $this->getNamedRoute($name);
        ...
        ... 
}
```

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -151,6 +151,26 @@ class Router extends RouteCollector implements RouterInterface
     }
 
     /**
+     * Get named route object
+     *
+     * @param string $name        Route name
+     *
+     * @return Route
+     *
+     * @throws RuntimeException   If named route does not exist
+     */
+    public function getNamedRoute($name)
+    {
+        if (is_null($this->namedRoutes)) {
+            $this->buildNameIndex();
+        }
+        if (!isset($this->namedRoutes[$name])) {
+            throw new RuntimeException('Named route does not exist for name: ' . $name);
+        }
+        return $this->namedRoutes[$name];
+    }
+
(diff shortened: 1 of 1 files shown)


geggleto:
:+1: I like this as I do use it in Slim 2, will help migration projects

AlexStansfield:
Nice, this will come in very handy for what I'm working on :+1:

silentworks:
I wonder if this should also be defined in the RouterInterface. I personally think it should and would prefer that be added to the PR before we do a merge.

akrabat:
Thanks @yurevichcv!

@silentworks I added `getNamedRoute` to the interface during merge.

PR #1378, opened 2015-07-24
Various small fixes

micheh:
This pull request fixes some small potential issues. I used a separate commit for each type (so you have the option to cherry-pick/squash the ones you like).
- Resolved undefined method calls: There were some issues where a method of a Slim implementation was called, which would not work with another pure PSR-7 implementation. Note: The NotFoundHandler still uses the `Uri->getBasePath()`, but there is no useful equivalent in PSR-7.
- Added missing php document tag
- Used explicit `return null` statements instead of simply `return`
- Reordered checks in `if` statements to use the least expensive expression first
- Used a more explicit `InvalidArgumentException` instead of a general `Exception`

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -10,6 +10,7 @@ namespace Slim;
 
 use Exception;
 use Closure;
+use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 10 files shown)


akrabat:
Thanks!

PR #1381, opened 2015-07-25
Fix using CallableResolvable as group middleware

### LINKED ISSUE #1377: Slim3 middleware on route groups doesn't work
The problem exists when adding middleware to route groups. Example below:

```
class Test {
    public function run($request, $response, $next){
        $response->write('+');
        $response->write('-');
        return $response;
    }
}
```
#### Fails

```
$app = new \Slim\App();
$app->group('/admin', function () use ($app) {
    $app->get('/test', function($request, $response, $args) {
        return $response->write('works!');
    });
})->add('Test:run');
$app->run();
```
#### Works

But if you add same middleware to single route everything works fine.

```
$app = new \Slim\App();
$app->group('/admin', function () use ($app) {
    $app->get('/test', function($request, $response, $args) {
        return $response->write('works!');
    })->add('Test:run');
});
$app->run();
```

opengeek:
The container was not available for resolving a CallableResolvable when invoking a RouteGroup

Should resolve #1377

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -264,7 +264,9 @@ class App
      */
     public function group($pattern, $callable)
     {
+        /** @var RouteGroup $group */
         $group = $this->container->get('router')->pushGroup($pattern, $callable);
+        $group->setContainer($this->container);
         $group($this);
         $this->container->get('router')->popGroup();
         return $group;


Slim/App.php
```diff
@@ -263,7 +263,7 @@ public function map(array $methods, $pattern, $callable)
      */
     public function group($pattern, $callable)
     {
-        $group = $this->container->get('router')->pushGroup($pattern, $callable);
+        $group = $this->container->get('router')->pushGroup($this->container, $pattern, $callable);
```
silentworks: Just add `$group->setContainer($this->container);` after `$group = $this->container->get('router')->pushGroup($pattern, $callable);` and that should fix the issue. Sorry only realised this after you made all these changes.

PR #1384, opened 2015-07-27
App responder fix

yurevichcv:
As far as $app->respond() function applies changes to immutable $response object - it should return new actual state of this object. Otherwise return value of $app->run() is not valid/actual.


yurevichcv:
Posted alternative version for this issue - #1385, which I personally like much more

akrabat:
Closed in favour of #1385.

PR #1385, opened 2015-07-27
App responder fix and logic separation (alternative solution)

yurevichcv:
Alternative solution for my PR #1384 

I guess this one is much better for many reasons:
1. Good logic separation: `respond()` method should not modify Response object, it should just output headers and body. Instead I've added `finalize()` method, which finalizes response and returns finalized value.
2. Good for Unit testing of `run()` function, where you don't need `respond` method in most cases.
3. Now App itself is "reusable" in Unit tests. For huge projects it is critical. For example unit tests for my project run almost for 40 minutes if I create App object from scratch all the time (with all Middleware, Redis/ElasticSearch/Doctrine/Acl/S3/etc settings). But if I make App object static and reusable for all my Test Cases and only inject new response/request objects - unit tests run only 15 minutes. Before it was not possible to "reuse" App because of static variable `$responded = true` inside `respond()` method.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -17,6 +17,7 @@ use Psr\Http\Message\ResponseInterface;
 use Interop\Container\ContainerInterface;
 use FastRoute\Dispatcher;
 use Slim\Exception\Exception as SlimException;
+use Slim\Http\Response;
 use Slim\Http\Uri;
 use Slim\Http\Headers;
 use Slim\Http\Body;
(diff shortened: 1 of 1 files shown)


geggleto:
Oh so much yes on this... :100: \* :+1:

yurevichcv:
The only thing I don't like is code duplication here:

```
$statusCode = $response->getStatusCode();
$hasBody = ($statusCode !== 204 && $statusCode !== 304);
```

If you know how to fix it, please, let me know.

Slim/App.php
```diff
@@ -145,6 +145,34 @@ public function get($pattern, $callable)
...
+        }
+
+        if (is_callable([$route, 'setOutputBuffering'])) {
+            $route->setOutputBuffering($this->container->get('settings')['outputBuffering']);
+        }
+
+        return $route;
+    }
+
+    /**
```
akrabat: Please put this method back under `any`

yurevichcv: Done. Sorry, my IDE sometimes does this trick...

Issue #1386, opened 2015-07-27
$request->getBody() returns empty stream


bota0mg:
When I use _$request->getParsedBody()_ everything works fine, but when I use _$request->getBody()_ it returns just empty stream. According to geggleto it looks like the stream gets detached at some point.

demo: [source](http://ix.io/jXD), [output](http://postimg.org/image/eigc7tnbb/)

JoeBengalen:
Maybe something with the pointer already being at the end after parsing the body?

bota0mg:
_rewind()_ing didn't help so I don't think that's the problem

geggleto:
https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Stream.php#L181

shows that the contents of the stream are returned, thus theres nothing left in the string.
slim uses getParsedBody internally which is why if its a parsable item the body is left empty.
Rewinding does not help because PHP is dumping the content as it reads it as its basically a byte buffer.

Only in a file implementation could you rewind since the data is on disk and not in memory.

I do believe this to be a bug in our implementation.

Upon Request class initialization the stream should be read in full and stored in a protected attribute, then it can be returned at will via __toString and getContents()

Outfl3sh:
But how to make *getBody()* functional and remain PSR-7 compliant? *getParsedBody()*, which consumes the original stream, gets called all over the place and *getBody()* should return *StreamInterface*. Storing the contents in the Request class won't help either because *string* is just not *StreamInterface*. Maybe some kind of cached *Stream*, but i don't know if that's a proper way to do things. In the end I never really used PHP more than needed.

bota0mg:
As @geggleto mentioned according to [php.net](http://php.net/manual/en/wrappers.php.php) **php://input can be read only once**, because right now there is no support for a seek operation. I think there is a really simple solution of just dumping the whole body into a string and then creating a stream from it. Not really a clean solution, but I can't seem to find a better one.

geggleto:
For reference: https://github.com/phly/http/blob/master/src/PhpInputStream.php#L38

geggleto:
//cc @akrabat

akrabat:
@Outfl3sh The tip of the 3.x branch should fix this issue for you.

bota0mg:
Great, thanks for the fix.

PR #1389, opened 2015-07-28
#1386 fix

bota0mg:


geggleto:
The one problem with this fix is that it reads the entire input stream into memory again... Given post_max sizes this is not ideal. However there does need to be a solution for this issue. We might need an additional wrapper around the current stream api implementation of php.

bota0mg:
I'm afraid that I don't have either time and skills for that, at least not right now.

geggleto:
That's no problem, thank you for letting us know!

JoeBengalen:
Or look at phly:

https://github.com/phly/http/blob/master/src/PhpInputStream.php

He puts every read a cache property so it can be read again from there if needed.

PR #1390, opened 2015-07-28
Update Slim.php

mikicaivosevic:
Refactor view method if else


geggleto:
It is unlikely that this will merged as we are only accepting security fixes to the 2.x branch.

silentworks:
Thank you for your contribution but unfortunately I wont be able to merge it for reasons stated above, we are currently only accepting security fixes for 2.x branch.

PR #1393, opened 2015-07-29
Introduce a reusable RequestBody and ensure Body::__toString() rewinds properly

opengeek:
This introduces a reusable RequestBody implementation that by default copies the `php://input` stream, which can only be read one time, to a `php://temp` stream that can be re-wound and re-read.

Also makes sure `Body::__toString()` does not throw any exceptions.

This should address issue #1386.

### DIFF
--- a/Slim/Http/RequestBody.php
+++ b/Slim/Http/RequestBody.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Slim Framework (http://slimframework.com)
+ *
+ * @link      https://github.com/codeguy/Slim
+ * @copyright Copyright (c) 2011-2015 Josh Lockhart
+ * @license   https://github.com/codeguy/Slim/blob/master/LICENSE (MIT License)
+ */
+namespace Slim\Http;
+
+/**
+ * Provides a PSR-7 implementation of a reusable raw request body
+ */
+class RequestBody extends Body
+{
+    /**
+     * Create a new RequestBody.
+     */
+    public function __construct()
+    {
+        $stream = fopen('php://temp', 'w+');
+        stream_copy_to_stream(fopen('php://input', 'r'), $stream);
+        rewind($stream);
+
+        parent::__construct($stream);
+    }
+}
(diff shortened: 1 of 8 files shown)


JoeBengalen:
Will `stream_copy_to_stream` be faster than appending each string being read to a cache property like phly does?
https://github.com/phly/http/blob/master/src/PhpInputStream.php

opengeek:
I tried both approaches and did some profiling with them but could not find any significant difference with small, basic payloads. The approach of keeping it an actual stream rather than caching the content should provide more flexibility within middleware usage however, with only slightly more memory usage. Copying it to the temp stream and discarding the original input stream should theoretically save memory usage however, with larger payloads since it uses temp files after 5Mb of memory is used.

I may try to find a way to test this with some larger payloads to better understand the impact. And now I wonder if I should call this implementation `ServerRequestBody`.

Slim/Http/RequestBody.php
```diff
@@ -0,0 +1,37 @@
...
+class RequestBody extends Body
+{
+    protected static $tempStream;
+
+    /**
+     * Create a new RequestBody.
+     *
+     * @param resource|null $stream
+     */
+    public function __construct($stream = null)
```
akrabat: I don't think we need to accept a  `$stream` parameter here. This is a specialist class solely for handling `php://input`

Slim/Http/RequestBody.php
```diff
@@ -0,0 +1,37 @@
...
+ * @license   https://github.com/codeguy/Slim/blob/master/LICENSE (MIT License)
+ */
+namespace Slim\Http;
+
+/**
+ * Provides a PSR-7 implementation of a reusable raw request body
+ */
+class RequestBody extends Body
+{
+    protected static $tempStream;
```
akrabat: Unsure why we need to keep the tempStream around?

akrabat:
Can we simplify `RequestBody` to something like:

``` php
class RequestBody extends Body
{
    /**
     * Create a new RequestBody.
     */
    public function __construct()
    {
        $stream = fopen('php://temp', 'w+');
        stream_copy_to_stream(fopen('php://input', 'r'), $stream);
        rewind($stream);

        parent::__construct($stream);
    }
}
```

i.e. what do we gain with your current implementation over this?

PR #1398, opened 2015-07-30
Updated version number references for Slim 2

### LINKED ISSUE #1397: Slim 2 version mismatch
According to Composer, installed version of Slim 2 is `2.6.2`:

> slim/slim                      **2.6.2**              Slim Framework, a PHP micro framework

According to [the code](https://github.com/slimphp/Slim/blob/e44a9198cbc05b75f720b0d9d12725e29eb3c310/Slim/Slim.php#L57), version is still listed as `2.6.1`

pixelchutes:
Fixes #1397

### DIFF
--- a/Slim/Environment.php
+++ b/Slim/Environment.php
@@ -6,7 +6,7 @@
  * @copyright   2011 Josh Lockhart
  * @link        http://www.slimframework.com
  * @license     http://www.slimframework.com/license
- * @version     2.6.1
+ * @version     2.6.2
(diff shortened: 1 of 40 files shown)


akrabat:
I'm not sure what to do about this. As 2.6.2 is released, we can't retro-fit it, so presumably it should be set to 2.6.3?

pixelchutes:
Makes sense to me. PR updated to target next 2.x patch release.

PR #1400, opened 2015-08-03
PSR-2 code style without errors

micheh:
Fix the remaining PSR-2 code style errors.
I also added PHP Code Sniffer to Travis. The build for PHP 5.6 will fail if errors are detected (warnings are ignored).


prisis:
Or we can use styleci.io :) to check if all pull request are PSR-2

PR #1405, opened 2015-08-04
Add method to get the route name

micheh:
Added a method to the `Request` object to get the route name of the request. This is very useful if you map multiple routes to the same callback and have to decide which route was called.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1350,4 +1350,21 @@ class Request implements ServerRequestInterface
 
         return isset($this->serverParams['REMOTE_ADDR']) ? $this->serverParams['REMOTE_ADDR'] : null;
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


silentworks:
At what point of your application would you need this? You can get the currently called route using

``` php
$route = $request->getAttribute('route');
$route->getName();
```

micheh:
@silentworks I use this in the route callable.

I didn't know that the `route` attribute exists, only saw the `routeInfo`. Can you tell me where this is added to the request?
Edit: Nevermind, found it in `Route::prepare`. Great!

silentworks:
Take a look here https://github.com/slimphp/Slim/blob/3.x/Slim/Route.php#L274 and it gets called here https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L447

Issue #1380, opened 2015-07-25
relative path in withPath generates double slashes in URI. 


asaokamei:
Hi, 

When setting a relative path using `withPath` method when no basePath is present (i.e. just '/'), the generated URI contains double slashes. 

example:

``` php
$environment = \Slim\Http\Environment::mock([
    'SCRIPT_NAME' => '/index.php',
    'REQUEST_URI' => '/bar',
]);
$uri = \Slim\Http\Uri::createFromEnvironment($environment);
$this->assertEquals('http://localhost/test', (string) $uri->withPath('test'));
```

the current result is 

```
http://localhost//test
```

where it should be 

```
http://localhost/test
```

thanks,

micheh:
@asaokamei Your example creates the expected URI without double slashes. Do you have this issue with the current version?

asaokamei:
@micheh Yes, it still fails with the current 3.x branch. 
So, I made a PR #1413; I hope that was OK...

Here's the output from the phpunit. 

``` sh
vendor/bin/phpunit 
PHPUnit 4.7.7 by Sebastian Bergmann and contributors.

...............................................................  63 / 357 ( 17%)
............................................................... 126 / 357 ( 35%)
............................................................... 189 / 357 ( 52%)
............................................................... 252 / 357 ( 70%)
............................................................... 315 / 357 ( 88%)
.........F................................

Time: 906 ms, Memory: 7.75Mb

There was 1 failure:

1) Slim\Tests\Http\UriTest::testWithPathWhenBaseRootIsEmpty
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://localhost/test'
+'http://localhost//test'

/Users/asao/vm/ubuntu1310/shared/www/Slim/tests/Http/UriTest.php:582
/Users/asao/vm/ubuntu1310/shared/www/Slim/vendor/phpunit/phpunit/phpunit:36
```

FYI, the PHP is version 5.6.5

``` sh
 php -v
PHP 5.6.5 (cli) (built: Feb 12 2015 10:39:21) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies
    with Xdebug v2.2.5, Copyright (c) 2002-2014, by Derick Rethans
```

micheh:
[...]

Issue #1375, opened 2015-07-22
Uri::createFromEnvironment missing basePath?


JoeBengalen:
Checkout the code below. It seems rather weird to me that the `basePath` disappears if you do not update the path. I would expect the same request uri to be returned when casting to string as was set in the environment.

Therefore I think the `Uri::createFromEnvironment()` should not include the root slash in `path` when a `basePath` is present. Am I right on this?

``` php
$uri = Uri::createFromEnvironment(
    Environment::mock([
        'SCRIPT_NAME' => '/foo/index.php',
        'REQUEST_URI' => '/foo/bar',
    ])
);

var_dump($uri);
// 'basePath' => '/foo'
// 'path'     => '/bar'

// The behaviour of the `__toString()` method seems fine:

var_dump((string) $uri->withPath('/baz')); // -> http://localhost/baz
// Results as expected: without the basepath, as it should not be included in a path with root (/)

var_dump((string) $uri->withPath('baz')); // -> http://localhost/foo/baz
// Results as expected: including the basepath, as it should be included in a path without root (/)

// But when casting back to string without any modifications the basePath is gone!
var_dump((string) $uri); // -> http://localhost/bar
// This is because the Uri::createFromEnvironment() method sets the path with the root (/)

// When removing that root slash the basePath is back again:
var_dump((string) $uri->withPath('bar')); // -> http://localhost/foo/bar
```

PR #1411, opened 2015-08-06
Correct path with Uri::createFromEnvironment & base path

micheh:
If you use `Uri::createFromEnvironment` the path is always set as an absolute path. This causes an issue if you have a base path, as the absolute path will ignore it (issue #1375).

``` php
'SCRIPT_NAME' => '/foo/index.php',
'REQUEST_URI' => '/foo/bar',
```

The current implementation creates the following values from the environment above:

```
Base: /foo
Path: /bar (absolute)
Result: /bar (wrong)
```

This pull request fixes this by using a relative path if the base path is available.

```
Base: /foo
Path: bar (relative)
Result: /foo/bar (correct)
```

Using a relative path is in line with PSR-7:

> If the path is intended to be domain-relative rather than path relative then it must begin with a slash ("/"). Paths not starting with a slash ("/") are assumed to be relative to some base path known to the application or consumer.

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -195,12 +195,13 @@ class Uri implements UriInterface
         $virtualPath = $requestUri;
         if (strpos($requestUri, $requestScriptName) === 0) {
             $basePath = $requestScriptName;
-            $virtualPath = substr($requestUri, strlen($requestScriptName));
-        } elseif (strpos($requestUri, $requestScriptDir) === 0) {
+        } elseif ($requestScriptDir !== '/' && strpos($requestUri, $requestScriptDir) === 0) {
             $basePath = $requestScriptDir;
-            $virtualPath = substr($requestUri, strlen($requestScriptDir));
         }
-        $virtualPath = '/' . ltrim($virtualPath, '/');
+
+        if ($basePath) {
+            $virtualPath = ltrim(substr($requestUri, strlen($basePath)), '/');
+        }
 
         // Query string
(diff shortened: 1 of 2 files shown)


micheh:
Integrated test from #1413 (also fixes #1380). cc @asaokamei

akrabat:
Given a url of `http://localhost/slim3/develop-testbed/public/bar` where `index.php` is in the `public` directory, I get Page Not Found for this route:

``` php
$app->get('/bar', function ($request, $response, $args) {
    return $response->write("<p>HERE</p>");
})->setName('bar');
```

This worked before applying this PR.

Note, however, that if I set up a vhost, so that I have `http://develop-testbed.slim3.dev/bar` as my URL,  then the route callable is correctly called.

(Edited to say what actually happens!)

akrabat:
I would guess that the fix is to update the `Router`'s `dispatch` method to always ensure that there's a leading `/` on the URI to be dispatched. Maybe something like:

``` diff
diff --git a/Slim/Router.php b/Slim/Router.php
index cca8634..c3c6c3d 100644
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -134,9 +134,10 @@ class Router extends RouteCollector implements RouterInterface

         $dispatcher = new GroupCountBasedDispatcher($this->getData());

+        $uri = '/' . ltrim($request->getUri()->getPath(), '/');
         return $dispatcher->dispatch(
             $request->getMethod(),
-            $request->getUri()->getPath()
+            $uri
         );
     }
```

PR #1412, opened 2015-08-06
Introduce common Message class

micheh:
The `Request` and `Response` class each implemented the methods of the `MessageInterface` on their own, resulting in many duplicated lines of code.

This pull requests extracts these methods into a common `Message` class.


geggleto:
This is a non-starter. It is required by PSR-7.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message.md

PR #1415, opened 2015-08-07
Fix ltrim

JoeBengalen:
Code should remove all slashes except for one.


micheh:
Just an idea: What if we move this code to `filterPath`?

JoeBengalen:
That means that the result `getPath()` will be different too. I think following the psr that should not happen? Only specific in the `__toString()` method.

PR #1417, opened 2015-08-07
Add parameter name based invocation strategy

micheh:
This pull request adds a parameter name based invocation strategy (comparable to Slim 2). The strategy will use Reflection to inspect the parameter names of the callable and map the argument names of the route to the parameter names. The special parameter names `request`, `req`, `response` and `res` are mapped to the `Request`/`Response` objects.

It supports both (anonymous) functions and class methods. You can order the parameters any way you like and even use optional parameters. Due to the Reflection you probably wouldn't use this in a high performance application, but it will add a nice and concise way for simpler usages.

Example:

``` php
$app->get('/hello/{name}', function ($name, $response, $from = 'Josh') {
    $greeting = sprintf('%s said: "Hello, %s!"', $from, $name);
    return $response->write($greeting);
});

// Call /hello/Michel => Josh said: "Hello, Michel!"
```

### DIFF
--- a/Slim/Handlers/Strategies/NameBased.php
+++ b/Slim/Handlers/Strategies/NameBased.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * Slim Framework (http://slimframework.com)
+ *
+ * @link      https://github.com/codeguy/Slim
+ * @copyright Copyright (c) 2011-2015 Josh Lockhart
+ * @license   https://github.com/codeguy/Slim/blob/master/LICENSE (MIT License)
+ */
+namespace Slim\Handlers\Strategies;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use ReflectionFunction;
+use ReflectionMethod;
+use RuntimeException;
+use Slim\Interfaces\InvocationStrategyInterface;
+
+/**
+ * Route callback strategy which maps the route argument names to the callable parameter names.
+ */
+class NameBased implements InvocationStrategyInterface
+{
+    /**
+     * Inspect the signature of the route callable and match the route argument names with the
+     * parameter names of the callable.
+     *
+     * @param array|callable         $callable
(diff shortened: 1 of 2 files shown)


JoeBengalen:
Kinda like the idea of the flexible parameters, although Reflection may impact the performance a bit.
Therefore I think it should not be the default invocation strategy, but having the option available would be nice.

Also maybe a better name would be `ParameterBased`?

silentworks:
Why would I use this over https://github.com/slimphp/Slim/blob/3.x/Slim/Handlers/Strategies/RequestResponseArgs.php, which gives you similar setup to what we had in Slim 2 and less overheads?

akrabat:
I don't think that this needs to be in Slim itself, but instead offered as a separate package via composer.

micheh:
Use cases:
- Minimal callable signature and avoid unused parameters (for example if you don't need access to the request and/or response object or if they are already injected in another way, say a Controller constructor). The callable will also be easier to test.
- Custom parameter order (for example if you set the response as an optional last parameter to be able to call the callable in another context where the callable creates a new response itself).

Note: I don't propose to use this strategy by default, but in my opinion it is very convenient and it would also provide the same functionality right out of the box as the other micro frameworks.

Since you use Reflection only once per request, the performance hit won't be that serious. Interestingly, in my Apache Bench tests (`-c 100 -n 5000`) the Reflection strategy always had a slightly lower mean and median. Not sure why, though.

```
RequestReponseArgs :: Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        1    2   1.6      2      24
Processing:   211  686  80.3    680    1339
Waiting:      150  685  81.6    678    1338
Total:        215  689  80.3    682    1341

Reflection :: Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    3   5.6      0      31
Processing:   223  674  82.4    669    1154
Waiting:      194  673  83.8    663    1154
Total:        224  677  82.4    671    1154
```

PR #1421, opened 2015-08-10
Fix basePath handling

akrabat:
This PR fixes two issues:
- The basePath isn't cleared when calling `withPath('/bar')`
- __toString doesn't take `basePath` into account every time, which results in an incorrect URI being returned in a specific situation. (test included)

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -542,6 +542,11 @@ class Uri implements UriInterface
         $clone = clone $this;
         $clone->path = $this->filterPath($path);
 
+        // if the path is absolute, then clear basePath
+        if (substr($path, 0, 1) == '/') {
+            $clone->basePath = '';
+        }
+
         return $clone;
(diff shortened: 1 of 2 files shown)


Slim/Http/Uri.php
```diff
@@ -768,12 +773,7 @@ public function __toString()
         $query = $this->getQuery();
         $fragment = $this->getFragment();
 
-        if ($authority && substr($path, 0, 1) !== '/') {
-            $path = $basePath . '/' . $path;
-        }
-        if (!$authority && substr($path, 0, 2) === '//') {
-            $path = '/' . ltrim($path, '/');
-        }
+        $path = $basePath . '/'  . ltrim($path, '/');
```
campersau, line 776: There are two spaces after the first `'/'`.

akrabat, line 776: Fixed. Thanks.

PR #1423, opened 2015-08-10
Update the meta information when getMetadata() is called

### LINKED ISSUE #1399: PSR-7 Stream implementation returns stale data for mutable properties
When opening a `\Slim\Http\Stream` in Slim 3.x all of the stream meta data is cached when the stream resource is `attach()`ed. For potentially mutable stream metadata like `size`, this causes stale data to be returned from certain methods like `getSize()` once anything about the stream is modified.

akrabat:
This ensures that the metadata isn't stale if the stream has changed.
Also remove test for setting metadata in constructor as we no longer do this.

Fixes #1399

### DIFF
--- a/Slim/Http/Stream.php
+++ b/Slim/Http/Stream.php
@@ -100,6 +100,7 @@ class Stream implements StreamInterface
      */
     public function getMetadata($key = null)
     {
+        $this->meta = stream_get_meta_data($this->stream);
         if (is_null($key) === true) {
             return $this->meta;
         }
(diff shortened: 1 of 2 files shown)


campersau:
I think this does not fix #1399 completely because this only fixes `getMetadata` and not `getSize` which uses `fstat`.

akrabat:
Thanks @campersau - does that look better?

silentworks:
What happens If I call `getSize` before `getMetadata`? would this not result in stale size?

silentworks:
I should really read code more, ignore last comment.

PR #1425, opened 2015-08-10
Make PHP 5.5 the minimum supported version

akrabat:
Also update all other composer requirements to the "^" syntax.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
--- a/composer.json
+++ b/composer.json
@@ -28,11 +28,11 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "pimple/pimple": "3.*",
-        "psr/http-message": "1.0",
+        "php": ">=5.5.0",
+        "pimple/pimple": "^3.0",
+        "psr/http-message": "^1.0",
         "nikic/fast-route": "^0.6",
-        "container-interop/container-interop": "~1.1"
+        "container-interop/container-interop": "^1.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0"


JoeBengalen:
Why drop php 5.4?

akrabat:
Because of http://php.net/supported-versions.php - specifically: "Security Support Until": "14 Sep 2015"

nhymxu:
I don't agree. Yes, php.net unsupport this, but many many server/ hosting provide continue using PHP 5.4 for a long time. 

I hate Laravel 5.1 because this require high PHP version than shared hosting support

silentworks:
Why should we support an unsupported platform? If the PHP core team says they won't be supporting it after 14 Sep 2015, then why should anyone support it. 

The problem is not this change in the project nor is it Laravel 5.1, the problem is hosting companies being too lazy to make the switch.

lalop:
I'm with @nhymxu on that, if the current code base work on php5.4 I don't see any good reason to define as uncompatible.
If in the next development php 5.4 appear to cause any issue so we can make this move at this moment.

piotr-cz:
I'm for keeping min. requirement of php 5.4, at least until the codebase requires higher version.

As others pointed out, 5.4 is supported by the PHP core team till 14 Sep 2015, so ATM it is still actively supported and security support ends on 10 Jul 2016.

silentworks:
Security support end 14 Sep 2015, not 10 Jul 2016.

piotr-cz:
true

codeguy:
We're 5.5+ from here on out.

goldsky:
http://windows.php.net/

> PHP 5.4 series will be the last versions to support Windows XP and Windows 2003

Why we can't ignore Windows XP?
1. Most of big companies and industries (read: clients) don't have easy path to upgrade
2. Most of the users are still using IE8, especially in Asia (sucks but true)

So in our case, we, as developer, must try to set our system as low as the user/client can be, to experience their usage, even just to throw "Unsupported Browser" message.

If you drop the support <5.5 not because of its particular feature/function but rather its supporting reason, I'd like to ask you to consider it again.
Please.

silentworks:
I think this thread has discussed this a lot already and many PHP projects are moving towards this direction also, we cant be kept at ransom by hosts, operating system or any of the likes. We are sticking with 5.5 and won't be changing this anytime soon. If you require support below 5.5, there is still Slim 2 available which is only taking security updates at the moment.

piotr-cz:
It's still possible to install Slim3 on a local machine with PHP 5.5+ and deploy on server running PHP 5.4.

tuupola:
Or you can just run composer with `--ignore-platform-reqs` switch.

nandin-borjigin:
> It's still possible to install Slim3 on a local machine with PHP 5.5+ and deploy on server running PHP 5.4.
> It's incompatible now(Slim 3.4.2), it uses `::class` and it's a new feature in PHP 5.5

PR #1431, opened 2015-08-12
Update Stream.php

sunaryohadi:
Add functionality of Clear the stream/contents.

So I can for eg.: replace the content with minified in the middleware.


akrabat:
There are documented ways to do this with PSR-7 that are not that verbose, so I don't see the need for this method. There's also a risk of data loss if the stream is something unexpected, so I'd rather that the more explicit PSR-7 ways are used.

Issue #1432, opened 2015-08-12
Routes with optional parameters


ondrejkajinek:
When you create route with optional part containing a fixed part and a parameter (e.g., /items(/page/:page) ), urlFor function produces '/items/page/:page' when parameter is not given.

akrabat:
We're not updating Slim 2 any more.

PR #1433, opened 2015-08-12
Fix for #1432

ondrejkajinek:
urlFor changed to create correct urls when using fixed parts in optional route segment, e.g. /items(/page/:page) will produce /items when no parameter given, or /items/page/:10 when calling urlFor('items', [ 'page' => 10 ])


akrabat:
As per the [README](https://github.com/slimphp/Slim#how-to-contribute), we are only accepting security fixes for the 2.x branch.

PR #1435, opened 2015-08-12
Do not assume response body is seekable in App::respond. Fixes #1434

### LINKED ISSUE #1434: [Slim3] Do not assume response body stream is seekable
Not all streams are seekable, and we should not assume this here:

https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L352

This bit me when trying to send a Guzzle PumpStream which tosses an exception when you try to seek it. I'll send a PR shortly.

codeguy:
This checks to see if the Response StreamInterface body is in fact seekable before attempting to seek it.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -349,7 +349,9 @@ class App
             // Body
             if (!$this->isEmptyResponse($response)) {
                 $body = $response->getBody();
-                $body->rewind();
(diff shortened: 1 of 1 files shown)


geggleto:
//cc @akrabat @silentworks Looks good to me!

PR #1441, opened 2015-08-16
Uri in withRedirect

akrabat:
Casting the $url parameter to a string means that if a UriInterface is passed to withRedirect, it will work correctly.

This simply saves having to cast to a string when calling `withRedirect` in this scenario:

``` php
$uri = $request->getUri()->withQuery('')->withPath($this->router->pathFor('foo'));
return $response->withRedirect($uri, 307);
```


akrabat:
@codeguy Updated.

PR #1445, opened 2015-08-17
Always resolve callables through the CallableResolver

mnapoli:
Always resolve callables through the CallableResolver, **even if it's not a `foo:bar` string**.

---

In default Slim's behavior, controller like `foo:bar` go through the `CallableResolver` so that `foo` is resolved using the container.

However I would like to allow other kind of syntaxes to write controller as services, e.g. `['Foo', 'bar']` (this being **not** a static method, so `Foo` is created by the container), `'My\Invokable\Controller'`, etc (basically this https://github.com/PHP-DI/Invoker#resolving-callables-from-a-container). To achieve this, I just have to write a custom `CallableResolver`.

However `Route` will call `CallableResolver` only if the controller is a string. I've changed that so that `CallableResolver` is always called, allowing users to customize completely how controllers are located.

### DIFF
--- a/Slim/CallableResolverAwareTrait.php
+++ b/Slim/CallableResolverAwareTrait.php
@@ -34,18 +34,14 @@ trait CallableResolverAwareTrait
      */
     protected function resolveCallable($callable)
     {
-        if (is_string($callable) && !is_callable($callable)) {
-            if ($this->container instanceof ContainerInterface) {
-                $container = $this->container;
-            } else {
-                throw new RuntimeException('Cannot resolve callable string');
-            }
-            /** @var CallableResolver $resolver */
-            $resolver = clone($container->get('callableResolver')); // we need a new one each time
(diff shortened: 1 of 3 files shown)


akrabat:
Seems reasonable to me. However, wouldn't it be more efficient to return `$callable` immediately if it is  `is_callable()` and so bypass the resolver?

mnapoli:
Makes sense, will update the PR!

mnapoli:
Done.

mnapoli:
Good god that was fast thanks! :)

akrabat:
You picked a day when I'm on holiday and thinking Slim :)

silentworks:
Yeah we all agreed it was good. Jump on irc at times if you want to discuss anything.

mnapoli:
@silentworks I need to think of that more often ;) Will keep it in mind, I might be working on such topics in the next days.

PR #1447, opened 2015-08-17
E_DEPRECATED Non-static method

EgorGruzdev:
When error_reporting = E_ALL calls handler_order lead to error E_DEPRECATED Non-static method

### DIFF
--- a/Slim/Handlers/Strategies/RequestResponse.php
+++ b/Slim/Handlers/Strategies/RequestResponse.php
@@ -29,7 +29,7 @@ class RequestResponse implements InvocationStrategyInterface
      * @return mixed
      */
     public function __invoke(
-        callable $callable,
+        $callable,
         ServerRequestInterface $request,
         ResponseInterface $response,
         array $routeArguments
--- a/Slim/Handlers/Strategies/RequestResponseArgs.php
+++ b/Slim/Handlers/Strategies/RequestResponseArgs.php
@@ -30,7 +30,7 @@ class RequestResponseArgs implements InvocationStrategyInterface
      * @return mixed
      */
     public function __invoke(
-        callable $callable,
+        $callable,
         ServerRequestInterface $request,
         ResponseInterface $response,
         array $routeArguments
--- a/Slim/Interfaces/InvocationStrategyInterface.php
+++ b/Slim/Interfaces/InvocationStrategyInterface.php
@@ -19,7 +19,7 @@ interface InvocationStrategyInterface
     /**
      * Invoke a route callable.
      *
-     * @param string               $callable The callable to invoke using the strategy.
+     * @param string|array|callable  $callable The callable to invoke using the strategy.
      * @param ServerRequestInterface $request The request object.
      * @param ResponseInterface      $response The response object.
      * @param array                  $routeArguments The route's placholder arguments
--- a/Slim/Interfaces/InvocationStrategyInterface.php
+++ b/Slim/Interfaces/InvocationStrategyInterface.php
@@ -19,12 +19,12 @@ interface InvocationStrategyInterface
     /**
      * Invoke a route callable.
      *
-     * @param callable               $callable The callable to invoke using the strategy.
+     * @param string               $callable The callable to invoke using the strategy.
      * @param ServerRequestInterface $request The request object.
      * @param ResponseInterface      $response The response object.
      * @param array                  $routeArguments The route's placholder arguments
      *
      * @return ResponseInterface|string The response from the callable.
      */
-    public function __invoke(callable $callable, ServerRequestInterface $request, ResponseInterface $response, array $routeArguments);
+    public function __invoke($callable, ServerRequestInterface $request, ResponseInterface $response, array $routeArguments);
 }


JoeBengalen:
Did you run this locally? Build failed:
`PHP Fatal error:  Declaration of Slim\Handlers\Strategies\RequestResponse::__invoke() must be compatible with Slim\Interfaces\InvocationStrategyInterface::__invoke($callable, Psr\Http\Message\ServerRequestInterface $request, Psr\Http\Message\ResponseInterface $response, array $routeArguments) in /home/travis/build/slimphp/Slim/Slim/Handlers/Strategies/RequestResponse.php on line 19`

EgorGruzdev:
Okay, I wanted to show that the use of "callable" leads to an error in the future (although not critical).

I am ready to correct all of the code which uses this interface, if it's interesting.

In my project, I made this inetfeys the scope of the framework.

JoeBengalen:
Well I do not fully understand what you are trying to solve here.

When would the `callable` lead to a deprecated warning?
And what is `handler_order`? I cannot find it as PHP function.

EgorGruzdev:
When the class is being Route on line 312 or 316 is a call to a function of class building InvocationStrategyInterface, it provides php depricated on callable reporting that function should be static if it is a function of class.

Example:

```
namespace App\DependencyInjection;

use Interop\Container\ContainerInterface;

use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;


class FoundHandler implements InvocationStrategyInterface
{
    /** @var \DI\Container */
    protected $container;
    public function __construct(ContainerInterface $container){
    }
    /**
     * Invoke a route callable.
     *
     * @param callable $callable The callable to invoke using the strategy.
     * @param ServerRequestInterface $request The request object.
     * @param ResponseInterface $response The response object.
     * @param array $routeArguments The route's placholder arguments
     *
     * @return ResponseInterface|string The response from the callable.
     */
    public function __invoke(callable $callable, ServerRequestInterface $request, ResponseInterface $response, array $routeArguments)
    {
        if(is_string($callable) && strpos($callable, '::') !== false){
            $callable = preg_split('/::/', $callable, 2);
        }

        //TODO формирование Request и Response, т.к. возможны subRequest
        $args = array(
            'request' => $request,
            'response' => $response
        );

        $args = array_replace($args, $routeArguments);

        return $this->container->call($callable, $args);
    }
}
```

Error:
<b>Strict Standards</b>:  Non-static method App\Controller\Test::hotelList() should not be called statically in <b>C:\OpenServer\domains\hotelengine\app\DependencyInjection\FoundHandler.php</b> on line <b>34</b><br />

P.S. error_reporting = E_ALL

JoeBengalen:
Ahh, I think I found the problem now.

Since you are using the syntax `class::method` PHP will return true for `is_callable()` whether the method can actually be called staticly or not. This is why `CallableResolverAwareTrait::resolveCallable()` returns the string instead of a `\Closure`.

Further on in the `InvocationStrategyInterface::invoke()` a `callable` is expected. Now PHP does not reconize the string as callable because the method cannot be called staticly.

So the root if the problem lies in that PHP thinks it can call a string formatted as `class::method` in is_callable(), but not when you typehint for `callable`.

EgorGruzdev:
I guess I do not understand something, but my calls have the form:

```
$app->group('/api/b2b/v1', function () use ($app) {
    $app->group('/hotel', function () use ($app) {
        $app->map(['GET', 'POST'], '/list[/{id:\d+}]', [\App\Controller\Test::class, 'hotelList']);
        $app->map(['GET', 'POST'], '/rates', [\App\Controller\Test::class, 'hotelRates']);
    });
});
```

JoeBengalen:
Ah well same thing. Although I do not understand the following code in you example as you are not passing a string:

``` php
if(is_string($callable) && strpos($callable, '::') !== false){
    $callable = preg_split('/::/', $callable, 2);
}
```

But back to the case:

PHP returns `true` for `is_callable([\App\Controller\Test::class, 'hotelList'])`, because it only checks if there is a method named `hotelList` in the class `App\Controller\Test`.

Bu when passed as argument to a method which expectecs a `callable` (typehint) PHP sees that `[\App\Controller\Test::class, 'hotelList']` is not actually callable, because the method is not static and cannot be called.

Do you see the difference?

So `CallableResolverAwareTrait::resolveCallable()` returns the array as it thinks it is callable. But when the same array is passed to the `FoundHandler` is finds the array is not callable at all, which is where the error comes from.

silentworks:
If you want to use PHP class name resolution with the default InvocationStrategies provided in Slim you will have to do it like:

``` php
$app->group('/api/b2b/v1', function () use ($app) {
    $app->group('/hotel', function () use ($app) {
        $app->map(['GET', 'POST'], '/list[/{id:\d+}]', App\Controller\Test::class . ':hotelList');
        $app->map(['GET', 'POST'], '/rates', App\Controller\Test::class . ':hotelRates');
    });
});
```

If you decide to implement your own Strategy I am afraid we can't help you much with that, but the signature of these Classes are unlikely to change at the moment.

opengeek:
I think this is just an oversight in the implementation of the default strategy. I have no problems with those type hints being removed. However, it should not be just `string` as the `@param` phpdoc. Maybe `string|array|callable`?

EgorGruzdev:
Thank you very much, I like the idea to make $callable string|array|callable.

JoeBengalen:
No, I think `InvocationStrategyInterface` MUST receive a callable. If you want a different notation you should solve that with a custom `CallableResolverInterface` implementation.

EgorGruzdev:
Thank you very much, so I did.

```
namespace App\DependencyInjection;
use RuntimeException;
use Interop\Container\ContainerInterface;
use Slim\Interfaces\CallableResolverInterface;

final class CallableResolver implements CallableResolverInterface
{

    /**
     * @var \DI\Container
     */
    private $container;

    /**
     * @var string
     */
    private $toResolve;

    /**
     * @var callable
     */
    private $resolved;

    /**
     * @param ContainerInterface $container
     * @param string             $toResolve
     */
    public function __construct(ContainerInterface $container, $toResolve = null)
    {
        $this->toResolve = $toResolve;
    }


    /**
     * Receive a string that is to be resolved to a callable
     *
     * @param  string $toResolve
     */
    public function setToResolve($toResolve)
    {
        $this->toResolve = $toResolve;
    }

    /**
     * Resolve toResolve into a closure that that the router can dispatch.
     *
     * If toResolve is of the format 'class:method', then try to extract 'class'
     * from the container otherwise instantiate it and then dispatch 'method'.
     *
     * @return \Closure
     *
     * @throws RuntimeException if the callable does not exist
     * @throws RuntimeException if the callable is not resolvable
     */
    private function resolve()
    {
        // if it's callable, then it's already resolved
        if (is_callable($this->toResolve)) {
            $this->resolved = $this->toResolve;

            // check for slim callable as "class:method"
        } elseif (is_string($this->toResolve)) {
            $callable_pattern = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
            if (preg_match($callable_pattern, $this->toResolve, $matches)) {
                $class = $matches[1];
                $method = $matches[2];

                if ($this->container->has($class)) {
                    $this->resolved = [$this->container->get($class), $method];
                } else {
                    if (!class_exists($class)) {
                        throw new RuntimeException(sprintf('Callable %s does not exist', $class));
                    }
                    $this->resolved = [$this->container->get($class), $method];
                }
                if (!is_callable($this->resolved)) {
                    throw new RuntimeException(sprintf('%s is not resolvable', $this->toResolve));
                }
            } else {
                throw new RuntimeException(sprintf('%s is not resolvable', $this->toResolve));
            }
        }
    }

    /**
     * Invoke the resolved callable.
     *
     * @return \Psr\Http\Message\ResponseInterface
     */
    public function __invoke()
    {
        if (!isset($this->resolved)) {
            $this->resolve();
        }

        $args =  func_get_args();

        return  $this->container->call($this->resolved, ['request' => $args[0], 'response' => $args[1]] + $args[2]);
    }
}
```

akrabat:
As per https://github.com/slimphp/Slim/pull/1447#issuecomment-132149192, this should be solved via a CallableResolver.

PR #1448, opened 2015-08-17
Refactored CallableResolver into a stateless service focused on resolving callables only

mnapoli:
Per discussions on IRC, I refactored CallableResolver into a stateless service focused on resolving callables only (not wrapping and invoking them).

This will lead to less object instantiations (since the service is now stateless instead of wrapping every middleware/controller callable), as well as better separation of concerns since the class' responsibility is now only to resolve callables. It isn't responsible of invoking callables anymore (responsibility which was spread between here and also the invokation strategy and middleware stack too IIRC).

One detail is that I have moved the `$callable->bindTo($container)` calls into the CallableResolver itself since it belongs there IMO (because it's about preparing callables for being invokable correctly). That avoids having "callable resolution" code all over the place (App and Route).

Feedback welcome. I'll probably be able to answer and update the PR in 2 days.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -14,8 +14,7 @@ use Slim\Interfaces\CallableResolverInterface;
 
 /**
  * This class resolves a string of the format 'class:method' into a closure
- * that can be dispatched. It is itself invokable as it lazily resolves the string
- * when it is invoked.
+ * that can be dispatched.
  */
 final class CallableResolver implements CallableResolverInterface
 {
@@ -25,34 +24,11 @@ final class CallableResolver implements CallableResolverInterface
     private $container;
 
     /**
-     * @var string
-     */
-    private $toResolve;
-
-    /**
-     * @var callable
-     */
-    private $resolved;
-
-    /**
      * @param ContainerInterface $container
-     * @param string             $toResolve
      */
-    public function __construct(ContainerInterface $container, $toResolve = null)
+    public function __construct(ContainerInterface $container)
     {
-        $this->toResolve = $toResolve;
         $this->container = $container;
-    }
-
-
-    /**
-     * Receive a string that is to be resolved to a callable
-     *
-     * @param  string $toResolve
-     */
-    public function setToResolve($toResolve)
-    {
-        $this->toResolve = $toResolve;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


tests/RouteTest.php
```diff
@@ -153,6 +154,8 @@ public function testAddMiddlewareAsString()
         $route = $this->routeFactory();
 
         $container = new Container();
+        $container['MiddlewareStub'] = new MiddlewareStub();
+
```
mnapoli, line 158: Could someone explain to me how it was working before? I'm afraid I introduced a regression because now I have to register the middleware in the container (and it's not necessary on 3.x).

mnapoli:
Also the next step IMO is to remove the `CallableResolverAwareTrait` (again, per discussions on IRC), but I wanted to do it in a separate pull request.

tests/RouteTest.php
JoeBengalen, line 158: It worked before because the `CallableResolver` was never really invoked, so it did not try to find the `MiddlewareStub`.

Slim/CallableResolver.php
```diff
@@ -25,87 +24,58 @@
...
-     *
-     * @param  string $toResolve
-     */
-    public function setToResolve($toResolve)
-    {
-        $this->toResolve = $toResolve;
-    }
-
     /**
      * Resolve toResolve into a closure that that the router can dispatch.
```
JoeBengalen, line 35: Docblock is outdated

JoeBengalen:
Maybe a final check should happen in `CallableResolver::resolve()` if `$resolved` is indeed callable. 

Currently, anything you pass that is not a string or callable will be returned without modification. And since the `resolve()` claims to return a callable at all times I think an Exception should be thrown if the resulting `$resolved` is not callable.

Also I agree on getting rid of the `CallableResolverAwareTrait`, since its not doing much anymore. Just performing a check (that is not even needed anymore?) and calling `CallableResolver::resolve()` on the passed argument.

One problem with your current implementation though! Closures will never be bound to the container since it wont get to the `CallableResolver`. `CallableResolverAwareTrait::resolveCallable()` will just return the Closure.

Apart from that I think this is an improvement :) Makes much more sense for the `CallableResolver` to be stateless (as you called it).

tests/RouteTest.php
mnapoli, line 158: OK so that's an improvement then ;)

Slim/CallableResolver.php
mnapoli, line 35: The new signature of the method is `public function resolve($toResolve)` so it makes sense right?

mnapoli:
Thanks for the review!

> Maybe a final check should happen in CallableResolver::resolve() if $resolved is indeed callable.

Done in ba08a05, an exception will be thrown if the CallableResolver cannot return a real callable.

> Closures will never be bound to the container 

Oh yeah good catch, I fixed it in e87d996. I also added a test for that to make sure it's staying.

JoeBengalen:
@ppetermann, good question ... It didn't change in this PR but was already bound to container instead of app. No sure where and why it was changed.

All given examples in de docs assume closures are indeed bound to app.

Edit; found the confusion: Middleware is bound to the container, route callable bound to app.
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L237

So this PR should keep that same behaviour.

Currently this PR will break this behaviour.
First in `App` the route callable will be bound to app: https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L237

But in the `CallableResolver` it will be bound to the container (and overwriting the previous bound app)
https://github.com/mnapoli/Slim/blob/simpler_callable_handling/Slim/CallableResolver.php#L74

So in the end route callable will be found to container instead of app, which if wrong.

Edit2:
As @silentworks noted: `RouteGroup` binds the closure to app too:
https://github.com/slimphp/Slim/blob/3.x/Slim/RouteGroup.php#L61

juliangut:
`CallableResolver` is used to resolve both middlewares and routes, so either bindTo is moved out from CallableResolver or it needs a parameter to be told what to bind to?

mnapoli:
Thank you for noticing that. I've restored the way it was before regarding binding Closures to app or container.

akrabat:
Looks good to me.

akrabat:
Thanks @mnapoli

Issue #1449, opened 2015-08-18
Case insenitive routes v3


neilmillard:
Issue 697 for slim v2 gave an option for case insensitive routes.
could we have this for v3?
this could be via the settings array with 
$request->getUri()->withPath(strtolower($request->getUri()->getPath()))
in the request class, or some code added to the Uri class.
what would the preferred way to code it?

akrabat:
You can quite easily create your own middleware that does this.

Something like this (untested):

``` php
$app->add(function($request, $response, $next) {
    $uri = $request->getUri();
    $uri = $uri->withPath(strtolower($uri->getPath());
    $request = $request->withUri($uri);

    return $next($request, $response);
});
```

neilmillard:
I would expect this to work. I'm missing something obvious maybe?

``` php
$app->add( function (\Slim\Http\Request $request, \Slim\Http\Response $response, callable $next)use($container){
    $basePath = $request->getUri()->getBasePath();
    $uri = $request->getUri()->withBasePath(strtolower($basePath));
    $newreq = $request->withUri($uri);
    $response = $next($newreq,$response);

    return $response;
});
```

JoeBengalen:
Is it on purpose you update the `basePath` instead of the `path`?

neilmillard:
yes, the issue I have is Slim running from a subdirectory and for some reason (IIS setup) it is returning 404 if the case is used.

neilmillard:
I know the basePath code has been changed slightly since I ```composer update```

JoeBengalen:
So what url are you trying to change?

Edit; For me this works fine

``` php
// requesting: http://localhost/dev/Slim/hello/me
// where my index is in dev\Slim\

$app->add(function (\Slim\Http\Request $request, \Slim\Http\Response $response, callable $next) {
    var_dump($request->getUri()->getBasePath());
    // -> /dev/Slim

    var_dump($request->getUri()->getPath());
    // -> hello/me

    var_dump((string) $request->getUri());
    // -> http://localhost/dev/Slim/hello/me

    $basePath = $request->getUri()->getBasePath();
    $uri = $request->getUri()->withBasePath(strtolower($basePath));
    $newreq = $request->withUri($uri);
    $response = $next($newreq,$response);

    var_dump((string) $newreq->getUri());
    // -> http://localhost/dev/slim/hello/me

    return $response;
});
```

[...]

PR #1450, opened 2015-08-18
compare basepath case insensitive. Solves #1449

JoeBengalen:
Solves #1449

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -195,7 +195,7 @@ class Uri implements UriInterface
         $virtualPath = $requestUri;
         if (strpos($requestUri, $requestScriptName) === 0) {
             $basePath = $requestScriptName;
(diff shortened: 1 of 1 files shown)


JoeBengalen:
Wait a second with this ...

Slim/Http/Uri.php
```diff
@@ -195,7 +195,7 @@ public static function createFromEnvironment(Environment $env)
         $virtualPath = $requestUri;
         if (strpos($requestUri, $requestScriptName) === 0) {
```
akrabat: Should this one be `stripos` too?

JoeBengalen: Seems logic to do so, but doesn't change anything for me. For some weird reason when I access `http://localhost/dev/slim/index.php/hello/me` script name is `/dev/slim/index.php` , but when I go to  
`http://localhost/dev/Slim/index.php/hello/me` script name is `/dev/Slim/index.php`.

But just to be sure I'll update it

PR #1451, opened 2015-08-18
Remove Request::isAjax()

JoeBengalen:
Remove the `Request::isAjax()` method as it was identical to `Request::isXhr()`. 2 methods for the same check can only confuse people, so I would like to see one of them removed.

Since the check is if header `X-Requested-With` is `XMLHttpRequest` (xhr) it seems best to remove the `isAjax()` method.


codeguy:
Agreed. Will merge when checks complete.

Issue #1430, opened 2015-08-12
[Slim3] Allow callable


harikt:
Hi, 

I have noticed the resolve https://github.com/slimphp/Slim/blob/3.x/Slim/CallableResolver.php#L77 is not checking if it is callable.

So even if you create a route like 

``` php
class LoginAction
{
    public function __invoke(...)
    {
    }
}

$app->get('/login', 'App\Controllers\Login\LoginAction')
    ->setName('login');
```

is not calling and throwing exception could not resolve. It can be resolved getting from the container and checking is callable ?

luckyzid:
I like now route pattern. but now you will also be following pattern

```php
private function resolve()
{
    // if it's callable, then it's already resolved
    if (is_callable($this->toResolve)) {
        $this->resolved = $this->toResolve;

    // check for slim callable as "class:method"
    } elseif (is_string($this->toResolve)) {
        $callable_pattern = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
        if (preg_match($callable_pattern, $this->toResolve, $matches)) {
            $class = $matches[1];
            $method = $matches[2];

            if ($this->container->has($class)) {
                $this->resolved = [$this->container->get($class), $method];
            } else {
                if (!class_exists($class)) {
                    throw new RuntimeException(sprintf('Callable %s does not exist', $class));
                }
                $this->resolved = [new $class, $method];
            }
            if (!is_callable($this->resolved)) {
                throw new RuntimeException(sprintf('%s is not resolvable', $this->toResolve));
            }
        } else {
            throw new RuntimeException(sprintf('%s is not resolvable', $this->toResolve));
        }
    }
}
```

```php
private function resolve()
{
    // if it's callable, then it's already resolved
    if (is_callable($this->toResolve)) {
        $this->resolved = $this->toResolve;

    // check for slim callable as "class:method"
    } elseif (is_string($this->toResolve)) {
        $callable_pattern = '!^([^\:]+)(\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$)?!';
        if (preg_match($callable_pattern, $this->toResolve, $matches)) {
            $class = $matches[1];
            $method = isset($matches[2]) ? $matches[2] : '__invoke';

            if ($this->container->has($class)) {
                $this->resolved = [$this->container->get($class), $method];
            } else {
                if (!class_exists($class)) {
[...]

PR #1455, opened 2015-08-20
Resolve callable classes

akrabat:
Update CallableResolver to check if the string to resolve is in the container or a class that can be invoked.

This addresses #1430

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -49,20 +49,33 @@ final class CallableResolver implements CallableResolverInterface
         if (!is_callable($toResolve) && is_string($toResolve)) {
             // check for slim callable as "class:method"
             $callablePattern = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
-            if (!preg_match($callablePattern, $toResolve, $matches)) {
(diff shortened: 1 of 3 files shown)


silentworks:
You finish with this PR @akrabat

akrabat:
Yep

Slim/CallableResolver.php
```diff
@@ -49,20 +49,33 @@ public function resolve($toResolve)
...
             } else {
-                if (!class_exists($class)) {
-                    throw new RuntimeException(sprintf('Callable %s does not exist', $class));
+                // check if string is something in the DIC that's callable or is a class name which
+                // has an __invoke() method
+                $class = $toResolve;
+                if ($this->container->has($class)) {
+                    $resolved = $this->container->get($class);
+                } else {
+                    if (!class_exists($class)) {
```
stevenwadejr, line 71: I know this is closed and merged, but what about also checking to make sure the class actually does have an `__invoke` method?

```
if (!class_exists($class) || !(new \ReflectionClass($class))->hasMethod('__invoke')) {
    throw new RuntimeException(sprintf('Callable %s does not exist', $class));
}
```

silentworks, line 71: At the moment if the class doesn't have an invoke method you will get `Class is not resolvable`, so you would hit line 78. I am happy with it being just that rather than firing up a Reflector class just to do the check first.

PR #1457, opened 2015-08-20
Pass settings to Pimple

### LINKED ISSUE #1407: Slim\Container isn't passing settings array to Pimple
We're unable to leverage full Pimple due to a key pass-thru missing in Slim\Container's constructor:

``` php
    /**
     * Create new container
     *
     * @param array $userSettings Associative array of application settings
     */
    public function __construct(array $userSettings = [])
    {
        parent::__construct();
        [...]
    } 
```

Slim\Container does not pass $userSettings to the parent (Pimple) class, resulting in a failed Pimple initialization. This doesn't work:

``` php
// config.php
return [
    // Slim configuration.
    'env' => 'production',
    // DI classes
    'logger' => PhpConsole\Handler::getInstance(),
];

// index.php
$app = new \Slim\App('config.php');
$app->logger->debug('test'); // Call fails, but is successful if I add $userSettings to parent::__construct().
```

Or should this be handled some other way within Slim? The point here is to use DI to pass in environment dependent classes, not hard code them into my index.php file.

akrabat:
This allows a user to configure the container via settings.

Fixes #1407

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -61,12 +61,13 @@ final class Container extends PimpleContainer implements ContainerInterface
     /**
      * Create new container
      *
-     * @param array $userSettings Associative array of application settings
+     * @param array $values The parameters or objects.
      */
-    public function __construct(array $userSettings = [])
+    public function __construct(array $values = [])
     {
-        parent::__construct();
+        parent::__construct($values);
 
+        $userSettings = isset($values['settings']) ? $values['settings'] : [];
         $this->registerDefaultServices($userSettings);
     }
 
@@ -104,9 +105,11 @@ final class Container extends PimpleContainer implements ContainerInterface
          *
          * @return EnvironmentInterface
          */
-        $this['environment'] = function ($c) {
-            return new Environment($_SERVER);
-        };
+        if (!isset($this['environment'])) {
+            $this['environment'] = function ($c) {
+                return new Environment($_SERVER);
+            };
+        }
 
         /**
          * PSR-7 Request object
@@ -115,9 +118,11 @@ final class Container extends PimpleContainer implements ContainerInterface
          *
          * @return ServerRequestInterface
          */
-        $this['request'] = function ($c) {
-            return Request::createFromEnvironment($c['environment']);
-        };
+        if (!isset($this['request'])) {
+            $this['request'] = function ($c) {
+                return Request::createFromEnvironment($c['environment']);
+            };
+        }
 
         /**
          * PSR-7 Response object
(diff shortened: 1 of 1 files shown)


geggleto:
You are a rockstar :) :+1:

codeguy:
:+1:

Slim/Container.php
```diff
@@ -61,12 +61,13 @@
...
      * Create new container
      *
-     * @param array $userSettings Associative array of application settings
+     * @param array $values The parameters or objects.
      */
-    public function __construct(array $userSettings = [])
+    public function __construct(array $values = [])
     {
-        parent::__construct();
+        parent::__construct($values);
```
silentworks, line 68: Is it me or would we be passing settings to Pimple then replacing it right after on line 71?

geggleto, line 68: Yeah it does appear that it will get passed in twice. but only the ['settings'] part of the user provided array.

akrabat, line 68: It's laziness on my part to avoid changing the default settings merging. Will see if I can tidy it up.

silentworks, line 68: Nah leave it, its probably quicker to replace it twice than to pop it off the array. Will merge, you can cleanup later if you wish.

yurevichcv:
Unfortunately, this PR doesn't have backward compatibility. Before I was able to "override" container values before `$app->run()`. 

Before I was able to do the following in unit tests:

```
$app = new \Slim\App();

Configurator::configureApp($app); //seudo-code

$c = $app->getContainer();
$c['environment'] = $c->factory(function ($c) {
....
});
$c['request'] = $c->factory(function ($c) {
....
});
```

Now it is not possible, getting following error:

```
RuntimeException : Cannot override frozen service "environment".
```

And main problem is that now you freeze almost all container variables during router initialisation:

Before it was (worked perfectly):

```
 $this['router'] = function ($c) {
     return new Router();
};
```

Not it is:

```
if (!isset($this['router'])) {
            $this['router'] = function ($c) {
                $router = new Router();

                $uri = $c['request']->getUri();

                if (is_callable([$uri, 'getBasePath'])) {
                    $router->setBasePath($uri->getBasePath());
                }

                return $router;
            };
}
```

As you can see, now you access $c['request'], which automatically freezes almost all essential components in container.

yurevichcv:
Main idea is that essential container components, such as `environment`, `request` and `response` should not be frozen before app run.

Use case:
1. I want to configure my application (including all routes, acl, domain services, etc)
2. Save it to static property or to cache.
3. Override container's `environment`, `request` and `response` with factories.
4. Run many tests against the same instance of application.

akrabat:
@yurevichcv Please open a new issue as it may be related to #1443 rather than this PR.

silentworks:
This has been the case before this change was made, you probably just didn't notice it before. Services are frozen from the minute they get called, this is how Pimple works.

yurevichcv:
Yes, sorry, you are right. This PR caused the problem - #1443

yurevichcv:
@silentworks Yes, I know how Pimple works. But `environment`, `request` and `response` should not be frozen before app run. This limits flexibility in testing. `request` is the most essential entry point, and it should be called only once, further it should just be "travelling" between middleware and routes.

The same issue was in Twig-View extension, where I also sumbmitted my PR against `request` freezing during initialization phase - https://github.com/slimphp/Twig-View/pull/30

geggleto:
@yurevichcv you are free to overwrite the defaults at any time during execution by simply overwriting the default container instance like this.

```
unset($container['environment']);
$container['environment'] = new Environment();
```

Issue #1465, opened 2015-08-26
Most of container variables appear to be "frozen" even before app run. 


yurevichcv:
This issue is caused by PR #1443

Unfortunately, this PR doesn't have backward compatibility. Before I was able to "override" container values before `$app->run()`. 

Before I was able to do the following in unit tests:

```
$app = new \Slim\App();

Configurator::configureApp($app); //seudo-code

$c = $app->getContainer();
$c['environment'] = $c->factory(function ($c) {
....
});
$c['request'] = $c->factory(function ($c) {
....
});
```

Now it is not possible, getting following error:

```
RuntimeException : Cannot override frozen service "environment".
```

And main problem is that now you freeze almost all container variables during router initialisation:

Before it was (worked perfectly):

```
 $this['router'] = function ($c) {
     return new Router();
};
```

Not it is:

```
if (!isset($this['router'])) {
            $this['router'] = function ($c) {
                $router = new Router();

                $uri = $c['request']->getUri();

                if (is_callable([$uri, 'getBasePath'])) {
                    $router->setBasePath($uri->getBasePath());
                }

                return $router;
            };
}
```

As you can see, now you access $c['request'], which automatically freezes almost all essential components in container.

Main idea is that essential container components, such as `environment`, `request` and `response` should not be frozen before app run.

Use case:
1. I want to configure my application (including all routes, acl, domain services, etc)
2. Save it to static property or to cache.
3. Override container's `environment`, `request` and `response` with factories.
4. Run many tests against the same instance of application.

yurevichcv:
The same issue was with Twig-View extension, which was "freezing" `request` object even before `$app->run()`: https://github.com/slimphp/Twig-View/pull/30

JoeBengalen:
When does it freeze? If wont freeze as long as you do not get anything from the container. Following code works fine:

``` php
$app = new Slim\App();

$c = $app->getContainer();

$c['environment'] = function ($c) {
    return new Slim\Http\Environment($_SERVER);
};

$c['router'] = function () {
    return new \Slim\Router();
};

$app->get('/', function () {
    echo 'home';
});
```

Indeed if you use the `router` you cannot override any used service anymore. Following code gives your error:

``` php
$app = new Slim\App();

$c = $app->getContainer();

$app->get('/', function () {
    echo 'home';
});

$c['environment'] = function ($c) {
    return new Slim\Http\Environment($_SERVER);
};

$c['router'] = function () {
    return new \Slim\Router();
};
```

[...]

PR #1466, opened 2015-08-26
Set router's basePath on dispatch

akrabat:
Move setting of the router's basePath to dispatch() so that adding routes doesn't freeze the environment and request container items.

To allow setting of the base path prior to dispatch, we only set it if it's currently null.

This addresses #1465


yurevichcv:
:+1:

yurevichcv:
@akrabat nice solution

akrabat:
This solution doesn't allow use of `pathFor` in middleware. Closing in favour of #1467.

PR #1467, opened 2015-08-26
Set router's basePath when App::run() is called

akrabat:
Move setting of the router's basePath to App::run() so that adding routes doesn't freeze the environment and request container items.

I've moved the setting to `run()` so that middleware is able to call the router's `pathFor`.

This addresses #1465


yurevichcv:
Hi, tested your PR in my app. All tests are passing. :+1:

silentworks:
Happy to hear that @yurevichcv, will get this merged in soon.

PR #1472, opened 2015-08-31
Added a check to ensure that the 'determineRouteBeforeAppMiddleware' …

chantron:
…settings key was set before seeing if it's set to true. This will help when using a 3rd party DI container that doesn't know to set it by default.

(I was playing around with https://github.com/akrabat/slim3-abstract-action-factory and PHP kept throwing "Notice: Undefined index: determineRouteBeforeAppMiddleware".)

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -298,7 +298,9 @@ class App
 
 
         // Dispatch the Router first if the setting for this is on
-        if ($this->container->get('settings')['determineRouteBeforeAppMiddleware'] === true) {
+        if (isset($this->container->get('settings')['determineRouteBeforeAppMiddleware']) &&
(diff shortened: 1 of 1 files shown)


chantron:
I'm thinking I jumped the gun on this. If so, my bad!

JoeBengalen:
In the docs you can read which settings must be defined:
http://www.slimframework.com/docs/concepts/di.html

chantron:
Gatdang! I read that over and over again and didn't see it under settings until just now. I'm closing this!

akrabat:
I've now tagged 1.0.2 of [akrabat/rka-slim-zfsm-container](https://github.com/akrabat/rka-slim-zfsm-container/releases/tag/1.0.2) which is the actual solution to this problem.

Issue #1475, opened 2015-09-01
Remove hard dependency on App in RouteGroup invocation?


designermonkey:
Can we remove the hard dependency on the `App` class in the `RouteGroup::__invoke` method please?

I am trying to build a `SlimeRouteMapper` class which allows me to use a nested array to define the routes and callbacks, but this at no point interacts with the `App` class, and instead mimics the `App::group` behaviour for nested routes.

Most of the other facets of Slim are only dependant on key requirements to their internal functionality, so it would make sense to me to allow any class to be passed to the method, so that it can be bound to anything.

Either that, or allow it's footprint to be `App $app = null` so I can get away with ignoring it (as I want to really). I know my callback isn't a closure, so doesn't need binding to anything, as it's an existing class and method.

Any thoughts @silentworks @akrabat @gmanricks @codeguy ?

designermonkey:
Sorry, shouldn't add a milestone until it's decided on.

designermonkey:
Looking at it a little more, the invoke method always assumes that the returned callable is a closure irrespective. I can't pass `[[$this, 'groupCallback'], [$routable]]` for example, to my group as a callable, which the resolver resolves fine, as it isn't called with `call_user_func_array` which it really should be.

geggleto:
I assume the reason why we bind the App is to get the closure... you could only then remove the App dependency and replace it with a Container-Interop dependency... I am unsure about the side-effects if you were to ignore at minimum passing the container which slim uses.

akrabat:
I see no reason why the signature can't be `public function __invoke(App $app = null)`

When you send in the PR, don't forget to update the `if` to check for $app being not null before we try to bind to it.

designermonkey:
Certainly can.

designermonkey:
Thanks.

PR #1484, opened 2015-09-04
Fix 1475

geggleto:
#1475


akrabat:
Why are there 10 commits for 4 line change?

PR #1485, opened 2015-09-04
Checking for / in route group route url and removing it

### LINKED ISSUE #1324: Empty string as root route for group
Slim3 route now no longer support just slash as root route:

``` php
$app->group('/foo/bar', function () use ($app) {
    $app->get('/', 'FooBar:action');
    $app->get('/bar', 'FooBar:actionbar');
});
```

In order to get it work you have to keep string empty which is wrong while all other routes begins with slash

``` php
$app->group('/foo/bar', function () use ($app) {
    $app->get('', 'FooBar:action');
    $app->get('/bar', 'FooBar:actionbar');
});
```

silentworks:
Fixes #1324

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -118,8 +118,9 @@ class Router extends RouteCollector implements RouterInterface
 
         // Prepend parent group pattern(s)
         if ($this->routeGroups) {
-            if (strlen($pattern) === 1 && strpos($pattern, '/') === 0) {
-                $pattern = ltrim($pattern, '/');
+            // If any route in the group only has / we remove it
+            if ($pattern === '/') {
+                $pattern = '';
             }
             $pattern = $this->processGroups() . $pattern;
         }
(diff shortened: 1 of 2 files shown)


Slim/Router.php
```diff
@@ -118,6 +118,9 @@ public function map($methods, $pattern, $handler)
 
         // Prepend parent group pattern(s)
         if ($this->routeGroups) {
+            if (strlen($pattern) === 1 && strpos($pattern, '/') === 0) {
```
akrabat: Why not `if ($pattern === '/') {` ?

akrabat: Add a comment on what you're testing for here please :)

Slim/Router.php
```diff
@@ -118,6 +118,9 @@ public function map($methods, $pattern, $handler)
...
+                $pattern = ltrim($pattern, '/');
+            }
```
akrabat, line 124: Can't this be: 

``` php
if ($pattern === '/') {
    $pattern = '';
}
```

?

PR #1489, opened 2015-09-05
Return arrays where possible for getParsedBody()

akrabat:
This fixes the weird case that if you POST, then getParsedBody() will have an array in it ($_POST), but if you PUT or have json, then getParsedBody() will have an object.

Even more weirdly, if you POST with the hidden _METHOD=PUT, you get an array in your put() handler, but if you PUT to it then your put() handler gets an object!

It's much more consistent to try to always make it an array and ensures that we can also handle keys with hyphens much more sanely as they are horrible to deal with when in a stdClass.


silentworks:
Yep I agree.

PR #1490, opened 2015-09-06
Update error handlers to be content-type aware

akrabat:
Ensure that Error, NotAllowed and NotFound return JSON, XML or HTML as determined by the request's Accept header. This makes Slim much more API friendly as the correct type of error will be returned without having to write three special handlers just to return JSON (or XML).


akrabat:
I added two new test files & coverage decreased?!

PR #1496, opened 2015-09-10
Added missing type for parameter

duckoder:
Added type for $value parameter in singleton method. 
It's not really important, but I think it will be better with this.


akrabat:
We're only accepting security fixes for Slim 2 now.

PR #1497, opened 2015-09-12
[Proof of Concept] ServiceConfig for all required services

silentworks:
I had a thinker over this and thought its actually bad that we have to re-implement every required service of the framework any time we decide to use a new `ContainerInterface`. It feels like it defeats the purpose of having the container swappable and also I don't like the idea of the core framework being reliant on the Container.

I know this change introduces another Container per say, but I think separating internal dependencies from external dependencies is better than having everything in the same Container. It is easy enough to get the internal dependencies if necessary from the `ServiceConfig` which I am introducing.

I have also removed the `bindTo` from the Route\RouteGroup classes, this way we don't need to be passing the `Container` around purely for this feature, it felt a bit heavy requiring the `Container` in each constructor just to have access to it for `bindTo`.

I still need to remove the `settings` from the container to make it completely unaware of the App's requirements.


ppetermann:
While I agree that it is kind of annoying to setup the defaults if you add your own container (as i did), and a benefit of your solution is that you don't have to update the defaults in your projects if those change, i feel that this is a bad idea.

The job of the container is to solve all dependencies, having two containers makes it confusing to the user, as he will expect to replace even the default dependencies through the container, and as it means for a user that he will have to look into two different places to find where a dependency is coming from.

In the end what this is causing is a duplication of logic.

In my opinion this gets even worse with the idea that a Class is used as configuration for those services, hence a change of _configuration_ ends up being a change of _logic_.

I think a much better solution would be to provide an aid for people to setup their containers with the default values (for example a class that returns an array with key / default value combinations).

https://gist.github.com/ppetermann/7d8f6690a74b6e8e9e0b

akrabat:
My initial thought is that I dislike having two containers in play as I have that in a Slim 2 application and it's a right old pain as I keep forgetting which one has which thing in it.

jaapverloop:
> I had a thinker over this and thought its actually bad that we have to re-implement every required service of the framework any time we decide to use a new ContainerInterface.

I agree with you! But I agree with @ppetermann too!
The idea of 2 containers feels a bit strange.

The config service could extend the container interface and consume a real implementation of the container interface at the same time (proxy).

``` php
interface ServiceConfigInterface extends ContainerInterface
{}

class ServiceConfig implements ServiceConfigInterface
{
    public function __construct(ContainerInterface $container = null, array $settings = [])
    {
        if (null === $container) {
            // We still need a small wrapper around Pimple, because is does not
            // implement the container interface.
            $container = new Container();
        }

        $this->settings = $settings;
    }

    public function get($id)
    {
        return $this->container->get($id);
    }

    public function has($id)
    {
        return $this->container->has($id);
    }
}
```

ppetermann:
uhm, doesn't that then completely miss the point?

jaapverloop:
Yeah, you're absolutely right!
The app class of this PR depends on both, I tried to remove the dependency on the container interface.

codeguy:
Whenever I see this argument come back up, I always think about using a composite container. In the Slim 3 App constructor, we accept an external ContainerInterop container. We then create a CompositeContainer instance in the constructor, pass in an instance of Slim/Container that provides default dependencies. We then pass in the user's external container into the composite container. When we invoke `has` or `get` on the composite container, it defers to its children containers (first to the external one, then to its default one). Seems like it addresses all issues, right?

akrabat:
Generally, I don't like having two containers in play even if one is subservient. However, if we do have to address this, then a composite container feels more workable to me.

codeguy:
I generally don't like having multiple containers. But I don't think there's anything _wrong_ with it, either. If we were to do so, a composite container is what I'd recommend because it provides a simple, expected interface to its child containers.

akrabat:
I've had a play with a composite container and it's fine (though a little slower) as long as your "inner" container never needs anything from the "outer" container as the factory in the inner container doesn't have access to the outer one.

It's do-able though, even if I don't like it and think it's unnecessary :)

silentworks:
I just had a look at Zend Expressive and they are doing a configure it yourself kinda setup, I think if we include defaults for the Pimple container we include and write documentation for implementing any other Container you want to use. This is the current approach of Zend Expressive http://zend-expressive.readthedocs.org/en/stable/container/pimple/.

So I say leave it as is, anyone wanting to change to another Container just needs instruction on what services needs to be implemented. I will open a issue on the documentation website.

Thanks for all the comments on here, it was really good learning from you all.

juliangut:
The difference is Slim3 ships with Pimple bundled in its core while Expressive does not bundle any container in particular, rather provide documentation for the three they like the most (I guess), Zend ServiceManager, Aura.DI and Pimple, meaning it is container unopinionated as you have to require the dependency specifically

ppetermann:
@juliangut not sure how that makes a difference? being able to replace the container and having the documentation seems to enable everyone to go his own way (including myself)

silentworks:
@juliangut we have always shipped with a Container, so this is nothing new, Pimple does not follow Container Interop, we could go on to create another class just to put the Pimple services in, but we don't see any value in this as it would still be only useful to Pimple. I think out approach is about ease of use for new developers, if you are a advanced developer then you can change and reconfigure you services if you wish. But we are still catering for beginners as much as we can.

Issue #1498, opened 2015-09-13
errorHandling detailed display mode


akrabat:
The detailed error message and stack traces in the errorHandler should not be displayed unless a setting is enabled by the user. This setting should "fail safe" and default to not display this data.

In general, I'd prefer a specific setting rather than a generic "debug" setting that controls lots of different things across the framework.

jaapverloop:
As discussed on IRC, I leave a comment here with the same question.

Is there a specific reason why the error handler does not include the type (class) for xml and json? I expected the same info, no matter the requested format. If it's because of security, we could use this setting you mentioned to hide this kind of information.

akrabat:
No reason at all - it was just missed. Feel free to submit a PR.

designermonkey:
I was talking about this the other day for a project I'm on. We are going for a `?debug` addition to a url to see the output.

gabriel403:
@akrabat can we close this now?

PR #1499, opened 2015-09-15
Add some little `improvements` to the error handler

jaapverloop:
Most important changes:
- add missing exception type to json/xml format (mentioned in #1498)
- renamed the root element to **error** for clarity (xml format)
- wrapped the json output in a root element like the xml format to make it more consistent

I found that heredoc, used in `renderXmlErrorMessage`, is a little bit limited to work with (calling functions for example). So I've changed the code to make it easier to work with (IMHO) for future changes.

Not sure whether we should wrap some xml elements like trace in `<![CDATA[ ]]>` tags.

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -148,7 +148,7 @@ class Error
             ];
         } while ($exception = $exception->getPrevious());
 
-        return json_encode($error);
+        return json_encode($error, JSON_PRETTY_PRINT);
     }
 
     /**
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -162,21 +162,17 @@ class Error
         $xml = "<root>\n  <message>Slim Application Error</message>\n";
 
         do {
-            $type = get_class($exception);
-            $xml .= <<<EOT
-  <exception>
-    <type>{$type}</type>
-    <code>{$exception->getCode()}</code>
-    <message>{$exception->getMessage()}</message>
-    <file>{$exception->getFile()}</file>
-    <line>{$exception->getLine()}</line>
-    <trace>{$exception->getTraceAsString()}</trace>
-  </exception>
-
-EOT;
+            $xml .= "  <exception>\n";
+            $xml .= "    <type>" . get_class($exception) . "</type>\n";
+            $xml .= "    <code>" . $exception->getCode() . "</code>\n";
+            $xml .= "    <message>" . $exception->getMessage() . "</message>\n";
+            $xml .= "    <file>" . $exception->getFile() . "</file>\n";
+            $xml .= "    <line>" . $exception->getLine() . "</line>\n";
+            $xml .= "    <trace>" . $exception->getTraceAsString() . "</trace>\n";
+            $xml .= "  </exception>\n";
         } while ($exception = $exception->getPrevious());
 
-        $xml .="</root>";
+        $xml .= "<root>";
 
         return $xml;
     }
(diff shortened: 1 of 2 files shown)


JoeBengalen:
This will put the XML error on a single line instead of being nicely formatted

jaapverloop:
That's true, but easy to fix in this setup if desired. The reason why is simple (albeit stupid): the json format is not pretty printed too.

So if you want it to look nice, I will do it!

JoeBengalen:
I do not know if it matters. I just wanted to point it out.

Personally I never use xml so I cant really be bothered. ;)

silentworks:
Travis says you are trying to convert a Array to string or vice-versa

jaapverloop:
@JoeBengalen OK, thanks for your review! Lets wait what others think about this.

@silentworks I noticed, see https://github.com/sebastianbergmann/phpunit/issues/1855

juliangut:
@jaapverloop don't know why but so many sprintf one after another doesn't seem right to me

akrabat:
Re-ran the one failed Travis build and it passed.

akrabat:
> add missing exception type to json/xml format (mentioned in #1498)

Thanks!

> renamed the root element to error for clarity (xml format)

I like this.

> wrapped the json output in a root element like the xml format to make it more consistent

I really don't see the point in this. The root element in XML is requirement of the XML specification. It's not needed for JSON.

> I found that heredoc, used in renderXmlErrorMessage, is a little bit limited to work with (calling functions for example). So I've changed the code to make it easier to work with (IMHO) for future changes.

Not a fan of this. I find it much easier to parse the heredoc than to deal with a series of sprintf() statements as they don't read left-to-right. If you're determined to get rid of the heredoc, then please use string concatenation.

> Not sure whether we should wrap some xml elements like trace in <![CDATA[ ]]> tags.

That's a good point. I suspect that we should do this for all the elements.

jaapverloop:
@akrabat thanks for your comments.

> I really don't see the point in this. The root element in XML is requirement of the XML specification. It's not needed for JSON.

Not a problem to me, but as I said, for consistency. Beside that, whoops, the populair error handler, does it the same way.

https://github.com/filp/whoops/blob/master/docs/API%20Documentation.md#handler-json

> Not a fan of this. I find it much easier to parse the heredoc than to deal with a series of sprintf() statements as they don't read left-to-right. If you're determined to get rid of the heredoc, then please use string concatenation.

Point taken. I will update this PR with string concatenation.

> That's a good point. I suspect that we should do this for all the elements.

Ok, working on it.

Other points/questions:
- Is pretty output for xml desired? If yes, we could use the `JSON_PRETTY_PRINT` flag for json as well.
- Why is the trace splitted in separate divs (html format)? Is the `pre` tag not enough? Beside that, the closing `div` tags are missing because it replaces with `'\n'` instead of `"\n"`.
- The values used in html format should be escaped too right? Any html elements in a trace are not visible for example.

Slim/Handlers/Error.php
```diff
@@ -158,27 +159,36 @@ private function renderJsonErrorMessage(Exception $exception)
...
-  <exception>
-    <code>{$exception->getCode()}</code>
-    <message>{$exception->getMessage()}</message>
-    <file>{$exception->getFile()}</file>
-    <line>{$exception->getLine()}</line>
-    <trace>{$exception->getTraceAsString()}</trace>
-  </exception>
-
-EOT;
+            $xml .= "  <exception>\n";
```
piotr-cz, line 165: Why not keep heredoc format? it's so readable

PR #1504, opened 2015-09-21
App output length equal to specified content-length (if specified explicitly)

yurevichcv:
This feature is very important when working with external streams and stream filters. 

For example, I'm downloading encrypted file from S3 and automatically decrypt it using stream_filter. Decryptor automatically pad my output with `"\0"` statements to fit block size. This way final downloaded file is corrupted. I can not apply `trim` function, as I don't have access to the whole stream content. This way I just want my app to output exact amount of bytes, without padded elements.

I prepared special unit test, where you can see the issue.


yurevichcv:
Added fix for situation when `BODY_SIZE % CHUNK_SIZE = 0`

akrabat:
Thank you

PR #1514, opened 2015-10-01
Removed redundant code block in CallableResolver::resolve()

oanhnn:
In this PR: 
- Removed redundant code block in `CallableResolver::resolve()`

I think this block is unnecessary.

```
if ($conditions) {
    // to do something
    check();
} else {
    // to do something
}
check();
```

Function `check()` will be executed double if `$conditions` is true.


silentworks:
Your comment doesn't match with the PR you sent. Closing this out until correct code is sent.

PR #1515, opened 2015-10-01
Removed redundant code block in CallableResolver::resolve()

oanhnn:
In this PR:
- Removed redundant code block in `CallableResolver::resolve()`

Sorry, my english is not good!

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -74,9 +74,6 @@ final class CallableResolver implements CallableResolverInterface
                     $resolved = new $class;
                 }
             }
-            if (!is_callable($resolved)) {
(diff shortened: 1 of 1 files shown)


silentworks:
No its run only once, its run agains the newly assigned `$resolved` variable.

JoeBengalen:
But on line 84 it will check again?
https://github.com/slimphp/Slim/pull/1515/files#diff-bf51d44b5bda6198e439b2067fc9cfb3L84

silentworks:
Haha, I was looking at the patch-2 branch submitted by @oanhnn and was wondering how comes I only see that check happening only once. Then I looked at the 3.x branch and realise you are both correct it is being called twice.

PR #1517, opened 2015-10-02
Assign the resolved early

silentworks:


akrabat:
Why this change?

oanhnn:
Are you sure that it will run faster? Assign command always is executed even `!is_callable($toResolve) && is_string($toResolve)` is `true`. With old code, it isn't executed.

silentworks:
@oanhnn not sure if its that much faster but its doing the same as previous code minus using the else.

PR #1518, opened 2015-10-02
Update Cookies.php

hz61p1:


JoeBengalen:
Why this change? What would is solve? Looks a bit weird to be. You cast the value to a string after you check `is_string()`?

Also this change does not force the value to be an array, which could break `array_replace()`. On the other hand; the old way could crash if the value cannot be casted to a string.

codeguy:
Not sure what this tries to fix. Closing. Feel free to provide a better explanation and I'll reconsider.

Issue #1513, opened 2015-09-30
Responses handling


designermonkey:
As I build an app with Slim 3, I have been musing over how the response handling works, and I'd like to put forward an idea (or get my understanding sorted out).

Currently, we have the ability to let a route that is not found in the router return a 404 handler directly, but what about if we can't find relevant data in our data layer? Surely that too is a 404, and should be handled by Slim too?

This got me thinking about these handlers we have. Shouldn't they be partnered with Exceptions? `NotFoundException` is throwable inside my application, and is 'handled' by Slim to display the registered `NotFound` handler.

The same would apply to `NotAllowed` and the generic `Error` (at least I think `Error` is already right?)

Exceptions are the only thing that successfully bubble up the chain, and allow me to concentrate more on building the app, than worrying about having to instantiate the right handler.

jensscherbl:
> NotFoundException is throwable inside my application, and is 'handled' by Slim to display the registered NotFound handler.

I handled this by returning a `notFoundHandler` response in my controller action.

```
if (!$data) {

    return call_user_func($this->notFoundHandler, $request, $response);
}
```

This was pre-beta a few month ago and a bit ugly in my opinion, not sure if this is still up to date.

But yeah, what @designermonkey suggests is pretty much what I'd expect as well.

kminek:
you can easily achieve what you want. set your custom exception handler via `set_exception_handler` (`$this` refers to some generic error class which holds reference to app object):

``` php
    set_exception_handler([$this, 'handleException']);
```

and then ALSO use this handler in Slim's callbacks:

``` php
    public function errorHandler()
    {
        $self = $this;
        return function ($request, $response, \Exception $e) use ($self) {
            return $self->handleException($e, $request, $response);
        };
    }
```

this way you'll have single place to handle exceptions and perform desired logic (based on custom exception class type for instance - like RecordNotFoundException or something like that)

designermonkey:
I understand this, yes, but I am talking about the actual architecture of Slim itself. Error handling is already dealt with in Slim's execution, here: https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L306-L318

I'm aiming this discussion at having more Exception classes, and expanding the catch list for the relevant exception types, passing them to the right handler during `App::run`s operation.

Slim's `App::__invoke` method would also just be required to thrown a `NotAllowedException` (which would accept the correct allowed methods) or `NotFoundException` which would also be caught by the catch list and handled accordingly.

[...]

PR #1519, opened 2015-10-02
Change to exception based route response

designermonkey:
As discussed in #1513, these changes allow us to throw specific exceptions within our app to generate the right response using the handlers.

Of course the existing exception behaviour remains the same.

There are two tests that fail, and I'm not sure what wants to be done with them, any suggestions from anyone, and I'll get to that.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -16,7 +16,9 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Interop\Container\ContainerInterface;
 use FastRoute\Dispatcher;
-use Slim\Exception\Exception as SlimException;
+use Slim\Exception\SlimException;
+use Slim\Exception\MethodNotAllowedException;
+use Slim\Exception\NotFoundException;
 use Slim\Http\Uri;
 use Slim\Http\Headers;
 use Slim\Http\Body;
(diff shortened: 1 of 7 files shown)


akrabat:
Thanks! I put back the ability to return a response in __invoke so that App's `__invoke` continues to act like Middleware.

i.e. it's possible that App may be at the bottom of a middleware chain run by [Stratigility](https://github.com/zendframework/zend-stratigility) or [Relay](http://relayphp.com) rather than one controlled by `App::run()`, so I want `__invoke()` to follow the correct convention and return a Response.

This also had the side effect of making the current tests pass, so I've added additional tests to check that the exceptions are thrown in `__invoke` if the handlers are not in the container.

designermonkey:
Great, thanks!

jensscherbl:
Nice, thanks.

> i.e. it's possible that App may be at the bottom of a middleware chain run by Stratigility or Relay rather 

Any use case where something like this would make sense? Always thought it would make more sense for Slim itself to use something like Stratigility or Relay internally.

akrabat:
The case I can think of is when you're integrating many applications together.

silentworks:
@jensscherbl my current use case is I have a old Codeigniter app which I have created a middleware for and I am using Zend Diactoros to pipe the Codeigniter app as a middleware and also adding Slim 3 as a middleware to it. So in my front controller I will look in Slim for the path to see if it matches, if not I tell Slim to go to the next middleware which is the Codeigniter one and see if it matched there.

jensscherbl:
Thanks for the answers. So this is mostly a practical thing for working with legacy projects, or for using middleware that's not directly compatible with Slim?

Issue #1325, opened 2015-06-28
Problems with HTTP_HOST containing IPv6 adress


Robbilie:
when the said server var contains an ipv6 adress it splits the adress after the first  ":" (like before the port in ipv4 adress) which does not make any sense with an ipv6 adress

akrabat:
Hi,

Are you seeing this in version 2 or the upcoming version 3?

Robbilie:
"slim/slim": "~2.6.2"

PR #1524, opened 2015-10-06
Fix address truncation if IPv6 address in HTTP_HOST

gkralik:
If `HTTP_HOST` contains an IPv6 address (eg `[2001:db8::1]`), the current implementation of `Request::getHost()` truncates this address after the first colon, yielding `[2001` as the host instead of `[2001:db8::1]` (see #1325).

This PR fixes the `Request::getHost()` behaviour to check if an IPv6 address is given in `HTTP_HOST` and to split only on the colon after the IPv6 address.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -474,10 +474,14 @@ class Request
     public function getHost()
     {
         if (isset($this->env['HTTP_HOST'])) {
-            if (strpos($this->env['HTTP_HOST'], ':') !== false) {
-                $hostParts = explode(':', $this->env['HTTP_HOST']);
+            if(preg_match('/^(\[[a-fA-F0-9:.]+\])(:\d+)?\z/', $this->env['HTTP_HOST'], $matches)) {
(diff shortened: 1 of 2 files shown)


akrabat:
Do we need the equivalent for Slim 3?

gkralik:
I'll check that...

gkralik:
Yes, but it's in `\Slim\Http\Uri` in Slim 3.
Will take care of it and create another PR.

akrabat:
I've merged this incase anyone needs this functionality in Slim 2. We probably won't do another release though unless a security fix is required.

silentworks:
I am inclined to consider this a critical bug and think there should be a patch release for it.

akrabat:
Fine with me. If we're going to do a new Slim2 patch release, we should probably review the other slim2 PRs and see if any of those are worth adding in too.

silentworks:
@akrabat I agree.

lalop:
@akrabat @silentworks any news about this path release ?

Issue #1508, opened 2015-09-29
Slim 2.x: Mishandling of array in post


r3wt:
Repro:

``` php
$app = new \Slim\Slim();

$app->get('/post_test',function() use($app){
    if($app->request->isPost()){
        pretty($app->request()->post());
        pretty($_POST);
    }else{
        $html =
        '
        <!DOCTYPE html>
        <html>
        <head>
            <meta charset="utf-8"/>
            <title>test</title>
        </head>
        <body>
        <form method="post" style="width:150px;">
            <select multiple name="regions[]" style="width:100%;display:block">
        ';
        for($i=0;$i<20;$i++){
            $html .='<option value="'.$i.'">'.$i.'</option>';
        }
        $html .= '</select></br><input type="submit" value="submit" style="width:100%; display: block"></form></body></html>';
        echo $html;
    }
});

function pretty($what)
{
    echo '<pre>'.print_r($what,true).'</pre>';
}
```

output:

```
//slim $app->request()->post()
Array
(
    [regions[]] => 3
)
//php $_POST
Array
(
    [regions] => Array
        (
            [0] => 1
            [1] => 2
            [2] => 3
        )

)
```

r3wt:
bizzzump

gkralik:
With Slim 2.6.2:

``` php
<?php

require 'vendor/autoload.php';

$app = new \Slim\Slim();

$app->map('/post_test',function() use ($app) {
    if($app->request->isPost()){
        pretty($app->request()->post());
        pretty($_POST);
    }else{
        $html =
        '
        <!DOCTYPE html>
        <html>
        <head>
            <meta charset="utf-8"/>
            <title>test</title>
        </head>
        <body>
        <form method="post" style="width:150px;">
            <select multiple name="regions[]" style="width:100%;display:block">
        ';
        for($i=0;$i<20;$i++){
            $html .='<option value="'.$i.'">'.$i.'</option>';
        }
        $html .= '</select></br><input type="submit" value="submit" style="width:100%; display: block"></form></body></html>';
        echo $html;
    }
    })->via('GET', 'POST');

$app->run();

function pretty($what)
{
    echo '<pre>'.print_r($what,true).'</pre>';
}
```

produces:

```
Array
(
    [regions] => Array
        (
            [0] => 1
            [1] => 2
        )

)
Array
(
    [regions] => Array
        (
            [0] => 1
            [1] => 2
        )
)
```

[...]

PR #1526, opened 2015-10-06
[HHVM] mb_parse_str() fails to parse array POST parameters

gkralik:
Using HHVM 3.9.1, a POST data string of `foo[]=bar&foo[]=baz` does not parse into the correct array structure using `mb_parse_str()`. It works correctly when using `parse_str()`.

``` php
mb_parse_str('foo[]=bar&foo[]=baz&abc=123', $out)
/*
On HHVM, this produces:
array(2) {
  ["foo[]"]=>
  string(3) "baz"
  ["abc"]=>
  string(3) "123"
}

but the expected output is:
/*
array(2) {
  ["foo"]=>
  array(2) {
    [0]=>
    string(3) "bar"
    [1]=>
    string(3) "baz"
  }
  ["abc"]=>
  string(3) "123"
}
*/
```

This PR adds a failing test for this behaviour.

### DIFF
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -260,6 +260,26 @@ class RequestTest extends PHPUnit_Framework_TestCase
     }
 
     /**
+     * Test fetch POST params with arrays
+     */
+    public function testPostArray()
+    {
+        $env = \Slim\Environment::mock(array(
+            'REQUEST_METHOD' => 'POST',
+            'slim.input' => 'foo[]=bar&foo[]=baz&abc=123',
+            'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
+            'CONTENT_LENGTH' => 27
+        ));
+        $req = new \Slim\Http\Request($env);
+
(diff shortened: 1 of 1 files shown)


gkralik:
A simple fix would be to check for HHVM and don't call `mb_parse_str()`. But this would (possibly) break multibyte support.

See #1508 for the related issue.

gkralik:
Maybe something like this:

``` php
if (function_exists('mb_parse_str') && !defined('HHVM_VERSION') 
    && !isset($this->env['slim.tests.ignore_multibyte']) 
) { 
    mb_parse_str($this->env['slim.input'], $output);
} else {
    parse_str($this->env['slim.input'], $output);
}
```

At least it doesn't break any existing tests....

gkralik:
upstream bug for reference: facebook/hhvm#6340

akrabat:
In general, we are only accepting critical bugs or security fixes for Slim 2. HHVM support doesn't meet these criteria.

PR #1531, opened 2015-10-07
Check body is writeable before writing

### LINKED ISSUE #1530: Do not assume Response is writable
In Route.php line 340, we should not assume Response body is writable.

https://github.com/slimphp/Slim/blob/3.x/Slim/Route.php#L340

gabriel403:
Fixes #1530 

Interested if this is the right approach or if we should be throwing an exception/creating a new body/setting fire to the nearest postbox if it's not writeable?

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -333,11 +333,15 @@ class Route extends Routable implements RouteInterface
             if ($this->outputBuffering === 'prepend') {
                 // prepend output buffer content
                 $body = new Http\Body(fopen('php://temp', 'r+'));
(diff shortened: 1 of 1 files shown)


codeguy:
Some streams are not writable (e.g. Guzzle's PumpStream). Which is not an exceptional situation. But it becomes one when Slim tries to write to it (I found this out yesterday).

codeguy:
We should respect the Response returned by the route callback. We can ask it if it's writable, and if it says yes, we write to it. If it isn't writable, we don't.

PR #1534, opened 2015-10-09
[3.x] Fix IPv6 address truncation in Uri factory

gkralik:
Fix the handling of IPv6 addresses in `HTTP_HOST` in `Uri::createFromEnvironment()`.
For details, see #1524

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -179,12 +179,19 @@ class Uri implements UriInterface
         }
 
         // Authority: Port
-        $pos = strpos($host, ':');
-        if ($pos !== false) {
-            $port = (int)substr($host, $pos + 1);
-            $host = strstr($host, ':', true);
+        $port = (int)$env->get('SERVER_PORT', 80);
+        if(preg_match('/^(\[[a-fA-F0-9:.]+\])(:\d+)?\z/', $host, $matches)) {
+            $host = $matches[1];
+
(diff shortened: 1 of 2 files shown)


akrabat:
Looks like you have some PHPCS failures: https://travis-ci.org/slimphp/Slim/jobs/84453270

http://i.19ft.com/20097d12.png

gkralik:
Just pushed a fix ;)

Slim/Http/Uri.php
```diff
@@ -179,12 +179,19 @@ public static function createFromEnvironment(Environment $env)
         }
 
         // Authority: Port
-        $pos = strpos($host, ':');
-        if ($pos !== false) {
-            $port = (int)substr($host, $pos + 1);
-            $host = strstr($host, ':', true);
+        $port = (int)$env->get('SERVER_PORT', 80);
+        if (preg_match('/^(\[[a-fA-F0-9:.]+\])(:\d+)?\z/', $host, $matches)) {
```
akrabat, line 183: Are you sure that the square brackets are always there? https://bugs.php.net/bug.php?id=43632 implies not.

Edit - experimenting locally, I think you're right that are always there.

gkralik, line 183: AFAIK they need to be there for `HTTP_HOST`. They shouldn't be in `SERVER_NAME` though (and that's where some servers choke).

PR #1537, opened 2015-10-13
Current route now updated during routing and Dispatched route is stored

pdscopes:
Fixing bug where Router->currentRoute was not being updated during Sim run.

Bug only appears when there is more than one matching route AND the first matching route calls $app->pass().

This meant that hook functions could not accurately tell which route was about to be dispatched/had just been dispatched.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -95,6 +95,15 @@ class Router
     }
 
     /**
+     * Sets the current route to the given route
+     * @param Route|null $route
+     */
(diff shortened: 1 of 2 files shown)


pdscopes:
Improved the reliability of current route being restored to it's initial value after the routing has been completed.

And added extra functionality to know which route was dispatched.

akrabat:
As Slim 3 is released, we are only fixing Security issues on Slim 2.

PR #1538, opened 2015-10-15
made test use psr-7 compliant interface

### LINKED ISSUE #1506: PSR-7 Compatibility
Is Slim 3.0 actually PSR-7 Swappable? It is not.

Results of commenting out non-psr-7 methods on Slim Classes
Tests: 362, Assertions: 231, Errors: 213.

Gist log of my PHPUnit test
https://gist.github.com/geggleto/c4b73336ef9507e24c48

1) Slim\Http\Request::registerMediaTypeParser
2) Slim\Http\Uri::getBasePath()
3) Slim\Http\Uri::withBasePath() 
4) Slim\Http\Response::write()
5) Response::withRedirect()
6) Response::isEmpty()
7) Response::isInformational()
8) Response::isOk()
9) Response::isSuccessful() 
10) ... a lot of these are failing because they are just tests and not that they are used internally.

ppetermann:
to write to response, fixes #1506

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -959,7 +959,7 @@ class AppTest extends \PHPUnit_Framework_TestCase
         $method->setAccessible(true);
 
         $response = new Response();
-        $response->write('foo');
+        $response->getBody()->write('foo');
 
         $response = $method->invoke(new App(), $response);
 


geggleto:
:+1:

PR #1541, opened 2015-10-17
Proxy non-existent method calls to the container

akrabat:
This is a convenience method that allows for calling container methods as if they were methods on App.

e.g.:

``` php
$app->get('/', function ($req, $res)
{
    return $this->notFoundHandler($request, $response);
}
```

Is this a good idea?

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -129,6 +129,24 @@ class App
         return $this->container->has($name);
     }
 
+    /**
+     * Calling a non-existant method on App checks to see if there's an item
+     * in the container than is callable and if so, calls it.
+     *
+     * @param  string $method
+     * @param  array $args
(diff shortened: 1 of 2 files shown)


ppetermann:
you mean so people use app as a service locator?

akrabat:
App already works as a service locator: `$app->notFoundHandler` already works.

This PR would allow for calling the located object if it's a function in one line of code rather than two. 

i.e. it replaces:

``` php
    $notFoundHandler = $this->notFoundHandler;
    return $notFoundHandler($request, $response);
```

with

``` php
    return $this->notFoundHandler($request, $response);
```

As I said, it's a convenience thing.

tuupola:
I like. Makes the clone look much cleaner.

Issue #1521, opened 2015-10-04
App->subRequest not working directly with Router->pathFor


pokmot:
I am working in a sub-folder, so getBasePath() = '/MM/systems/admin'

$App->subRequest('POST', $router->pathFor('myNamedRoute')) produces:
/MM/systems/admin/myroute

However, subRequest is only expecting 'myroute'

To make this work, I have to do a str_replace:

$App->subRequest('POST', str_replace($request->getUri()->getBasePath(), '', $router->pathFor('myNameRoute')));

akrabat:
Hmm.
Does the router need a different method for this? or a parameter on pathFor() ?

pokmot:
My gut feeling is that something like $router->internalPathFor() would be appropriate; it would make it clear whether the path is returned with or without getBasePath() prepended.

akrabat:
I quite like `internalPathFor()` too.

pokmot:
I've made a suggested code change (#1551) - unfortunately I won't have a chance to update any unit tests, but hope it helps/is useful.

akrabat:
ok. I'll sort out.

silentworks:
This is now fixed with the new `$router->relativePathFor()` method.

PR #1545, opened 2015-10-19
router->pathfor fullpath param for subRequests. #1521

ssotomayor:
I assume with this you could do something like this:

```
$app->get('/', function ($request, $response, $args) use ($router) {
    $sub = $this->subRequest('GET', $router->pathFor('hello', [], [], false));
})->setName('home');

$app->get('/hello', function ($request, $response, $args) use ($router) {
    $response->write("hellorequest ");
    return $response;
})->setName('hello');
```

However i'm not sure $router-pathFor is expected to be used with app->subRequest, since it simply returns the path. Perhaps an owner can jump in and clarify, or maybe you can use this change.

Best.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -243,13 +243,14 @@ class Router extends RouteCollector implements RouterInterface
      * @param string $name        Route name
      * @param array  $data        Named argument replacement data
      * @param array  $queryParams Optional query string parameters
+     * @param boolean $fullPath   Optional param to return the full path or the last part for subRequests
      *
      * @return string
      *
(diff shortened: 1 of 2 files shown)


akrabat:
Can you explain the issue that you're fixing please?

ssotomayor:
Hi @akrabat i was referring to issue #1521. Altough now that i think about it, perhaps it would be better to have another method to return the route to work with subRequests.

akrabat:
Superseded by #1552 which implements `internalPathFor()`

PR #1551, opened 2015-10-22
Update Router.php adding internalPathFor

pokmot:
Suggested code update for issue #1521

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -279,7 +279,7 @@ class Router implements RouterInterface
     }
 
     /**
-     * Build the path for a named route
+     * Build the path for a named route - without basePath so it can be passed to subRequest
      *
      * @param string $name        Route name
      * @param array  $data        Named argument replacement data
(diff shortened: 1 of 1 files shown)


akrabat:
We still need pathFor().

The implementation of pathFor() should be:

``` php
    /**
     * Build the path for a named route including the base path
     *
     * @param string $name        Route name
     * @param array  $data        Named argument replacement data
     * @param array  $queryParams Optional query string parameters
     *
     * @return string
     *
     * @throws RuntimeException         If named route does not exist
     * @throws InvalidArgumentException If required data not provided
     */
    public function pathFor($name, array $data = [], array $queryParams = [])
    {
        $url = $this->internalPathFor($name, $data, $queryParams);

        if ($this->basePath) {
            $url = $this->basePath . $url;
        }

        return $url;
    }
```

Also, we need unit tests for `internalPathFor`

akrabat:
Superseded by #1552

PR #1552, opened 2015-10-22
Implement internalPathFor()

### LINKED ISSUE #1521: App->subRequest not working directly with Router->pathFor

akrabat:
This method is mostly useful when using `subRequest` which does not
expect the base path.

Closes #1521

### DIFF
--- a/Slim/Interfaces/RouterInterface.php
+++ b/Slim/Interfaces/RouterInterface.php
@@ -87,7 +87,21 @@ interface RouterInterface
     public function lookupRoute($identifier);
 
     /**
-     * Build the path for a named route
+     * Build the path for a named route excluding the base path
+     *
+     * @param string $name        Route name
+     * @param array  $data        Named argument replacement data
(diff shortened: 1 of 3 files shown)


silentworks:
Would this not be relative and absolute pathFor, I am not sure about the current name as its a public method saying its internal, when its accessible externally.

akrabat:
Valid point.

I'm not especially keen on renaming `pathFor` though on the grounds that this is the one that everyone should be using 99% of the time. Happy to call the other one `relativePathFor`.

akrabat:
Renamed.

PR #1558, opened 2015-10-28
remove container references and update documentation

juliangut:
Container is not needed in most of default services registration callbacks
Plus use `Interop\Container\ContainerInterface::get`instead of ArrayAccess

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -90,82 +90,74 @@ final class Container extends PimpleContainer implements ContainerInterface
          * This service MUST return an array or an
          * instance of \ArrayAccess.
          *
-         * @param Container $c
-         *
          * @return array|\ArrayAccess
          */
-        $this['settings'] = function ($c) use ($userSettings, $defaultSettings) {
+        $this['settings'] = function () use ($userSettings, $defaultSettings) {
             return array_merge($defaultSettings, $userSettings);
         };
 
-        /**
-         * This service MUST return a shared instance
-         * of \Slim\Interfaces\Http\EnvironmentInterface.
-         *
-         * @param Container $c
-         *
-         * @return EnvironmentInterface
-         */
         if (!isset($this['environment'])) {
-            $this['environment'] = function ($c) {
+            /**
+             * This service MUST return a shared instance
+             * of \Slim\Interfaces\Http\EnvironmentInterface.
+             *
+             * @return EnvironmentInterface
+             */
+            $this['environment'] = function () {
                 return new Environment($_SERVER);
             };
         }
 
-        /**
-         * PSR-7 Request object
-         *
-         * @param Container $c
-         *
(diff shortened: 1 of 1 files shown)


silentworks:
ArrayAccess is correct in these cases since you would already be in the context of Pimple inside of the closure. The Container being passed in via the registration doesn't matter too much wether it is being used or not.

silentworks:
Sorry about closing this earlier, can you send only the changes for the removal of unused `$c` variable please and I will get this merged in.

Slim/Container.php
```diff
@@ -186,60 +178,55 @@ private function registerDefaultServices($userSettings)
          */
         if (!isset($this['errorHandler'])) {
             $this['errorHandler'] = function ($c) {
-                $displayDetails = $c->get('settings')['displayErrorDetails'];
```
juliangut, line 189: In here `ContainerInterface` was being used instead of ArrayAccess
I just unify the way the container is being used on the rest of the callbacks to this syntax

Should I replace this line to use ArrayAccess then?

akrabat:
The Pimple documentation always shows that the factory closure takes a `$c` parameter even though it may not be used. Why are we removing it?

I've no particular opinion on changing use of `get()` to array access.

juliangut:
@akrabat I always avoid "unused parameters/variables" cause they are considered code smells.

In Slim docs (and Pimple's) it's stated that closures are passed container as parameter so that is clear to anyone registering services, so IMO there is no reason to keep these unused parameters in an internal closure just to act "as documentación"

juliangut:
Sorry for that "documentación" (documentation in spanish, easy one right?) Android spell checker!!

juliangut:
About getting services from container it doesn't matter if `$c->get()` or `$c[]` is used anytime it is consistent

silentworks:
Does `$c->get()` work in that context though? I thought Pimple was injecting itself and not the extended `Slim\Container` class.

akrabat:
@silentworks: Yes. See https://github.com/slimphp/Slim/blob/3.x/Slim/Container.php#L189

JoeBengalen:
Pimple will inject `$this`, so in our case our own extended version of the container:
https://github.com/silexphp/Pimple/blob/master/src/Pimple/Container.php#L113

juliangut:
@silentworks In my first comment on this I pointed to a line in the current Container implementation that is already using get() syntax, and it's working right now.

silentworks:
Ok in terms of consistence, if we are going with `$c->get()`, we will need to update all the other repositories README.md that make use of ArrayAccess instead. 

I am all for using `$c->get()`.

juliangut:
ArrayAccess is still a valid way to access services from Slim\Container (although I'd always go for ContainerInterface::get). I just made container use consistent **inside** Slim\Container.

PR #1559, opened 2015-10-29
Remove Request::getIp()

### LINKED ISSUE #1528: Request::getIp() trusts client headers and PSR-7 helper methods
https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L1078

It would be trivial for a client to send an X-Forwarded-For header that tricked the application into thinking it was from somewhere else. If this is going to be done, a trusted proxies whitelist must be established - and the method should check to see that the request comes from one of these before trusting the contents of the header.

My personal opinion is that adding methods like this to the PSR-7 implementation makes no sense. The whole point of using PSR-7 is that you are writing code which isn't coupled to an implementation of the standard. I think it makes more sense to create a seperate helper class which takes the PSR-7 message as an input.

akrabat:
This method was using the 'X-Forwarded-For' header which is unsafe as it's client settable. A better solution is to use middleware that sets an "ip" attribute onto the Request object.

Resolves #1528


silentworks:
I am going to leave this on hold for the minute. We need to check the impact this is going to have and find a viable solution to suggest to people first.

akrabat:
This is a stab at doing it in middleware: https://github.com/akrabat/rka-ip-address-middleware

piotr-cz:
Good catch.

piotr-cz:
For the exact same reason the check for secure connection shouldn't be based on the `HTTP_X_FORWARDED_PROTO` header inside the [`Uri`](https://github.com/slimphp/Slim/blob/8bd1002bb69fc1a0c354e72c3195bc56d7dd22c1/Slim/Http/Uri.php#L161) class (more details are [here](https://github.com/joomla/joomla-framework/pull/212)).

Anyway base path resolution is wrong determining when using server rewrites (the `L` flag in .htaccess file).
There should be an option to define base path (and possibly secure connection) as an application setting which can be read from an environment variable.

akrabat:
> Anyway base path resolution is wrong determining when using server rewrites (the L flag in .htaccess file).
> There should be an option to define base path (and possibly secure connection) as an application setting which can be read from an environment variable.

Please open as a separate issue with the apache rewrite settings the show the problem.

akrabat:
See https://github.com/slimphp/Slim/issues/1567 for the Uri() usage of `HTTP_X_FORWARDED_PROTO`.

akrabat:
If we remove them, then https://travis-ci.org/akrabat/rka-scheme-and-host-detection-middleware would work as a middleware replacement.

Issue #1563, opened 2015-10-30
Changing Slim's settings during execution


kminek:
The fact that `settings` shared service returns just a plain array prevents modyfing application settings 'state' during execution (for instance inside route handlers). Such ability may be desired in many cases. Here is an example:

``` php
$app = new Slim\App;
$app->get('/', function () {
    $settings = $this->getContainer()->get('settings');
    $settings['displayErrorDetails'] = true;
    $settings['responseChunkSize'] = ...;
    ...
    throw new Exception('Something went wrong');
});
$app->run();
```

My advise is to wrap settings with object with array access. It could be done with minimal effort. It looks like Slim already provides generic collection class: `Slim\Http\Collection`. I would suggest moving this class under `Slim\Collection` namespace (as collections can be used not only in HTTP context) and define `settings` service like following:

``` php
$this['settings'] = function () use ($userSettings, $defaultSettings) {
    return new Collection(array_merge($defaultSettings, $userSettings));
};
```

silentworks:
You are correct indeed, not sure why we didn't do that. Let me get @akrabat to comment here if he can see any reason not to do this.

akrabat:
Fixed in #1564

PR #1564, opened 2015-10-30
Moved Collection to top level namespace

silentworks:
Updated settings to use Collection instead of a raw array

This PR should solve #1563


akrabat:
Moving Collection will make splitting out Slim\Http to its own rep slightly harder later. Not insurmountable, though.

PR #1573, opened 2015-11-04
Allow overriding of 'settings' service in Container

akrabat:
We allow every other service to be overridden, so this PR allows settings to be overridden too.


akrabat:
Can't get this to work without executing the passed in settings factory, so it would be better for someone to replace our settings factory with their own after `Container` is instantiated.

PR #1574, opened 2015-11-04
determineContentType to use array_intersect

juliangut:
In case the first 'Accept' content type on the list is not in the known content types array using array_intersect improves speed in comparison with traversing the list

By the way, couldn't determineContentType be placed on `Slim\Handlers\AbstractHandler` and use $request as parameter instead? this way it can be used by users as a base to extend their handlers upon. Just a thought

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -24,6 +24,18 @@ class Error
     protected $displayErrorDetails;
 
     /**
+     * Known handled content types
+     *
+     * @var array
+     */
+    protected $knownContentTypes = [
+        'application/json',
+        'application/xml',
+        'text/xml',
+        'text/html',
+    ];
+
+    /**
      * Constructor
      *
      * @param boolean $displayErrorDetails Set to true to display full details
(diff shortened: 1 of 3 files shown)


oanhnn:
Why not do you use traits script for reused ?

akrabat:
> By the way, couldn't determineContentType be placed on Slim\Handlers\AbstractHandler and use $request as parameter instead? this way it can be used by users as a base to extend their handlers upon. 

For application-workflow use-cases determineContentType isn't really comprehensive enough as the Accept header specification is more complicated. It's much better to use the `willdurand/Negotiation` package instead. For the error case here, determineContentType is "good enough", but I don't want to encourage it's usage by users.

juliangut:
I agree with your point

oanhnn:
I have a test about this method and old method to determine content type.
https://gist.github.com/oanhnn/bddebdccb100b337200a
I will create a PR to:
- create trait for determine content type
- implement determine content type by `isset()` function.
  What are you think about it?

Issue #1577, opened 2015-11-07
Route Group issue


silentworks:
There is an issue with Route Group in Slim 3 which have risen since adding in making the first route in a group being able to use `/` as its path. I have explained it more in the link along with an example.

https://gist.github.com/silentworks/0a9b03b69861dcb1d828

PR #1578, opened 2015-11-07
Better handling of slash in routegroup patterns

akrabat:
This PR handles route group patterns and slashes between them in a better manner. See issue #1577

When processing route groups, ensure that each group pattern is separated by a slash. However also ensure that there is never a double slash between groups and also that we never end up with an empty
pattern.

**Note**
This also means that a route with a trailing slash is different from a route without one. If we want `/foo` and `/foo/` to route to the same place, then I can update, or the route can be defined as:

``` php
$app->get('/foo[/]', function($request, $response, $args){/*...*/});
```

I have also significantly updated the [tests](https://github.com/akrabat/Slim/blob/better-handling-of-slash-in-routegroup-patterns/tests/AppTest.php#L151-L767) to cover all the cases I could think of:

Single routes:

| Route URI | Final pattern registered with router |
| --- | --- |
| `'/foo'` | `/foo` |
| `'/foo/'` | `/foo/` |
| `'foo'` | `/foo` |
| `'/'` | `/` |
| `''` | `/` |

Route groups:

| Group 1 | Group 2 (if exists) | Route URI | Final pattern registered with router |
| --- | --- | --- | --- |
| `'/foo'` |  | `'/bar'` | `/foo/bar` |
| `'/foo'` |  | `'/bar/'` | `/foo/bar/` |
| `'/foo'` |  | `'/'` | `/foo/` |
| `'/foo'` |  | `''` | `/foo` |
| `'/foo'` | `'/baz'` | `'/'` | `/foo/baz/` |
| `'/foo'` | `'/baz'` | `''` | `/foo/baz` |
| `'/foo'` | `'/baz'` | `'/bar'` | `/foo/baz/bar` |
| `'/foo'` | `'/baz'` | `'/bar/'` | `/foo/baz/bar/` |
| `'/foo'` | `'/'` | `'/bar'` | `/foo/bar` |
| `'/foo'` | `'/'` | `'bar'` | `/foo/bar` |
| `'/foo'` | `''` | `'/bar'` | `/foo/bar` |
| `'/foo'` | `''` | `'bar'` | `/foo/bar` |
| `'/'` |  | `'/bar'` | `/bar` |
| `'/'` |  | `'/bar/'` | `/bar/` |
| `'/'` |  | `'/'` | `/` |
| `'/'` |  | `''` | `/` |
| `'/'` | `'/baz'` | `'/'` | `/baz/` |
| `'/'` | `'/baz'` | `''` | `/baz` |
| `'/'` | `'/baz'` | `'/bar'` | `/baz/bar` |
| `'/'` | `'/baz'` | `'/bar/'` | `/baz/bar/` |
| `'/'` | `'/'` | `'/bar'` | `/bar` |
| `'/'` | `'/'` | `'bar'` | `/bar` |
| `'/'` | `''` | `'/bar'` | `/bar` |
| `'/'` | `''` | `'bar'` | `/bar` |
| `''` |  | `'/bar'` | `/bar` |
| `''` |  | `'/bar/'` | `/bar/` |
| `''` |  | `'/'` | `/` |
| `''` |  | `''` | `/` |
| `''` | `'/baz'` | `'/'` | `/baz/` |
| `''` | `'/baz'` | `''` | `/baz` |
| `''` | `'/baz'` | `'/bar'` | `/baz/bar` |
| `''` | `'/baz'` | `'/bar/'` | `/baz/bar/` |
| `''` | `'/'` | `'/bar'` | `/bar` |
| `''` | `'/'` | `'bar'` | `/bar` |
| `''` | `''` | `'/bar'` | `/bar` |
| `''` | `''` | `'bar'` | `/bar` |

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -122,14 +122,18 @@ class Router implements RouterInterface
             throw new InvalidArgumentException('Route pattern must be a string');
         }
 
+        // pattern must start with a / if not empty
+        if ($pattern) {
+            $pattern = '/' . ltrim($pattern, '/');
+        }
+
         // Prepend parent group pattern(s)
         if ($this->routeGroups) {
-            // If any route in the group only has / we remove it
-            if ($pattern === '/') {
-                $pattern = '';
-            }
             $pattern = $this->processGroups() . $pattern;
         }
+
+        // Complete pattern must start with a /
+        $pattern = '/' . ltrim($pattern, '/');
 
         // Add route
         $route = new Route($methods, $pattern, $handler, $this->routeGroups, $this->routeCounter);
@@ -235,7 +239,8 @@ class Router implements RouterInterface
     {
         $pattern = "";
         foreach ($this->routeGroups as $group) {
-            $pattern .= $group->getPattern();
+            // The route group's pattern doesn't end with a /
+            $pattern .= rtrim($group->getPattern(), '/');
         }
         return $pattern;
     }
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -582,7 +582,6 @@ class AppTest extends \PHPUnit_Framework_TestCase
         $this->assertAttributeEquals('/bar', 'pattern', $router->lookupRoute('route0'));
     }
 
-//HERE
     public function testEmptyGroupWithSegmentRouteThatDoesNotEndInASlash()
     {
         $app = new App();
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -149,6 +149,69 @@ class AppTest extends \PHPUnit_Framework_TestCase
     }
 
     /********************************************************************************
+     * Route Patterns
+     *******************************************************************************/
+    public function testSegmentRouteThatDoesNotEndInASlash()
+    {
+        $app = new App();
+        $app->get('/foo', function ($req, $res) {
+            // Do something
+        });
+        /** @var \Slim\Router $router */
+        $router = $app->router;
+        $router->finalize();
+        $this->assertAttributeEquals('/foo', 'pattern', $router->lookupRoute('route0'));
+    }
+
+    public function testSegmentRouteThatEndsInASlash()
+    {
+        $app = new App();
+        $app->get('/foo/', function ($req, $res) {
+            // Do something
+        });
+        /** @var \Slim\Router $router */
+        $router = $app->router;
+        $router->finalize();
+        $this->assertAttributeEquals('/foo/', 'pattern', $router->lookupRoute('route0'));
+    }
+
+    public function testSegmentRouteThatDoesNotStartWithASlash()
+    {
+        $app = new App();
+        $app->get('foo', function ($req, $res) {
+            // Do something
+        });
+        /** @var \Slim\Router $router */
+        $router = $app->router;
+        $router->finalize();
+        $this->assertAttributeEquals('/foo', 'pattern', $router->lookupRoute('route0'));
+    }
+
+    public function testSingleSlashRoute()
+    {
+        $app = new App();
+        $app->get('/', function ($req, $res) {
+            // Do something
+        });
+        /** @var \Slim\Router $router */
+        $router = $app->router;
+        $router->finalize();
+        $this->assertAttributeEquals('/', 'pattern', $router->lookupRoute('route0'));
+    }
+
+    public function testEmptyRoute()
+    {
+        $app = new App();
+        $app->get('', function ($req, $res) {
+            // Do something
+        });
+        /** @var \Slim\Router $router */
+        $router = $app->router;
+        $router->finalize();
+        $this->assertAttributeEquals('/', 'pattern', $router->lookupRoute('route0'));
+    }
+
+    /********************************************************************************
      * Route Groups
      *******************************************************************************/
     public function testGroupSegmentWithSegmentRouteThatDoesNotEndInASlash()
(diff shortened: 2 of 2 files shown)


akrabat:
Thinking about it, I wonder if we should just provide middleware to redirect the trailing slash if someone wants that?

``` php
    $uri = $request->getUri();
    $path = $uri->getPath();
    if ($path != '/' && substr($path, -1) == '/') {
        // permanently redirect paths with a trailing slash
        // to their non-trailing counter part
        $uri = $uri->withPath(substr($path, 0, -1));
        return $response->withRedirect((string)$uri, 301);
    }

});
```

harikt:
Good work @akrabat . And I do agree with your thought about making it a middleware.

geekish:
I've been using the trailing slash middleware in [oscarotero/psr7-middlewares](//github.com/oscarotero/psr7-middlewares) for this reason. :)

designermonkey:
I'm really confused as to what's happened here. We added the functionality to allow a group's callback to be registered with `'/'` or `''`, yet now the former is broken and tries to match a trailing slash.

Is this therefore a bug?

designermonkey:
Reading this more, it is desired behaviour from the above list, but this breaks something that already worked.

If the plan is to add a middleware, can this be done sooner rather than later? Also, to use the middleware, this should now be an option in the settings to allow trailing slashes. I was never comfortable with the assumption that it was the preferred option. Also, I was sure there were issues around FastRoute and trailing slashes?

designermonkey:
Actually, the middleware alone doesn't solve this issue if you have been declaring routes within a group having the slash. I now have to spend time fixing this in my code, after spending time trying to debug what we thought was an error in our code.

I have to say (and I'll stop shortly), that this is now not consistent. A route should always start with a base left-hand slash, whether inside a group or not, so inside a group, this should mean 'the root of that group' not be appended to the end of that group pattern. If that is desired then the group pattern should be `/foo/` not `/foo`.

The samples above are too lenient for setting a standard. No route should ever be allowed to be registered without a left slash, or an empty string, it's just wrong, and we came to that conclusion before.

I have to say this has wound me up more than it should have. The code is in an RC which should be most stable, and a piece of core functionality has changed that other's code relies on.

Bad form.

tuupola:
Personally I think route groups feature is can of worms. It saves a few keystrokes but has been a big source of confusion already since Slim 2.

jaapverloop:
I agree with @tuupola

silentworks:
@designermonkey just to note RC1 was broken in the way it handled route group, when I upgraded my live app I noticed that the route mapping was failing due to this and this fix in RC2 fixed that issue, however due to regression we didn't notice the original issue came back about. We will be doing a RC3 to fix the original issue while fixing issues that were caused in RC1. Here is a code snippet below outlining what was wrong in RC1.

``` php
<?php

$app = new Slim\App();

$app->group('/', function () {
    $this->get('/', function($req, $res) {
        return $res->getBody()->write('Welcome Home.');
    });

    $this->get('/about', function($req, $res) {
        return $res->getBody()->write('About me.');
    });
});

$app->run();
```

If you visited `/` it would all work correctly, but if you try and visit `/about` you would notice it all failed. I tested this with nginx and also with php build in development server. The reason it failed is because the route for `about` is actually registered as `//about` which in turn makes it an absolute path rather than a relative path to webroot. In order to get this to work in RC1 we would have to make our routes in the group not start with a leading `/`.

Will fix all this for RC3 and ask for more feedback from everyone before making a release of RC3 as we want to go final after RC3.

designermonkey:
Re the middleware proposed. This does not work if you set Slim to determine the route early. This is why I was hoping for a solution like we had, where the trailing slash is automatically removed.

I appreciate all the work going in to this, but it kind of hits me that: An app will either allow trailing slashes, or not. So given that definition, we could always:
- Remove any double slashes `//` (As you spotted @silentworks)
- Based on a config option append or remove a trailing slash.

This needs to be simple, as has been stated by others, this is way too complex already. With that in mind, the tests outlined shouldn't care about trailing slashes, as a config app wide setting should deal with that.

silentworks:
@designermonkey move discussions over to #1597.

akrabat:
@designermonkey: I've been away, so am just getting back into trying to understand the problem that you are having.

> A route should always start with a base left-hand slash, whether inside a group or not, so inside a group, this should mean 'the root of that group' not be appended to the end of that group pattern. If that is desired then the group pattern should be /foo/ not /foo.

Current behaviour (RC2):

``` php
$app->group('/user', function () use ($app) {
        // matches /user/
    });
    $app->get('', function ($request, $response, $args) {
        // matches /user
    });
});

$app->group('/user/', function () use ($app) {
        // matches /user/
    });
    $app->get('', function ($request, $response, $args) {
        // matches /user
    });
});
```

i.e. the final `/` on a group is ignored and the URI path in the final `map` determines how the trailing `/` works.

If I've read your comments correctly, I think you want this behaviour if a trailing slash is required on a URI path:

``` php
$app->group('/user/', function () use ($app) {
        // matches /user/
    });
    $app->get('', function ($request, $response, $args) {
        // error condition
    });
});
```

and if you want to match `/user`, then you do:

``` php
$app->group('/user', function () use ($app) {
        // matches /user
    });
    $app->get('', function ($request, $response, $args) {
        // error condition
    });
});
```

i.e. the any double trailing `/` are removed from the final URI path, but any path definition must have at least a single `/` in it.

Is my understanding of your request correct?

(updated for typos!)

PR #1581, opened 2015-11-10
Make http methods case insensitive. Convert to uppercase

### LINKED ISSUE #1580: Slim3 map is case sensitive
Now map is case sensitive `$app->map(['GET', 'POST']` != `$app->map(['get', 'post']`
Maybe it could be insensitive?

dopesong:
Close #1580

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -134,6 +134,9 @@ class Router implements RouterInterface
 
         // Complete pattern must start with a /
         $pattern = '/' . ltrim($pattern, '/');
+        
+        // According to RFC methods are defined in uppercase (See RFC 7231)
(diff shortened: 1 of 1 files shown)


geggleto:
:+1: meh we discussed this in IRC it, I approve of keeping it simple.

silentworks:
Will get this merged as soon as we decide on what the next milestone will be.

akrabat:
If this is all we have, I'm minded to go with a 3.0 final. Thoughts?

geggleto:
I want to write up more stuff regarding the getIp() debacle. It actually broke my production environment, and I need to come up with a good way to do it.

PR #1582, opened 2015-11-12
Update determine content type method in error handler class

oanhnn:
As my [comment](https://github.com/slimphp/Slim/pull/1574#issuecomment-155992343) , this pull request includes:
- Change determineContentType to use `isset()` for best performance.
- Move determineContentType to a trait file


JoeBengalen:
Why the term `Aware` in the trait name?

oanhnn:
> Why the term `Aware` in the trait name?

The trait named `DetermineContentTypeAwareTrait` , like `MiddlewareAwareTrait` and `CallableResolverAwareTrait` :smile:

akrabat:
As per https://github.com/slimphp/Slim/pull/1574#issuecomment-153971977, I don't want determineContentType easily usable.

silentworks:
I am going to close this one out as we don't want to promote the use of this in your own app. I know you made comments to wildurand Negotiation but that will probably not be read by many.

edit. why you so quick @akrabat :)

akrabat:
😀

oanhnn:
:worried: 
Using trait is better for maintenance and using `isset()` for best performance.
You don't want but it is better for end-users, who don't always use middleware for this work.

akrabat:
I'm happy to take the isset() change as  its own PR.

Issue #1588, opened 2015-11-15
Request for changing visibility and structure of methods in Error.php, NotAllowed.php and NotFound.php


pokmot:
I love the new JSON and XML error displays, but for HTML I am triggering Smarty to display the error within our template.

As Error.php is using private methods, I have to copy and paste most of the code.

It would be great if I instead could extend the Error class and only override renderHtmlErrorMessage - but as this is a private function I can't. Would this be possible to change to protected instead?

NotAllowed.php and NotFound.php generates the output inside the switch statement instead of using functions like Error.php - it would be similarly great if the output was defined by a protected function rather inline statements, as these could easily be overridden by extending the class.

I do appreciate that this makes BC more of an issue in the future if these methods change. If this is an acceptable change I am happy to submit a PR for them.

RyanNerd:
Assuming you are using Slim V3  (Slim V2 is only accepting security updates)

You can already do this by overriding the error class via the Pimple container:

``` PHP
    // $sc = Slim DI Container (Pimple)
    $sc = $app->getContainer();
    unset($sc['errorHandler']);
    $sc['errorHandler'] = function() {new MyErrorHandler .... };    
```

pokmot:
Yes, I already do override the error handler. The issue here is that I currently need to recreate the whole class, and I would prefer to extend the existing class and just override a single method in the class to provide the output I need. That way I would benefit automatically from the framework improving the error handling class, whilst retaining the ability to customize the message.

silentworks:
@pokmot send separate PR's for the issues you mentioned above. One for making private methods protected and another for moving the inline errors to their own protected methods.

RyanNerd:
@pokmot I see now. Sorry I misunderstood your intent.

pokmot:
Hold on before merging - I just realised it may be very beneficial to have the Request object available to the rendering method.

silentworks:
@pokmot make that a separate PR, also I don't think these methods should take anything more than the exception as an argument. At the point where you need more flexibility, create your own Error handling class.

pokmot:
Agreed.

pokmot:
@silentworks Question - the request method is used to display the homeUrl in NotFound.php - what would be the best way to handle this?

Should we make the render message generic rather than including the homeUrl and then expect people to override it? I could easily override the __invoke, set a protected $request variable in my class, and read this in the render method, so that's easily handled.

pokmot:
[...]

PR #1593, opened 2015-11-16
Making private methods protected

pokmot:
Changing visibility of rendering methods to be protected rather than private, allowing the class to be extended for customized messages. Issue #1588


pokmot:
Before merging - I will also make the Request object available to the rendering method

pokmot:
As discussed with @silentworks no further changes are planned for this PR

PR #1595, opened 2015-11-16
Changed inline output of NotAllowed to use protected functions

pokmot:
This way the class can be extended and messages customized by overriding the method.

Also reintroduced 'default:', line 66, in the switch test as otherwise $output could be undefined when reaching line 74.

Discussed in issue #1588

### DIFF
--- a/Slim/Handlers/NotAllowed.php
+++ b/Slim/Handlers/NotAllowed.php
@@ -54,40 +54,18 @@ class NotAllowed
             $contentType = $this->determineContentType($request);
             switch ($contentType) {
                 case 'application/json':
-                    $output = '{"message":"Method not allowed. Must be one of: ' . $allow . '"}';
+                    $output = $this->renderJsonNotAllowedMessage($methods);
                     break;
 
                 case 'text/xml':
                 case 'application/xml':
-                    $output = "<root><message>Method not allowed. Must be one of: $allow</message></root>";
+                    $output = $this->renderXmlNotAllowedMessage($methods);
                     break;
 
                 case 'text/html':
+                default:
                     $contentType = 'text/html';
-                    $output = <<<END
-<html>
-    <head>
-        <title>Method not allowed</title>
-        <style>
-            body{
-                margin:0;
-                padding:30px;
-                font:12px/1.5 Helvetica,Arial,Verdana,sans-serif;
-            }
-            h1{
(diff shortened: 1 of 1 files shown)


akrabat:
👍

Slim/Handlers/NotAllowed.php
```diff
@@ -119,4 +97,80 @@ private function determineContentType(ServerRequestInterface $request)
...
         return 'text/html';
     }
+
+    /**
+     * Render PLAIN not allowed message
+     *
+     * @param  array                  $methods
+     * @return string
+     */
+    protected function renderPlainNotAllowedMessage($methods)
```
juliangut, line 105: What is the use of this method?

silentworks, line 105: Good spot, @pokmot can this be removed please, if you need it in your extended class add it there.

pokmot:
Apologies - a copy and paste gone wrong. It is used for the output in line 49 so plain text output also can be customized.

juliangut:
Another thing, @pokmot comments that default is needed so that $output is not undefined BUT as $contentType is determied by an array_intersect $contentType will always have one of `$this->knownContentTypes` or "text/html" anyway. I'm not against `default` on switch statements of course but I consider it unnecessary when the possible values are all controlled

pokmot:
$this->knownContentTypes is a protected property and may be overridden. In that case, there may be instances where $output is left undefined unless there is a default.

As I am not planning to change $this->knownContentTypes I am happy to remove the default, but I thought it better to plan for someone who forgets/doesn't understand and therefore preventing the framework producing a notice.

silentworks:
I think if someone was to change `knownContentTypes` to something not in the switch, it should fail badly, easier for the developer to debug than something magical happening without them knowing why.

juliangut:
Muy thoughts exactly

Slim/Handlers/NotAllowed.php
dopesong, line 105: But this just changes hardcoded output line: $output = 'Allowed methods: ' . $allow; Easer to extend plain message method than whole __invoke.

silentworks, line 105: Yeah part of the code was missing from the PR initially, but now its all there this method is fine.

PR #1597, opened 2015-11-17
Regression fix for Route Group

silentworks:
These are the cases covered:

Single routes:

| Route URI | Final pattern registered with router |
| --- | --- |
| `'/foo'` | `/foo` |
| `'/foo/'` | `/foo/` |
| `'foo'` | `/foo` |
| `'/'` | `/` |
| `''` | `/` |

Route groups:

| Group 1 | Group 2 (if exists) | Route URI | Final pattern registered with router |
| --- | --- | --- | --- |
| `'/foo'` |  | `'/bar'` | `/foo/bar` |
| `'/foo'` |  | `'/bar/'` | `/foo/bar/` |
| `'/foo'` |  | `'/'` | `/foo` |
| `'/foo'` |  | `''` | `/foo` |
| `'/foo'` | `'/baz'` | `'/'` | `/foo/baz` |
| `'/foo'` | `'/baz'` | `''` | `/foo/baz` |
| `'/foo'` | `'/baz'` | `'/bar'` | `/foo/baz/bar` |
| `'/foo'` | `'/baz'` | `'/bar/'` | `/foo/baz/bar/` |
| `'/foo'` | `'/'` | `'/bar'` | `/foo/bar` |
| `'/foo'` | `'/'` | `'bar'` | `/foo/bar` |
| `'/foo'` | `''` | `'/bar'` | `/foo/bar` |
| `'/foo'` | `''` | `'bar'` | `/foo/bar` |
| `'/'` |  | `'/bar'` | `/bar` |
| `'/'` |  | `'/bar/'` | `/bar/` |
| `'/'` |  | `'/'` | `/` |
| `'/'` |  | `''` | `/` |
| `'/'` | `'/baz'` | `'/'` | `/baz` |
| `'/'` | `'/baz'` | `''` | `/baz` |
| `'/'` | `'/baz'` | `'/bar'` | `/baz/bar` |
| `'/'` | `'/baz'` | `'/bar/'` | `/baz/bar/` |
| `'/'` | `'/'` | `'/bar'` | `/bar` |
| `'/'` | `'/'` | `'bar'` | `/bar` |
| `'/'` | `''` | `'/bar'` | `/bar` |
| `'/'` | `''` | `'bar'` | `/bar` |
| `''` |  | `'/bar'` | `/bar` |
| `''` |  | `'/bar/'` | `/bar/` |
| `''` |  | `'/'` | `/` |
| `''` |  | `''` | `/` |
| `''` | `'/baz'` | `'/'` | `/baz` |
| `''` | `'/baz'` | `''` | `/baz` |
| `''` | `'/baz'` | `'/bar'` | `/baz/bar` |
| `''` | `'/baz'` | `'/bar/'` | `/baz/bar/` |
| `''` | `'/'` | `'/bar'` | `/bar` |
| `''` | `'/'` | `'bar'` | `/bar` |
| `''` | `''` | `'/bar'` | `/bar` |
| `''` | `''` | `'bar'` | `/bar` |

Routes and Route Groups are expected to always start with a slash `/`.

Related to #1578 

Feedback welcome @akrabat @designermonkey @geggleto @dopesong

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -122,13 +122,12 @@ class Router implements RouterInterface
             throw new InvalidArgumentException('Route pattern must be a string');
         }
 
-        // pattern must start with a / if not empty
-        if ($pattern) {
-            $pattern = '/' . ltrim($pattern, '/');
-        }
-
         // Prepend parent group pattern(s)
         if ($this->routeGroups) {
+            // If any route in the group only has / we remove it
+            if ($pattern === '/') {
+                $pattern = '';
+            }
             $pattern = $this->processGroups() . $pattern;
         }
 
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -253,7 +253,7 @@ class AppTest extends \PHPUnit_Framework_TestCase
         /** @var \Slim\Router $router */
         $router = $app->router;
         $router->finalize();
-        $this->assertAttributeEquals('/foo/', 'pattern', $router->lookupRoute('route0'));
+        $this->assertAttributeEquals('/foo', 'pattern', $router->lookupRoute('route0'));
     }
 
     public function testGroupSegmentWithEmptyRoute()
@@ -283,7 +283,7 @@ class AppTest extends \PHPUnit_Framework_TestCase
         /** @var \Slim\Router $router */
         $router = $app->router;
         $router->finalize();
-        $this->assertAttributeEquals('/foo/baz/', 'pattern', $router->lookupRoute('route0'));
+        $this->assertAttributeEquals('/foo/baz', 'pattern', $router->lookupRoute('route0'));
     }
 
     public function testTwoGroupSegmentsWithAnEmptyRoute()
@@ -350,22 +350,6 @@ class AppTest extends \PHPUnit_Framework_TestCase
         $this->assertAttributeEquals('/foo/bar', 'pattern', $router->lookupRoute('route0'));
     }
 
-    public function testGroupSegmentWithSingleSlashGroupAndSegmentRouteWithoutLeadingSlash()
-    {
-        $app = new App();
-        $app->group('/foo', function () use ($app) {
-            $app->group('/', function () use ($app) {
-                $app->get('bar', function ($req, $res) {
-                    // Do something
-                });
-            });
-        });
-        /** @var \Slim\Router $router */
-        $router = $app->router;
-        $router->finalize();
-        $this->assertAttributeEquals('/foo/bar', 'pattern', $router->lookupRoute('route0'));
-    }
-
     public function testGroupSegmentWithEmptyNestedGroupAndSegmentRoute()
     {
         $app = new App();
@@ -382,20 +366,18 @@ class AppTest extends \PHPUnit_Framework_TestCase
         $this->assertAttributeEquals('/foo/bar', 'pattern', $router->lookupRoute('route0'));
     }
 
-    public function testGroupSegmentWithEmptyNestedGroupAndSegmentRouteWithoutLeadingSlash()
+    public function testGroupSegmentWithOptionalSegmentRouteWithoutLeadingSlash()
     {
         $app = new App();
         $app->group('/foo', function () use ($app) {
-            $app->group('', function () use ($app) {
-                $app->get('bar', function ($req, $res) {
-                    // Do something
-                });
+            $app->get('[/{bar:[0-9]+}]', function ($req, $res) {
+                // Do something
             });
         });
         /** @var \Slim\Router $router */
         $router = $app->router;
         $router->finalize();
-        $this->assertAttributeEquals('/foo/bar', 'pattern', $router->lookupRoute('route0'));
+        $this->assertAttributeEquals('/foo[/{bar:[0-9]+}]', 'pattern', $router->lookupRoute('route0'));
     }
 
     public function testGroupSingleSlashWithSegmentRouteThatDoesNotEndInASlash()
@@ -460,22 +442,6 @@ class AppTest extends \PHPUnit_Framework_TestCase
         $app->group('/', function () use ($app) {
             $app->group('/baz', function () use ($app) {
                 $app->get('/', function ($req, $res) {
-                    // Do something
-                });
-            });
-        });
-        /** @var \Slim\Router $router */
-        $router = $app->router;
-        $router->finalize();
-        $this->assertAttributeEquals('/baz/', 'pattern', $router->lookupRoute('route0'));
-    }
-
-    public function testGroupSingleSlashWithNestedGroupSegmentWithAnEmptyRoute()
-    {
-        $app = new App();
-        $app->group('/', function () use ($app) {
-            $app->group('/baz', function () use ($app) {
-                $app->get('', function ($req, $res) {
                     // Do something
                 });
             });
@@ -651,7 +617,7 @@ class AppTest extends \PHPUnit_Framework_TestCase
         /** @var \Slim\Router $router */
         $router = $app->router;
         $router->finalize();
-        $this->assertAttributeEquals('/baz/', 'pattern', $router->lookupRoute('route0'));
+        $this->assertAttributeEquals('/baz', 'pattern', $router->lookupRoute('route0'));
     }
 
     public function testEmptyGroupWithNestedGroupSegmentWithAnEmptyRoute()


geggleto:
:+1:

juliangut:
There is one question I've been wondering for a while, maybe it's obvious but I don't get it. Why keeping the last slash on routes? Can't it be made optional by default?
I find myself using square brackets for them all the time

``` php
$app->get('/users[/]', function (...) {});
```

silentworks:
@juliangut with this PR you don't need to, but if you should create your route with that last slash then I guess that is your intention.

Also look at the original issue if you want a middleware to remove trailing slash at all times.

dopesong:
Just tested with my broken app. Everything works fine now :+1:

akrabat:
This PR removes the ability to create a final pattern of /foo/ within the /foo group?

geggleto:
@akrabat is that not covered in the '/foo'      '/bar/' /foo/bar/scenario ?

akrabat:
@geggleto: /foo/bar/ is different to /foo/

akrabat:
This has broken:

``` php
$app->get('/users/', function ($req, $res, $args) {
    // ....
});
```

So is the table at the top wrong?

Update: ignore this - poor test on my part

silentworks:
@akrabat tested that `/users/` and thats working correctly. Do note the endpoint would be `/users/` and not `/users`.

akrabat:
ah yeah - I didn't simplify down enough for that one.

theodorejb:
Why would anyone make a group containing only a blank string or slash? What's the point of a group in that case?

geggleto:
@theodorejb  you can apply a middleware to the entire group.

theodorejb:
@geggleto Ah, that makes sense. Why not always require a slash, though, if it means the same thing as a blank string?

My typical use case is to have a group like `'/users'` with a routes inside like GET `'/{id}'` and POST `'/'`. Previously the POST route resolved to `'/users'`, but in RC2 it changed to `'/users/'` and I had to make the POST route a blank string for my API to continue working as expected. If this pull request makes both the blank string and slash mean the same thing, why even allow blank strings? I'd rather enforce consistency in how routes are defined.

geekish:
How is that any different than app-level middleware then?

theodorejb:
@silentworks The description for this PR says "Routes and Route Groups are expected to always start with a slash", but above that there are a bunch of examples of routes and route groups with a blank string. If a slash is expected, shouldn't routes and route groups that don't start with a slash produce an error?

silentworks:
@theodorejb its more of a best practise than a forced expectation.

@geekish A completely valid use case for this would be:

``` php
<?php

$app->get('/hello', function ($req, $res) {
    return $res->getBody()->write('Hello');
});

$app->group('/world', function () {
    // I want these validated
    $app->group('', function () {
        $app->get('/yo', function ($req, $res) {
            return $res->getBody()->write('Yo');
        });
        $app->get('/howdy', function ($req, $res) {
            return $res->getBody()->write('Howdy');
        });
    })->add(new AuthMiddleware());

    // I don't want these validated
    $app->group('/place', function () {
        $app->get('/hello', function ($req, $res) {
            return $res->getBody()->write('Hello');
        });
    });
});
```

The difference is I won't need to inspect the path in order to know which routes to apply the middleware to.

geggleto:
@theodorejb This PR fixes that break. It wont be available until RC3. You can switch to bleeding edge to get it immediately though.

@geekish In the case of a group being simply '/' there really is not any difference at all.

akrabat:
This is one of my test cases for an API end point:
- `/users/`      <= collection
- `/users/{id}` <= single resource

It's set up as:

``` php
$app->group('/users/', function () {
    $this->get('', 'UserCollection::list');
    $this->post('', 'UserCollection::add');
    $this->get('{id}', 'UserResource::load');
    $this->put('{id}', 'UserResource::update');
    $this->delete('{id}', 'UserResource::delete');
});
```

Is the intention that this will not be allowed and that the collection must be `/user`?

silentworks:
@akrabat that code doesn't work in this PR nor in RC2, you would have to do

``` php
$app->group('/users/', function () {
    $this->get('/', 'UserCollection::list');
    $this->post('/', 'UserCollection::add');
    $this->get('/{id}', 'UserResource::load');
    $this->put('/{id}', 'UserResource::update');
    $this->delete('/{id}', 'UserResource::delete');
});
```

for it to work in RC2

akrabat:
FWiW, I'm not against this PR, but I worry that we're introducing a limitation.

juliangut:
I have to say I'm a little confused about how this gets used in the end, I'm currently developing against RC2 and I'm trying to make my self up on how this PR would change my current routes, and it seems I'm not the only one.

@silentworks the trailing slash on the group route seems off to me, in my mind it will be resolved by this url for example: `GET .../users//10`, I have to work against my own mind to understand there are no double slashes. In your table above the groups doesn't have trailing slashes.

Wouldn't this work?

``` php
$app->group('/users', function () {
    $this->get('/{id}', 'UserResource::load');
});
```

On the other hand the use of an "empty" group to apply route middleware is really cool

designermonkey:
Instead of trying to cater for all of the things, why don't we lay down some ground rules instead.
1. All route definitions _must_ start with a slash.
2. All route definitions _must not_ end in a slash.
3. Trailing slashes are handled globally by a configuration choice (on|off).
4. Deviation from the above throws a relevant exception.

This covers single routes, route groups, nested groups et al. Instead of trying to cater for everyone, it's a basic obvious simple ruleset definition.

We simply should not be allowing any omission of a required slash from a definition. If there is the propensity for them to double up (like root + group) then we should just string replace doubles with singles.

tuupola:
What @designermonkey suggested sounds good.  Current situation feels messy.

dopesong:
+1 for @designermonkey suggestion

juliangut:
Thanks @designermonkey, basic rules is what I felt was needed here, +1

akrabat:
I have sites in production where collections end in / and resources don't. 

Edit: I don't particularly mind what happens in groups, though would prefer not to limit, but do need flexibility on final URL created.

geggleto:
I thought the original slim acted as a string concatinator for URIs before passing to FastRoute.

Why can't we just simply go back to that, and eliminate all the magic that we are trying to fix.

Let us assume that if you want to do something like:

```
group('/foo/' {
$get('')
$post('')
$put('')
});
```

You know what you are doing.

We should stay out of the whole battle between what is proper or not. It is pretty clear through these discussions that everyone has a different idea on what is "proper". 

This also has an added benefit of being easier to understand and document. Is it "purely perfect?" No, but who cares lets just make Slim easy to use.

My 2cents.

geekish:
What @designermonkey said, except I think that 1) yes routes should begin with a slash, 2) routes may end with or without a slash, but groups should NOT end in slashes, since they are prefixes. Trailing slashes should be handled via middleware. Just my opinion though.

theodorejb:
I have to agree with @geggleto on this. Too much magic makes the behavior more confusing and difficult to predict (not to mention error/bug prone). By default I think Slim should just concatenate the strings for nested groups and routes. Middleware could optionally be used to enforce conventions for different projects.

mathmarques:
I'm with @geggleto too. Creating rules for URI(or route) sounds like limiting something when we can just concatenate the route as strings. It's simple, intuitive and flexible.

pokmot:
@geggleto made a great point. One of the reasons I love Slim is that it allows me to do what I want, rather than forcing me to do it the way the framework wants.

akrabat:
My understanding of RC2 is that we concatenate, but remove double slashes. 

i.e:

``` php
$app->group('/bar/', function () use ($app) {
        // matches /bar/, not /bar//
    });

    $app->group('/foo/', function () use ($app) {
            // matches /bar/foo/, not /bar//foo//
        });
    });
});
```

Personally, I think this is reasonable as I can't think of a case when you would want to match double `/`.

theodorejb:
So do we want to keep the current behavior in RC2, then, rather than this pull request? Personally I am okay with this - I have already updated my apps to work with RC2 and it is working well.

designermonkey:
Well, I count it as a 4 to 3 vote for my ideas so far. And the pull request too makes it 5 in a way.

dopesong:
This case is broken:

``` php
$app->group('/v1/users/{id:[0-9]+}', function () use ($app) {
    $app->group('/notes', function () use ($app) {
        $app->get('[/{count:count}]', 'App\Modules\V1\Users\Notes\NotesController:getUserNotes');
        $app->post('/', 'App\Modules\V1\Users\Notes\NotesController:addUserNote');
   });
});
```

POST:
v1/users/1/notes - not working
v1/users/1/notes/ - working

Returns

Method not allowed. Must be one of: GET

codeguy:
Sorry for jumping in late on this thread. Can someone clearly list the use-cases that are _not_ already solved with the existing codebase (excluding this PR)?

codeguy:
I've talked with @akrabat about this. I've decided that we're going to make Slim as dumb as possible when it comes to group and route paths... this means we will concatenate the group and route paths exactly as they are defined by the end user. Most of this behavior is already in place. @akrabat is making a few small tweaks to further support this decision.

The side-effect of this decision is that we allow `$app->get('', ...)`. This looks funny, but it is predictable. I would prefer predictable over magic.

saltnfat:
@codeguy Does this mean trailing slashes are not explicitly supported unless defined by the user?  E.g. The two routes below are equivalent
/user
/user/

akrabat:
@jwyuen `/user` and `/user/` are two different routes. If you want them to be the same, you have two choices:
1. `$app->get('/user[/]', …`
2. Add this [middleware](http://www.slimframework.com/docs/cookbook/route-patterns.html) from the docs.

saltnfat:
@akrabat Ah I see.  Thanks!

PR #1599, opened 2015-11-19
Patch not found handler

pokmot:
This is my suggested solution for issue #1588 

With this, my class extending NotFound is very easy to write:

```
class NotFound extends \Slim\Handlers\NotFound
{
  protected function renderHtmlNotFoundOutput(ServerRequestInterface $request, ResponseInterface $response)
  {

    $contentType = 'text/html';
    return Container::get('view')->render($response, 'not-found.twig')
                         ->withStatus(404)
                         ->withHeader('Content-Type', $contentType);

  }
}
```

### DIFF
--- a/Slim/Handlers/NotAllowed.php
+++ b/Slim/Handlers/NotAllowed.php
@@ -61,7 +61,6 @@ class NotAllowed
                     break;
 
                 case 'text/html':
-                    $contentType = 'text/html';
                     $output = $this->renderHtmlNotAllowedMessage($methods);
                     break;
             }
--- a/Slim/Handlers/NotFound.php
+++ b/Slim/Handlers/NotFound.php
@@ -42,22 +42,93 @@ class NotFound
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response)
     {
-
-        $contentType = $this->determineContentType($request);
+       $contentType = $this->determineContentType($request);
         switch ($contentType) {
             case 'application/json':
-                $output = '{"message":"Not found"}';
-                break;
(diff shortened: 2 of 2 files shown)


Slim/Handlers/NotAllowed.php
```diff
@@ -61,7 +61,6 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
                     break;
 
                 case 'text/html':
-                    $contentType = 'text/html';
```
akrabat, line 64: Why have you removed this line?

akrabat:
Why set the status, header and body in each render method? I would have expected each render method to have returned a string and then have:

``` php
        $body = new Body(fopen('php://temp', 'r+'));
        $body->write($output);

        return $response->withStatus(404)
                        ->withHeader('Content-Type', $contentType)
                        ->withBody($body);
```

at the bottom of `__invoke`.

Slim/Handlers/NotAllowed.php
juliangut, line 64: why check $contentType in the switch and then reassign it to the same value?.

``` php
switch ($contentType) {
    case 'text/html':
        $contentType = 'text/html';
}
```

It was previously discussed in another issue that switch's default is not needed; that line was there because of that old default

juliangut:
I agree with @akrabat moving headers/body out of __invoke is unnecessary duplication

Slim/Handlers/NotAllowed.php
akrabat, line 64: Ah yes - $contentType is guaranteed to be one of the switch values.

Issue #1590, opened 2015-11-15
Issue with "index.php" appearing in Pretty Rewrite URLs


pokmot:
This can of course be resolved with middleware, however as this has an effect on potentially all sites using URL rewrites I thought I would raise it now.

If, when using URL rewrite, you access the site by including the PHP file name in the url (e.g. index.php), getBasePath includes the file name in the result. When using this as a base href="" tag, all my static files are now being routed through the PHP file:

```
In HTML:
<base href="{{ BasePath }}">
<link rel="stylesheet" href="css/style.css"/>

Sample:
$uri = \Slim\Http\Uri::createFromEnvironment(
    \Slim\Http\Environment::mock(
        [
            'SCRIPT_NAME' => '/foo/index.php',
            'REQUEST_URI' => '/foo/index.php',
        ])
);
print $uri->getBasePath();

// Output: /foo/index.php
// Asset path becomes: /foo/index.php/css/style.css **not working**


$uri = \Slim\Http\Uri::createFromEnvironment(
    \Slim\Http\Environment::mock(
        [
            'SCRIPT_NAME' => '/foo/index.php',
            'REQUEST_URI' => '/foo/',
        ])
);
print $uri->getBasePath();

// Output: /foo
// Asset path becomes: /foo/css/style.css **works great**
```

The issue is the comparison made on line 197 of Uri.php - as the request uri and the script name is the same, getBasePath will return it including the script name.

I do understand this is desired behavior for those not using URL rewriting, however for those that do, this is an issue.

My main issue is that the behavior is inconsistent and depends whether it's loading a script directly or using URL rewriting. If it always included the script name I would simply use dirname(). If it never included it I could easily add index.php where needed.

With this scenario I need to do the same comparison as done on line 197 in Uri.php, and if they match I use dirname to get the path on its own,

Any thoughts on how this is best resolved? Middleware or Slim code change? Or use a redirect in .htaccess?

pokmot:
I currently resolve this issue in the following way:

```
// Register Twig View helper
$container['view'] = function ($c) use ($app) {
  $view = new Twig(
      $app->settings['view']['template_path'], [
          'cache' => $app->settings['view']['twig']['cache'],
          'auto_reload' => $app->settings['view']['twig']['auto_reload'],
          'strict_variables' => $app->settings['view']['twig']['strict_variables'],
      ]
  );

  // We need to adjust the base URL a little when the script name is included
[...]

PR #1606, opened 2015-11-22
Corrected wrong getBasePath for issue #1590

pokmot:
Issue #1590 The following produces an incorrect output:

```
$uri = \Slim\Http\Uri::createFromEnvironment(
    \Slim\Http\Environment::mock(
        [
            'SCRIPT_NAME' => '/foo/index.php',
            'REQUEST_URI' => '/foo/index.phpXYZ',
        ])
);
print $uri->getBasePath();

// Produces /foo/index.php - correct output should be /foo/    
```

The issue is the use of stripos on line 197 in Uri.php - as long as you include the file name of the currently executing script you can add anything you like - and the file is returned.

The change is using strcasecmp instead of stripos which produces correct output

```
if (strcasecmp($requestUri, $requestScriptName) === 0) {
```

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -194,7 +194,7 @@ class Uri implements UriInterface
         $requestUri = parse_url($env->get('REQUEST_URI'), PHP_URL_PATH);
         $basePath = '';
         $virtualPath = $requestUri;
-        if (stripos($requestUri, $requestScriptName) === 0) {
+        if (strcasecmp($requestUri, $requestScriptName) === 0) {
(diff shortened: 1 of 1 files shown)


akrabat:
Are you able to add a test to prove that this changes solves it?

codeguy:
This looks good, but I also want a test to prove the changes.

akrabat:
I've added a [unit test](https://github.com/slimphp/Slim/blob/3.x/tests/Http/UriTest.php#L618-L633).

codeguy:
Thanks!

PR #1618, opened 2015-11-25
Add the method add to RouteInterface and RouteGroupInterface

mathmarques:
A class that implements RouteInterface or RouteGroupInterface must have the method add() to prepend middlewares


mathmarques:
I think RouteInterface should also declare the add method.
Should I use this PR to change it too?

PR #1620, opened 2015-11-26
Helper methods for request params

dopesong:
Added helper methods to fetch param value from just request body or query string.


akrabat:
As per our conversation on IRC, I'm happy to consider these for 3.1.

Issue #1622, opened 2015-11-26
Route Middleware stack


mathmarques:
Suppose we have this:

``` php
$app->add(new \App\Middleware\MidTest('App Level Middleware1'));
$app->add(new \App\Middleware\MidTest('App Level Middleware2'));

$app->group('/foo', function() {

    $this->group('/bar', function() {

        $this->get('/', function () {

            return $this->response->getBody()->write("Page inside Group /foo Group /bar Get /! <br>");

        })->add(new \App\Middleware\MidTest('Get / Middleware 1'))->add(new \App\Middleware\MidTest('Get / Middleware 2'));

    })->add(new \App\Middleware\MidTest('Group /bar Middleware 1'))->add(new \App\Middleware\MidTest('Group /bar Middleware 2'));

})->add(new \App\Middleware\MidTest('Group /foo Middleware 1'))->add(new \App\Middleware\MidTest('Group /foo Middleware 2'));
```

App Level Middlewares are always executed first, since they are on app level. Ok.

Group /foo Middlewares will execute before Group /bar Middlewares. Ok.

The question:
Why Get / Middlewares will execute before Group /foo Middlewares ?
Shouldn't be executed after the Group /bar Middlewares?

I was expecting this order:

```
App Level Middleware2
App Level Middleware1
Group /foo Middleware 2
Group /foo Middleware 1
Group /bar Middleware 2
Group /bar Middleware 1
Get / Middleware 2
Get / Middleware 1
Page inside Group /foo Group /bar Get /!
```

But I got:

```
App Level Middleware2 
App Level Middleware1
Get / Middleware 2
Get / Middleware 1
Group /foo Middleware 2
Group /foo Middleware 1
Group /bar Middleware 2
Group /bar Middleware 1
Page inside Group /foo Group /bar Get /! 
```

I fixed this changing the \Slim\Route->finalize() to:

``` php
public function finalize()
    {
        $groupMiddleware = [];
        foreach ($this->getGroups() as $group) {
            foreach ($group->getMiddleware() as $middleware) {
                array_unshift($groupMiddleware, $middleware);
            }
        }

        $this->middleware = array_merge($this->middleware, $groupMiddleware);

        foreach ($this->getMiddleware() as $middleware) {
            $this->addMiddleware($middleware);
        }
    }
```

If this is a bug(what I think it is), I can make PR.
Commits here: https://github.com/mathmarques/Slim/commits/fix-route-middleware

geggleto:
[...]

PR #1625, opened 2015-11-28
Fix route middleware stacks

mathmarques:
See https://github.com/slimphp/Slim/issues/1622

Removed the array_unshift from RouteGroup->add(), since Route->add() don't use it. Consistency on middleware arrays.


codeguy:
Does this work for you? Can you add a unit test for this change?

mathmarques:
Yes, works fine. I'll try to add some tests 'cause I'm not that good on phpunit stuffs...

mathmarques:
Added tests

codeguy:
Thanks!

Issue #1627, opened 2015-11-28
Question on \Slim\Router->finalize()


mathmarques:
Why Slim needs to call all Routes->finalize() via \Slim\Router->finalize()?
I tried to look for a reason but I did not find.

My question is because I'm trying to work around this issue:
https://github.com/slimphp/Slim/issues/1604

Why not 'finalize' only the dispatched Route?

codeguy:
I think this is a fair assessment. I don't see why we have to finalize all routes instead of the one matching route. However, I don't think this is paramount for the 3.0.0 RTM release. Will move to the 3.1.0 milestone.

codeguy:
There are some edge cases such as sub requests, etc. We'd want to save the "finalized" state as a member variable in each route instance, and then finalize routes lazily on invocation. But a problem to be addressed later in 3.1.0 for sure.

mathmarques:
I thought about sub requests and I moved the finalized state to Route class on my "fix" for Lazy resolving of route middlewares:
https://github.com/mathmarques/Slim/commit/99a85f9f1fbabfde730e160898da0024e56c7bb3

@codeguy Is it too early for create a PR for change this finalize behavior? :)

codeguy:
Not too early. Just be prepared to potentially rebase your PR in the future.

codeguy:
If this is running successfully, we may as well just merge this into RC3 /cc @akrabat

silentworks:
@codeguy can we get this tested on an app that makes heavy use of middleware's first, both `Route` and `App` level middleware.

codeguy:
@silentworks For sure.

mathmarques:
I just need to work on removing the \Slim\Router->finalize() (A lot os unit tests use this function and I just remove the content of it)

I'll make the changes and open a PR.

akrabat:
Fixed by #1632.

PR #1632, opened 2015-11-28
Lazy \Slim\Route->finalize()

mathmarques:
Issue: https://github.com/slimphp/Slim/issues/1627

P.S: This do not make route middleware resolvable lazy, but with this we can easily do it :)

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -112,20 +112,22 @@ class Route extends Routable implements RouteInterface
      */
     public function finalize()
     {
-        if (!$this->finalized) {
-            $groupMiddleware = [];
-            foreach ($this->getGroups() as $group) {
-                $groupMiddleware = array_merge($group->getMiddleware(), $groupMiddleware);
-            }
-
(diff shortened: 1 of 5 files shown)


akrabat:
Looks okay to me. Needs testing by @silentworks - assuming he has an app that uses a lot of middleware.

Slim/Route.php
```diff
@@ -110,15 +112,19 @@ public function add($callable)
      */
     public function finalize()
     {
-        $groupMiddleware = [];
-        foreach ($this->getGroups() as $group) {
-            $groupMiddleware = array_merge($group->getMiddleware(), $groupMiddleware);
-        }
+        if (!$this->finalized) {
```
akrabat: Please change this to:

``` php
if ($this->finalized) {
    return;
}
```

rather than indenting the entire method

akrabat:
I've done more testing locally and can't break it. Taking for RC3 as it doesn't actually do the lazy resolving so is less risky.

PR #1642, opened 2015-12-03
Added ServerRequestInterface and ResponseInterface for custom functions in Response and Request class.

irfanevrens:
Extended functionality is become more clear with new interfaces.


akrabat:
We are more likely to solve this by decoration of the PSR-7 objects with Slim ones in the same way that Stratigility does.

irfanevrens:
There is no perfect decoration and will not indeed. PSR-7 learn it's decoration rules from php ecosystem, such as Slim.

PR #1645, opened 2015-12-04
Add proper tags to container docblock

geggleto:
Add property read's to the container doc block


akrabat:
A standard: http://www.phpdoc.org/docs/latest/references/phpdoc/tags/property-read.html

codeguy:
Nice. Merging in.

PR #1653, opened 2015-12-06
Fixed some typos.

irfanevrens:


Slim/Http/Headers.php
```diff
@@ -9,7 +9,6 @@
 namespace Slim\Http;
 
 use Slim\Collection;
-use Slim\Http\Environment;
```
akrabat, line 12: We use this for type hinting on line 51 of this file:

``` php
public static function createFromEnvironment(Environment $environment)
```

akrabat:
You delete `Slim/Interfaces/ContainerInterface.php` however, I can't find this file in https://github.com/slimphp/Slim/tree/3.x/Slim/Interfaces ?

Slim/Http/Headers.php
irfanevrens, line 12: Environment and Headers are in the same namespace, Slim\Http.

akrabat, line 12: Doh! So they are :)

irfanevrens:
Added on 43b693480028d48f387a620a34a9cafdfebcc767 then removed on 02b629c63632c02c7b0a44b340130bd85791f4eb. It was my local file.

akrabat:
Thanks!

PR #1654, opened 2015-12-06
Added container interface for "@property-read"

irfanevrens:


ppetermann:
Don't!

this patch doesn't improve anything, but means you can't just plugin interop- compliant containers.

irfanevrens:
Used Interop Container Interface many times in code base. This is not good for "code completion". If we use our extended Interface we can set properties for "code completion" on it.

akrabat:
We would need to implement our own Container decoration around the user supplied interop-container before we could type-hint the main codebase against a Slim specific interface.

This isn't going to happen for 3.0.0.

ppetermann:
Decorating would work, but it would mean to teach users to access the mandatory slim stuff in a different way then everything else in the container. 

IMHO the @property-read hints shouldn't have been for the container in the first place, the correct way to access those values is

``` php

/* @var \Slim\Interfaces\RouterInterface $router */
$router = $container->get('router');

```

as this is using the interop interface and gives you completion on $router.

silentworks:
I agree with @ppetermann, but at the same time we need to keep the ease of use for the end user. Some of this stuff will probably be better handled in the `slim-skeleton` as I think pushing all of this into Slim core is going to get messy.

PR #1655, opened 2015-12-06
Tidy up method handling

### LINKED ISSUE #1637: Slim3: XML _METHOD override throws InvalidArgumentException
Given the test XML:

``` xml
<test><person><name>Josh</name></person><_METHOD>HEAD</_METHOD></test>
```

I would expect Slim's method override (in Slim\Http\Request) to change the request type from the original to "HEAD". However, passing that string in as the request body (with an application/xml;charset=utf8 content-type) results in an error:

```
InvalidArgumentException: Unsupported HTTP method; must be a string, received SimpleXMLElement

/vagrant/..../vendor/slim/slim/Slim/Http/Request.php:308
/vagrant/..../vendor/slim/slim/Slim/Http/Request.php:243
```

A "quick fix" for this seems to be changing \Slim\Http\Request::getMethod

``` php
$this->method = $this->filterMethod($body->_METHOD);
```

to

``` php
$this->method = $this->filterMethod((string)$body->_METHOD);
```

I have not provided a PR as I'm not 100% confident that is how you are expecting people to override the request method in XML and I'm not confident the fix is the best solution (I dislike typecasting unless really really necessary) [If I get time at the weekend, I'll see if I can do something though]

[PHP7, composer installed]

akrabat:
A couple of bug fixes:
- Ensure that the _METHOD key is always a string so that it works with XML.
- Ensure that Request's `withMethod()` works as the user expects(!)

Fixes #1637

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -240,9 +240,9 @@ class Request extends Message implements ServerRequestInterface
                 $body = $this->getParsedBody();
 
                 if (is_object($body) && property_exists($body, '_METHOD')) {
-                    $this->method = $this->filterMethod($body->_METHOD);
+                    $this->method = $this->filterMethod((string)$body->_METHOD);
                 } elseif (is_array($body) && isset($body['_METHOD'])) {
-                    $this->method = $this->filterMethod($body['_METHOD']);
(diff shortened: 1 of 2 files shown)


geggleto:
Interesting... :)  Good catch!

PR #1663, opened 2015-12-09
ContainerInterop compliance improvement

Sam-Burns:
Strict adherence to ContainerInterop standard w.r.t. exceptions

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -8,6 +8,7 @@
  */
 namespace Slim;
 
+use Doctrine\Instantiator\Exception\InvalidArgumentException;
 use Interop\Container\ContainerInterface;
 use Interop\Container\Exception\ContainerException;
 use Pimple\Container as PimpleContainer;
@@ -266,9 +267,28 @@ final class Container extends PimpleContainer implements ContainerInterface
         }
         try {
             return $this->offsetGet($id);
-        } catch (\Exception $exception) {
-            throw new SlimContainerException(sprintf('Container error while retrieving "%s"', $id), null, $exception);
+        } catch (\InvalidArgumentException $exception) {
+            if ($this->exceptionThrownByContainer($exception)) {
+                throw new SlimContainerException(
+                    sprintf('Container error while retrieving "%s"', $id),
+                    null,
+                    $exception
+                );
+            }
         }
+    }
+
+    /**
+     * Tests whether an exception needs to be recast for compliance with Container-Interop.  This will be if the
+     * exception was thrown by Pimple.
+     *
+     * @param \InvalidArgumentException $exception
+     *
+     * @return bool
+     */
+    private function exceptionThrownByContainer(\InvalidArgumentException $exception)
+    {
+        return preg_match('/^Identifier ".*" is not defined.$/', $exception->getMessage());
     }
 
     /**
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -8,7 +8,6 @@
  */
 namespace Slim;
 
-use Doctrine\Instantiator\Exception\InvalidArgumentException;
 use Interop\Container\ContainerInterface;
 use Interop\Container\Exception\ContainerException;
 use Pimple\Container as PimpleContainer;
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -14,6 +14,7 @@ use Pimple\Container as PimpleContainer;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Slim\Exception\ContainerValueNotFoundException;
+use Slim\Exception\ContainerException as SlimContainerException;
 use Slim\Handlers\Error;
 use Slim\Handlers\NotFound;
 use Slim\Handlers\NotAllowed;
@@ -263,7 +264,11 @@ final class Container extends PimpleContainer implements ContainerInterface
         if (!$this->offsetExists($id)) {
             throw new ContainerValueNotFoundException(sprintf('Identifier "%s" is not defined.', $id));
         }
-        return $this->offsetGet($id);
+        try {
+            return $this->offsetGet($id);
+        } catch (\Exception $exception) {
+            throw new SlimContainerException(sprintf('Container error while retrieving "%s"', $id), null, $exception);
+        }
     }
 
     /**
--- a/Slim/Exception/ContainerException.php
+++ b/Slim/Exception/ContainerException.php
@@ -1,0 +1,20 @@
+<?php
+/**
+ * Slim Framework (http://slimframework.com)
+ *
+ * @link      https://github.com/codeguy/Slim
+ * @copyright Copyright (c) 2011-2015 Josh Lockhart
+ * @license   https://github.com/codeguy/Slim/blob/master/LICENSE (MIT License)
+ */
+namespace Slim\Exception;
+
+use RuntimeException;
+use Interop\Container\Exception\ContainerException as InteropContainerException;
+
+/**
+ * Container Exception
+ */
+class ContainerException extends RuntimeException implements InteropContainerException
+{
+
+}
--- a/Slim/Exception/ContainerException.php
+++ b/Slim/Exception/ContainerException.php
@@ -8,13 +8,13 @@
  */
 namespace Slim\Exception;
 
-use RuntimeException;
+use InvalidArgumentException;
 use Interop\Container\Exception\ContainerException as InteropContainerException;
 
 /**
  * Container Exception
  */
-class ContainerException extends RuntimeException implements InteropContainerException
+class ContainerException extends InvalidArgumentException implements InteropContainerException
 {
 
 }
--- a/tests/ContainerTest.php
+++ b/tests/ContainerTest.php
@@ -139,4 +139,24 @@ class ContainerTest extends \PHPUnit_Framework_TestCase
         $c->get('settings')['httpVersion'] = '1.2';
         $this->assertSame('1.2', $c->get('settings')['httpVersion']);
     }
+
+    public function testGetMagicMethod()
+    {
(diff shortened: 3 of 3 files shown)


Sam-Burns:
Referring to the [ContainerInterop specification on exceptions](https://github.com/container-interop/container-interop/blob/master/docs/ContainerInterface.md#12-exceptions)

Slim/Container.php
```diff
@@ -263,7 +264,11 @@ public function get($id)
         if (!$this->offsetExists($id)) {
             throw new ContainerValueNotFoundException(sprintf('Identifier "%s" is not defined.', $id));
         }
-        return $this->offsetGet($id);
+        try {
```
geggleto, line 267: The containers job is to retrieve the value and that's it. If the value throws an exception it's not the containers responsibility to deal with that.

silentworks:
I think this part of the spec is a bit confusing, because the second sentence seems to call for the enforcement of the first sentence. 

Not sure why you would need both or why the container should be handling this part of exception handling rather than the end user doing it themselves.

//cc @mnapoli

Sam-Burns:
The [docblocks](https://github.com/slimphp/Slim/blob/fb0e176a2ff6cea2fce0a946d3c60683e6d834c1/Slim/Container.php#L257) copied into Slim from the Interop `ContainerInterface` already suggest we are throwing an exception which complies with the standard, but none exists in the repo.

The intention behind having a consistent type of `ContainerException` is to allow the user clarity that it was specifically their DI config that caused the problem.

The PR goes further, adding the service ID to the message and making the previous exception accessible via `getPrevious()` - allowing easier debugging.

mnapoli:
Hi!

Here is the spec:

> Exceptions directly thrown by the container MUST implement the Interop\Container\Exception\ContainerException.
> 
> A call to the get method with a non-existing id SHOULD throw a Interop\Container\Exception\NotFoundException.

Exceptions thrown by factories (i.e. closures) are not thrown by the container, so I don't see a reason to catch them and rethrow them as `ContainerException`.

> I think this part of the spec is a bit confusing, because the second sentence seems to call for the enforcement of the first sentence.

I'm not sure I understand what you mean. If you are saying that the second sentence conflicts with the first, I don't think it's the case because `NotFound` extends `ContainerException`.

silentworks:
@mnapoli thanks for clearing that up. I didn't realise `NotFound` extended `ContainerException` hence why I found it a bit confusing.

geggleto:
> Exceptions thrown by factories (i.e. closures) are not thrown by the container, so I don't see a reason to catch them and rethrow them as ContainerException.

@silentworks @mnapoli 

https://github.com/slimphp/Slim/pull/1663/files#diff-1fe12fa9ad45fa920bf98a8afcf0939eR270

This catches and rethrows the exceptions. I think we should remove this, thoughts?

Sam-Burns:
I've rewritten it to take this into account.  Exceptions thrown explicitly by the container are now recast, but exceptions thrown in a factory closure are not.  This is captured by the unit tests added.

The only thing that can actually go wrong while retrieving a dependency (where `has()` would return `true`) is if there is a problem with the DI configuration.  At present, if a dependency factory closure tries to obtain another, non-existent, dependency from the container, it throws an `\InvalidArgumentException`.  The spec says that if the container explicitly throws an exception, it has to be an implementation of `ContainerException`, so something has to be done here.

Slim/Container.php
```diff
@@ -8,12 +8,14 @@
  */
 namespace Slim;
 
+use Doctrine\Instantiator\Exception\InvalidArgumentException;
```
akrabat: This is a dependency we would not want.

Slim/Container.php
```diff
@@ -263,7 +265,30 @@ public function get($id)
         if (!$this->offsetExists($id)) {
             throw new ContainerValueNotFoundException(sprintf('Identifier "%s" is not defined.', $id));
         }
-        return $this->offsetGet($id);
+        try {
+            return $this->offsetGet($id);
+        } catch (\InvalidArgumentException $exception) {
+            if ($this->exceptionThrownByContainer($exception)) {
+                throw new SlimContainerException(
```
akrabat, line 271: This is a BC break. Currently, calling code can expect to catch an `\InvalidArgumentException`, but this code changes that to a `\RuntimeException`.

Sam-Burns, line 271: @akrabat The docblock says it is a `Interop\Container\Exception\ContainerException`.  What would your suggestion be?  Would you prefer to see `Slim\Exception\ContainerException` extend `\InvalidArgumentException` to preserve backwards compatibility?  I can change it if so.

mnapoli, line 271: How about `Slim\Exception\ContainerEntryNotFoundException extends \InvalidArgumentException implements Interop\Container\NotFoundException` ? It's a "not found" kind of exception so it should extend `NotFoundException`.

silentworks, line 271: But we already have a `ContainerValueNotFoundException` that extends `Interop\Container\NotFoundException` on line 266.

mnapoli, line 271: Oh well then that's it!

akrabat, line 271: > @akrabat The docblock says it is a Interop\Container\Exception\ContainerException. What would your suggestion be? Would you prefer to see Slim\Exception\ContainerException extend \InvalidArgumentException to preserve backwards compatibility? I can change it if so.

In this particular case, the only thing I especially care about is not changing anyone's code which calls get() within a try/catch block.

Sam-Burns:
@akrabat I have changed the new exception class to preserve BC as discussed

silentworks:
I am still not seeing the value in this, it feels like its been over-engineered for something that the developer should deal with themselves. Wouldn't this example below solve this issue?

``` php
$container = new Container;
$container['foo'] = function (ContainerInterface $container) {
    try {
        return $container['doesnt-exist'];
    } catch (\Exception $e) {
        // deal with it here
    }
};
$container->get('foo');
```

Sam-Burns:
@silentworks The issue the class is trying to fix is that things like [this line](https://github.com/silexphp/Pimple/blob/v3.0.2/src/Pimple/Container.php#L96) in Pimple aren't compliant with the Container-Interop standard.  Wouldn't it be reasonable for a user to expect to be able to catch `Interop\Container\Exception\ContainerException` and handle or log that differently from other exceptions, if we're saying the container complies with the standard?

silentworks:
Sorry for going back and forth on this one @Sam-Burns but we want to get to the best possible solution on this, I wonder if we should do as suggested on https://github.com/zendframework/zend-expressive/issues/129#issuecomment-157516173 until the actual Pimple team decides to adapt container interop.

akrabat:
Where are we on this one?

vlakoff:
One step forward from _Slim_ to [_Enterprise software_](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) (pun intended).

vlakoff:
Seriously, IMHO, don't put band-aids like this just to forcibly adhere to some standard. It's a mess, just use Pimple as it is, it will adhere to the standard when time comes.

mnapoli:
@vlakoff:
- Slim already supports container-interop, that's not the question here
- this issue is about minor improvements about how the standard is implemented
- if you want to use Pimple, you already can
- try to be more constructive and respectful of other's work

vlakoff:
I just say catching exceptions and _trying_ to identify some of these to redefine them is not a good solution, IMHO.

I have nothing against this standard. It's just not implemented upstream (i.e. Pimple), so let's deal with it for now.

Slim/Container.php
```diff
@@ -263,7 +264,30 @@ public function get($id)
         if (!$this->offsetExists($id)) {
             throw new ContainerValueNotFoundException(sprintf('Identifier "%s" is not defined.', $id));
         }
-        return $this->offsetGet($id);
+        try {
+            return $this->offsetGet($id);
+        } catch (\InvalidArgumentException $exception) {
+            if ($this->exceptionThrownByContainer($exception)) {
```
vlakoff, line 270: If `(! $this->exceptionThrownByContainer($exception))` the exception is silently discarded, it should be re-thrown.

akrabat:
Merged in 70c6d6d4c1b8c8d9beb788db438ee7442c2762cb.

Thanks for this work; I'm sorry it took so long to get in.

Slim/Container.php
```diff
@@ -263,7 +264,30 @@ public function get($id)
...
+     * Tests whether an exception needs to be recast for compliance with Container-Interop.  This will be if the
+     * exception was thrown by Pimple.
+     *
+     * @param \InvalidArgumentException $exception
+     *
+     * @return bool
+     */
+    private function exceptionThrownByContainer(\InvalidArgumentException $exception)
+    {
+        return preg_match('/^Identifier ".*" is not defined.$/', $exception->getMessage());
```
vlakoff, line 290: (nitpick) Final dot should be escaped `\.`

akrabat, line 290: Good catch. Want to create a PR?

vlakoff:
As I noted above, to not potentially lose the exception:

``` php
try {
    return $this->offsetGet($id);
} catch (\InvalidArgumentException $exception) {
    if ($this->exceptionThrownByContainer($exception)) {
        throw new SlimContainerException(...);
    }
    throw $exception; <-- this should be added
}
```

akrabat:
Also good point. Not sure how I missed it. Please PR along with the escaped `.`

Slim/Container.php
vlakoff: I think that would be overkill.

vlakoff:
Maybe the trace could be used to make the detection robuster?

As a quick example:

```
$trace = $exception->getTrace();
$trace[0]['class'] === 'Pimple\Container';
$trace[0]['function'] === 'offsetGet';
```

PR #1664, opened 2015-12-09
Improved CallableResolver's flow.

irfanevrens:


silentworks:
What makes this an improvement over what was already there?

irfanevrens:
I simplified function body and deleted some "else statements" and separate logics. I do not know why coverage decreased.

akrabat:
Removing the else statements is nice, but I don't want it split into two methods. Please refactor so that it is back in the single `resolve` method.

silentworks:
This is not an issue at the moment as this change doesn't give any performance increase or the likes and it will be refactored in next major version.

PR #1671, opened 2015-12-11
add .editorconfig file to make sure editors works as intended

nmhnmh:
see more on http://editorconfig.org/


silentworks:
This would only be beneficial to contributors and not to the user using the end product. I am inclined to thinking that so far all contributors are following PSR2 coding standards and we won't benefit much from this.

PR #1676, opened 2015-12-15
Exclude files and folder from dist

feryardiant:
Described here [https://redd.it/2jzp6k](https://redd.it/2jzp6k)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# Enforce Unix newlines
+* text=lf
+
+# Exclude unused files
+# see: https://redd.it/2jzp6k
+/tests              export-ignore
+/.codeclimate.yml   export-ignore
+/.coveralls.yml     export-ignore
+/.gitattributes     export-ignore
+/.gitignore         export-ignore
+/.travis.yml        export-ignore
+/phpunit.xml.dist   export-ignore


.gitattributes
```diff
@@ -0,0 +1,12 @@
+# Enforce Unix newlines
+* text=lf
```
akrabat, line 2: Do we want this @codeguy? You're the resident Windows developer, so may have an opinion?

codeguy, line 2: Will look at this later today.

tuupola, line 2: Unix newlines are in PSR-2 standard which Slim says it follows. Everybody should use the same newlines. Otherwise diff will be borken.

codeguy, line 2: What @tuupola said. PSR-2 requires Unix LF line returns. So I give this a big :+1:

geggleto, line 2: ;( I'm on windows too.

feryardiant:
Thanks for merging.

Anyway, may I ask. Are you guys has plan to implement it for another repo under slimphp org?

.gitattributes
codeguy, line 2: Git translates line returns on pull and commit. So it's not really a big deal, right?

silentworks:
@feryardiant if you see any other repo that would benefit from this, send a PR.

feryardiant:
Got it :+1:

PR #1684, opened 2015-12-17
runtime exception when responding with json fails to encode

zzgab:
Sometimes you need to respond `->withJson($data)` where `$data` is obtained from a source (database for example) containing characters that he PHP json lib fails to encode.

For example, the **é** symbol is very basic in French (_liberté, égalité, fraternité_ ;) ), and legacy data sources may contain this letter in the form of `chr(233)` character code. (This is not the case if the data source is utf8-encoded of course).

`json_encode` silently fails when the `$data` contains this char code (as well as many other diacritics). The result is an empty string in the `Response` body.

I thought that it is important to raise an exception when this occurs.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -301,7 +301,7 @@ class Response extends Message implements ResponseInterface
         $body->write(json_encode($data, $encodingOptions));
 
         // Ensure that the json encoding passed successfully
-        if( ($jsonLastError = json_last_error()) != JSON_ERROR_NONE ) {
+        if (($jsonLastError = json_last_error()) != JSON_ERROR_NONE) {
             throw new \RuntimeException(json_last_error_msg(), $jsonLastError);
         }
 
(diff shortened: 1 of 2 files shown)


Slim/Http/Response.php
```diff
@@ -299,6 +300,11 @@ public function withJson($data, $status = 200, $encodingOptions = 0)
         $body->rewind();
         $body->write(json_encode($data, $encodingOptions));
 
+        // Ensure that the json encoding passed successfully
+        if (($jsonLastError = json_last_error()) != JSON_ERROR_NONE) {
```
AndrewCarterUK: This code could be a bit faster.

`json_encode` returns false on failure (so you know if there's an error without calling `json_last_error` every time).

I'd try something like

``` php
$body->rewind();

$json = json_encode($data, $encodingOptions);

if (false === $json) {
    throw new \RuntimeException(json_last_error_msg(), json_last_error());
}

$body->write($json);
```

zzgab: Sure, I see your point. Thank you. PR modified.

zzgab:
Thanks

PR #1685, opened 2015-12-17
Code Coverage

geggleto:
- Increase code coverage to 100% for ContainerTest.php
- Increase speed of tests by using phpunit setUp instead of creating a bunch of containers per test case


geggleto:
hmm.

PR #1688, opened 2015-12-18
Added App::process() to expose Request -> Response functionality

AndrewCarterUK:
### The Problem

It's currently very difficult to give the application a request and retrieve a response. To have a request that isn't created from the PHP super globals you need to patch the container.

A project of mine ([PHPFastCGI](http://github.com/PHPFastCGI/FastCGIDaemon)) has struggled to provide a clean adapter to the Slim framework for this reason.
### The Solution

Split `App::run()` into two methods by creating `App::process()`. The latter acts as a simple map from request to response. 
### Downsides

The are two that I can think of:
1. It adds a new method to the public API that needs to be maintained and supported.
2. Confusion over the difference between `App:process()` and `App::__invoke()`. Both have the same signature and do similar things, but ultimately at a different level.

---
### Questions

Assuming that people think this is a good idea:
- How would you like me to test this? I haven't really added any features. Any test of `App::run()` now covers `App:process()`. I'm not adverse to adding unit tests, I just thought it was worth asking here what people think they should look like rather than just adding something pointless for the sake of it.
- Would you like me to add documentation (and where)?

Interested to hear thoughts :)

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -297,12 +297,38 @@ class App
         $request = $this->container->get('request');
         $response = $this->container->get('response');
 
+        $response = $this->process($request, $response);
+
+        $response = $this->finalize($response);
+
+        if (!$silent) {
+            $this->respond($response);
+        }
+
+        return $response;
+    }
+
+    /**
+     * Process a request
+     *
+     * This method traverses the application middleware stack and then returns the
+     * resultant Response object.
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return ResponseInterface
+     *
+     * @throws Exception
+     * @throws MethodNotAllowedException
+     * @throws NotFoundException
+     */
+    public function process(ServerRequestInterface $request, ResponseInterface $response)
+    {
         // Ensure basePath is set
         $router = $this->container->get('router');
         if (is_callable([$request->getUri(), 'getBasePath']) && is_callable([$router, 'setBasePath'])) {
             $router->setBasePath($request->getUri()->getBasePath());
         }
-
 
         // Dispatch the Router first if the setting for this is on
         if ($this->container->get('settings')['determineRouteBeforeAppMiddleware'] === true) {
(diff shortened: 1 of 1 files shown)


Slim/App.php
```diff
@@ -297,13 +297,39 @@ public function run($silent = false)
         $request = $this->container->get('request');
         $response = $this->container->get('response');
 
+        $response = $this->process($request, $response);
+
+        $response = $this->finalize($response);
```
mathmarques: This method should be on process(), shouldn't?

mathmarques: This:

``` php
$response = $this->finalize($response);
```

 should be on process(), shouldn't?

geggleto:
I don't see the problem you are solving with this PR. Is there something that __invoke doesn't do that you need?

`To have a request that isn't created from the PHP super globals you need to patch the container.` totally agree with you. https://github.com/geggleto/reactive-slim/blob/master/bin/app.php such a hack :)

mathmarques:
When using directly __invoke, you are ignoring the app level middlewares. I think this is the only problem.

geggleto:
I think we should fix that...

AndrewCarterUK:
@geggleto

`__invoke()` also doesn't have the exception handling logic. If `__invoke()` worked in such a way that doing something like:

``` php
$request = Request::createFromGlobals();
$response = $app($request, $response);
$app->send($response);
```

Was the same as doing:

``` php
$app->run();
```

Then that would solve the problem.

(Edit) - But that sort of change couldn't be made until 4.x because it would break B/C

Slim/App.php
AndrewCarterUK: Good spot!

geggleto:
@AndrewCarterUK there is 1 other way to do this.

```
$res = $app->run(true);
```

https://github.com/slimphp/Slim/blob/3.x/tests/AppTest.php#L1558-L1592

AndrewCarterUK:
@geggleto That only works if you've patched the container or want the request to be created from global variables.

In my project the request is received from a FastCGI connection.

(edit to clarify: That's not PHP-FPM, the application runs as its own FastCGI daemon)

akrabat:
Looks good to me and maybe is that start of removing request and response from the container.

adambro:
The same problem is for creating WebTest class that mocks Slim, creates a request and checks response object. For now I'm using __invoke which works, because I don't have any middleware yet.

stivni:
> Looks good to me and maybe is that start of removing request and response from the container.

@akrabat is there an issue for that?

geggleto:
@stivni its a 4.0 thing ;)

Issue #1673, opened 2015-12-13
Is $uploadedFile['tmp_name'] a typo?


upliu:
hello,
I am reading slim's source code. In file Slim\Http\UploadedFile.php:108 and line 118 in with the line 

```
isset($uploadedFile['tmp_name']) ? $uploadedFile['name'] : null,
```

Why isset($uploadedFile['tmp_name']) , I think it should be isset($uploadedFile['name']):

```
isset($uploadedFile['name']) ? $uploadedFile['name'] : null,
```

Some can help me ? thanks.

silentworks:
This is correct as it is. Read from the link below for reference on $_FILES in php.

http://php.net/manual/en/features.file-upload.post-method.php

PR #1693, opened 2015-12-20
Is this typos?

phpinfointh:
`isset($uploadedFile['tmp_name']) ? $uploadedFile['name'] : null`
should it be
`isset($uploadedFile['name']) ? $uploadedFile['name'] : null`

### DIFF
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -105,7 +105,7 @@ class UploadedFile implements UploadedFileInterface
             if (!is_array($uploadedFile['error'])) {
                 $parsed[$field] = new static(
                     $uploadedFile['tmp_name'],
-                    isset($uploadedFile['tmp_name']) ? $uploadedFile['name'] : null,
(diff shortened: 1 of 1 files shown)


mathmarques:
No, see: https://github.com/slimphp/Slim/issues/1673

akrabat:
Looks like a bug to me. I don't understand from the manual page why it isn't. Input from @silentworks required!

silentworks:
This is what the manual page states.
- $_FILES['userfile']['name']: The original name of the file on the client machine.
- $_FILES['userfile']['tmp_name']: The temporary filename of the file in which the uploaded file was stored on the server.

To me this says that the `tmp_name` is greater to check against as this is an actual place the files is store on the server and if it wasn't this would return none from what the docs says, while there is no mention on if `name` would even be a thing at this stage.

But saying that, after reading this comment http://php.net/manual/en/features.file-upload.post-method.php#104349 it might make more sense to check `name`.

phpinfointh:
The reason I thought this was a typo is ...
If someone uses array data that `name` key is missing (maybe mocking data), this will result in PHP notice.

PR #1694, opened 2015-12-20
Option to send multiple response headers as comma separated single header

palanik:
[RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) specifies that response headers with multiple values sent as comma separated single header or as multiple headers to be the treated the same. Current implementation of Slim 3.x sends as multiple headers only.

Although this may not be an issue with modern browsers, I ran into a problem with a legacy client application that cannot handle duplicate headers. (No problem handling comma-separated values).

Adding a new container settings - `singleHeader` gives the flexibility to send response header in either format.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -365,9 +365,16 @@ class App
             ));
 
             // Headers
-            foreach ($response->getHeaders() as $name => $values) {
-                foreach ($values as $value) {
-                    header(sprintf('%s: %s', $name, $value), false);
+            if ($this->container->get('settings')['singleHeader'] === true) {
+                foreach ($response->getHeaders() as $name => $values) {
+                    header(sprintf('%s: %s', $name, $response->getHeaderLine($name)), false);
+                }
+            }
+            else {
+                foreach ($response->getHeaders() as $name => $values) {
+                    foreach ($values as $value) {
+                        header(sprintf('%s: %s', $name, $value), false);
+                    }
                 }
             }
         }
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -68,6 +68,7 @@ class Container extends PimpleContainer implements ContainerInterface
         'outputBuffering' => 'append',
         'determineRouteBeforeAppMiddleware' => false,
         'displayErrorDetails' => false,
+        'singleHeader' => false,
     ];
 
     /**


akrabat:
Reading up on this, it seems to be preferred to send one header only with multiple values as comma separated.

Should be just change our implementation to do that?

akrabat:
Ah… `Set-Cookie` must be sent as multiple separate headers.

From [RFC6265](https://www.ietf.org/rfc/rfc6265.txt):

> An origin server can include multiple Set-Cookie header fields in a single response. 

and

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field.  The usual mechanism for folding HTTP headers fields (i.e., as defined in [RFC2616]) might change the semantics of the Set-Cookie header field because the %x2C (",") character is used by Set-Cookie in a way that conflicts with such folding.

So we'll need an exception for that when `singleHeader` is true

SvenRtbg:
I'd think that dealing with a client that is unable to properly parse HTTP does not warrant implementing a configuration settings for ALL Slim applications to change behavior for ALL clients.

akrabat:
I agree with SvenRtbg.

palanik:
I have to disagree. It is not my intention to affect ALL Slim applications. That is why I specified this as an _optional_ settings.

The option is also not arbitrary. It is part of the standard spec. & I wanted the flexibility.

As I stated earlier, if I could change the client, I would have.
You know that not all clients of Slim server are browsers. 

A simple mistake by a fellow programmer without foresight affects us all.

Thanks for looking into this.

akrabat:
My preference is to send one header with multiple values where possible, as my reading of RFC2616 is that this is preferred over sending multiple headers with the same name. 

However, as I commented on 1st Jan, this cannot be a blanket thing as there are headers where this doesn't apply, such as Set-Cookie and may be others? Similarly, they may be headers where there can only be one value?

I don't want a setting; I want Slim to do the right thing.

tuupola:
One could always write a middleware which rewrites the headers to format which the legacy client can understand.

``` php
$app = new Slim\App

    $something = $response->getHeaderLine("X-Something");
    return $response = $response->withHeader("X-Something", $something);
});

$app->get("/header", function($request, $response, $arguments) {
    return $response
        ->withAddedHeader("X-Something", "foo")
        ->withAddedHeader("X-Something", "bar");
});

$app->run();
```

Yields

``` php
$ curl --include http://localhost:8080/header
HTTP/1.1 200 OK
X-Something: foo,bar
```

palanik:
@akrabat Respect your decision. Thanks.

@tuupola Thanks. Will try it out.

PR #1698, opened 2015-12-22
Request protocol version

abbert:
Property protocolVersion of Request defaults to 1.1. This change makes it change to the actual request protocol version. Is str_replace('HTTP/' sufficient?

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -178,6 +178,10 @@ class Request extends Message implements ServerRequestInterface
         $this->body = $body;
         $this->uploadedFiles = $uploadedFiles;
 
+        if (isset($serverParams['SERVER_PROTOCOL'])) {
+            $this->protocolVersion = str_replace('HTTP/', '', $serverParams['SERVER_PROTOCOL']);
+        }
+
         if (!$this->headers->has('Host') || $this->uri->getHost() !== '') {
             $this->headers->set('Host', $this->uri->getHost());
         }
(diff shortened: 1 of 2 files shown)


AndrewCarterUK:
:-1: in current form

The 'host' header isn't supported by HTTP/1.0 so you need to make sure you strip it in that scenario.

akrabat:
I don't understand what problem you are having.

The protocol is set via the `Environment` which merges the information from `$_SERVER`, so if it's set in `$_SERVER`, then we pick it up.

AndrewCarterUK:
@akrabat The protocol version is hard-coded to 1.1 in the message class (I think).

I don't actually think this is the wrong approach. The PSR-7 specification is only for the representation of HTTP/1.1 messages, there are features of it that are incompatible with HTTP/1.0 - so I think it's dangerous pretending to support it.

AndrewCarterUK:
Although that said, you can still do `withProtocolVersion()` :s

akrabat:
It's not very hardcoded! https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Message.php#L87-L101

abbert:
@AndrewCarterUK Yes indeed. That's why I thought it needed to be implemented. Also, this comment is from the MessageInterface: The version string MUST contain only the HTTP version number (e.g., "1.1", "1.0").

@akrabat when I delete the lines of code I added in the constructor of Request the test will fail..That means the protocolVersion is never overwritten with the version in the actual request

AndrewCarterUK:
Reading this I think @abbert is right - I can't see where the protocol version is extracted from the environment?

akrabat:
Yeah - I understand the issue now.

akrabat:
Thanks very much for this @abbert!

abbert:
No problem :)

PR #1708, opened 2015-12-29
Fix autoloader include in example/index.php file

michalbrauner:


akrabat:
The example is supposed to be standalone, so I've made this more explicit in #1712.

PR #1710, opened 2015-12-31
Save request/response state in middleware exceptions

johnstevenson:
I have gone about this the wrong way, in that I should have asked questions first, rather than making assumptions about certain aspects of the code. As such this PR may be completely unusable or unneeded, so feel free to close it or do with it as you wish.

The intention is to provide more consistent exception handling. For example, the `NotFound` and `MethodNotAllowed` exceptions supply the current request/response instances (which may have been modified by app middleware), whilst a general exception can only use the initial instances from the container.

The `SlimException` seems to provide a method of returning the current request/response instances, except that only the response is used (and it is expected to be the final one). Since this exception is currently undocumented (apart from being called a _Stop Exception_ in the comments), I have taken the following approach, which was inspired by which was inspired by PM Jones' https://github.com/relayphp/Relay.Relay 

**Stack**
Instead of using an SPL stack object, this PR introduces a `Slim\Stack\Stack` module that will catch an exception and wrap it in a new `StackException`, together with the last known request and response instances. This exception is then thrown and can be handled by `App::process()`. The design allows it to _bubble up_ linked stacks (like the route and the app kernels).

**BC breaks**
I have tried to keep these to zero, but there is the possibility of two edge-case breaks:
1. Code that uses a `try ... catch` block around a `next()` call to inner middleware would not be expecting a StackException.
2. Code that directly calls `App::__invoke()` may break if the route is not found or the method is not allowed because this function now throws the appropriate exception (rather than looking for and invoking an error handler).

**Tests**
The Stack module has its own independent tests. Several existing tests needed to be changed and I hope that my lack of knowledge has not resulted in any harmful oversights. Several changes deal with receiving a wrapped `StackException` and accessing items in the internal stack queue.

**Resolver**
The `Stack` class includes the functionality to resolve non-callable queue entries. This is not currently used but could be implemented in the future. The simplistic design uses an array of _resolvers_ (which should return `null` on failure) so that  user-supplied resolvers can be used in addition to the Slim `CallableResolver`. A final implementation would probably require more thought than this.

**Things I am not sure about**
1. If my changes to the tests have omitted something important in the desire to make them to work.
2. If the `Stack::hasHttpMessage()` method is needed. Its purpose is to allow user-supplied exceptions that implement  `getRequest()` and `getResponse()` methods to bypass the StackException wrapping. I am not sure if this just complicates things or not (in that it would be another exception to check for if your middleware try-catches a `next` call, if indeed this is something that would be done).
3. A general best-practice strategy for middleware exceptions using immutable PSR messages!

### DIFF
--- a/Slim/MiddlewareAwareTrait.php
+++ b/Slim/MiddlewareAwareTrait.php
@@ -9,11 +9,8 @@
 namespace Slim;
 
 use RuntimeException;
-use SplStack;
-use SplDoublyLinkedList;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
-use UnexpectedValueException;
 
 /**
  * Middleware
@@ -27,17 +24,9 @@ trait MiddlewareAwareTrait
     /**
      * Middleware call stack
      *
-     * @var  \SplStack
-     * @link http://php.net/manual/class.splstack.php
+     * @var  \Slim\Stack\Stack
      */
     protected $stack;
-
-    /**
-     * Middleware stack lock
-     *
-     * @var bool
-     */
-    protected $middlewareLock = false;
 
     /**
      * Add middleware
@@ -51,26 +40,14 @@ trait MiddlewareAwareTrait
      * @return static
      *
      * @throws RuntimeException         If middleware is added while the stack is dequeuing
-     * @throws UnexpectedValueException If the middleware doesn't return an instance of \Psr\Http\Message\ResponseInterface
      */
     protected function addMiddleware(callable $callable)
     {
-        if ($this->middlewareLock) {
-            throw new RuntimeException('Middleware can’t be added once the stack is dequeuing');
-        }
-
         if (is_null($this->stack)) {
             $this->seedMiddlewareStack();
         }
-        $next = $this->stack->top();
-        $this->stack[] = function (ServerRequestInterface $req, ResponseInterface $res) use ($callable, $next) {
-            $result = call_user_func($callable, $req, $res, $next);
-            if ($result instanceof ResponseInterface === false) {
-                throw new UnexpectedValueException('Middleware must return instance of \Psr\Http\Message\ResponseInterface');
-            }
 
-            return $result;
-        };
+        $this->stack->add($callable);
 
         return $this;
     }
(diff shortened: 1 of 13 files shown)


Slim/MiddlewareAwareTrait.php
```diff
@@ -27,19 +24,11 @@
...
+     * @var  \Slim\Stack\Stack
      */
     protected $stack;
 
     /**
-     * Middleware stack lock
-     *
-     * @var bool
-     */
-    protected $middlewareLock = false;
```
AndrewCarterUK, line 40: potential B/C break?

Slim/MiddlewareAwareTrait.php
```diff
@@ -27,19 +24,11 @@
     /**
      * Middleware call stack
      *
-     * @var  \SplStack
```
AndrewCarterUK, line 30: potential B/C break?

AndrewCarterUK:
Need to give a more thorough review, but I think you've got B/C breaks for any code using the MiddlewareAwareTrait as you remove a property and change the type of another.

edit:

PHPDoc labels it as an internal class. Not sure if this changes things?

https://github.com/johnstevenson/Slim/blob/stack/Slim/MiddlewareAwareTrait.php#L18

johnstevenson:
Thanks @AndrewCarterUK. I composed it like this (rather than decorating the existing MiddlewareAwareTrait properties in some way) because of the explicit nature of the PHPDoc comment:

>  This is an internal class that enables concentric middleware layers. This
> class is an implementation detail and is used only inside of the Slim
> application; it is not visible to—and should not be used by—end users.

As I said in my introductory note, I have gone about this completely the wrong way!

akrabat:
I think that `SlimException` was named incorrectly and we should have called it `StopException`. Oh well. - maybe we can rename and deprecate the old one?

What's the actual use-case we're trying to solve?

Is is that we can't write an `errorHandler` that knows about the current request/response?

akrabat:
I like the `handleException` refactor a lot on its own.

In `App.php`, it could look like something like this:

``` php
    public function process(ServerRequestInterface $request, ResponseInterface $response)
    {
        // Ensure basePath is set
        $router = $this->container->get('router');
        if (is_callable([$request->getUri(), 'getBasePath']) && is_callable([$router, 'setBasePath'])) {
            $router->setBasePath($request->getUri()->getBasePath());
        }

        // Dispatch the Router first if the setting for this is on
        if ($this->container->get('settings')['determineRouteBeforeAppMiddleware'] === true) {
            // Dispatch router (note: you won't be able to alter routes after this)
            $request = $this->dispatchRouterAndPrepareRoute($request, $router);
        }

        // Traverse middleware stack
        try {
            $response = $this->callMiddlewareStack($request, $response);
        } catch (SlimException $e) {
            // This is a Stop exception and contains the response
            $response = $e->getResponse();
        } catch (Exception $e) {
            $response = $this->handleException($e, $request, $response);
        }

        $response = $this->finalize($response);

        return $response;
    }

    /**
     * Call relevant handler from the Container if we can. If it doesn't exist,
     * then just re-throw.
     *
     * @param  Exception $e
     * @param  ServerRequestInterface $request
     * @param  Response $response
     *
     * @throws Exception
     * @return Response
     */
    protected function handleException($e, $request, $response)
    {
        if ($e instanceof MethodNotAllowedException) {
            $handler = 'notAllowedHandler';
            $params = [$e->getRequest(), $e->getResponse(), $e->getAllowedMethods()];
        } elseif ($e instanceof NotFoundException) {
            $handler = 'notFoundHandler';
            $params = [$e->getRequest(), $e->getResponse()];
        } else {
            // Other exception, use $request and $response params
            $handler = 'errorHandler';
            $params = [$request, $response, $e];
        }
        if ($this->container->has($handler)) {
            $callable = $this->container->get($handler);
            // Call the registered handler
            return call_user_func_array($callable, $params);
        }
        // No handlers found, so just throw the exception
        throw $e;
    }
```

(put here, so I don't lose it!)

johnstevenson:
> I think that SlimException was named incorrectly and we should have called it StopException ...

Possibly, although I could see that causing some confusion because it is used differently from the exceptions that extend it (in that it requires a final response to have been created), so maybe a more generic name is better. Who knows? Whatever, it would be nice if it was documented (which I don't mind helping with).

> Is is that we can't write an errorHandler that knows about the current request/response?

Correct. And that makes it inconsistent with the notFound and notAllowed handlers.

johnstevenson:
Glad you like the `handleException` refactor. It certainly makes things a lot more readable. I can do a separate PR for this on it own if you wish (if it becomes the final outcome of this PR) - or you can just incorporate it yourself.

akrabat:
Yes - let's get the `handleException()` change in as that's the "easy" bit. Please do a separate PR.

akrabat:
I'm sorry, but this is too invasive for Slim 3. Maybe the ideas in here need looking at when we do 4.0?

PR #1712, opened 2016-01-01
Add composer.json to example and update comments

akrabat:
The example is supposed to be a standalone thing, so make this clearer.


geggleto:
``` php
//Step 3:

The route pattern is relative to where your index.php [application entry point] is.
That is to say these routes are not abolsute.
'/example/index.php/myroute' is actually '/myroute'

```

silentworks:
Closing this one out as I think it will bring more confusion than clarity. Someone is likely to clone the repo and run composer update in `/example` and then ask why is there 2 copies of Slim.

PR #1715, opened 2016-01-02
Refactor exception handling

johnstevenson:
As per https://github.com/slimphp/Slim/pull/1710

Note that the `SlimException` needs to be in the handler, otherwise it erroneously catches its descendants (MethodNotAllowedException, NotFoundException).

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -339,29 +339,8 @@ class App
         // Traverse middleware stack
         try {
             $response = $this->callMiddlewareStack($request, $response);
-        } catch (MethodNotAllowedException $e) {
-            if (!$this->container->has('notAllowedHandler')) {
-                throw $e;
-            }
-            /** @var callable $notAllowedHandler */
-            $notAllowedHandler = $this->container->get('notAllowedHandler');
-            $response = $notAllowedHandler($e->getRequest(), $e->getResponse(), $e->getAllowedMethods());
-        } catch (NotFoundException $e) {
-            if (!$this->container->has('notFoundHandler')) {
-                throw $e;
-            }
-            /** @var callable $notFoundHandler */
-            $notFoundHandler = $this->container->get('notFoundHandler');
-            $response = $notFoundHandler($e->getRequest(), $e->getResponse());
-        } catch (SlimException $e) {
-            $response = $e->getResponse();
         } catch (Exception $e) {
-            if (!$this->container->has('errorHandler')) {
-                throw $e;
-            }
-            /** @var callable $errorHandler */
-            $errorHandler = $this->container->get('errorHandler');
-            $response = $errorHandler($request, $response, $e);
+            $response = $this->handleException($e, $request, $response);
         }
 
         $response = $this->finalize($response);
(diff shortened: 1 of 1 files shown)


AndrewCarterUK:
The comment on [line 592](https://github.com/johnstevenson/Slim/blob/refactor-exception/Slim/App.php#L592) would be good information for the phpdoc annotation on [line 567](https://github.com/johnstevenson/Slim/blob/refactor-exception/Slim/App.php#L567).

It's a protected method so it is potentially public facing and the situation where a method called 'handleException' actually throws an exception is worthy of slightly more formal documentation (IMO).

That's all I can really add here - it looks like a good refactor that improves readability.

johnstevenson:
Thanks @AndrewCarterUK. It does mention this in the general comment <a href="https://github.com/johnstevenson/Slim/blob/refactor-exception/Slim/App.php#L559">line 559</a>, but I agree that it is certainly worth including with the @throws annotation. I also need to correct the Response annotations to ResponseInterface so I'll do them all together.

> where a method called 'handleException' actually throws an exception

Good point - that bit went completely over my head. Do you think it would be worth changing the method name, and if so, to what?

AndrewCarterUK:
I don't think it requires a name change. Part of the job of handling an exception is possibly to throw it if you can't do anything else with it - it's just behaviour that is significant enough to be included in the main documentation for the method.

johnstevenson:
@akrabat. The refactor could also be used in `App::__invoke()`, although it is a bit clumsy in that it passes $request and $response params into `handleException` that are not actually used. Let me know if you want this included in the PR.

``` php
    /**
     * Invoke application
     *
     * This method implements the middleware interface. It receives
     * Request and Response objects, and it returns a Response object
     * after compiling the routes registered in the Router and dispatching
     * the Request object to the appropriate Route callback routine.
     *
     * @param  ServerRequestInterface $request  The most recent Request object
     * @param  ResponseInterface      $response The most recent Response object
     *
     * @return ResponseInterface
     * @throws MethodNotAllowedException
     * @throws NotFoundException
     */
    public function __invoke(ServerRequestInterface $request, ResponseInterface $response)
    {
        // Get the route info
        $routeInfo = $request->getAttribute('routeInfo');

        /** @var \Slim\Interfaces\RouterInterface $router */
        $router = $this->container->get('router');

        // If router hasn't been dispatched or the URI changed then dispatch
        if (null === $routeInfo || ($routeInfo['request'] !== [$request->getMethod(), (string) $request->getUri()])) {
            $request = $this->dispatchRouterAndPrepareRoute($request, $router);
            $routeInfo = $request->getAttribute('routeInfo');
        }

        // Run the route if it has been matched and return the response
        if ($routeInfo[0] === Dispatcher::FOUND) {
            $route = $router->lookupRoute($routeInfo[1]);
            return $route->run($request, $response);
        }

        // Route not matched so create an exception to pass to handler
        if ($routeInfo[0] === Dispatcher::METHOD_NOT_ALLOWED) {
            $exception = new MethodNotAllowedException($request, $response, $routeInfo[1]);
        } else {
            $exception = new NotFoundException($request, $response);
        }

        return $this->handleException($exception, $request, $response);
    }
```

PR #1635, opened 2015-11-29
Lazy resolve Route middleware

mathmarques:
See https://github.com/slimphp/Slim/issues/1604

Issue #1604, opened 2015-11-21
Lazy resolving of middleware


chriseskow:
I was playing around with Slim 3 and noticed that for group and route middleware, Slim will resolve the callable immediately when the middleware is added, instead of only when it is actually needed and called. This means if a particular route or group has a middleware defined by a DI container key which is potentially expensive to instantiate (e.g. because a database connection is injected into the constructor), it will be created on every request, not just requests that utilize that middleware.

Any chance that callable resolving for route- and group-level middleware could be deferred to when the middleware is executed, instead of when it is added? (App-level middleware shouldn't matter, since they run on every request anyway.)

geggleto:
the thing is, on routes and groups the callables need to be present in order to be added to the stack. Having this would require a significant rewrite of the middleware process, but it is possible.

I would imagine this could be done for 3.1

mathmarques:
@chriseskow
I made a modification to fix this here:
https://github.com/mathmarques/Slim/tree/lazy-resolving-middleware
The change:
https://github.com/mathmarques/Slim/commit/99a85f9f1fbabfde730e160898da0024e56c7bb3

I'm justing trying to figure out why Slim needs to finalize() all routes via Router->finalize(). We just need to finalize the dispatched route.

I leave the Router->finalize() because of the Tests.

akrabat:
Something for 3.1.0. See https://github.com/slimphp/Slim/issues/1627

akrabat:
Fixed via #1716

chriseskow:
Thanks for getting this done, guys! :sparkles:

PR #1716, opened 2016-01-02
added a deferredMiddleware responsible for resolving middleware on demand

jeremygiberson:
This is an alternative solution for [1604](https://github.com/slimphp/Slim/issues/1604). 

Differences to [1635](https://github.com/slimphp/Slim/pull/1635) include removal of resolving responsibility (duplicated code) from `App`, `Routable`,  and `Route` and the responsibility is placed into `DeferredMiddleware`. Also, middleware is not resolved until it has been invoked.

### DIFF
--- a/Slim/DeferredMiddleware.php
+++ b/Slim/DeferredMiddleware.php
@@ -16,10 +16,10 @@ class DeferredMiddleware
 
     /**
      * DeferredMiddleware constructor.
-     * @param $callable
-     * @param $container
+     * @param callable|string $callable
+     * @param ContainerInterface $container
      */
-    public function __construct($callable, $container)
+    public function __construct($callable, ContainerInterface $container = null)
     {
         $this->callable = $callable;
         $this->container = $container;
--- a/Slim/DeferredMiddleware.php
+++ b/Slim/DeferredMiddleware.php
@@ -3,7 +3,6 @@
 
 namespace Slim;
 
-
 use Closure;
 use Interop\Container\ContainerInterface;
 
--- a/Slim/DeferredMiddleware.php
+++ b/Slim/DeferredMiddleware.php
@@ -1,0 +1,40 @@
+<?php
+
+
+namespace Slim;
+
+
+use Closure;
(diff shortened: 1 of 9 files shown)


Slim/DeferredMiddleware.php
```diff
@@ -0,0 +1,39 @@
...
+    private $callable;
+    /** @var  ContainerInterface */
+    private $container;
+
+    /**
+     * DeferredMiddleware constructor.
+     * @param $callable
+     * @param $container
+     */
+    public function __construct($callable, $container)
```
geggleto: Needs type checks. 
callable?, container interface

tests/DeferredMiddlewareTest.php
```diff
@@ -0,0 +1,60 @@
...
+            ->expects($this->once())
+            ->method('foo');
+
+        $container = new Container();
+
+        $test = $this;
+
+        $closure = function () use ($container, $test, $assertCalled) {
+            $assertCalled->foo();
+            $test->assertSame($container, $this);
```
geggleto: Not 100% sure that this is a valid check

jeremygiberson: `$test->assertSame($container, $this);` makes sure that the closure has been bound to the container. 

`$assertCalled` sets up an expectation that `$assertCalled->foo()` must be called once. This makes sure the test fails if the closure is never executed (since assertSame would only execute if the closure was invoked). `assertSame` makes sure the two compared objects are the same instance.

Slim/DeferredMiddleware.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+
+namespace Slim;
+
+use Closure;
+use Interop\Container\ContainerInterface;
+
+class DeferredMiddleware
```
akrabat: Maybe we should name this `CallableResolver`?  It's not middleware, so I'm uncomfortable with the current name.

geggleto: :+1: Yeah I don't really like the name either.

DeferredCallable ?

PR #1720, opened 2016-01-05
array specialization in phpdoc

zzgab:
If this is in line with Slim's codebase philosophy, here is a minor change (phpdoc comments) about the returning type of `Request`'s `getUploadedFiles()` method. Since this method always returns an array (possibly empty) of `UploadedFileInterface` instances, this phpdoc change enables IDE analysis and auto-complete in call chaining.
Thanks.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -776,7 +776,7 @@ class Request extends Message implements ServerRequestInterface
      * These values MAY be prepared from $_FILES or the message body during
      * instantiation, or MAY be injected via withUploadedFiles().
      *
-     * @return array An array tree of UploadedFileInterface instances; an empty
+     * @return UploadedFileInterface[] An array tree of UploadedFileInterface instances; an empty
      *     array MUST be returned if no data is present.
      */
     public function getUploadedFiles()


geggleto:
Yes! more type hinting :D :+1:

JoeBengalen:
Nope, look at the interface: https://github.com/php-fig/http-message/blob/master/src/ServerRequestInterface.php#L137

If I remember correctly `array` was chosen over `UploadedFileInterface []`, because it can be a "tree", not just a list of.

zzgab:
@JoeBengalen What kind of tree shall we expect? Do you have an example in mind, that could help me understand the construct of such tree? Thanks a lot.

-- edit:
Don't bother: I re-read the doc http://www.php-fig.org/psr/psr-7/
And inded, they provide an example there.
Thanks.

JoeBengalen:
@gabrielzerbib Read the standard: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message.md#16-uploaded-files

Examples are given ;)

zzgab:
(yep :) )

PR #1726, opened 2016-01-09
Converted Routable::add to allow multiple middleware via an array

DJCMBear:
Depending on how the programmer codes this might be a nice addition to the framework, I have only done it for the add function of the Routable class for now but it can be applied to other functions throughout the framework.

Basically it will allow the framework to function as normal but also allows unlimited middleware elements to be added via one add function call.

The current way to add middleware:

``` php
$app = new \Slim\App;

$app -> get( "/", function ( $request, $response ) {
    return $response -> write( "Hello World!" );
} )
-> add( SomeClass::class )
-> add( SomeOtherClass::class )
-> add( [ 'SomeClass', 'method' ] )
-> add( [ $instance, 'method' ] );

$app -> run();
```

With my additions:

``` php
$app = new \Slim\App;

$app -> get( "/", function ( $request, $response ) {
    return $response -> write( "Hello World!" );
} ) -> add( [
    SomeClass::class,
    SomeOtherClass::class,
    [ 'SomeClass', 'method' ],
    [ $instance, 'method' ]
] );

$app -> run();
```

It isn't anything major but to me it looks cleaner and it doesn't affect how any current app code works, so all apps using v3 of Slim will still work in exactly the same way just with the added bonus of being able to use only one add function call if the developer wants to.

### DIFF
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -91,7 +91,7 @@ abstract class Routable
      */
     public function add($callable)
     {
-        foreach (func_get_args() As $callable) {
+        foreach (func_get_args() as $callable) {
             $callable = $this->resolveCallable($callable);
             if ($callable instanceof Closure) {
                 $callable = $callable->bindTo($this->container);
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -91,12 +91,14 @@ abstract class Routable
      */
     public function add($callable)
     {
-        $callable = $this->resolveCallable($callable);
-        if ($callable instanceof Closure) {
-            $callable = $callable->bindTo($this->container);
-        }
+        foreach (func_get_args() As $callable) {
+            $callable = $this->resolveCallable($callable);
+            if ($callable instanceof Closure) {
+                $callable = $callable->bindTo($this->container);
+            }
 
-        $this->middleware[] = $callable;
+            $this->middleware[] = $callable;
+        }
         return $this;
     }
 }
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -91,14 +91,22 @@ abstract class Routable
      */
     public function add($callable)
     {
-        foreach (func_get_args() as $callable) {
+        if (is_array($callable)) {
+            foreach ($callable as $closure) {
+                $closure = $this->resolveCallable($closure);
+                if ($closure instanceof Closure) {
+                    $closure = $closure->bindTo($this->container);
+                }
+                $this->middleware[] = $closure;
+            }
+        } else {
             $callable = $this->resolveCallable($callable);
             if ($callable instanceof Closure) {
                 $callable = $callable->bindTo($this->container);
             }
-
             $this->middleware[] = $callable;
         }
+        
         return $this;
     }
 }
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -91,22 +91,18 @@ abstract class Routable
      */
     public function add($callable)
     {
-        if (is_array($callable)) {
-            foreach ($callable as $closure) {
-                $closure = $this->resolveCallable($closure);
-                if ($closure instanceof Closure) {
-                    $closure = $closure->bindTo($this->container);
-                }
-                $this->middleware[] = $closure;
-            }
-        } else {
-            $callable = $this->resolveCallable($callable);
-            if ($callable instanceof Closure) {
-                $callable = $callable->bindTo($this->container);
-            }
-            $this->middleware[] = $callable;
+        if (!is_array($callable) || is_callable($callable)) {
+            $callable = [$callable];
         }
-        
+
+        foreach ($callable as $closure) {
+            $closure = $this->resolveCallable($closure);
+            if ($closure instanceof Closure) {
+                $closure = $closure->bindTo($this->container);
+            }
+            $this->middleware[] = $closure;
+        }
+
         return $this;
     }
 }
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -85,7 +85,7 @@ abstract class Routable
     /**
      * Prepend middleware to the middleware collection
      *
-     * @param mixed $callable The callback routine
+     * @param mixed $callable The callback routine or an array of callback routines
      *
      * @return static
      */
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -99,18 +99,25 @@ class App
      *
      * This method prepends new middleware to the app's middleware stack.
      *
-     * @param  mixed    $callable The callback routine
+     * @param  mixed    $callable The callback routine which can be an array of multiple routines
      *
      * @return static
      */
     public function add($callable)
     {
-        $callable = $this->resolveCallable($callable);
-        if ($callable instanceof Closure) {
-            $callable = $callable->bindTo($this->container);
+        if (!is_array($callable) || is_callable($callable)) {
+            $callable = [$callable];
         }
 
-        return $this->addMiddleware($callable);
+        foreach ($callable as $closure) {
+            $closure = $this->resolveCallable($closure);
+            if ($closure instanceof Closure) {
+                $closure = $closure->bindTo($this->container);
+            }
+            $this->addMiddleware( $closure );
+        }
+
+        return $this;
     }
 
     /**
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -99,7 +99,7 @@ class App
      *
      * This method prepends new middleware to the app's middleware stack.
      *
-     * @param  mixed    $callable The callback routine which can be an array of multiple routines
+     * @param mixed $callable The callback routine or an array of callback routines
      *
      * @return static
      */


Slim/Routable.php
```diff
@@ -91,12 +91,14 @@ public function setContainer(ContainerInterface $container)
      */
     public function add($callable)
```
flavioheleno, line 92: As you're using `func_get_args()` rather than the `$callable` parameter, I guess you should remove it from the function signature.

DJCMBear, line 92: I thought that but to keep in with Slim coding standards instead of allowing no arguments to be passed with no errors being thrown, keeping it in makes sure at least one argument is always present.

akrabat:
Thinking about this one, I think I'd prefer it if `add()` took an array rather than multiple arguments.

Thoughts @silentworks, @codeguy, @geggleto, anyone else?

tuupola:
Personally I also prefer arrays over multiple arguments.

DJCMBear:
I updated the code to accept an array instead of multiple arguments, hopefully it will be added as it would make adding multiple middleware easier.

akrabat:
Thanks. Please update the documentation in `add()`'s docblock.

vlakoff:
You could avoid code duplication by doing:

```
if (!is_array($callable)) {
    $callable = [$callable];
}
foreach ($callable as $current) { ... }
```

**But more importantly:**
- What if someone does `->add([$instance, 'method'])` or `->add(['class', 'method'])`? These callables are also recognized as arrays.
- `App->add()` should be kept sync.

akrabat:
@vlakoff good point about `->add([$instance, 'method'])`. I wonder why we don't have a failing test for this branch…

vlakoff:
To fix this PR:

```
if (!is_array($callable) || is_callable($callable)) {
    $callable = [$callable];
}
foreach ($callable as $current) { ... }
```

geggleto:
@akrabat I would be open to seeing something like `->addMiddlewareStack(array $stack)` where all the special logic goes.

I think adding this functionality to add will get a bit too confusing. Internally the addStack will end up calling the singular version.

akrabat:
ok - more discussion on this idea is required.

vlakoff:
As a reminder, don't forget to apply the same changes to `App::add` method.

silentworks:
Would this not be a BC break? because as someone mentioned above if I were already using `->add([$instance, 'method']);`, this PR would break my code.

akrabat:
@silentworks It shouldn't be a BC break with vlackoff's changes are noted in https://github.com/slimphp/Slim/pull/1726#issuecomment-180373096.

The question is whether we want it or whether to have a separate method as suggested by @geggleto.

DJCMBear:
To me personally I find it nice that you only have to remember one method name and can pass it one or more middleware routines, it also allows for quick code edits if you want to convert a single add routine into adding multiple routines without changing method names.

However on the other hand I can see why different method name would make things easier to read while skimming through the code and can see quickly that this method is being passed multiple routines where as add is just adding a single routine.

geggleto:
If someone could clarify. @akrabat @silentworks 

Is this currently valid?
`->add([$instance, 'method']);`

Or are we adding this in this PR as well as handling nested callables as well.

DJCMBear:
From my tests the current slim accepts `->add([$instance,'method']);` and is valid, the same goes for this branches code, this code only adds the ability to pass multiple middleware routines using one add method instead of using an add method call for each middleware routine.

vlakoff:
Adding a method would defeat the purpose of this PR, simplifying the API. Having a new method to remember would be worst than just doing `->add()->add()`.

I assume most of the time middlewares are added using strings,
and `->add(['FirstMiddleware', 'SecondMiddleware'])` looks pretty neat to me.

DJCMBear:
@vlakoff I agree with everything you said, the code change I am proposing fits with what you are saying but adds the extra ability to use more than just strings. it doesn't affect how the code works currently, it just extends the functionality for those who want to use just one add method while giving them the ability to still use the method in the same way.

The current code in this PR allows this to work...

> -> add( [
> &nbsp;&nbsp;&nbsp;&nbsp;SomeClass::class,
> &nbsp;&nbsp;&nbsp;&nbsp;SomeOtherClass::class,
> &nbsp;&nbsp;&nbsp;&nbsp;'SomeMethod',
> &nbsp;&nbsp;&nbsp;&nbsp;new ClassName,
> &nbsp;&nbsp;&nbsp;&nbsp;'SomeClass::SomeMethod',
> &nbsp;&nbsp;&nbsp;&nbsp;[ 'SomeClass', 'method' ],
> &nbsp;&nbsp;&nbsp;&nbsp;[ $instance, 'method' ],
> &nbsp;&nbsp;&nbsp;&nbsp;function ( [args here] ) { [code here] }
> ] )
> -> add( new ClassName )
> -> add( [ $instance, 'method' ] )
> -> add( 'SomeMethod' )
> -> add( SomeClass::class )
> -> add( 'SomeClass::SomeMethod' )
> -> add( [ SomeClass, 'method' ] )
> -> add( function ( [args here] ) { [code here] } );

vlakoff:
To make sure, does anyone think of a case where the `if (!is_array($callable) || is_callable($callable))` test would give wrong result?

Slim/Routable.php
```diff
@@ -85,18 +85,24 @@ public function setContainer(ContainerInterface $container)
     /**
      * Prepend middleware to the middleware collection
      *
-     * @param mixed $callable The callback routine
+     * @param mixed $callable The callback routine or an array of callback routines
```
vlakoff, line 88: description should be identical in both files

DJCMBear, line 88: I updated App.php to match docblock in Routable.php

JoeBengalen:
@vlakoff Since people are free to provide their own `CallableResolverInterface` implementation I'm not sure if you can ever be 100% sure if you statement will work. The given "callable" could be anything (type `mixed`)

Therefore I would discourage the array syntax. If you want to pass multiple middlewares within the same `add` method, stick with the earlier idea of multiple arguments.

vlakoff:
- That's a good point. This would break isolation and copy some of `CallableResolver` logic into App/Router. Here is some idea to handle this… but I don't like it:

```
// new method in shipped CallableResolver
// check if argument is supposedly a callable/middleware
// note the string validity is not checked (no RuntimeException)
public function isResolvable($toResolve)
{
    return is_callable($toResolve) || is_string($toResolve);
}

public function add($callable)
{
    if ($this->isResolvable($callable)) {
        $callable = [$callable];
    }

    foreach ($callable as $closure) {
        $closure = $this->resolveCallable($closure);
        if ($closure instanceof Closure) {
            $closure = $closure->bindTo($this->container);
        }
        $this->addMiddleware( $closure );
    }

    return $this;
}
```
- A caveat with the variadic approach, it avoids adding some method parameters in the future.
- FYI, Laravel allows passing an array of middlewares… but middlewares can only be defined as strings, so it's easier for them: `is_string($middleware) ? /*middleware*/ : /*array of middlewares*/`

DJCMBear:
Anyone else got a way like @vlakoff which he may like more, I am open to anything which everyone is happy with.

I just feel that allowing the developers who use the Slim framework to have the ability to create systems with it faster and easier while keeping everything working is the way to move forward with adding to / modifying the framework, even though it only removes a small number of characters when adding middleware to a route, those less characters would add up to less time spent to develop a system with Slim.

vlakoff:
Considering JoeBengalen's very relevant point above, I have changed my mind and would hold on this change. It's the price in exchange for being liberal in how we can define the callable, up to even allowing a custom resolver.

Thus, I would rather implement a `addMany` method, which would be simple and robust. Even better, I think the application code would be more readable this way, as we would be much more easily noticing the method call is adding several middlewares.

DJCMBear:
Would we still not be in the same position with `addMany` as we are with `add` if we decide to go that route, surly adding a new method which is using the current way this PR is using will just be adding to the problem we are trying to solve.

That is unless you are talking about switching from an array to multiple arguments like I originally opened this PR to request, if you are on about multiple arguments then surly modifying the `add` method to extend the number of arguments it can take is more easier to do than creating a whole new method.

tuupola:
My concern is that currently nothing is broken so no need to fix it. Changing current behaviour will make Slim code more complex and probably more error prone. Personally I have nothing against typing `->add(...)` more than once so I would leave it as it is.

DJCMBear:
@tuupola good point but just a thought to keep in mind, v2 had a much nicer way to add middleware just by doing `$app -> get('/','mw1','mw2',function);` so to say allowing multiple arguments to be passed to the `add` method isn't anything that Slim hasn't been able to do in the past. Maybe if this feature isn't added now it might be nice to see it be added at some point.

vlakoff:
TBH, just calling `->add()` several times doesn't look that bad. I even think it's clearer than calling it once with an array. See example in first message. So the only remaining option would be an `->addMany()` method, but is it worth it?

A bigger improvement would be middleware groups, [as Laravel did](https://laravel.com/docs/5.2/middleware#middleware-groups). But please open a separate issue if you want to discuss about it.

DJCMBear:
I love the idea of middleware groups, we currently can only do this via adding middleware the group method to cover everything within that group but you also need to pass a base path to the route, having a middleware group could solve that.

I will open a discussion about that a little bit later so others can give their thoughts on the subject. Should I mark this as closed or do you think this needs more of a discussion still?

geggleto:
As someone who helps lots of people in the Slim IRC channel, I think this might confuse people. So I am with @tuupola. The only way I could see this working is if we had a Collection object that would could pass to add, but if we go that far might as well look at something a bit more beefy like middleware groups.

silentworks:
I personally don't really have too much of an opinion on this as I tend to have only ever needed 1, max 2 middleware per route, normally I would add them to route groups or at the app level. I don't really mind typing `->add()` over and over, nor would I mind typing `->add(an_array_of_middlewares)`.

SvenRtbg:
@vlakoff 

I constructed a case where `is_callable()` will incorrectly fail:

https://3v4l.org/6GkRS

```
class Foo {
    function __invoke() {
        return 1;
    }

    function trim() {
       return 1;   
    }
    private function drum() {
        return 1;
    }
}
```

Assume that you want to add `Foo::__invoke` as a callable together with a call to the function `trim`, this detection would go wrong and detect a callable to `Foo::trim`.

vlakoff:
I don't see what is wrong in the snippet you posted (by any chance, didn't you confuse with __call?)

Nevertheless, can you please clarify by posting a `->middleware()` call?

Other points:
- I think `__invoke` is a "no, do not use it" thing. Maybe with the exception of middleware `$next()`.
- Looks like this is a very edge case. Unless you prove otherwise by showing an use in `->middleware()`.

Finally, as I stated above, considering the multiple ways of providing a callback, I think it would be better not to add complexity to the `middleware` method.

SvenRtbg:
> To make sure, does anyone think of a case where the if (!is_array($callable) || is_callable($callable)) test would give wrong result?

I think the discussion is about allowing either `add(SingleCallableMiddleware)` or `add([MultipleCallableMiddleware,MultipleCallableMiddleware])` and how to detect them. A valid callable can be an array itself, and elements of that array can also be callable themself.

If the user provides an array with a reference to an object that implements `__invoke`, and a string with a function that also is a valid method on the object as well as exists globally, you cannot distinguish between "use object's __invoke and the global function" and "use object's method".

My example probably failed to correctly display it, here's another one: https://3v4l.org/LGlf0

My point is: Don't overload methods with argument variants. Adding a new method that explicitly expects an array (or maybe any traversable) and adds all contained elements as another middleware is dead simple to write - messing with detection (that can go wrong in edge cases) is not.

geggleto:
:+1: @SvenRtbg

vlakoff:
Gotcha your example (even if it seems quite improbable). Nevertheless, we do agree overloading the `add` method isn't a good idea.

So, if we decide to go further, we could implement a method `addMany` or `addSeveral`, which could without problem accept both array and variadic.

akrabat:
I've been thinking about this and I now agree that overloading `add()` to do two things isn't the way forward. I'm not convinced that the additional utility of an `addMany()` is worth adding to core as we would then have two ways to do it which is more confusing than one way.

Issue #1731, opened 2016-01-13
Unable to set empty array to functions withQueryParams and withParsedBody


phm46:
If I set empty array to function withQueryParams and withParsedBody in middleware like this:

``` php
$app->add(function ($request, $response, $next) {
    var_dump($request->getQueryParams());
    var_dump($request->getParsedBody());

    $request = $request->withQueryParams(Array())->withParsedBody(Array());

    var_dump($request->getQueryParams());
    var_dump($request->getParsedBody());
    $response = $next($request, $response);
    return $next($request, $response);
});
```

it returns same values but it was supposed to return empty array.

AndrewCarterUK:
I couldn't reproduce your issue.

This code:

``` php
<?php

chdir(dirname(__DIR__));
require 'vendor/autoload.php';

use Slim\Http\Environment;
use Slim\Http\Request;

// Creating request with query params and parsed body
$request = Request::createFromEnvironment(Environment::mock());
$request = $request->withQueryParams(['foo' => 'bar'])->withParsedBody(['bar' => 'foo']);

echo 'With data:' . PHP_EOL;
var_dump($request->getQueryParams());
var_dump($request->getParsedBody());
echo PHP_EOL;

// Attempting to clear
$request = $request->withQueryParams(Array())->withParsedBody(Array());

echo 'Without data:' . PHP_EOL;
var_dump($request->getQueryParams());
var_dump($request->getParsedBody());
```

Gave:

```
With data:
array(1) {
  'foo' =>
  string(3) "bar"
}
array(1) {
  'bar' =>
  string(3) "foo"
}

Without data:
array(0) {
}
array(0) {
}
```

Could you modify this script to reproduce your issue?

phm46:
Yes, that is true. But this bug happens only if you use normal request with query and body params. If you manualy set these params with

``` php
$request = $request->withQueryParams(['foo' => 'bar'])->withParsedBody(['bar' => 'foo']);
```

it works corectly.
Try this code:

``` php
<?php

chdir(dirname(__DIR__));
require 'vendor/autoload.php';
$app = new \Slim\App;

use Slim\Http\Environment;
use Slim\Http\Request;

$app->get('/', function($request, $response, $args){
    echo 'With data:' . PHP_EOL;
    var_dump($request->getQueryParams());
    var_dump($request->getParsedBody());

    echo 'Without data:' . PHP_EOL;
    $request = $request->withQueryParams(Array())->withParsedBody(Array());
[...]

PR #1733, opened 2016-01-13
Fix getQueryParams

mathmarques:
Empty array returns false but an empty array is valid in this case.

Fix a part of: https://github.com/slimphp/Slim/issues/1731


akrabat:

akrabat:
Thanks!

PR #1734, opened 2016-01-13
Fix getParsedBody

mathmarques:
Fix a part of: https://github.com/slimphp/Slim/issues/1731

The only remaining problem is setting null to parsed body.
Using:

``` php
 $request = $request->withParsedBody(null);
```

Won't take effect.

The only way I see to fix this is setting a flag if the body is already parsed or not.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -92,7 +92,7 @@ class Request extends Message implements ServerRequestInterface
      *
      * @var null|array|object
      */
-    protected $bodyParsed;
+    protected $bodyParsed = false;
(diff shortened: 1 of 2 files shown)


AndrewCarterUK:
http://www.php-fig.org/psr/psr-7/

`null`, `array` and `object` are the allowed values.

You could use `false` as the 'not here' value but I think that would be messy - would be clearer to introduce a new property that acts as a flag (as you suggested).

akrabat:

akrabat:
Thank you.

PR #1737, opened 2016-01-14
withJson should not set http status code implicitly

GHKEN:
I thought that `withJson` set http status code confusing user.
e.g.
`$response->withStatus(500)->withJson([])`
and
`$response->withJson([])->withStatus(500)`
are not equally.:cry:
We should fix `withJson` to intuitive.
Thank you for your consideration.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -290,11 +290,12 @@ class Response extends Message implements ResponseInterface
      * response to the client.
      *
      * @param  mixed  $data   The data
+     * @param  int    $status The HTTP status code.
      * @param  int    $encodingOptions Json encoding options
      * @throws \RuntimeException
      * @return self
      */
-    public function withJson($data, $encodingOptions = 0)
+    public function withJson($data, $status = null, $encodingOptions = 0)
     {
         $body = $this->getBody();
         $body->rewind();
@@ -305,7 +306,11 @@ class Response extends Message implements ResponseInterface
             throw new \RuntimeException(json_last_error_msg(), json_last_error());
         }
 
-        return $this->withHeader('Content-Type', 'application/json;charset=utf-8');
+        $response_with_json = $this->withHeader('Content-Type', 'application/json;charset=utf-8');
+        if (isset($status)) {
+            return $response_with_json->withStatus($status);
+        }
+        return $response_with_json;
     }
 
     /**
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -306,11 +306,11 @@ class Response extends Message implements ResponseInterface
             throw new \RuntimeException(json_last_error_msg(), json_last_error());
         }
 
-        $response_with_json = $this->withHeader('Content-Type', 'application/json;charset=utf-8');
+        $responseWithJson = $this->withHeader('Content-Type', 'application/json;charset=utf-8');
         if (isset($status)) {
-            return $response_with_json->withStatus($status);
+            return $responseWithJson->withStatus($status);
         }
-        return $response_with_json;
+        return $responseWithJson;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


AndrewCarterUK:
Unfortunately this would be a backwards incompatible change (it would break the code of anyone using this method when they ran a composer update).

GHKEN:
Thank you for the comment.
I thought that default status code to null is better if we need backwards compatibility.

silentworks:
This is still a BC break. Also why wouldn't you do `$response->withJson([], 500)` if you need a different status code?

AndrewCarterUK:
@silentworks I can see the issue.

``` php
$response = $response->withStatusCode(404)->withJson([]);
```

This code should return a response with a 404 status code.

silentworks:
Please also note that `withJson` is a helper method, it should help you write less code for a specific purpose. 

@AndrewCarterUK that should return a response with 200, if the developer wants it to return a status of 404 either do:

``` php
$response = $response->withStatusCode(404)
    ->withHeader('Content-Type', 'application/json;charset=utf-8')
    ->getBody()->write(json_encode([]));
```

or

``` php
$response->withJson([], 404);
```

AndrewCarterUK:
It 'does' return that status code, I wouldn't say that it should.

`withJson` should just give you your response 'with json'. Changing the status code is a non-obvious side effect that I can see causing problems (it already has enough to trigger this PR).

Also, by default the status code would be 200, so it's a truly pointless feature if that was the intent. All it could do is overwrite an implicit change.

geggleto:
:-1: 
You are loading a message into the response and it should return 200 by default.

If you want to change it, it would only make sense if all methods that mimic this functionality were also changed.

The supplying 200 by default is a shortcut for our users.

akrabat:

akrabat:
Thanks.

Loic-Rameau:
```.json
{
            "name": "slim/slim",
            "version": "3.3.0",
            "source": {
                "type": "git",
                "url": "https://github.com/slimphp/Slim.git",
                "reference": "939f2e85d57508de9cff241d10091cd972f221c3"
            },
            "dist": {
                "type": "zip",
                "url": "https://api.github.com/repos/slimphp/Slim/zipball/939f2e85d57508de9cff241d10091cd972f221c3",
                "reference": "939f2e85d57508de9cff241d10091cd972f221c3",
                "shasum": ""
            },
            "require": {
                "container-interop/container-interop": "^1.1",
                "nikic/fast-route": "^0.6",
                "php": ">=5.5.0",
                "pimple/pimple": "^3.0",
                "psr/http-message": "^1.0"
            },
            "require-dev": {
                "phpunit/phpunit": "^4.0",
                "squizlabs/php_codesniffer": "^2.5"
            },
            "type": "library",
            "autoload": {
                "psr-4": {
                    "Slim\\": "Slim"
                }
            },
            "notification-url": "https://packagist.org/downloads/",
            "license": [
                "MIT"
            ],
            "authors": [
                {
                    "name": "Rob Allen",
                    "email": "rob@akrabat.com",
                    "homepage": "http://akrabat.com"
                },
                {
                    "name": "Josh Lockhart",
                    "email": "hello@joshlockhart.com",
                    "homepage": "https://joshlockhart.com"
                },
                {
                    "name": "Gabriel Manricks",
                    "email": "gmanricks@me.com",
                    "homepage": "http://gabrielmanricks.com"
                },
                {
                    "name": "Andrew Smith",
                    "email": "a.smith@silentworks.co.uk",
                    "homepage": "http://silentworks.co.uk"
                }
            ],
            "description": "Slim is a PHP micro framework that helps you quickly write simple yet powerful web applications and APIs",
            "homepage": "http://slimframework.com",
            "keywords": [
                "api",
                "framework",
                "micro",
                "router"
            ],
            "time": "2016-03-10 21:37:40"
        }
```

PR #1738, opened 2016-01-14
Improve Router.php namedRoutes caching

jacklynmatts:
pathFor() should function properly ( and not throw an exception) during the following sequence.

1) call pathFor()
2) map() a new named route
3) call pathFor() for the route added in step 2.

Therefore, getNamedRoute() should be more precise than is_null($this->namedRoutes).  Comparing the size of namedRoutes to routes, while not perfect, is better.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -58,13 +58,6 @@ class Router implements RouterInterface
     protected $routeCounter = 0;
 
     /**
-     * Named routes
-     *
-     * @var null|Route[]
-     */
-    protected $namedRoutes;
-
-    /**
      * Route groups
      *
      * @var RouteGroup[]
@@ -199,13 +192,12 @@ class Router implements RouterInterface
      */
     public function getNamedRoute($name)
     {
-        if (count($this->namedRoutes) != count($this->routes)) {
-            $this->buildNameIndex();
+        foreach ($this->routes as $route) {
+            if ($name == $route->getName()) {
+                return $route;
+            }
         }
-        if (!isset($this->namedRoutes[$name])) {
-            throw new RuntimeException('Named route does not exist for name: ' . $name);
-        }
-        return $this->namedRoutes[$name];
+        throw new RuntimeException('Named route does not exist for name: ' . $name);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


akrabat:
Shouldn't we be invalidating `namedRoutes` when `map` is called?

jacklynmatts:
That would be a better solution. Unfortunately, after thinking about this a little further, I'm not sure even that fully solves the problem.

It's absolutely possible to call `Route::setName` multiple times.  This means that the route name could change even after `map` is called - and I can't think of any way to observe that inside of `Router`. This is, to be a fair, an edge case.

I think the simplest way to solve that would be to completely eliminate `namedRoutes`, unless you can think of something else?

ghost:
hi
The only way for fixing this problem is removing `Route::setName` method and insert a feature for registering named routes like other routes in `App` class.
for more info see elegantweb/framework routing in github :D
[Route class](https://github.com/elegantweb/framework/blob/master/src/Elegant/Routing/Route.php) and [RouteCollection class](https://github.com/elegantweb/framework/blob/master/src/Elegant/Routing/RouteCollection.php#L117-L128)

jacklynmatts:
Thanks @n0o0bSina , perhaps I should have been more clear.  I wasn't suggesting that the named routes feature be removed.  Rather, I'm now suggesting that `getNamedRoute` loops through each `Route` object in `routes` on every call, as this seems the only way to ensure we're working with accurate, current information. This would result in removal of the `Router::namedRoutes` property, since it would be unused.

Sure, it's less performant, but that seems like a secondary concern to being correct. Does it make sense to just correct the error, and address any performance issues in another PR?

I've updated the pull request taking this conversation into consideration.

ghost:
@jacklynmatts there is more better way if you don't like standards :D You can inject `Router` class into each `Route` then set key in both `Route` and `Router`.

akrabat:
Ah - so the problem is that the `Router` holds the list of named routes, but we set the name on the `Route`, so we can't invalidate the cached list in `Router`.

akrabat:
Thanks for finding and fixing this.

jacklynmatts:
Thanks!

PR #1739, opened 2016-01-18
Ensure that responses with an indeterminate-length body have their content sent

jj101k:
In cases where the body getSize() returns null and no length has been provided via header, no content will be sent due to ceil(null / $chunkSize) == 0. This affects custom body objects as would be used to stream data in cases where waiting for the body to be complete would be prohibitively expensive and no size can be determined in advance, for example the results of "tail -0f /var/log/messages | head" or "du -sk /home/*".

This fixes the issue by doing the loop without chunk counting if the size is not known.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -385,16 +385,25 @@ class App
             if (!$contentLength) {
                 $contentLength = $body->getSize();
             }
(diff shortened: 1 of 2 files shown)


akrabat:
Good catch. Thank you!

PR #1745, opened 2016-01-26
Register the default services from a Provider.

mducharme:
By splitting the default Slim Container in 2 (The `Container` and its associated `Service Provider` to register services) this allows one to use a custom (Pimple-based) container and still auto-register all required service for Slim.

It does not solve the more generic problem of having a reusable provider (because container-interop does not have a "setter" method to register service) but it is, I believe, one step in the right direction. In other words, this won't work more for general-purpose container, but it at least allows any Pimple container to register Slim services without having to have a bridge to the SlimContainer.


vlakoff:
I agree having the services registered in the container class looked very strange at first glance.

Slim/ServiceProvider.php
```diff
@@ -0,0 +1,177 @@
...
+ * Slim's default Service Provider.
+ */
+class ServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * Register Slim's default services.
+     *
+     * @param Container $container
+     */
+    public function register(PimpleContainer $container)
```
akrabat: Please remove the type hint. Any DIC that uses array access with closures would in theory work with this...

Slim/ServiceProvider.php
```diff
@@ -0,0 +1,177 @@
...
+use Slim\Http\Response;
+use Slim\Interfaces\CallableResolverInterface;
+use Slim\Interfaces\Http\EnvironmentInterface;
+use Slim\Interfaces\InvocationStrategyInterface;
+use Slim\Interfaces\RouterInterface;
+
+/**
+ * Slim's default Service Provider.
+ */
+class ServiceProvider implements ServiceProviderInterface
```
akrabat: Can we rename to `DefaultServicesProvider`?

akrabat: Also, again no need to implement ServiceProviderInterface. We can simply instantiate this class and call `register()` directly.

Slim/ServiceProvider.php
juliangut: Wouldn't it be better to code against `ContainerInterface` instead

Additionally check for services with that interface instead of ArrayAccess?

``` php
if (!$container->has('environment')) {
}
```

mducharme: @juliangut I ported the service code exactly as it was previously in the container.

I don't disagree with the idea; I just think that it's a bit outside the scope of that PR.

akrabat:
Thanks.

PR #1764, opened 2016-02-11
Write to the error log if displayErrorDetails is false

### LINKED ISSUE #1761: Write to error_log when displayErrorDetails is false
When `displayErrorDetails` is false (the default), the ErrorHandler doesn't display any information. It should write to the `error_log` in this situation so that people who do not know about the `displayErrorDetails` setting can see that that Slim knows what went wrong. 

Bonus points for writing a note about the `displayErrorDetails` setting into the error_log too.

akrabat:
When displayErrorDetails is false, then the developer doesn't get any information on the exception. Add this information to the error log so that they can find it.

Closes #1761

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -72,6 +72,8 @@ class Error
                 break;
         }
 
+        $this->writeToErrorLog($exception);
+
         $body = new Body(fopen('php://temp', 'r+'));
         $body->write($output);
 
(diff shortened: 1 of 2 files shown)


akrabat:
ping @silentworks, @codeguy

Slim/Handlers/Error.php
```diff
@@ -81,6 +83,66 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+    /**
+     * Render exception as Text.
+     *
+     * @param Exception $exception
+     *
+     * @return string
+     */
+    protected function renderTextException(Exception $exception)
+    {
+        $html = sprintf('Type: %s' . PHP_EOL, get_class($exception));
```
campersau: Nit: Why is this variable named `$html` instead of `$text`? The docblock says`Render exception as Text.`

Slim/Handlers/Error.php
```diff
@@ -81,6 +83,66 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+        if (($file = $exception->getFile())) {
+            $html .= sprintf('File: %s' . PHP_EOL, $file);
+        }
+
+        if (($line = $exception->getLine())) {
+            $html .= sprintf('Line: %s' . PHP_EOL, $line);
+        }
+
+        if (($trace = $exception->getTraceAsString())) {
+            $html .= 'Trace: ';
```
campersau: NIT: Why not 

``` php
$html .= sprintf('Trace: %s', $trace);
```

as everywhere else?

akrabat:
@campersau Updated. Thank you!

geggleto:
@akrabat thoughts on just including monolog?

akrabat:
Don't want to. It's not the only PSR-3 logging library

silentworks:
Looks good to me.

akrabat:
/me waits for @silentworks to merge it…

PR #1765, opened 2016-02-11
Expose the Router's path builder

mcaskill:
This allows paths to be built from arbitrary routes or delegate complex patterns to other objects.

Useful for application's with flexible/catch-all routes like a hierarchical page system or for named routes that need to be language-locked, e.g.: `{slug:projet|fr/projet|project|en/project}/{id}`.

``` php
$url = $container['router']->parseRoutePattern(
    'fr/projet/{id}',
    [
        'id' => 'mon-profile-geek'
    ]
);
```

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -253,22 +253,18 @@ class Router implements RouterInterface
     }
 
     /**
-     * Build the path for a named route excluding the base path
+     * Build the path for a route pattern excluding the base path
      *
-     * @param string $name        Route name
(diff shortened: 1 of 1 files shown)


silentworks:
Is this not just `pathFor` without the using named routes? if this is the case why not just create named routes?

mcaskill:
Reviewing my application, indeed this was a silly proposition.

Keep up the great work with the framework!

PR #1767, opened 2016-02-14
Compound services before Pimple container creation

juliangut:
Compound default services and those provided to container creation before adding them to Pimple container __construct

Avoids the need to continuously verify for service existence by defining them in advance

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -9,10 +9,7 @@
 namespace Slim;
 
 use Interop\Container\ContainerInterface;
-use Interop\Container\Exception\ContainerException;
 use Pimple\Container as PimpleContainer;
-use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ServerRequestInterface;
 use Slim\Exception\ContainerValueNotFoundException;
 
 /**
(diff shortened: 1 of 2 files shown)


silentworks:
This PR could makes this unusable with other Pimple implementations.

juliangut:
I'm not aware of other Pimple implementations sorry, how would this break them?

silentworks:
By default if you want to add a ServiceProvider to Pimple you would do.

``` php
$container = new Pimple\Container();
$container->register(new Slim\DefaultServicesProvider());
```

With your change this would no longer be possible. I also cannot see any advantages in what you are doing.

juliangut:
How about..

``` php
$container = new Pimple\Container(Slim\DefaultServicesProvider::getDefaultServices());
```

Instead of creating the object and then injecting the dependencies I prefere the constructor DI when possible. So to accomplish that here default services (in an array) are merged with provided services in `$values` array and then create the container with them. This avoids the if, if, if, ... list of statements that is present in `DefaultServicesProvider::register`

It doesn't add any extra feature or fixes anything but adds a little bit of clarity (IMO) Haven't even tested if it's faster or not.

Anyway I understand that anyone used to Pimple that sees a `*Provider` class might expect it to work as you wrote

akrabat:
Sorry, but I don't see the advantage to this as I think it is less explicit than the `isset` checks. i'm closing it. Please reopen if there's a performance advantage that we should be considering.

Issue #1758, opened 2016-02-08
UploadedFile file property missing


geggleto:
In https://github.com/slimphp/Slim/blob/3.x/Slim/Http/UploadedFile.php#L143

It does not exist as a protected member.

silentworks:
Fixed.

PR #1768, opened 2016-02-14
add missed class property (issue #1758)

krydos:
as @geggleto noticed, the UploadedFile class doesn't have **$file** property defined. This property dynamically defined in constructor which is fine but doesn't look clear.

### DIFF
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -29,7 +29,6 @@ class UploadedFile implements UploadedFileInterface
      * @var string
      */
     protected $file;
-
     /**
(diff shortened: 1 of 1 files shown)


Slim/Http/UploadedFile.php
```diff
@@ -24,6 +24,13 @@
 class UploadedFile implements UploadedFileInterface
 {
     /**
+     * The client-provided full path to the file
+     *
+     * @var string
+     */
+    protected $file;
+
```
campersau: Nit: remove blank line here for consistency with the code below?

akrabat:
Thanks.

PR #1769, opened 2016-02-14
fix typo in error handler.

krydos:
s/Errror/Error - there was a typo in "Errror" word in method **writeToErrorLog**


akrabat:
Good catch. Thank you.

PR #1770, opened 2016-02-14
Support PHP7+ Errors

### LINKED ISSUE #1762: Catch PHP 7 Errors with the ErrorHandler
PHP 7 [Errors](http://php.net/Error) needs to be handled by the ErrorHandler

akrabat:
Implement a new error handler, `PhpError` which handles PHP 7's new `Throwable` interface.

This fixes #1762.

Note that I couldn't just use the current `Error` handler as I'd have had to have changed the method signatures which would have broken BC. I've chosen to catch `Throwable` in this new handler so that when we go to 4.0, we can simply remove the current `Error` handler and this one can then be used to catch both Exceptions and PHP 7+ Errors.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -9,6 +9,7 @@
 namespace Slim;
 
 use Exception;
+use Throwable;
 use Closure;
 use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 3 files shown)


akrabat:
Explicit ping for @codeguy & @silentworks :) Thoughts appreciated from everyone else!

geggleto:
I like it. But we as we build new features into Slim you can really start to see some of the short-comings.

1) Hard-coded content-types
2) No Handler Interfaces

These 2 issues force a user to extend the current class and override the methods instead of simply implementing an interface. I don't think its a big deal breaker, but these sorts of issues should be cataloged somewhere for 4.0

PR #1774, opened 2016-02-19
Added dynamic exception handling functionality

vrxj81:
This change enables developers to develop error handlers for every time of exception they want.
In the settings they add handlers in the form "ExceptionType" => "handler".
Example, I use Eloquent ORM and i want to handle the ModelNotFoundException: 

```
<?php
return [
    'settings' => [
        'displayErrorDetails' => true,

        // Renderer settings
        'renderer' => [
            'template_path' => __DIR__ . '/../templates/',
        ],

        // Monolog settings
        'logger' => [
            'name' => 'slim-app',
            'path' => __DIR__ . '/../logs/app.log',
        ],
        // Custom Error Handlers
        'handlers' => [
            'ModelNotFoundException' => 'modelNotFoundHandler',
         ],
```

Then the developer injects the custom errorhandler like describer in the Slim Docs: http://www.slimframework.com/docs/handlers/error.html

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -590,8 +590,15 @@ class App
             // This is a Stop exception and contains the response
             return $e->getResponse();
         } else {
-            // Other exception, use $request and $response params
-            $handler = 'errorHandler';
+            // Determine if a handler is defined for the exception
+            $handlers = $this->container->get('settings')['handlers'];
+            if (in_array(get_class($e, $handlers))
+            {
+              $handler = $settings['handlers'][get_class($e)];
+            } else {
+              // Other exception, use $request and $response params
+              $handler = 'errorHandler';
+            }
             $params = [$request, $response, $e];
         }
 
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -592,7 +592,7 @@ class App
         } else {
             // Determine if a handler is defined for the exception
             $handlers = $this->container->get('settings')['handlers'];
-            if (in_array(get_class($e, $handlers))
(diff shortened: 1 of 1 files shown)


silentworks:
Any reason why you wouldn't just handle this in a custom handler? Also you are missing a `)` on line 595.

vrxj81:
1 customer handler for 10(or more) different Exception? I'd like to seperate the handling.

silentworks:
Nothing stopping you from separating them. I think exceptions should be handled in user land code, so you can try and catch them where necessary otherwise create a custom handler that passes it off to the correct handler.

geggleto:
:-1: You should be catching your own errors, it's not Slim's job.

vrxj81:
I know, but it offers a mechanism.

akrabat:
This could be done in Middleware and so I'm loathe to add to core. FWiW, the middleware looks like this:

``` php
    try {
    } catch (\Exception $e) {
        $handler = $this->get('settings')['handlers'][get_class($e)] ?? null;
        if ($this->has($handler)) {
            return $this->get($handler)($request, $response, $e);
        }
        throw $e;
    }
});
```

(Also, I love PHP 7!)

alexweissman:
so, what _would_ be the proper way to use Monolog with Slim 3?  Right now I'm experimenting with injecting a Monolog `Logger` object into a custom error handler, and writing to the `Logger` instead of using `error_log`.

akrabat:
One way: https://akrabat.com/logging-errors-in-slim-3/

PR #1777, opened 2016-02-23
fix renderJsonErrorMessage -  error variable was overriding it self

nogo:
In the function renderJsonErrorMessage error was defined as parameter and array. So it was overriding itself.


akrabat:
Good catch. Thank you.

PR #1779, opened 2016-02-26
Added application/vnd.api+json media type by default

Big-Shark:
Hello, I am started a new project today, slim + json api, and I had a problem because slim returned empty data, and I spent many time searching a reason, and I think you need add this format by default.

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -30,6 +30,7 @@ class Error
      */
     protected $knownContentTypes = [
         'application/json',
+        'application/vnd.api+json',
         'application/xml',
         'text/xml',
         'text/html',
@@ -59,6 +60,7 @@ class Error
         $contentType = $this->determineContentType($request);
         switch ($contentType) {
             case 'application/json':
+            case 'application/vnd.api+json':
                 $output = $this->renderJsonErrorMessage($exception);
                 break;
 
--- a/Slim/Handlers/NotAllowed.php
+++ b/Slim/Handlers/NotAllowed.php
@@ -27,6 +27,7 @@ class NotAllowed
      */
     protected $knownContentTypes = [
         'application/json',
+        'application/vnd.api+json',
         'application/xml',
         'text/xml',
         'text/html',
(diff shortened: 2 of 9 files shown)


geggleto:
Mind adding text/javascript  as well?

akrabat:
Also, please update ether Error and PhpError handlers.

geggleto:
Needs Unit tests for these new content types

Big-Shark:
@geggleto @akrabat Fix Error and PhpError, done. Add tests, done.

tuupola:
I think this is a wrong way to do it. There are gazillion of custom vendor mediatypes. You cannot add them all to Slim core. The `application/vnd.api+json` is not an general mediatype for JSON based APIs. Instead it is specific for a vendor called [JSON API](http://jsonapi.org/). 

More clear example is `application/vnd.github.v3+json` which can is parsed as vendor prefix `vnd`, vendor `github` version `3` datatype `json`.

So instead of adding all different vendor prefixed mediatypes to Slim core correct way would be to just parse the vendor mediatypes into their general mediatype equivalents. For example `application/vnd.api+json` becomes `application/json` and `application/vnd.foobar+xml` becomes  `application/xml`.

akrabat:
Good point.

akrabat:
The problem is that you can't just convert the mediatype in the request in middleware as you'll need it to understand the parsed data you get back out of `getParsedBody()`.

Do we need to update the way the media parser works so that `+json` and `+xml` are detected and handled correctly?

Big-Shark:
@akrabat Yes, I think @tuupola offered a good option, but you need time for rewrite media parser, and why would you don't merge this commit, and start rewrite for next version?

tuupola:
@akrabat I think you are correct. Media parser should be updated.

akrabat:
@Big-Shark,

Until we sort out the way we are most happy with, you can use middleware to register the parser you need:

``` php
    $request->registerMediaTypeParser('tapplication/vnd.api+json', function ($input) {
        return json_decode($input, true);
    });

});

```

Big-Shark:
@akrabat hahaha, this is my code now, but I think as long as we try to make decisions, many people may face with this problem, as jsonapi quite popular format.

akrabat:
Closed in favour of #1781

PR #1780, opened 2016-02-27
Deprecate Request and Response in Container

akrabat:
This PR adds deprecation notices when you retrieve the Request, Response or Environment from the Container. It's very confusing to have these in the Container as they are "stale" and should never be used.

It updates `App::run()` so that it no longer uses the container for these objects, but instead creates them itself. I've also updated `run()`'s signature so that you can now pass in a request and a response object if you want to use your own.

Similarly, I've updated `App::subRequest` to create it's own Environment and Response objects.

Would appreciate people's thoughts on this.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -659,16 +659,14 @@ class App
 
             // register our new environment with the container for BC
             if ($this->container instanceof \ArrayAccess) {
-                if (!$this->container->has('environment')) {
-                    $container['environment'] = function () use ($environment) {
-                        trigger_error(
-                            'Retrieving the environment from the container is deprecated; '
-                            . 'update your code to use the one within the Request object.',
-                            E_USER_DEPRECATED
-                        );
-                        return $environment;
-                    };
-                }
+                $container['environment'] = function () use ($environment) {
+                    trigger_error(
+                        'Retrieving the environment from the container is deprecated; '
+                        . 'update your code to use the one within the Request object.',
+                        E_USER_DEPRECATED
+                    );
+                    return $environment;
+                };
             }
         }
 
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -498,9 +498,7 @@ class App
         $request = new Request($method, $uri, $headers, $cookies, $serverParams, $body);
 
         if (!$response) {
-            $headers = new Headers(['Content-Type' => 'text/html; charset=UTF-8']);
-            $response = new Response(200, $headers);
-            $response = $response->withProtocolVersion($this->container->get('settings')['httpVersion']);
+            $response = $this->createResponse();
         }
 
         return $this($request, $response);
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -721,5 +721,7 @@ class App
                 return $response;
             };
         }
+
+        return $response;
     }
 }
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -293,13 +293,11 @@ class App
     public function run($silent = false, ServerRequestInterface $request = null, ResponseInterface $response = null)
     {
         if (!$request) {
-            $request = Request::createFromEnvironment(new Environment($_SERVER));
+            $request = $this->createRequest();
         }
 
         if (!$response) {
-            $headers = new Headers(['Content-Type' => 'text/html; charset=UTF-8']);
-            $response = new Response(200, $headers);
-            $response = $response->withProtocolVersion($this->container->get('settings')['httpVersion']);
+            $response = $this->createResponse();
         }
 
         $response = $this->process($request, $response);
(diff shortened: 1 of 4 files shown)


mathmarques:
:+1: 
Would this not be a BC break?
If someone changes the container request and response object before calling run(), they would expect that Slim uses these objects, but instead, Slim will create a new request and response.

akrabat:
That's a good question. They would have to change their code to send the request and response in via `run()`.

Maybe we need a new method then.

mathmarques:
I guess everything should work now! 
Good job :+1:

mnapoli:
I'm [still](https://github.com/slimphp/Slim/pull/1179#issuecomment-94328618) definitely :+1: on that ;) Those are value objects, not stateless services, so IMO they shouldn't be in the container.

akrabat:
@mnapoli I'm annoyed with myself for not thinking of changing the signature to run() back then!

mathmarques:
@akrabat Wouldn't be good to call `createResponse()` instead of this:
https://github.com/akrabat/Slim/blob/deprecate-request-response-env-from-container/Slim/App.php#L501-L503

akrabat:
Yes - I've updated.

Slim/DefaultServicesProvider.php
```diff
@@ -37,47 +37,6 @@ class DefaultServicesProvider
...
-             *
-             * @return ResponseInterface
-             */
-            $container['response'] = function ($container) {
-                $headers = new Headers(['Content-Type' => 'text/html; charset=UTF-8']);
-                $response = new Response(200, $headers);
-
-                return $response->withProtocolVersion($container->get('settings')['httpVersion']);
-            };
-        }
```
mnapoli, line 79: Why not leave these here but not actually use them in `createRequest()` and `createResponse()`? That way you keep BC, you can tag these entries as deprecated and `App` is much simpler.

Slim/App.php
```diff
@@ -481,7 +488,7 @@ public function subRequest(
         $bodyContent = '',
         ResponseInterface $response = null
     ) {
-        $env = $this->container->get('environment');
+        $env = new Environment($_SERVER);
```
mnapoli, line 491: Do you think it would make sense to have something like `Environment::fromGlobal()` that would do the same thing?

Slim/DefaultServicesProvider.php
akrabat, line 79: If someone has added their own request/response to the container before calling run() they will expect that their container will be used.  The easiest way to tell that this has happened is to remove our default implementations.

Slim/App.php
akrabat, line 491: Maybe, but that would be a different PR.

Slim/DefaultServicesProvider.php
mnapoli, line 79: Right sorry that was probably what you meant before but I didn't understand it that way.

Slim/App.php
mnapoli, line 491: :+1:

Slim/App.php
```diff
@@ -630,4 +637,91 @@ protected function handlePhpError(Throwable $e, ServerRequestInterface $request,
...
+        }
+
+        if ($this->container->has('environment')) {
+            $environment = $this->container->get('environment');
+        } else {
+            $environment = new Environment($_SERVER);
+
+            // register our new environment with the container for BC
+            if ($this->container instanceof \ArrayAccess) {
+                if (!$this->container->has('environment')) {
```
vlakoff: Do we need this test? The `else` line 657 tells us the container doesn't have environment.

akrabat: Good point. Removed.

juliangut:
How would this affect slim/twig-view `TwigExtension`? It extracts the request from container to retrieve basePath, how would this be done instead?

``` php
// Register Twig View helper
$container['view'] = function ($c) {
    $view = new \Slim\Views\Twig('path/to/templates', [
        'cache' => 'path/to/cache'
    ]);

    // Instantiate and add Slim specific extension
    $basePath = rtrim(str_ireplace('index.php', '', $c['request']->getUri()->getBasePath()), '/');
    $view->addExtension(new Slim\Views\TwigExtension($c['router'], $basePath));

    return $view;
};
```

mathmarques:
We still be causing a BC break. If someone try to get request/response from container before call `run`.
Good catch @juliangut, but in that case it won't generate error unless you use view before call `run`(no problem using inside middleware/route callback).

akrabat:
Still more thinking required!

geggleto:
I love this.
Yes plz.
:+1:

geggleto:
Should we not put into the container actual Factories for Request/Response/Environment?

Slim/App.php
geggleto, line 491: @akrabat this should be put elsewhere so we can test it properly.

Slim/App.php
```diff
@@ -288,10 +290,15 @@ public function group($pattern, $callable)
      * @throws MethodNotAllowedException
      * @throws NotFoundException
      */
-    public function run($silent = false)
+    public function run($silent = false, ServerRequestInterface $request = null, ResponseInterface $response = null)
     {
-        $request = $this->container->get('request');
-        $response = $this->container->get('response');
+        if (!$request) {
+            $request = $this->createRequest();
```
geggleto, line 296: I really think these should be put on the actual Http Object.

akrabat:
> How would this affect slim/twig-view TwigExtension? It extracts the request from container to retrieve basePath, how would this be done instead?

It won't affect it. This PR registers request and response with the container so that when the view is retrieved from the container, the objects are there.

However, we need a new pattern as a deprecation notice will be raised.

juliangut:
> However, we need a new pattern as a deprecation notice will be raised.

That was what I was refering to :smile:, I see a deprecation notice when there was none as a way of affecting behavior

akrabat:
Needs to be done in 4.0.

Issue #1683, opened 2015-12-16
Add to Media Parsers


geggleto:
Weird issue in IE.
http://www.entwicklungsgedanken.de/2008/06/06/problems-with-internet-explorer-and-applicationjson/
Add text/javascript for JSON

codeguy:
We'll address this soon. I want to discuss future road map with @silentworks and @akrabat first.

geggleto:
@codeguy I found a work around for this issue this morning.

By overriding the default container factory for request, it is possible to register a media parser before the App class is created.

Example:

``` php
//...
// Required to manually create the container
$container = new \Slim\Container();
// Create new factory 
$container['request'] = function ($c) {
    // construct request from environment
    $request = \Slim\Http\Request::createFromEnvironment($c['env']);
    // add media parser
    $request->registerMediaTypeParser("text/javascript", function ($input) {
        return json_decode($input, true);
    });
};

// Construct App
$app = new Slim\App($container);
//...
```

PR #1781, opened 2016-02-27
Support structured suffix body parsers (+json/+xml)

akrabat:
Add body parsers for structured syntax suffixes (RFC 6839) - that is, media types with the format {anything}+json and {anything}+xml

This is a more generic solution to that addressed in https://github.com/slimphp/Slim/pull/1779. 

We'll have to sort out the error handlers in a separate PR.


akrabat:
ping @silentworks, @codeguy

geggleto:
Do you think I could get #1683  included since you are mucking about? :)

PR #1782, opened 2016-02-27
Support +json & +xml accept headers when creating error responses

akrabat:
Support structured suffix body parsers (+json/+xml) for error handling.

This means that if the request has an Accept header of "application/hal+xml" and an error condition occurs, then the response will have an content type of "application/xml" rather than "text/html" as it is today.

This is the flip side to #1781

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -295,6 +295,14 @@ class Error
             return $selectedContentTypes[0];
         }
 
+        // handle +json and +xml specially
+        if (preg_match('/\+(json|xml)/', $acceptHeader, $matches)) {
+            $mediaType = 'application/' . $matches[1];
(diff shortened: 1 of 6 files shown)


akrabat:
ping @silentworks, @codeguy

Big-Shark:
@akrabat :+1:  Cool

Big-Shark:
@akrabat https://github.com/akrabat/Slim/commit/4f4c65d91a25bfec459f3ac73a2707522ea5a3f1 DRY?
Maybe we should use trait?

akrabat:
It's intentional. If we create a trait, others will use it and this code is not suitable for that. 

If you want to deal the the Accept header for non-error handling, you should use Will Durrand's [Negotiation](https://github.com/willdurand/Negotiation) package.

nagaraj1245:
which is the main slim framework file?..

Issue #1785, opened 2016-02-28
The actual callable is not passed to InvocationStrategy since introduction of DeferredCallable in Slim 3.2.0


astax-t:
Since introduction of `DeferredCallable` in Slim 3.2.0, the actual controller's callable is not given to the class implementing InvocationStrategy (the one set in `$container['foundHandler']`)

This makes it impossible to use reflection to find what are actual expected parameters.

I'm trying to use a custom `InvocationStrategy` that would match controller parameters names to route parameters - similar to the way it's done in Symfony HttpKernel. This was working in Slim 3.1.0, but not in 3.2.0.

astax-t:
The pull request https://github.com/slimphp/Slim/pull/1783 would help to mitigate this, if accepted. I'd be able to retrieve the resolved callable and get its parameters using reflection.

But really it would be better to call the custom InvocationStrategy from DeferredCallable instead of just passing arguments given to `DeferredCallable ::__invoke()`

irfanevrens:
I am agree with @astax-t and I also use a custon InvocationStrategy.

michaelboke:
Having the same issue since 3.2 my auto wiring isn't working anymore.

@irfanevrens i added a pull request on #1783 that now passes the resolvedCallable result to the foundHandler. Restoring the old behavior.

akrabat:
Can you provide an example of code that's no longer working please?

astax-t:
Sure. It's not the minimal code to replicate the problem, but pretty much copied from HttpFoundation:

```
class CompatibilityInvocationStrategy implements InvocationStrategyInterface
{
    public function __invoke(
        callable $callable,
        ServerRequestInterface $request,
        ResponseInterface $response,
        array $routeArguments
    )
    {
        if (is_array($callable))
        {
            $r = new \ReflectionMethod($callable[0], $callable[1]);
        } elseif (is_object($callable) && !$callable instanceof \Closure)
        {
            /** @noinspection PhpParamsInspection */
            $r = new \ReflectionObject($callable);
            $r = $r->getMethod('__invoke');
        } else
        {
            $r = new \ReflectionFunction($callable);
        }

        $arguments = $this->doGetArguments($request, $response, $routeArguments, $r->getParameters());

        $result = call_user_func_array($callable, $arguments);

        return $result;
    }

    protected function doGetArguments($request, $response, $attributes, array $parameters)
    {
        $arguments = array();
        foreach ($parameters as $param)
        {
[...]

PR #1787, opened 2016-02-29
Resolve deferred callable before passing it to foundHandler for invocation

astax-t:
Fixes issue #1785

Includes https://github.com/slimphp/Slim/pull/1783

### DIFF
--- a/Slim/DeferredCallable.php
+++ b/Slim/DeferredCallable.php
@@ -1,6 +1,11 @@
 <?php
-
-
+/**
+ * Slim Framework (http://slimframework.com)
+ *
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2016 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/3.x/LICENSE.md (MIT License)
+ */
 namespace Slim;
 
 use Closure;
@@ -10,12 +15,15 @@ class DeferredCallable
 {
     use CallableResolverAwareTrait;
 
+    /** @var callable|string */
     private $callable;
+
     /** @var  ContainerInterface */
     private $container;
 
     /**
      * DeferredMiddleware constructor.
+     *
      * @param callable|string $callable
      * @param ContainerInterface $container
      */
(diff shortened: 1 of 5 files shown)


Slim/DeferredCallable.php
```diff
@@ -27,13 +35,15 @@ public function __construct($callable, ContainerInterface $container = null)
 
     public function __invoke()
     {
+        return call_user_func_array($this->getResolvedCallable(), func_get_args());
+    }
+
+    public function getResolvedCallable()
+    {
         $callable = $this->resolveCallable($this->callable);
```
akrabat, line 43: Should we do this only once? i.e. should `getResolvedCallable()` store the resolved callable into `$this->resolvedCallable` and then return that rather than re-resolving?

Slim/Route.php
```diff
@@ -314,13 +314,19 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
         /** @var InvocationStrategyInterface $handler */
         $handler = isset($this->container) ? $this->container->get('foundHandler') : new RequestResponse();
 
+        if ($this->callable instanceof DeferredCallable) {
+            $callable = $this->callable->getResolvedCallable();
+        } else {
+            $callable = $this->callable;
+        }
```
akrabat, line 321: Why is this required? i.e. if the problem is that you can't get at the real callable in the invocation strategy, then this is solved by the availability of `getResolvedCallable()` which could be called in the the invocation strategy's `__invoke()` if required?

akrabat:
I think #1789 is the correct way to solve this.

Slim/Route.php
astax-t, line 321: It's possible, of course, but `DeferredCallable` is an implementation detail and my custom `InvocationStrategy` doesn't need to know about it. So it's reasonable to expect it to get the original callable as the first argument but not some wrapper.

Alternatively, this should be documented.

akrabat, line 321: You are correct, which is why #1789 is a nicer solution as we don't actually need a `DeferredCallable` at all for the route callable.

astax-t:
I agree, https://github.com/slimphp/Slim/pull/1789 is better way

PR #1800, opened 2016-03-06
More flexibility for URL parameters

aimeos:
Use query parameter in route path if no URL parameter is explicitly available.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -276,7 +276,7 @@ class Router implements RouterInterface
         // The most specific is last, so we look for that first.
         $routeDatas = array_reverse($routeDatas);
 
-        $segments = [];
+        $segments = $segmentKeys = [];
         foreach ($routeDatas as $routeData) {
             foreach ($routeData as $item) {
                 if (is_string($item)) {
@@ -295,6 +295,7 @@ class Router implements RouterInterface
                     break;
                 }
                 $segments[] = $data[$item[0]];
+                $segmentKeys[$item[0]] = true;
             }
             if (!empty($segments)) {
                 // we found all the parameters for this route data, no need to check
@@ -308,7 +309,7 @@ class Router implements RouterInterface
         }
         $url = implode('', $segments);
 
-        $params = array_merge(array_diff($data, $segments), $queryParams);
+        $params = array_merge(array_diff_key($data, $segmentKeys), $queryParams);
         if ($params) {
             $url .= '?' . http_build_query($params);
         }
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -256,19 +256,18 @@ class Router implements RouterInterface
      * Build the path for a named route excluding the base path
      *
      * @param string $name        Route name
-     * @param array  $params      URL and query parameters
-     * @param array  $deprecated  Only for BC. Use $params instead
+     * @param array  $data        Named argument replacement data
+     * @param array  $queryParams Optional query string parameters
      *
      * @return string
      *
      * @throws RuntimeException         If named route does not exist
      * @throws InvalidArgumentException If required data not provided
      */
-    public function relativePathFor($name, array $params = [], array $deprecated = [])
+    public function relativePathFor($name, array $data = [], array $queryParams = [])
     {
         $route = $this->getNamedRoute($name);
         $pattern = $route->getPattern();
-        $params = $params + $deprecated;
 
         $routeDatas = $this->routeParser->parse($pattern);
         // $routeDatas is an array of all possible routes that can be made. There is
@@ -277,7 +276,6 @@ class Router implements RouterInterface
         // The most specific is last, so we look for that first.
         $routeDatas = array_reverse($routeDatas);
 
-        $args = [];
         $segments = [];
         foreach ($routeDatas as $routeData) {
             foreach ($routeData as $item) {
@@ -288,19 +286,15 @@ class Router implements RouterInterface
                 }
 
                 // This segment has a parameter: first element is the name
-                if (array_key_exists($item[0], $params)) {
-                    $segments[] = $params[$item[0]];
-                    $args[$item[0]] = null;
-                    continue;
+                if (!array_key_exists($item[0], $data)) {
+                    // we don't have a data element for this segment: cancel
+                    // testing this routeData item, so that we can try a less
+                    // specific routeData item.
+                    $segments = [];
+                    $segmentName = $item[0];
+                    break;
                 }
-
-                // we don't have a data element for this segment: cancel
-                // testing this routeData item, so that we can try a less
-                // specific routeData item.
-                $args = [];
-                $segments = [];
-                $segmentName = $item[0];
-                break;
+                $segments[] = $data[$item[0]];
             }
             if (!empty($segments)) {
                 // we found all the parameters for this route data, no need to check
@@ -313,10 +307,10 @@ class Router implements RouterInterface
             throw new InvalidArgumentException('Missing data for URL segment: ' . $segmentName);
         }
         $url = implode('', $segments);
-        $queryParams = array_diff_key($params, $args);
 
-        if ($queryParams) {
-            $url .= '?' . http_build_query($queryParams);
+        $params = array_merge(array_diff($data, $segments), $queryParams);
+        if ($params) {
+            $url .= '?' . http_build_query($params);
         }
 
         return $url;
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -256,18 +256,19 @@ class Router implements RouterInterface
      * Build the path for a named route excluding the base path
      *
      * @param string $name        Route name
-     * @param array  $data        Named argument replacement data
-     * @param array  $queryParams Optional query string parameters
+     * @param array  $params      URL and query parameters
+     * @param array  $deprecated  Only for BC. Use $params instead
      *
      * @return string
      *
      * @throws RuntimeException         If named route does not exist
      * @throws InvalidArgumentException If required data not provided
      */
-    public function relativePathFor($name, array $data = [], array $queryParams = [])
+    public function relativePathFor($name, array $params = [], array $deprecated = [])
     {
         $route = $this->getNamedRoute($name);
         $pattern = $route->getPattern();
+        $params = $params + $deprecated;
 
         $routeDatas = $this->routeParser->parse($pattern);
         // $routeDatas is an array of all possible routes that can be made. There is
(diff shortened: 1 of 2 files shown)


silentworks:
We had this in Slim 2, maybe an oversight that we missed it out in Slim 3. I will check the code again and look at some issues to see if there was a reason its not in Slim 3.

Slim/Router.php
```diff
@@ -276,7 +276,7 @@ public function relativePathFor($name, array $data = [], array $queryParams = []
         // The most specific is last, so we look for that first.
         $routeDatas = array_reverse($routeDatas);
 
-        $segments = [];
+        $segments = $params = [];
```
SvenRtbg: Is there any benefit in keeping initialization that short? Reading this somehow suggests that $segments and $params have something important in common, but all I see after looking at it a bit longer is that they are simply set to an empty array. What about using two lines instead of one?

OptimusCrime: This syntax is odd to me. It works in Javascript, but in PHP it just looks wrong.

aimeos:
Any news for this issue besides if the initialization should be different?

silentworks:
After reading through the code more now I understand what you are trying to do, what is the use case for this as we intended for it to work the way it currently does as its more clear that `$data` will replace placeholders in `path` while `$queryParams` will be added at the end.

I think the current setup is as flexible as it gets, your change is limiting that flexibility by checking in both arrays while not necessary. If the developer wants the placeholder replaced then place the variable in the `$data` array otherwise if they want it added to the end do so in the `$queryParams` array.

akrabat:
I agree with @silentworks. At the moment there are two parameters, `$data` and `$queryParams`, which are each responsible for one thing only which makes sense.

aimeos:
For us, the separation of `$data` and `$queryParams` is rather artificial because it limits what our users (developers using the Aimeos e-commerce library) can do. This is most apparent in the generated URLs if you want to create an online shop based on Aimeos.

We have a few fixed parameters like site, locale and currency which fits very well into the `$data` array to create URLs like `http://localhost/shop1/en/`. But then, there are dynamic parameters too like the `f_catid`, `f_name` and `f_sort` for the list pages or `d_prodid`, `d_name` and others for the detail pages. These parameters do not only depend on the page but also on the used parameters in the concrete implementation.

To get nice looking URLs, the default routes provided in the aimeos-slim package to to fit for most cases but can be adapted by the developer, e.g.

```
/list[/{f_catid:[0-9]+}[/{f_name}]]
/detail/{d_prodid:[0-9]+}[/{d_name}]
```

As we don't have any control over the needed parameters for the `$data` array, we simply push all query parameter into `$data`. But if they are not used in the URL, they must be added as query string as well. There's no known way to find out if a parameter will be used as URL or query parameter so our URLs currently look like

```
/list/1/Home?f_catid=1&f_name=Home&f_sort=price
/detail/1/Test?d_prodid=1&d_name=Test&l_pos=0
```

which are very ugly and no one would like to use that in his online shop. We can't leave out the query parameters completely because then no product sorting would be possible and some parts wouldn't work at all.

Our use case is probably not the most common one where a single developer has full control over everything but we think it's a valid use case. In the current form, nobody will build an e-commerce application based on Slim and use the Aimeos components due to the URLs.

We can only ask you to rethink your decision. There's no compelling advantage of separating URL and query parameters for URL generation, but only major drawbacks like shown above. Also, no other framework does such a separation.

akrabat:
Thanks for the use-case and detailed explanation.

If we were to go down this route, then we'd deprecate the use of `$queryParams` so that the approved use of `relativePathFor` would be `relativePathFor($name, $params=[])`. That would imply that `pathFor`'s documentation would also need changing.

I've reopened for discussion on this. Certainly, I don' t want `$data` and `$queryParams` to be interchangeable.

aimeos:
In that case, the implementation can be simplified :-)

akrabat:
Thoughts, @silentworks ?

aimeos:
@silentworks What are your thoughts? This change is rather important for us :-)

tests/RouterTest.php
```diff
@@ -162,6 +162,38 @@ public function testPathForWithQueryParameters()
...
+        $methods = ['GET'];
+        $pattern = '/hello/{name}';
+        $callable = function ($request, $response, $args) {
+            echo sprintf('Hello %s', $args['name']);
+        };
+        $route = $this->router->map($methods, $pattern, $callable);
+        $route->setName('foo');
+
+        $this->assertEquals(
+            '/hello/josh?a=b',
```
campersau, line 175: What would happen if the url looked like this: `/hello/josh?name=test` ?

aimeos, line 175: That's an edge case that can't be supported if the implementation should be more compliant to other frameworks.

silentworks, line 175: Thats an edge case we can live with. It would be also concerning if you were using `{name}` for both placeholder and `$_GET` argument, in that case I would think its an API design flaw.

OptimusCrime, line 175: Still, I am all in favor of having a consistent behavior. What would happen today?

silentworks, line 175: @OptimusCrime it would be kinda a breaking change if this were to get merged, because the developer wouldn't expect any undefined placeholder to become a query string at the end of the path. I don't think this would cause much harm to any project, but it might not be what the developer expected. I however think this is something we can live with as this is expected behaviour in some other frameworks.

geggleto, line 175: The use case `/route/{name}` where `/hello/josh?name=test` should throw a router exception.

geggleto, line 175: I am very much +0 on this because adding the query params after the pathFor is rather trivial while not ideal.

After reviewing this, I would want this to be a separate method to prevent BC breaks and we can fully document it's pro's and con's... At the next point release we could mark this old method as being depreciated but still give our user base some time to convert over. In addition we should prepare web site content on how to migrate and what the differences are.

//ping @akrabat @silentworks

aimeos:
@silentworks @akrabat Any decisions how to proceed with this issue?

silentworks:
@aimeos This is a breaking change that will have to wait until 4.0, it would lead to an unexpected behaviour for the developer, potentially exposing data they didn't wish to expose in the querystring.

An alternative would be to provide a setting option to turn this **on** and this would be set to **off** by default. This I could merge in right away.

aimeos:
@silentworks Wouldn't an additional optional parameter for the `relativePathFor()` and `pathFor()` methods be a better alternative? The router class shouldn't depend on any external configuration settings IMHO and using a `setCombinedParameter()` seems to be only the second best option.

silentworks:
@aimeos what would be the default behaviour with this optional parameter?

aimeos:
@silentworks The current one to avoid BC breaks. Using the new behavior has to be enabled explicitely.

silentworks:
@aimeos I am happy with it working like that. Update PR and we will get it merged in if no-one has any objections. @codeguy @akrabat

aimeos:
@silentworks @codeguy @akrabat Maybe we've found a very elegant way to extend the functionality without the need to change the interface. It uses only a subtle, undocumented difference when handling route attributes: Instead of dropping surplus route attributes silently, they are added as query parameters.

You now have these behaviors (route, data, queryParams -> result):
- /{name} + ['name' => 'josh'] + [] -> /josh (same as before)
- /{name} + ['name' => 'josh'] + ['a' => 'b'] -> /josh?a=b (same as before)
- /{name} + ['name' => 'josh', 'c' => 'd'] + ['c' => 'e'] -> /josh?c=e (same as before)
- /{name} + ['name' => 'josh', 'c' => 'd'] + [] -> /josh?c=d (new, before: /josh only)
- /{name}/{c} + ['name' => 'josh'] + ['c' => 'd'] -> exception (same as before)

What do you think? If it's still too much of a change, we can still add the additional parameter to activate the new behavior.

aimeos:
@silentworks Are you fine with the approach?

KuroThing:
While i'm not entirely sure if my opinion on this will change anything, but i'm better off expressing it than not. Fore warning, this is in reference to the edge case of the second iteration. I do not know (or care to check, since if it still exists, the comment will be relevant) if the third iteration will have this edge case as well.

I'm personally against such a change, because at the core level you are essentially denying the usage of certain Query Parameters, purely because they have the same name as something in the Route Parts, as shown in the edge case of `/user/{name}` having problems with `/user/john?name=test` in the second iteration.

In relation to the third iteration, one of the tested examples was not the edge case of a Route Param and Query Param having the same name like `/{name} + ["name" => "john"] + ["name" => "test"] = ??`. Only missing a route part and trying to provide it in query params was shown.

Even if the worst case scenario was avoided and @geggleto's (imo) horrible suggestion of "throwing a Router Exception because the Request data had a key that is the same a routes part key" isn't implemented, the edge case of which gets what still exists with the second iteration still exists, however if the fore mentioned test solves this problem with the third iteration, and more extensive tests are conduct when provided keys that exist in the route parts as query params are done, than i see no problem, however since the entire problem about this issue is that providing the data for both the Route Part and Query Params produces duplicates in the url, as expected, i doubt that it will work as it currently does (Which would make it a BC change, since it would break existing applications that have the same key in both route parst and query params).

While this is kind of irrelevant, it does explain my opinion on this matter, and why i'm against it. The way the routes are setup for this example usage, in my opinion, goes against the entire routing philosophy of slim, which has always been SEO friendly urls, even if urls that were not SEO friendly were supported it was never documentation, shown, or even mentioned anywhere. So while this is a perfectly good example of using optional parts, trying to polyfill them with Query Params to make them optional goes against what i believe, and what appears to be the routing philosophy of slim.

I'm sure i've said some things here that some people will not like, and i'm arguing the entire point of this Pull Request that the core maintainers have pretty much accepted, but none the less thank you for reading my comment, and opinion on this matter. 

Also @aimeos sorry, but i, personally, don't think Slim will benefit from adding this, and "just because everyone else is doing it you should too" is not a valid reason, and i do believe it's just peer pressure. Maybe you should consider another approach, like extending the Router, and providing your modified relativePathFor function to uses, which would just entail using `$container->extend(callable)` to inject it into Slim.

tuupola:
> There's no known way to find out if a parameter will be used as URL or query parameter so our URLs currently look like.

Instead of altering current behaviour wouldn't it be better to add a way to get the placeholder names for a given route. This would solve the root of the problem since developer would have information which parameters belong to `$data` and which belong to `$queryParams`.

akrabat:
A quick survey suggests that Symfony, Yii2 and Laravel will automatically convert extra route parameters to query parameters while ZF2 and CakePHP require you to separate them out.

silentworks:
So this is a matter of implicit vs explicit. I am always for the explicit way of doing things, but I see no harm in this being one of those occasions we do it implicitly.

akrabat:
I think I prefer explicit too.

However:

> Instead of altering current behaviour wouldn't it be better to add a way to get the placeholder names for a given route. This would solve the root of the problem since developer would have information which parameters belong to $data and which belong to $queryParams.

Should be implemented.

Issue #1803, opened 2016-03-07
update Request Body won`t update parsedBody


DavidWiesner:
Hi,

if I modify the body of a Request the parsedBody won`t be parsed again.
## Steps to reproduce
1. create Request with empty Body
2. set a new body `withBody` with parsable content
3. check result of `getParsedBody` 

``` php
$body = 'foo=bar&abc=123';
$env = \Slim\Http\Environment::mock(
    [
        'REQUEST_METHOD' => 'POST',
        'QUERY_STRING' => 'one=1&two=2&three=3',
        'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
        'CONTENT_LENGTH' => 15,
    ]
);

$slimRequest = \Slim\Http\Request::createFromEnvironment($env);
$bodyStream = $slimRequest->getBody();
$bodyStream->write($body);
$bodyStream->rewind();
$slimRequest = $slimRequest->withBody($bodyStream);

$this->assertSame(15, $slimRequest->getContentLength());
$this->assertSame('application/x-www-form-urlencoded', $slimRequest->getContentType());
$this->assertSame('123', $slimRequest->getParsedBodyParam('abc'));
$this->assertSame('bar', $slimRequest->getParsedBodyParam('foo'));
```
## Expected Result
- `getParsedBodyParams` contains array with keys `abc` and `foo` with values `123` and `bar`
  ## Actaual Result
- `getParsedBodyParams` is an empty array
## Possible Fix
- override `withBody` in `Request` and set `bodyParsed = false` after clone

akrabat:
To avoid any BC break, it looks like the ability to manually force reparsing of the body is the way forward. Something like `$request->reparseBody()`.

akrabat:
@DavidWiesner does #1811 address this issue for your case?

DavidWiesner:
@akrabat that's isn't a good solution! Set `withBody` than schould call `reparseBody` automatically. In  #1804 I also described `getBody()->write()`  should also invalidate the content of `parsedBody` automatically. You can also set a new ContentType with `withHeader` which also should result in re-parsing the body. The easiest solution would be to parse the body every time when `getParsedBody` will called. That would leed to a discussion how often `getParsedBody` will called and how expensive this will be in contrast to more and more guards and checks.

akrabat:
@DavidWiesner I spoke with the authors of PSR-7 and the spec deliberately doesn't link Body with ParsedBody. 

[...]

PR #1804, opened 2016-03-07
Fix $request->withBody

mathmarques:
`$request->withBody` must set `bodyParsed = false` 'cause body will have to be parsed again on `$request->getParsedBody()`

Fix https://github.com/slimphp/Slim/issues/1803

### DIFF
--- a/Slim/Http/Message.php
+++ b/Slim/Http/Message.php
@@ -290,6 +290,10 @@ abstract class Message implements MessageInterface
         $clone = clone $this;
         $clone->body = $body;
 
+        if (property_exists($clone, 'bodyParsed')) {
+            $clone->bodyParsed = false;
+        }
+
         return $clone;
     }
 }
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -772,6 +772,29 @@ class RequestTest extends \PHPUnit_Framework_TestCase
         $this->assertNull($request->getParsedBody());
     }
 
+    public function testGetParsedBodyAfterWithBody()
+    {
+        $uri = Uri::createFromString('https://example.com:443/?one=1');
+        $headers = new Headers([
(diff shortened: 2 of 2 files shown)


Slim/Http/Message.php
```diff
@@ -290,6 +290,10 @@ public function withBody(StreamInterface $body)
         $clone = clone $this;
         $clone->body = $body;
 
+        if (property_exists($clone, 'bodyParsed')) {
```
geggleto, line 293: this guard is not required. if we are making a new body then we should reset the Slim request.

mathmarques, line 293: `Response` extends `Message` too. If we remove this line won't throw error but will declare a new property on response object.
Isn't this a problem?

geggleto, line 293: Ignore my previous comment. for whatever reason we only parse on the Request object and not all messages.

DavidWiesner:
Indeed there is another issue with parseBody and this solution. I can modify the existing body with getBody()->write(). So one option to solve also this might be to add an guard that check the modification of the body. E.g.: calculate a hash value and compare to the previous hash of the Body. I don't know if there is another option to check if the body was modified.

DavidWiesner:
Another option to also resolve the issue with directly modifying the body. Is to make RequestBody immutable. So isWriteable return false and only allow to set a new body using withBody

geggleto:
The `getBody()->write()` thing follows the PSR spec

DavidWiesner:
Ok, then you have to check if the body was modified in parseBody

Slim/Http/Message.php
akrabat, line 293: I think that this should be in `Request::withBody()`

akrabat:
As discussed on irc with @mathmarques, this will be a BC break. We need a different solution.

PR #1805, opened 2016-03-08
Implements Mobile_Detect lib for Response's write

zguillez:
Allow different write responses for calls from **Desktop** or **Mobile** (phone, tablet, iOS and Android):
$response->write("All devices");
$response->writeDesktop("Only from desktop");
$response->writeMobile("Only from mobile");
$response->writePhone("Only from a phone");
$response->writeTablet("Only from a tablet");
$response->writeIOS("Only from iOS");
$response->writeAndroid("Only from Android");

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,14 @@ require 'vendor/autoload.php';
 $app = new Slim\App();
 
 $app->get('/hello/{name}', function ($request, $response, $args) {
-    $response->write("Hello, " . $args['name']);
-    return $response;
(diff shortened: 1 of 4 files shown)


akrabat:
I'm not sure that this is a core Slim functionality.

geggleto:
This should be in its own composer project.

tuupola:
I like the feature but it does not belong to core.

akrabat:
Not a core feature.

PR #1808, opened 2016-03-10
Ensure getParsedBody() return null if body is not parsed

### LINKED ISSUE #1807: Request::getParsedBody returns false but must null|array|object
I send POST request from jQuery via XMLHttpRequest. In case of empty request body there is no Content-Type header. In such case Request::getMediaType returns null and Request::getParsedBody returns false.
This was broke after 3.2.0 update (https://github.com/slimphp/Slim/pull/1734)

akrabat:
PSR-7 states that getParsedBody() must return null, an array or an object. If we can't parse the body as there's no media type parser available, then return null.

Fixes #1807

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -978,9 +978,10 @@ class Request extends Message implements ServerRequestInterface
                 );
             }
             $this->bodyParsed = $parsed;
+            return $this->bodyParsed;
         }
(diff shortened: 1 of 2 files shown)


geggleto:
:+1: looks good.

mathmarques:
:+1:

PR #1817, opened 2016-03-13
Use stack trace to guess if exception was thrown by Pimple

vlakoff:
Follow-up to #1663. Maybe this method has caveats, so it needs to be reviewed.

ping @Sam-Burns @mnapoli :)

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -144,7 +144,9 @@ class Container extends PimpleContainer implements ContainerInterface
      */
     private function exceptionThrownByContainer(\InvalidArgumentException $exception)
     {
-        return preg_match('/^Identifier ".*" is not defined\.$/', $exception->getMessage());
+        $trace = $exception->getTrace()[0];
+
+        return $trace['class'] === 'Pimple\Container' && $trace['function'] === 'offsetGet';
     }
 
     /**


Sam-Burns:
:+1:

geggleto:
This PR requires a unit test.

vlakoff:
Actually this seems to be already tested, in [`ContainerTest.php`](https://github.com/slimphp/Slim/blob/3.x/tests/ContainerTest.php) see `testGetWithDiConfigErrorThrownAsContainerValueNotFoundException` and `testGetWithDiConfigErrorThrownAsInvalidArgumentException` whereweareexpoectinga ContainerExpception.

geggleto:
Another side thought, In the unlikely event that someone extends the container and overrides `offsetGet` this will no longer work.

I don't see why would we would want to change it. What was the point of this PR?

Slim/Container.php
```diff
@@ -144,7 +144,9 @@ public function get($id)
      */
     private function exceptionThrownByContainer(\InvalidArgumentException $exception)
     {
-        return preg_match('/^Identifier ".*" is not defined\.$/', $exception->getMessage());
+        $trace = $exception->getTrace()[0];
+
+        return $trace['class'] === 'Pimple\Container' && $trace['function'] === 'offsetGet';
```
akrabat, line 149: I think that this won't work if someone overrides Container and then extends offsetGet()…

akrabat:
@geggleto The current code relied on string matching which may be more brittle

vlakoff:
I had thought about a test like `is_a($trace['class'], 'Pimple\Container', true)`, but currently Slim Container "hard extends" Pimple Container and cannot extend a different class.

akrabat:
But someone can extend `Container` and then implement `offsetGet`…

vlakoff:
Got your point.
- If this custom `offsetGet` makes use of Pimple and the latter throws an exception, it will be detected as well.
- If this custom `offsetGet` throws exceptions by itself, they should extend Interop. Otherwise, there is just no way to properly detect them.

PR #1830, opened 2016-03-29
Abstract handlers

juliangut:


akrabat:
This level of change requires a discussion on what the current problem is and why this is the best solution to that problem.

mcaskill:
The current error handlers are well made and very useful since they can respond to different content-type requests. One downside is that they also duplicate a lot of code between. The presence of an abstract handler allows this to be mitigated. Also, having a common base class (or interface) allows for better testing and customization in projects.

This pull request provides a better basis for developers to extend and customize their own handlers while still benefitting from Slim's without having to duplicate too much code.

I had plans on submitting a similar pull request for abstract handlers. We've implemented our own for the Charcoal web framework.
- App
  - [`\Charcoal\App\Handler`](https://github.com/locomotivemtl/charcoal-app/tree/master/src/Charcoal/App/Handler)
  - [`\Charcoal\App\ServiceProvider\AppServiceProvider`](https://github.com/locomotivemtl/charcoal-app/blob/master/src/Charcoal/App/ServiceProvider/AppServiceProvider.php#L67)
  - [`app/handler/default.mustache`](https://github.com/locomotivemtl/charcoal-app/blob/master/templates/charcoal/app/handler/default.mustache)
- Admin
  - [`\Charcoal\Admin\Template\HandlerTemplate`](https://github.com/locomotivemtl/charcoal-admin/blob/master/src/Charcoal/Admin/Template/HandlerTemplate.php)
  - [`admin/handler.mustache`](https://github.com/locomotivemtl/charcoal-admin/blob/master/templates/charcoal/admin/template/handler.mustache)

In our case, the enhanced handlers tie into our view renderer and allows a developer to define template and view controller options. The `NotFound` handler is the best example of a view that ideally needs to be branded.

This enhancement also allows us to more easily customize those same handlers when dealing with a module, such as an administration dashboard that would have its own templating features (using Pimple's extend feature).

juliangut:
@mcaskill is right

There is no "actual" problem trying to be solved but the duplication of code, I've tried to abstract as much code as possible without changing the error rendering methods.

Also providing abstracted handlers allow developers to focus on their real impletentation and not care about commonalities such as writing to log or detecting content type.

I didn't consider abstracting the `render*Message` methods on error handlers but now that I review it I think it should be done the same way `writeToErrorLog` was, in the end \Error and \Exception are both \Throwable

PR #1833, opened 2016-04-04
Allow dispatching of the router in middleware

dangerousdan:
Currently, the router is not dispatched until after the app middleware.
If we set `determineRouteBeforeAppMiddleware` in the settings, the router is dispatched straight away.
There seems to be no middle ground.

If we move App::dispatchRouterAndPrepareRoute into a trait, this allows us to dispatch the route in middleware without duplicating the contents of that method.

``` php
<?php
class RouteDispatcherMiddleware
{
    use Slim\DispatchRouterTrait;

    protected $router;

    public function __construct($router)
    {
        $this->router = $router;
    }

    public function __invoke($request, $response, $next)
    {
        if (! $request->getAttribute('route')) {
            $this->dispatchRouterAndPrepareRoute($request, $this->router);
        }
    }
}
```

This means users can get the best of both.

``` php
$app->add(new MiddlewareThatRequiresAccessToTheRoute());
$app->add(new RouteDispatcherMiddleware($app->getContainer()->get('router');
$app->add(new MiddlewareThatManipulatesTheRequest());
```

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -48,6 +48,7 @@ use Slim\Interfaces\RouterInterface;
  */
 class App
 {
+    use DispatchRouterTrait;
     use MiddlewareAwareTrait;
 
     /**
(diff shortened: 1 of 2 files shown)


geggleto:
Can you explain why this is required?

dangerousdan:
So that we can reuse the trait in middleware to dispatch the router.
This is useful as it enables us to have outer middleware that manipulates the route, and inner middleware which has access to the chosen route.

It's possible to do this already by copying the contents of app::determineRouteBeforeAppMiddleware, but this isn't DRY.

akrabat:
I like the concept. I need to think about it a bit in terms of how to do it.

akrabat:
I think this functionality needs to be implemented as middleware that App calls if it hasn't bene called already. I'm not sure that this is possible in 3.x or whether it needs to be a 4.0 thing.

PR #1834, opened 2016-04-05
define precise interface for the di container

ckressibucher:
In the docblock of `Slim\DefaultServicesProvider::register`, there's a line saying: `@param Container $container A DI container implementing ArrayAccess and container-interop`. The `Container`s FQCN is `Slim\Container` which extends the PimpleContainer. So, it _does_ implement `ArrayAccess` and the interop `ContainerInterface`, but _not only_ this, is also adds a (IMHO) unneccessary constraint to the Slim Container. This makes it very hard to use a different container. (It's possible because the constraint is only a comment, but it's not "clean" style).

I declared a dedicated interface which actually does exactly what the comment "A DI container implementing ArrayAccess and conatiner-interop" says. And added it also as a type hint to the method declaration.

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -8,13 +8,13 @@
  */
 namespace Slim;
 
-use Interop\Container\ContainerInterface;
 use Interop\Container\Exception\ContainerException;
 use Pimple\Container as PimpleContainer;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Slim\Exception\ContainerValueNotFoundException;
 use Slim\Exception\ContainerException as SlimContainerException;
+use Slim\Interfaces\ArrayContainerInterface;
 
 /**
  * Slim's default DI container is Pimple.
@@ -44,7 +44,7 @@ use Slim\Exception\ContainerException as SlimContainerException;
  * @property-read callable notAllowedHandler
  * @property-read \Slim\Interfaces\CallableResolverInterface callableResolver
  */
-class Container extends PimpleContainer implements ContainerInterface
+class Container extends PimpleContainer implements ArrayContainerInterface
 {
     /**
      * Default settings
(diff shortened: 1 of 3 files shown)


geggleto:
:+1: No idea where this will land but I like it. Likely the next major release.

JoeBengalen:
Pimple already has ArrayAccess implemented

https://github.com/silexphp/Pimple/blob/master/src/Pimple/Container.php#L34

Edit: Nvm I need to read first ;)

I am against this though, I don't want to implement `ArrayAccess` on a different container.

ckressibucher:
@JoeBengalen I do want to implement it on a different container, and likely there are others too. I don't know the slim framework very well, but at first sight there are no problems introduced by this change, at least the tests are all passing.

If the change introduces problems, then the comment should be fixed (because I just did what the comment says...)

akrabat:
I disagree with this. Slim doesn't need ArrayAccess at all - only the service registration does.

ckressibucher:
Ok, didn't know that, then I agree with @akrabat 

Nonetheless, I'd be thankful for a minimal interface with a setter to be able to reuse the service provider. (The interop container only provides `get` and `has` methods).

So my suggestion would be:
1. Write an interface `Slim WritableContainerInterface extends \Interop\Container\ContainerInterface` and define an addittional `offsetSet`.
2. Rewrite the service provider to use that interface

In my project, to use a different container, I would then just need to write a Bridge class which implements this simple interface (whereas I now need to implement / extend the Slim container).

The other Slim code would not need to use this additional interface, as long as they only need the `get` and `has` methods defined by the interop container.
What do you think about this?

silentworks:
This has come up before and we have settled for keeping it the way it currently works. I think on a per project basis the developer can decide which container they want to use, not all containers use ArrayAccess for setting and if we were going to create an interface in order for you to create a bridge class, you might as well just do that in your own code rather than having this in the core of Slim. 

This is the reason container interop spec didn't define a way to `set` but only `get` and `has`.

PR #1839, opened 2016-04-08
Fixes slim.errors setting key ignored on instantiation. 

ghost:
Can now pass in slim.errors array key that references a resource and Slim default logger will be used. This behaviour now matches the 2.x documentation.


akrabat:
We are only applying security fixes to Slim 2.

PR #1842, opened 2016-04-10
change private to protected for extending

irfanevrens:


akrabat:
Not needed any more as it's been refactored.

PR #1848, opened 2016-04-14
add `psr/http-message-implementation` to composer

lalop:
Define `psr/http-message-implementation` as provided by Slim

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,9 @@
         "squizlabs/php_codesniffer": "^2.5",
         "phpunit/phpunit": "^4.0"
     },
+    "provide": {
+        "psr/http-message-implementation": "1.0"
+    },
     "autoload": {
         "psr-4": {
             "Slim\\": "Slim"


akrabat:
I'm not sure we should do this as Slim is more than just a PSR 7 implementation.

This would make more sense when we have a viable `Slim\Http` repo I think?

lalop:
This is usefull for composer, if a package need `psr/http-message-implementation` compoesr will ask to add a dependencie to complete the requirement

akrabat:
Yes. I'm just not sure that Slim 3 is the right answer to that question.

lalop:
Here is a list of librairy requiring `psr/http-message-implementation`  https://packagist.org/packages/psr/http-message-implementation/dependents

To use them with Slim we will need to install another implementation since Slim doesn't declare itself as providing it.

akrabat:
Aha! I understand now.

Issue #1686, opened 2015-12-17
Slim 4.0 Road Map


codeguy:
Here are some intiial thoughts:
- [BC break] Extract `Slim\Http` into separate PSR-7 component (ONLY PSR-7 methods)
- [BC break] Create custom decorator classes to hold Slim-specific additions to PSR-7
- Accommodate additional media type parsers in request
- Address issues reported by Phan
- Improve test coverage

Additional ideas:
- [BC break] Remove Request, Response and Env from container (@akrabat)
- [BC break] Don't set content-length by default. Provide a middleware for people to use if they want to. (@akrabat)
- Make it faster (or at least don't slow it down!) (@Cyrille37)
- [BC break] Removing Method-Override and providing a middleware (@geggleto)
- [BC break] Decide if we need `subRequest` at all (@codeguy/@silentworks/@geggleto)
- Lazy resolve route middlewares (#1635) We already do this with route callable(only resolve the matched one). (@mathmarques)
- Look at `__invoke` and `run` on `App` and see if we can consolidate functionality and make them easier to unit test. (@geggleto)
- [BC break] Consolidate error handlers for exceptions and PHP 7 errors (@akrabat)
- Can we avoid statics in `Slim\Http`? See #1850 (@Raistlfiren)
- [BC break] Look at `pathFor()`'s parameters? see #1800 (@akrabat)
- [BC break] Do something about `basePath` from `Request`. Maybe middleware solution instead? (@akrabat)
- [BC break] Remove `determineRouteBeforeDispatch` and make routing a middleware that can be placed where you want it to be. If you don't add it yourself, we would add it immediately before dispatch. (@akrabat)
- Use `getallheaders()` to get the Authorisation header. (irc channel)
- Move documentation into main repo and publish to website on each new release. That way we can insist on documentation with the PR.
- [BC break] Either rename `Slim\Exception\SlimException` or remove it.

Please add additional thoughts below.

geggleto:
Removing Method-Override and providing a middleware

silentworks:
`subRequest` needs to use the request object from the container and not instantiate its own.

mathmarques:
Lazy resolve route middlewares (https://github.com/slimphp/Slim/pull/1635)
We already do this with route callable(only resolve the matched one).

geggleto:
[...]

PR #1850, opened 2016-04-20
(Thoughts?) Feature - Allow users to Specify their own Http Classes Easily

Raistlfiren:
The problem - Currently, you are limited to using Slims classes to parse requests for the Uri, Headers, Cookies, and UploadedFiles. https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L139

Solution - Allow the user to specify the class they want to use in order to handle the request type in the settings file. For example, a user declares the setting => `'uploadedFileClass' => '<Their new UploadedFileClass>'`, then the users class will be used instead of the default. 

Now users can go about extending the UploadedFile class and adding methods that they want. 

**I did break all tests, just curious if anyone else wanted the same functionality**

### DIFF
--- a/Slim/DefaultServicesProvider.php
+++ b/Slim/DefaultServicesProvider.php
@@ -58,7 +58,7 @@ class DefaultServicesProvider
              * @return ServerRequestInterface
              */
             $container['request'] = function ($container) {
-                return Request::createFromEnvironment($container->get('environment'));
+                return Request::createFromEnvironment($container->get('environment'), $container->get('settings'));
             };
         }
 
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -130,10 +130,11 @@ class Request extends Message implements ServerRequestInterface
      * Environment object
      *
      * @param  Environment $environment The Slim application Environment
+     * @param  Collection $settings The Slim application Settings
      *
      * @return self
      */
-    public static function createFromEnvironment(Environment $environment)
+    public static function createFromEnvironment(Environment $environment, Collection $settings)
     {
         $method = $environment['REQUEST_METHOD'];
         $uri = Uri::createFromEnvironment($environment);
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -13,6 +13,7 @@ use InvalidArgumentException;
 use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UploadedFileInterface;
 use Slim\Collection;
+use Slim\Interfaces\Http\RequestBuilderInterface;
 
 /**
  * Represents Uploaded Files.
@@ -22,7 +23,7 @@ use Slim\Collection;
  * @link https://github.com/php-fig/http-message/blob/master/src/UploadedFileInterface.php
  * @link https://github.com/php-fig/http-message/blob/master/src/StreamInterface.php
  */
-class UploadedFile implements UploadedFileInterface
+class UploadedFile implements UploadedFileInterface, RequestBuilderInterface
 {
     /**
      * The client-provided full path to the file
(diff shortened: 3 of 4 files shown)


JoeBengalen:
Slim is psr7 compliant, you can easily use a different http implementation.

Raistlfiren:
I am mostly just wanting to extend off of the default classes and add my own methods. Is there an easy way to go about doing that?

JoeBengalen:
Just overwrite the `request` and `response` factories in the container.

Or you could call the `process` and `respond` methods of `App` instead of `run`

geggleto:
:-1: you can already do this by doing what @JoeBengalen said.
Slim accepts any PSR-7 compatible Request/Response strategy.

> I am mostly just wanting to extend off of the default classes and add my own methods. Is there an easy way to go about doing that?

Currently, there is no easy way.

Raistlfiren:
True, but that is a lot of work to just add a few methods to the UploadedFile class. For instance, my scenario is say I want to add the method getExtension to UploadedFile or maybe I want to use Symfony's UploadedFile class. There isn't an easy way of switching that out without overwriting the whole request object. Should it really be that hard to do that?

Raistlfiren:
> Currently, there is no easy way.

Exactly, this would enable us to overwrite that class easily.

geggleto:
It's on the roadmap :)

Raistlfiren:
but bUT BUT, I neeeeeed it now ;)

geggleto:
@Raistlfiren you can extend the Slim objects and place them into the container after you create the $app object then :)

akrabat:
I'm missing what the problem is. Just create some middleware:

``` php

    $uploadedFiles = MyUploadedFilesObject::createFrom($_FILES);
    $request = $request->withUploadedFiles($uploadedFiles);

});
```

Raistlfiren:
@akrabat I suppose my problem/annoyance is that I can't override the default classes without replacing the `Request` object in the container or using middleware.

Raistlfiren:
The problem with using the middleware as you displayed is that you are not only executing the previous UploadedFiles code/class as well, but now you are using your own class to parse $_FILES. In essence you are duplicating code with that middleware.

---Edit---
In this instance the middleware would be a hack in order to resolve a bigger problem, right?

akrabat:
In that case, you just have to override `createFromEnvironment`:

``` php
class MyUploadedFile extends UploadedFile
{
}

class MyRequest extends Request
{
    public static function createFromEnvironment(Environment $environment)
    {
        $method = $environment['REQUEST_METHOD'];
        $uri = Uri::createFromEnvironment($environment);
        $headers = Headers::createFromEnvironment($environment);
        $cookies = Cookies::parseHeader($headers->get('Cookie', []));
        $serverParams = $environment->all();
        $body = new RequestBody();
        $uploadedFiles = MyUploadedFile::createFromEnvironment($environment);

        $request = new static($method, $uri, $headers, $cookies, $serverParams, $body, $uploadedFiles);

        if ($method === 'POST' &&
            in_array($request->getMediaType(), ['application/x-www-form-urlencoded', 'multipart/form-data'])
        ) {
            // parsed body must be $_POST
            $request = $request->withParsedBody($_POST);
        }
        return $request;
    }
}

$settings = [
    'settings' => [
        'displayErrorDetails' => true,
    ],
    'request' => function ($c) {
        return MyRequest::createFromEnvironment($container->get('environment'));
    },
];
$app = new \Slim\App($settings);
```

Raistlfiren:
@akrabat Absolutely, which does work just fine. I am just proposing that we get rid of this step and allow users to define their own `MyUploadedFile` class.

akrabat:
The ability to override a specific class within the Http classes via settings is not going to be accepted in Slim 3. This may change in Slim 4 when we look to create a base pure PSR-7 implementation with our enhancements in a separate system.

akrabat:
> @akrabat Absolutely, which does work just fine. I am just proposing that we get rid of this step and allow users to define their own MyUploadedFile class.

From my point of view, I don't want the additional complexity for a rare situation the can be handled by a 23 line class. 

Sorry.

Raistlfiren:
@akrabat Is that really all that complex? In my sample, which I know is incomplete, allows you to do what I am asking for in a few lines of code. It requires the static build method to look for that unique setting and call a different class. 

There aren't to many classes that you need to override with Slim 3. Most of them can be overwritten pretty easily like the Request object that you displayed. Those Http classes are just more difficult because they are called statically.

akrabat:
I've added a note to #1686 about this issue for when we design `Slim\Http` for 4.0.

PR #1854, opened 2016-04-30
Add explanation for running example.

thinkerou:


akrabat:
Thank you! I moved it to the example directory as the main readme as that's where I feel it belongs.

Issue #1851, opened 2016-04-20
[RFC] CORS Support


geggleto:
Looking towards 4.0... I was wondering about including CORS support built-in for Slim users and I want to hear your feedback on it.

Ideally it would be something that you can either turn ON or OFF, so that you are not forced to use it.

The idea comes from the few projects that already address this issue.
[1] https://github.com/palanik/CorsSlim
[2] http://joshuasampia.com/2015/11/05/cors-slim-php-setup/
... etc

silentworks:
I don't think this belongs in the core of Slim, but can be part of Slim Skeleton.

akrabat:
Are there any things within Slim that make supporting CORS harder than it could be?

mathmarques:
I don't think we should add this on Slim.
A 'simple' middleware could add CORS Support.
Maybe we only need a Middleware/Addon section on slim website to index these useful things

geggleto:
@mathmarques @akrabat 

The problem with CORS is that in the pre-flighted requests you need to know what endpoints are registered.

Example:
Resource: `/v1/user` 
- Supports: GET, POST, PUT

The preflight OPTIONS `/v1/user` should return a header with 
`Access-Control-Allow-Methods: POST, GET, PUT`

Doing this in a middleware... I have no idea how that could be done.

It makes most sense to do it at the router level, as it can inspect the Dispatcher to find routes LIKE this one and construct a HTTP METHOD List for this resource.

geggleto:
@silentworks Not sure if it should land in core myself, but its something that I deal with frequently. Perhaps a CorsRouter might be more useful.

lalop:
I'm maybe wrong but any middleware can access to the routes registred in the router, no ?

geggleto:
@lalop true... in order to figure out what ROUTE you are in you will need to use the determineRouteBeforeDispatch setting.

akrabat:
Should we be organising the information in the router, so that getting this information out is reasonably easy?

lalop:
Adding a route middleware and the route information as attribute to the request should do the trick I suppose

tuupola:
I think CORS does not belong to core.There already are couple of middlewares,  including [mine](https://github.com/tuupola/cors-middleware) which uses [neomerx/cors-psr7](https://github.com/neomerx/cors-psr7) for heavy lifting. However the OPTIONS problem @geggleto mentioned is real.

[...]

PR #1857, opened 2016-05-04
Adding Cors support

geggleto:
Related to #1851


geggleto:
This will add the ability to pull from the router the HTTP Verbs associated with a given pattern.

Patterns can be retrieved from a Request Object in a middleare when the `determineRouteBeforeAppMiddleware` setting is set to `true`.

This then will lead to the ability to use a middleware to accurately provide CORS Support.

``` php
///....

$app->add(function ($req, $res, $next) {
    $route = $req->getAttribute('route');
    $router = $this->router;

    $verbs = $router->getMethods($route->getPattern());

   $res = $next($req, $res);

   $res->withHeader('Access-Control-Allow-Methods', implode(',', $verbs))
         ->withHeader('Access-Control-Allow-Origin', '*');

    return $res;
});


```

dopesong:
In my opinion - there is no point for class like this. There could be two options:
- Add this functionality to main router.
- Move it into separate repo as extension.

Also your example is lack of information how it should be used, because right now you need to change default router - what adds additional complexity.

geggleto:
I don't think a separate repo is the proper place.

The example to use it is trivial...
`
$container['router'] = function ($c) {
     return new CorsRouter();
};
`

tuupola:
I also think this functionality belongs to main router.

tuupola:
I found a way for middleware to get the defined verbs for current routes **pattern** with the code below.  This works currently without any changes to the router. Note that this is not always the the same as all the defined verbs for current routes **uri**. However it is close enough for most use cases.

``` php
require __DIR__ . "/vendor/autoload.php";

$app = new Slim\App([
    "settings"  => [
        "determineRouteBeforeAppMiddleware" => true,
    ]
]);

    $pattern = $request->getAttribute("route")->getPattern();

    $methods = [];
    foreach ($this->router->getRoutes() as $route) {
        if ($pattern === $route->getPattern()) {
            $methods = array_merge_recursive($methods, $route->getMethods());
        }
    }

    $response = $response->write(print_r($methods, true));
});

$app->get("/api/{id}", function($request, $response, $arguments) {
});

$app->post("/api/{id}", function($request, $response, $arguments) {
});

$app->map(["DELETE", "PATCH"], "/api/{id}", function($request, $response, $arguments) {
});

$app->run();
```

```
%curl --include http://localhost:8080/api/xxx
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
Content-Length: 76

Array
(
    [0] => GET
    [1] => POST
    [2] => DELETE
    [3] => PATCH
)
```

geggleto:
very cool @tuupola that should go into some sort of service provider then, and we can close this PR

akrabat:
So we just need some middleware?

dopesong:
Correct

dopesong:
@tuupola have you checked how it works with routes groups?

tuupola:
Just tested and seems to work.

akrabat:
Closing as solution is possible without this change.

PR #1866, opened 2016-05-09
Error handlers throws exception if content-type not supported

dopesong:


akrabat:
Please make it throw an exception as `determineContentType` should mean that the default never happens.

akrabat:
Needs tests.

dopesong:
Will be added + additional to increase coverage 👍

geggleto:
@dopesong Don't forget to squash the commit :D

Issue #1784, opened 2016-02-28
Cannot get Authorization header in middleware


humpedli:
The problem is that Slim PHP 3 `$request->getHeaders()` function is not returning Authorization header parameter while the default `getallheaders()` PHP function is working correctly. When I use this Slim PHP function in a normal route, it's returning Authorization header parameter correctly, the issue exists only in the middleware.

**Request headers:**

```
Accept:application/json, text/plain, */*
Accept-Encoding:gzip, deflate, sdch
Accept-Language:hu-HU,hu;q=0.8,en-US;q=0.6,en;q=0.4,it;q=0.2
Authorization:Bearer *****
Connection:keep-alive
Cookie:*****
Host:*****
Referer:*****
User-Agent:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.116 Safari/537.36
```

**Code snippet:**

```
public function __invoke($request, $response, $next) {
    $slimHeaders = $request->getHeaders();
    var_dump($slimHeaders);

    $phpHeaders = getallheaders();
    var_dump($phpHeaders);
}
```

**$slimHeaders dump:**

```
array(8) {
  ["Host"]=>
  array(1) {
    [0]=>
    string(5) "*****"
  }
  ["HTTP_CONNECTION"]=>
  array(1) {
    [0]=>
    string(10) "keep-alive"
  }
  ["HTTP_ACCEPT"]=>
  array(1) {
    [0]=>
    string(33) "application/json, text/plain, */*"
  }
  ["HTTP_USER_AGENT"]=>
  array(1) {
    [0]=>
    string(121) "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.116 Safari/537.36"
  }
  ["HTTP_REFERER"]=>
  array(1) {
    [0]=>
    string(5) "*****"
  }
  ["HTTP_ACCEPT_ENCODING"]=>
  array(1) {
    [0]=>
    string(19) "gzip, deflate, sdch"
  }
  ["HTTP_ACCEPT_LANGUAGE"]=>
  array(1) {
    [0]=>
    string(44) "hu-HU,hu;q=0.8,en-US;q=0.6,en;q=0.4,it;q=0.2"
  }
  ["HTTP_COOKIE"]=>
  array(1) {
    [0]=>
[...]

Issue #1616, opened 2015-11-25
Authorization not in header


bmutinda:
Hey guys,
With Slim2 I was able to get the value supplied in the Authorization in the headers but with Slim3 i can't get it.
I can get other header parameters though.
This is what I have tried

``` php
$request->getHeader("Authorization")
```

Is there any other way to access this?

geggleto:
That is how you read a header field.

tuupola:
Call to `$request->getHeader("Authorization")` works fine. However if you are using Apache and send something else than valid `Authorization: Basic` header PHP won't have access to it unless you add the following to your `.htaccess` file.

`RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization}]`

bmutinda:
@tuupola Thanks it works. I am using apache. Does this also fix if am using nginx?  
For apache server, this is how I had hacked it :dart: 

``` php
$headers = apache_request_headers();
$headers["Authorization"];
```

akrabat:
@bmutinda The missing `Authorization` header appears to be an Apache only thing.

ienev:
On Apache 2.4 (before 2.4.13) with PHP-FPM this rewrite would work

`RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization}]`

but it will result in a REDIRECT_HTTP_AUTHORIZATION header which is still not accessible by 

`$request->getHeader("Authorization")`

Apache 2.4.13 introduced a CGIPassAuth directive that would allow passing the Authorization header to the script although I haven't tested this myself.

For now the only workarounds are to either access $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] or use a custom header name.

tuupola:
Just FYI, https://github.com/slimphp/Slim/pull/898 was Slim 2 pull request regarding this. Did not know about CGIPassAuth. Thanks for the tip!

ienev:
Thanks I was only searching for Slim 3. Any chance a similar PR would be accepted for 3?

ianrussel:
@akrabat 

I have this issue and cant solve it. I am using nginx and Slim 3 heres my route

`$app->get('/api/private/offers', function($request, $response, $next){

    //$headers = apache_request_headers();
    $headers = $request->getHeaders();
    var_dump($headers);`
})

Here is the output

    `array (size=11)
     'HTTP_COOKIE' => 
     array (size=1)
[...]

PR #1867, opened 2016-05-09
Fall back to getallheaders() when available

tuupola:
When sending something else than valid `Authorization: Basic` header PHP will not set the `HTTP_AUTHORIZATION` environment. Because of this the `Authorization` header is not available via `$request->getHeader("Authorization")`. See #1784, #1616.

This is problematic when you are using `Authorization: Bearer` or something similar.  Usual workaround is to add the following rewrite rule.

`RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization}]`

However even without the workaround, the header is however available via `getallheaders()`. This PR adds a fallback call to  `getallheaders()` when it is available and `HTTP_AUTHORIZATION` is not set.


Slim/Http/Headers.php
```diff
@@ -64,6 +65,30 @@ public static function createFromEnvironment(Environment $environment)
...
     /**
+     * If HTTP_AUTHORIZATION does not exist tries to get it from
+     * getallheaders() when available.
+     *
+     * @param Environment $environment The Slim application Environment
+     *
+     * @return Environment
+     */
+
+    public static function normalizeEnvironment(Environment $environment)
```
akrabat: I would prefer a more specific method name.

PR #1872, opened 2016-05-12
gh-1838 added checks for pipe streams

### LINKED ISSUE #1838: Return a popen stream in the response
Hello,

I am trying to return a popen stream like this:

``` php
$stream = \GuzzleHttp\Psr7\stream_for(
    popen(
        'ls', 'r'
    )
);
return $response->withBody($stream);
```

But I see two problems with it:
- For some reason, App::finalize() seems to replace $response->getBody()->stream with an _Unknown resource_.
- `$response->getBody()->getSize()` always returns 0 in this case so App::respond() does not output anything.

llvdl:
Fixes #1838

### DIFF
--- a/Slim/Http/Stream.php
+++ b/Slim/Http/Stream.php
@@ -317,7 +317,7 @@ class Stream implements StreamInterface
             $this->seekable = false;
             if ($this->isAttached()) {
                 $meta = $this->getMetadata();
-                $this->seekable = $meta['seekable'] && !$this->isPipe();
+                $this->seekable = $meta['seekable'];
             }
         }
 
--- a/Slim/Http/Stream.php
+++ b/Slim/Http/Stream.php
@@ -80,7 +80,7 @@ class Stream implements StreamInterface
     /**
      * Is this stream a pipe?
      *
-     * @var type
+     * @var bool
      */
     protected $pipe;
 
--- a/Slim/Http/Stream.php
+++ b/Slim/Http/Stream.php
@@ -20,6 +20,11 @@ use RuntimeException;
 class Stream implements StreamInterface
 {
     /**
+     * This is octal as per header stat.h
+     */
+    const FSTAT_MODE_S_IFIFO = 0010000;
+    
+    /**
      * Resource modes
      *
      * @var  array
@@ -71,6 +76,13 @@ class Stream implements StreamInterface
      * @var null|int
      */
     protected $size;
+    
+    /**
+     * Is this stream a pipe?
+     *
+     * @var type
+     */
+    protected $pipe;
 
     /**
      * Create a new Stream.
@@ -158,6 +170,7 @@ class Stream implements StreamInterface
         $this->writable = null;
         $this->seekable = null;
         $this->size = null;
+        $this->pipe = null;
 
         return $oldResource;
     }
@@ -196,7 +209,11 @@ class Stream implements StreamInterface
     public function close()
     {
         if ($this->isAttached() === true) {
-            fclose($this->stream);
+            if ($this->isPipe()) {
+                pclose($this->stream);
+            } else {
+                fclose($this->stream);
+            }
         }
 
         $this->detach();
@@ -211,7 +228,7 @@ class Stream implements StreamInterface
     {
         if (!$this->size && $this->isAttached() === true) {
             $stats = fstat($this->stream);
-            $this->size = isset($stats['size']) ? $stats['size'] : null;
+            $this->size = isset($stats['size']) && !$this->isPipe() ? $stats['size'] : null;
         }
 
         return $this->size;
@@ -226,7 +243,7 @@ class Stream implements StreamInterface
      */
     public function tell()
     {
-        if (!$this->isAttached() || ($position = ftell($this->stream)) === false) {
+        if (!$this->isAttached() || ($position = ftell($this->stream)) === false || $this->isPipe()) {
             throw new RuntimeException('Could not get the position of the pointer in stream');
         }
 
@@ -300,7 +317,7 @@ class Stream implements StreamInterface
             $this->seekable = false;
             if ($this->isAttached()) {
                 $meta = $this->getMetadata();
-                $this->seekable = $meta['seekable'];
+                $this->seekable = $meta['seekable'] && !$this->isPipe();
             }
         }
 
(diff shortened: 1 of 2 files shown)


ghost:
Since Slim's implementation has methods beyond what's in StreamInterface anyway, I'd add isSeekableForward() and seekForward() for completeness. If one's dealing with pipes, he/she might need them anyway.

```
    /**
     * Returns whether or not the stream is seekable forward.
     *
     * Note: This method is not part of the PSR-7 standard
     * and is provided for Stream's compatibility with pipes.
     *
     * @return bool
     */
    public function isSeekableForward()
    {
        return $this->isSeekable() || $this->isPipe();
    }

    /**
     * Seek forward in the stream.
     *
     * Note: This method is not part of the PSR-7 standard
     * and is provided for Stream's compatibility with pipes.
     *
     * @param int $offset Stream offset
     * @throws RuntimeException on failure.
     */
    public function seekForward($offset)
    {
        // Note that fseek returns 0 on success!
        if (!$this->isSeekableForward() || fseek($this->stream, $offset, SEEK_CUR) === -1) {
            throw new RuntimeException('Could not seek forward in stream');
        }
    }
```

I'd also change isSeekable() implementation - there's no need for isPipe() call, I think, as 'seekable' metadata is guaranteed to be 1 in case of seekable handle and something else (typically 0 but undef for pipes) in other cases.

```
    public function isSeekable()
    {
        if ($this->seekable === null) {
            $this->seekable = false;
            if ($this->isAttached()) {
                $meta = $this->getMetadata();
                $this->seekable = ($meta['seekable'] === 1);
            }
        }

        return $this->seekable;
    }
```

HTH.

llvdl:
Thanks for the feedback, @ddalek.

I replaced `isSeekable()` with your version, but then multiple unit tests fail, so I reverted to the version I submitted before.

ghost:
Argh, I misinterpreted print_r() output. My comparison was unnecessary, the value returned is bool, not int. This will teach me not to rely on print_r(). :)

Still, 'seekable' meta should be false for pipes (despite SEEK_CUR working) and `&& !$this->isPipe()` in isSeekable() adds the cost of fstat() to the first isSeekable() call on verifiably seekable streams. But that's just a tiny perf nitpick (if I'm not talking out of my back again).

llvdl:
For some reason the tests fail on the hhvm machine: `RuntimeException: Could not seek forward in stream` is raised when trying to read from the pipe.

llvdl:
I removed the check for `&& !$this->isPipe()` and the test still pass, so it seems you are right, @ddalek .

ghost:
Cursory search through their issues shows quite a few historical incompatibilities around files and streams, including a new one around fseek() so I'm not terribly surprised. I'll take the time over the weekend to see if there's anything that can and should be done in Slim or if it's HHVM problem.

akrabat:
Tests fail on HHVM.

llvdl:
@I created a virtual machine with hhvm and could reproduce the failing tests. The cause was `isReadable()` returning `false` in hhvm. I added a unit test and a check for `isPipe()` and the tests 

I removed the seekForward implementation for now, as it doesn't seem to work for pipes in PHP. 

**Update:** the reason `isReadable()` returned `false` with hhvm is that the mode is empty when using `stream_get_meta_data()` on a resource opened with `popen()`:

``` php
<?php
$fh = popen('echo 1', 'r');
$meta = stream_get_meta_data($fh);
var_dump($meta['mode']);

stream_get_contents($fh);
pclose($fh);
```

Result using PHP (version 7.0.4): `string(1) "r"`
Result using hhvm (version 3.13.1): `string(0) ""`

Apparently, the mode for the stream cannot be determined using `stream_get_meta_data()` in hhvm in the case of pipes. For what it's worth, the [manual](http://php.net/manual/en/function.stream-get-meta-data.php) does not mention streams using `popen()`, so the behaviour of hhvm is technically not incorrect.

Also, although the metadata provided by hhvm using `stream_get_meta_data()` on a pipe states that the stream is seekable, a `fseek($fh, 1, SEEK_CUR)` seems to always return an error (`-1`) and does not change the stream pointer.

With the current behaviour of hhvm, I think it will be hard to get pipes supported by `Stream` correctly.

**Update**: I have reported the issue at hhvm: https://github.com/facebook/hhvm/issues/7089

Slim/Http/Stream.php
```diff
@@ -73,6 +78,13 @@ class Stream implements StreamInterface
     protected $size;
 
     /**
+     * Is this stream a pipe?
+     *
+     * @var type
+     */
+    protected $pipe;
```
akrabat: Please name this property `isPipe`.

Slim/Http/Stream.php
```diff
@@ -20,6 +20,11 @@
 class Stream implements StreamInterface
 {
     /**
+     * This is octal as per header stat.h
```
akrabat, line 25: Can you add info on what this particular mode means?

Slim/Http/Stream.php
```diff
@@ -73,6 +78,13 @@ class Stream implements StreamInterface
     protected $size;
 
     /**
+     * Is this stream a pipe?
+     *
+     * @var type
```
akrabat: This is a `boolean` I think

Slim/Http/Stream.php
```diff
@@ -158,6 +170,7 @@ public function detach()
         $this->writable = null;
         $this->seekable = null;
         $this->size = null;
+        $this->pipe = null;
```
akrabat: Should default to `false`.

Slim/Http/Stream.php
llvdl: That would collide with the `isPipe` method

Slim/Http/Stream.php
llvdl: The value is cached. If `null` the value is determined and then set in `isPipe()`.

Slim/Http/Stream.php
akrabat: No. PHP can tell the difference between a property and a method with the same name…

Slim/Http/Stream.php
akrabat: Ah. In that case. When is it cleared?

llvdl: In `Stream::detach()`

Slim/Http/Stream.php
llvdl: I didn't know that :)

Slim/Http/Stream.php
llvdl, line 25: I expanded the comment

PR #1875, opened 2016-05-16
Fix infinite loop issue created in 36cdaa2f

epitomus:


akrabat:
Tests fail on HHVM....

PR #1880, opened 2016-05-16
Adding Test for urlFor

danielgsims:
- Tested that urlFor triggers deprecation error
- Tested that urlFor aliases pathFor

### DIFF
--- a/tests/RouterTest.php
+++ b/tests/RouterTest.php
@@ -400,4 +400,40 @@ class RouterTest extends \PHPUnit_Framework_TestCase
         $dispatcher2 = $method->invoke($this->router);
         $this->assertSame($dispatcher2, $dispatcher);
     }
+
+    /**
+     * @expectedException \RuntimeException
+     */
+    public function testLookupRouteThrowsExceptionIfRouteNotFound()
+    {
+        $this->router->lookupRoute("thisIsMissing");
+    }
+
+    /**
+     * PHPUnit will convert the triggered error into a Framework error,
+     *
+     * @expectedException \PHPUnit_Framework_Error
+     */
+    public function testUrlForTriggersError()
+    {
+        $this->router->urlFor('foo', ['name' => 'josh'], ['a' => 'b', 'c' => 'd']);
+    }
+
+    /**
+     * Test that the router urlFor will proxy into a pathFor method.
+     */
+    public function testUrlForAliasesPathFor()
+    {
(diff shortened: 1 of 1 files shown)


danielgsims:
Testing this message gives us full coverage on the Router. I wasn't sure how others would feel by using a partial mock for testing the map urlFor -> pathFor. I thought it was okay because we don't care about that functionality, because it's tested in pathFor, we just want to make sure the alias works. I'd appreciate any feedback if anyone disagrees.

danielgsims:
@akrabat Do you know anything about the HVVM Travis build? It looks like the error was converted into an exception, so the test goes on to run the subsequent code which shouldn't run. Any ideas?

dopesong:
For proper test you should expect for error, not to suppress it.

Example how to do so:

``` php
<?php
class MyTest extends PHPUnit_Framework_TestCase
{
    private $errors;

    protected function setUp() {
        $this->errors = array();
        set_error_handler(array($this, "errorHandler"));
    }

    public function errorHandler($errno, $errstr, $errfile, $errline, $errcontext) {
        $this->errors[] = compact("errno", "errstr", "errfile",
            "errline", "errcontext");
    }

    public function assertError($errstr, $errno) {
        foreach ($this->errors as $error) {
            if ($error["errstr"] === $errstr
                && $error["errno"] === $errno) {
                return;
            }
        }
        $this->fail("Error with level " . $errno .
            " and message '" . $errstr . "' not found in ", 
            var_export($this->errors, TRUE));
    }

    public function testDoStuff() {
        // execute code that triggers a warning
        $this->assertError("Message for the expected error",
            E_USER_WARNING);
    }
}
```

danielgsims:
@dopesong - Thanks for the feedback!

To address the '@' - There were two tests, one that tests whether the error is triggered, and another to test the functionality afterwards. I think the suppression operator was necessary because PHPUnit will convert the error to an exception through PHPUnit, right?

Also, with that strategy, how does PHPUnit keep that error handler attached? Just for the life of the one test case, the file, or the whole shebang? Do we need to use restore_error_handler in there?

dopesong:
With that strategy you can check if exact error triggered. Also you are correct that error handler must be restored, because set_error_handler is global.

Better to expect for error and value returned than suppress it.

danielgsims:
I made some changes based on @dopesong's feedback. Closing, please see #1882

PR #1882, opened 2016-05-17
Testing Router Methods

danielgsims:
- Testing sad path from route lookup
- Testing that urlFor proxies pathFor but triggers warning


danielgsims:
@dopesong - Writing the error handler this way helped us get around the weird error-conversion issue with the HVVM build on travis too

dopesong:
Glad to see that ;)

PR #1883, opened 2016-05-17
Container converts not UnexpectedValueException but InvalidArgumentException

dopesong:
Tests correct exception


danielgsims:
You shouldn't need the annotation and calling set expected exception.

dopesong:
I know. Trying to understand what is wrong with HHVM.

PR #1884, opened 2016-05-17
Check if getReasonPhrase returns empty message

dopesong:


tests/Http/ResponseTest.php
```diff
@@ -114,11 +114,13 @@ public function testWithStatusEmptyReasonPhrase()
     public function testGetReasonPhrase()
     {
         $response = new Response();
+        $responseWithNoMessage = new Response(310);
         $responseStatus = new ReflectionProperty($response, 'status');
         $responseStatus->setAccessible(true);
         $responseStatus->setValue($response, '404');
 
         $this->assertEquals('Not Found', $response->getReasonPhrase());
+        $this->assertEquals('', $responseWithNoMessage->getReasonPhrase());
```
akrabat: Please add this one as it's own test method. 
Also, why does the 404 test use Reflection to set the Status?!

PR #1886, opened 2016-05-20
Bumping version

danielgsims:
I opened this as a question really. Is the intention to bump the release with future versions? If so, it's grown stale. Thanks!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -55,7 +55,7 @@ class App
      *
      * @var string
      */
-    const VERSION = '3.0.0';
+    const VERSION = '3.4.0';
 
     /**
      * Container


geggleto:
we should probably drop this const :D /cc @akrabat

danielgsims:
I did a quick search on github, and I found some folks using it :(

geggleto:
Oh god no......

akrabat:
I've updated `VERSION` for the 3.4.2 release and have set it on 3.x`to a`-dev`number. Don't trust the number on`3.x` though...

codeguy:
This should be automated or removed imo.

akrabat:
Yes - automation would be nice. We can't remove until 4.0…

danielgsims:
Thanks @akrabat!

PR #1887, opened 2016-05-20
Testing header functionality in App respond.

danielgsims:
- Created a Global Until class to wrap php global functions
  that are difficult to mock
- Added Global Until to Services Provider
- Created Mock Global Util
- Tested that respond method sets headers as expected

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -353,10 +353,12 @@ class App
      */
     public function respond(ResponseInterface $response)
     {
+        $globalUtil = $this->container->get('globalUtil');
+
         // Send response
-        if (!headers_sent()) {
+        if (!$globalUtil->headersSent()) {
             // Status
-            header(sprintf(
+            $globalUtil->header(sprintf(
                 'HTTP/%s %s %s',
                 $response->getProtocolVersion(),
                 $response->getStatusCode(),
@@ -366,7 +368,7 @@ class App
             // Headers
             foreach ($response->getHeaders() as $name => $values) {
                 foreach ($values as $value) {
-                    header(sprintf('%s: %s', $name, $value), false);
+                    $globalUtil->header(sprintf('%s: %s', $name, $value), false);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


danielgsims:
I decided to add a new class to mock these methods instead of making them private variables within App. I thought it made for better testing, and increases coverage on App. We can accomplish testing the functionality if we want to create internal function wrappers for headers, headers_sent, and connection status. We would just test it using a partial mock, but I think this is better.

I decided to go with a concrete Mock GlobalUtil class instead of using PHPUnit to try to build a mock. I tried both ways, and the tests turned out more readable and less brittle with a real class to test against. Otherwise, I ended up with a lot of lines like this:

```
 $this->expects($this->at(1))->method('header')...
```

And I broke them pretty easily. I'd appreciate any feedback.

akrabat:
I don't know how to solve this, but adding a new wrapper class used in `App` isn't the right solution. Sorry.

danielgsims:
That's fair, that's why I was asking for some feedback. There are two other options I can think of:

One would be creating those private functions in app. Another would be stashing wrappers to those functions in pimple.

Solution A, which lets you test using a partial mock and is less intrusive:

```
class App
{
    private function headersSent(...)
    {
          return headers_sent
     }

    public function respond()
    {
         if (!$this->headersSent()) {
         }
    }
}
```

Solution B, put the functions in pimple, and change them with mocks at test time

```
$container['headers_sent'] = $container->protect(function() {
     return headers_sent();
});
```

I'll open both and post a diff link to get some feedback to see which one feels better

danielgsims:
@akrabat or @silentworks 

Here is a gist of the idea

Solution A:
https://github.com/slimphp/Slim/compare/3.x...danielgsims:headers-app?expand=1

Solution B:
https://github.com/slimphp/Slim/compare/3.x...danielgsims:headers-pimple?expand=1

I like Solution A better because it's simpler and would work better in IDEs, but solution B gets you some dependency inversion. What do you guys think?

PR #1889, opened 2016-05-22
Small fixes

llvdl:


danielgsims:
There may be a conflict with type hinting the methods as an array. The FastRoute RouteCollector::addRoute method accepts a string or array of strings of http methods. This is the only place I think we use the Route::getMethods function. So technically, even though the phpdoc documents an array of strings, if you use a string everything should work. This could be a BC break.

llvdl:
@danielgsims, thanks for looking into this. The framework seems to consistently call `Route` with an array of methods, but the `Route` constructor is public, so this could be BC.

I have removed the typehint and added a check in the constructor to convert the parameter to an array if a string is passed. This way `Route::getMethods` always returns an array, as stated in the docblock. The docblock of the constructor is changed so it is clear that either a string or an array can be used.

danielgsims:
Maybe go make a note on the Slim 4 road map discussion!

Slim/Route.php
```diff
@@ -72,15 +72,15 @@ class Route extends Routable implements RouteInterface
...
+     * @param string|string[]   $methods The route HTTP methods
+     * @param string            $pattern The route pattern
+     * @param callable          $callable The route callable
+     * @param RouteGroup[]      $groups The parent route groups
+     * @param int               $identifier The route identifier
      */
     public function __construct($methods, $pattern, $callable, $groups = [], $identifier = 0)
     {
-        $this->methods  = $methods;
+        $this->methods  = is_string($methods) ? [$methods] : $methods;
```
akrabat, line 83: Why add support for `string`?

llvdl, line 83: Because FastRoute supports both a string parameter and an array of string. Because there is no `array` typehint on `$methods`, `string` was already supported. I updated the docblock to reflect this. 

Also, when setting `Route::$methods` in the constructor, the value is set to an array, so `Route::getMethods()` always returns an array as stated in the docblock.

PR #1896, opened 2016-05-27
Env REQUEST_URI can contain query params.

adambro:
This is a typical scenario on tests, where REQUEST_URI contains path and
all params to perform a request. No need to set QUERY_STRING in
environment separately.


akrabat:
Which tests don't set up the Request object correctly?

adambro:
I've meant my own test cases where I use the mock factory to create
request. It is not about tests from framework itself.

Anyway REQUEST_URI can contain query string parameters so would be nice to
account for it.

Issue #1890, opened 2016-05-22
Undocumented "slim.files" option


llvdl:
In the function `Slim\Http\UploadedFile::createFromEnvironment` I found the following lines:

``` php
        if (is_array($env['slim.files']) && $env->has('slim.files')) {
            return $env['slim.files'];
        } // else ...
```

The "slim.files" option doesn't seem to be documented and isn't covered by any tests. Is there a reason for keeping it in? And if so, what is to the use case?

Raistlfiren:
`slim.files` is used for testing purposes. You can view the comment here - https://github.com/slimphp/Slim/issues/1099#issuecomment-83735868

llvdl:
Is this for testing an application that uses Slim?

`slim.files` is not used by Slims internal tests. If I remove the `slim.files` part, the tests still succeed.

akrabat:
Now used in the unit test `testCreateFromEnvironmentFromUserData`.

PR #1899, opened 2016-05-28
gh-1890 removed "slim.tests" environment option in UploadFile that wa…

llvdl:
…s not used in tests

### DIFF
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -75,21 +75,12 @@ class UploadedFile implements UploadedFileInterface
     protected $moved = false;
 
     /**
-     * Create a normalized tree of UploadedFile instances from the Environment.
-     *
-     * @param Environment $env The environment
-     *
-     * @return array|null A normalized tree of UploadedFile instances or null if none are provided.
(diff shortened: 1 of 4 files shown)


dopesong:
You removed `slim.files`, not `slim.tests` :)

JoeBengalen:
bc break

Slim/Http/UploadedFile.php
```diff
@@ -75,21 +75,12 @@ class UploadedFile implements UploadedFileInterface
...
-     * Create a normalized tree of UploadedFile instances from the Environment.
-     *
-     * @param Environment $env The environment
-     *
-     * @return array|null A normalized tree of UploadedFile instances or null if none are provided.
+     * Parses the $_FILES global
+     * @return array A normalized tree of UploadedFile instances or null if none are provided.
      */
-    public static function createFromEnvironment(Environment $env)
+    public static function createFromGlobals()
```
akrabat, line 81: We can't change public method signatures or how they operate within the 3.x series.

akrabat:
We can't remove `slim.files` as an application may be using it.

PR #1901, opened 2016-05-30
Move valid protocol versions to a class static var

pine3ree:
Just a matter of taste: i think it looks uglier as a function static var. 
Also, in this way it's also available to all Message(s)

### DIFF
--- a/Slim/Http/Message.php
+++ b/Slim/Http/Message.php
@@ -32,6 +32,17 @@ abstract class Message implements MessageInterface
     protected $protocolVersion = '1.1';
 
     /**
+     * A map of valid protocol versions
+     *
+     * @var array
+     */
+    protected static $validProtocolVersions = [
+        '1.0' => true,
+        '1.1' => true,
+        '2.0' => true,
+    ];
+
+    /**
      * Headers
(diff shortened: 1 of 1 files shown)


akrabat:
Why does coverage go down?

OptimusCrime:
Looks like an error in coveralls or something. These two files have identical coverage:
https://coveralls.io/builds/6341055/source?filename=Slim%2FDefaultServicesProvider.php
https://coveralls.io/builds/6390776/source?filename=Slim%2FDefaultServicesProvider.php

So does these:
https://coveralls.io/builds/6341055/source?filename=Slim%2FApp.php
https://coveralls.io/builds/6390776/source?filename=Slim%2FApp.php

The first of each is the current coverage in master (the last push) and the last is the coverage for this pull request.

pine3ree:
@akrabat I was wondering the same thing :-)

llvdl:
The lines of code with the `$valid` static variable that where in the function `withProtocolVersion` had code coverage. Those lines have been taken out and the total count of lines with code coverage has decreased because of this (static class variables don't count for code coverage).

PR #1904, opened 2016-06-05
Update Route.php

afief:
if route callback returns an array or an object, then alter the response into json.


akrabat:
This is application level, not Slim level.

silentworks:
As stated above, we don't think this should be part of the framework itself. You can do this in your application via a Middleware or other means if you wish. On that basis I am closing this PR.

PR #1906, opened 2016-06-08
added a method to create Route instance

ghost:
## By doing so one can
- create a custom Router class inheriting from Slim\Router
- redefine $container['router'] to return this custom Router instance
- override createRoute method to return a custom Route (inheriting from Slim\Route)
- add some extra methods in custom Route class (ex: getIsSecured())

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -153,7 +153,7 @@ class Router implements RouterInterface
         $methods = array_map("strtoupper", $methods);
 
         // Add route
-        $route = new Route($methods, $pattern, $handler, $this->routeGroups, $this->routeCounter);
+        $route = $this->createRoute($methods, $pattern, $handler);
         $this->routes[$route->getIdentifier()] = $route;
         $this->routeCounter++;
 
@@ -177,6 +177,14 @@ class Router implements RouterInterface
             $request->getMethod(),
             $uri
         );
+    }
+
+    /**
+     * @return Slim\Route
+     */
+    protected function createRoute($methods, $pattern, $callable)
+    {
+        return new Route($methods, $pattern, $callable, $this->routeGroups, $this->routeCounter);
     }
 
     /**
@@ -244,7 +252,7 @@ class Router implements RouterInterface
         }
         throw new RuntimeException('Named route does not exist for name: ' . $name);
     }
-    
+
     /**
      * Remove named route
      *


marcelbonnet:
@vicolachips44 , I also was looking for a way to set my own implementation of `\Slim\Route` . Good one, I'll try it later. 

But, @codeguy , @silentworks  is this ok for merge? It would be much better to use this approach as a Slim official release, so we would not break compatibility in near future official releases, please!

akrabat:
Why not just override `map`?

marcelbonnet:
@akrabat , thanks for your question! See the piece of code below. IMHO, it push us to change to much core code just to insert a new function to the inherited class. 

I am concerned about overriding map:
- what if changes occour to Router::map in near future? It would break the approach
- we have to create two new classes, an inheritance of Router and Route, because we do not have direct access to change the way Router instantiate a type of Route (see below)
- and all this work just to instantiate another Route (like the example, RouteB) ?

That is why I agree with this merge request. Seems reasonable. Correct me if I'm wrong.

``` php
class RouteB extends \Slim\Route{

    public function allow($roles){
        //my new function
    }
}

class RouterB extends \Slim\Router {
    /**
     *
     * {@inheritDoc}
     * @see \Slim\Router::map()
     */
    public function map($methods, $pattern, $handler)
    {
        if (!is_string($pattern)) {
            throw new InvalidArgumentException('Route pattern must be a string');
        }

        // Prepend parent group pattern(s)
        if ($this->routeGroups) {
            $pattern = $this->processGroups() . $pattern;
        }

        // According to RFC methods are defined in uppercase (See RFC 7231)
        $methods = array_map("strtoupper", $methods);

        /* ******************************************************
                 * Is this approach ok just to instantiate RouteB instead of Slim\Route ????? 
                 *******************************************************
                */
        $route = new RouteB ($methods, $pattern, $handler, $this->routeGroups, $this->routeCounter); 
        $this->routes[$route->getIdentifier()] = $route;
        $this->routeCounter++;
        return $route;
    }
}

$container['router'] = new RouterB();
```

akrabat:
> we have to create two new classes, an inheritance of Router and Route, because we do not have direct access to change the way Router instantiate a type of Route (see below)

You'll still have to do this.

marcelbonnet:
> You'll still have to do this.

but the point is: All we need is just to tell `Router` what kind of custom `Route` implementation we want. We don't want to change other parts of `Router::map`. 

If  `Router::map` make a call to a new function `Router::createRoute($methods, $pattern, $callable)` , it is done, no more `map` code replication :)

PR #1919, opened 2016-06-30
Fix namespaces usage. Correct phpdocs

vkhramtsov:
Fixes:
- Remove import for classes from global namespace
- Remove imports for classes which actually in the same namespace
- Correct code due to previous point
- Remove obsolete `use`s
- Order uses alphabetically

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -8,22 +8,19 @@
  */
 namespace Slim;
 
-use Exception;
-use Throwable;
-use Closure;
+use FastRoute\Dispatcher;
+use Interop\Container\ContainerInterface;
 use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
-use Interop\Container\ContainerInterface;
-use FastRoute\Dispatcher;
-use Slim\Exception\SlimException;
+use Psr\Http\Message\ServerRequestInterface;
 use Slim\Exception\MethodNotAllowedException;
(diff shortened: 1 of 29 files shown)


silentworks:
Thank you for wanting to contribute to the framework, but I can't see any benefit in what has been done in this PR. On that basis, I will be rejecting this PR at this point in time. We look forward to your future contributions.

vkhramtsov:
@silentworks  I think it's better to use code style. Sometimes you are using classes from global namespace with adding them to `use` (like `Exception`), sometime you use just `\Exception` without importing (also you can compare `Collection` class and `CollectionInterface`). It looks like you use different code style in different files.
Also I see obsolete `use` statements in some files. I think that it's better to remove them instead of keeping garbage in source code. What do you think?

opengeek:
For what it is worth, in my opinion we should always import classes, even from global namespace. Otherwise, I agree that unused `use` statements should be removed and utilized ones organized alphabetically.

tuupola:
I also agree unused `use` statements should be removed. I think the problem with the PR was that it does too many things at once. I would break it down to separate PR for example for PHPDoc fixes and one for removing the `use` statements.

akrabat:
>  I would break it down to separate PR for example for PHPDoc fixes and one for removing the use statements.

I agree. Smaller focussed PRs are easier for us to digest and understand what's going on and why.

vkhramtsov:
And again, should I remove uses for classes from global namespace and add `\` for this classes in code or just remove obsolete `use` statements? I'm asking this because after adding `\` in type hints, I have to update phpDoc blocks.
May I sort `use` statements by alphabet?

Thank you.

PR #1920, opened 2016-06-30
Added more tests for UploadedFiles

dopesong:


tests/Http/UploadedFilesTest.php
```diff
@@ -185,6 +276,45 @@ public function providerCreateFromEnvironment()
                     ],
                 ]
             ],
+            // array of files as multidimensional array: <input name="avatars[]" type="file">
+            [
+                // $_FILES array
```
akrabat, line 281: $_FILES doesn't look like this at all

akrabat: http://php.net/manual/en/features.file-upload.multiple.php explains the format when you have two files uploaded with the same name as an array.

akrabat, line 281: Ignore me - I'm thinking of the case where the name is `foo[bar][]`… see https://github.com/php-fig/fig-standards/pull/746/files

dopesong, line 281: /ignore @akrabat

PR #1921, opened 2016-07-01
Add parameter to Container::get() to allow it to return false if a re…

robertgdev:
…quested entry is not found

There may be instances where you wish to check if a requested container entry already exists. While this is currently possible with a try/catch block, this patch makes this easier by allowing the function to return false instead of throwing an exception. This change is backwards compatible and should not introduce any issues.


prisis:
You dont follow the container Interface 

https://github.com/container-interop/container-interop

akrabat:
As @prisis point out, this method is defined by [container-interop](https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/ContainerInterface.php#L16-L26) and so can't be changed.

PR #1928, opened 2016-07-05
Add getCookieParam method

martynbiz:
Consistent with getParam and getQueryParam

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -670,6 +670,26 @@ class Request extends Message implements ServerRequestInterface
      ******************************************************************************/
 
     /**
+     * Fetch cookie value from cookies sent by the client to the server.
+     *
+     * Note: This method is not part of the PSR-7 standard.
+     *
+     * @param      $key
+     * @param null $default
+     *
+     * @return null
+     */
+    public function getCookieParam($key, $default = null)
+    {
+        $getCookies = $this->getCookieParams();
+        $result = $default;
+        if (isset($getCookies[$key])) {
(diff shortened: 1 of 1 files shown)


dopesong:
Tests missing :)

AlexDanault:
I'd also add that getCookieParam should be after getCookiesParams and not before, to be consistent with getQueryParam (which is after getQueryParams) and getParsedBodyParam (also after getParsedBodyParams)

martynbiz:
Hi, thanks for the feedback. Sorry, I'm a little new to creating PR. I hope I've done it correctly here - https://github.com/slimphp/Slim/pull/1929 - let me know if there is anything I've missed or done incorrectly. Thanks :)

dopesong:
So you don't need to create new PR, just add new commits to existing branch and PR will be updated automatically. :)

Slim/Http/Request.php
```diff
@@ -670,6 +670,26 @@ public function getContentLength()
...
+    public function getCookieParam($key, $default = null)
+    {
+        $getCookies = $this->getCookieParams();
+        $result = $default;
+        if (isset($getCookies[$key])) {
+            $result = $getCookies[$key];
+        }
+        return $result;
+    }
+
```
jony4, line 691: How about this?

```
    public function getCookieParam($key)
    {
        $cookies = $this->cookies;
        if (isset($cookies[$key])) {
            return $cookies[$key];
        }
        return false;
    }

```

akrabat:
Replaced by #1929

PR #1929, opened 2016-07-05
Add getCookieParam method

martynbiz:
Added new getCookieParam method with tests

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -670,27 +670,6 @@ class Request extends Message implements ServerRequestInterface
      ******************************************************************************/
 
     /**
-     * Fetch cookie value from cookies sent by the client to the server.
-     *
-     * Note: This method is not part of the PSR-7 standard.
-     *
-     * @param      $key
-     * @param null $default
-     *
-     * @return null
-     */
-    public function getCookieParam($key, $default = null)
-    {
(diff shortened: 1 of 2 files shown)


martynbiz:
What's went wrong here? I see a few tests are being skipped, but that was the case before I made changes. Is that what's prevent 1 check from passing?

akrabat:
@martynbiz: https://i.19ft.com/60f75283.png

(https://travis-ci.org/slimphp/Slim/jobs/142474362#L299-L330)

martynbiz:
Thanks for the screenshot, I think it's OK now?

akrabat:
Thanks!

Slim/Http/Request.php
```diff
@@ -685,6 +685,27 @@ public function getCookieParams()
     }
 
     /**
+     * Fetch cookie value from cookies sent by the client to the server.
+     *
+     * Note: This method is not part of the PSR-7 standard.
+     *
+     * @param     $key
```
basert, line 692: Can the phpdoc get fixed?

akrabat, line 692: @basert, This is what I committed: https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L687L696

Nearly got it right :)

PR #1946, opened 2016-08-04
Fix url mismatch when the base path contains a space

JoeBengalen:
Related to this issue:
http://discourse.slimframework.com/t/how-to-use-routes-well/573

When the application has a base path that contains a space the base path will not be set correctly. This is because in the `REQUEST_URI` the space is rawurlencoded and in the `SCRIPT_NAME` it is not.

This behavour can be seen in his example, and I also tested this om my pc.

``` php
array(
 'REQUEST_URI' => "/PHP/Lumy%203/tests/",
 'SCRIPT_NAME' => "/PHP/Lumy 3/tests/index.php",
);
```

I could not simply `rawurlencode` the entire `SCRIPT_NAME` as slashes must remain untouched.

To fix this particular issue I just replaced the space, but there may be other characters that should be encoded.

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -197,7 +197,8 @@ class Uri implements UriInterface
         }
 
         // Path
-        $requestScriptName = parse_url($env->get('SCRIPT_NAME'), PHP_URL_PATH);
+        $envScriptName = str_replace(' ', '%20', $env->get('SCRIPT_NAME'));
+        $requestScriptName = parse_url($envScriptName, PHP_URL_PATH);
         $requestScriptDir = dirname($requestScriptName);
 
         // parse_url() requires a full URL. As we don't extract the domain name or scheme,
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -197,13 +197,13 @@ class Uri implements UriInterface
         }
 
         // Path
-        $envScriptName = str_replace(' ', '%20', $env->get('SCRIPT_NAME'));
-        $requestScriptName = parse_url($envScriptName, PHP_URL_PATH);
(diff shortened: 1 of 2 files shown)


dopesong:
Why not to rawurlencode script name and `str_replace('%2F', '/', $scriptName)` instead?

JoeBengalen:
It could work. I don't know if I will break anything with rawurlencode, other than the slashes.

akrabat:
BasePath is a real nuisance, isn't it!

akrabat:
I'm not convinced that the REQUEST_URI is simply rawurlencoded - it appears to percent encode any non-ASCII characaters:

Given a directory of `Rob's%20tésting`, I see this in `$_SERVER`:

```
'REQUEST_URI' => string '/Rob's%20te%cc%81sting/public/' (length=30)
'SCRIPT_NAME' => string '/Rob's tésting/public/index.php' (length=33)
```

However, the `rawurlencode` of that string is:

```
Rob%27s%20testing
```

As such, maybe we need to go the other way and `rawurldecode` REQUEST_URI and then test?

akrabat:
i.e. would it be better to add a `$requestUri = rawurldecode($requestUri);`?

JoeBengalen:
I can try that

It has some weird behavior that request _uri 😒

akrabat:
@JoeBengalen Did you get a chance to look at this?

JoeBengalen:
whoops, I kinda forgot about this. Will do so tonight!

JoeBengalen:
Updated!

Note that `getBasePath` encodes the `'` into `%27`

akrabat:
> Note that getBasePath encodes the ' into %27

Is this a change?

akrabat:
Thanks for your patience @JoeBengalen! 

I've decided that this is a bug-fix and so is good to go.

PR #1956, opened 2016-08-13
Add getServerParam method

joelhy:
Add getServerParam method to `Slim\Http\Request`.

It's inconvenient to use server params without a method to get single parameter from $_SERVER.
Just like getCookieParam() to getCookieParams() , getParsedBodyParam() to getParsedBodyParams().


akrabat:
Thanks. Would you be able to create a PR for the [docs](https://github.com/slimphp/Slim-Website) to document this? I suspect that `getServerParams` also needs documenting!

Issue #1962, opened 2016-08-16
Replacing Route "callable"


geggleto:
It came to mind that it might be required to replace a "callable" on a Route.

We should add a `Route::setCallable(callable/string $callable)`

This would be a step forward in fixing/adding #1938

geggleto:
For example: If I have a website A that implements a bunch of end-points... and I install package B that overrides some of the end-points... currently that process is a non-starter.

akrabat:
Fixed with #1963

PR #1963, opened 2016-08-16
Related to #1962. Implementation with UnitTest

geggleto:
https://github.com/slimphp/Slim/pull/1962

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -122,6 +122,16 @@ class Route extends Routable implements RouteInterface
     }
 
     /**
+     * This method enables you to override the Route's callable
+     *
+     * @param string|\Closure $callable
+     */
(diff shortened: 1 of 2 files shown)


akrabat:
Please put `setCallable()` immediately after `getCallable()`.

Also, while we're here, can we add a protected property for $this->callable?

mathmarques:
@akrabat $callable is already protected, [see here](https://github.com/slimphp/Slim/blob/3.x/Slim/Routable.php#L28).

geggleto:
@mathmarques makes more sense to surface it for readability... I thought it was super weird that it was buried.

akrabat:
ah! - it's in the parent…

geggleto:
@akrabat I moved the method as well ;)

PR #1967, opened 2016-08-20
Add leading slash if missing in route patterns

vlakoff:
This PR allows to omit the leading slash when defining routes.

For the record, I have a lot of routes like so:

``` php
$app->get('/about', function () {
    return view('about');
});
```

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -143,6 +143,9 @@ class Router implements RouterInterface
             throw new InvalidArgumentException('Route pattern must be a string');
         }
 
+        // Ensure pattern has a leading slash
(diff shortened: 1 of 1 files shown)


vlakoff:
Wow, that's an impressive testing suite!

mathmarques:
Route pattern is just a simple concatenation. [See this](https://github.com/slimphp/Slim/pull/1626) for more information.

Slim used to do the way you proposed, but it's better without all the "magic".

vlakoff:
Yes this "no magic" approach seems to have many advantages. I'll live with the leading slash.

Thanks for pointing this out.

Issue #1818, opened 2016-03-14
Response withJSON method overwriting not replacing body


AmibeWebsites:
Should rewind completely replace the response body?

I'm using withJSON and is seems that instead of replacing the body it's overwriting, and presume fopen uses r+ instead of w+.

```
return $response->withJSON('abcd')->withJSON('xyz');
// returns 'xyzd'
```

AmibeWebsites:
Looking at the code, rewind is doing exactly what it's supposed to do: "Seek to the beginning of the stream."

So if the intention of the Response withJSON method is to replace the response body (as I believe it is) then it's not working as intended.

So perhaps replace:

```
        $body = $this->getBody();
        $body->rewind();
```

with

```
        $this->body = $body = new Body(fopen('php://temp', 'w+'));
```

Not sure if anything would be lost this way, like body meta, that should be added back in?

And should clone be used instead, like this?

```
        $clone = clone $this;
        $clone->body = new Body(fopen('php://temp', 'w+'));
        $clone->body->write($json = json_encode($data, $encodingOptions));
```

geggleto:
Well you can do `->withBody()` ... https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Message.php#L287-L294

atraianovschi:
Hello,

I don't see why this issue was closed because the current method `->withJson()` is just overwriting the stream and not replacing the current body with the new JSON. And I really think this is the approach that everyone is expecting from this method. 

For example I have a middleware that should filter the response based on some rights. The `->withJson()` method works perfectly when I create the response and when it goes to the middleware I should filter that response and replace it with the filtered version of it, with same method `->withJson()`.

If the method withBody is replacing the content, and withStatus is also replacing the status, why withJson should not replace the content ?

akrabat:
I agree with @deddus - `withJson` should instantiate a new `Body` as the current Body could be anything…

robertprice:
I've put a pull request in for this. The code is along the lines @dudewithamood suggested.

PR #1975, opened 2016-09-01
Ensure withJson returns a new Response

robertprice:
Fixes issue #1818. The withJson method now clones the Response, and
updates this with a new header, and a new Body.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -306,16 +306,16 @@ class Response extends Message implements ResponseInterface
      */
     public function withJson($data, $status = null, $encodingOptions = 0)
     {
-        $body = $this->getBody();
-        $body->rewind();
-        $body->write($json = json_encode($data, $encodingOptions));
+        $clone = clone $this;
+        $clone->body = new Body(fopen('php://temp', 'w+'));
+        $clone->body->write($json = json_encode($data, $encodingOptions));
 
         // Ensure that the json encoding passed successfully
         if ($json === false) {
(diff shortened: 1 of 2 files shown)


Slim/Http/Response.php
```diff
@@ -306,16 +306,16 @@ public function withRedirect($url, $status = null)
      */
     public function withJson($data, $status = null, $encodingOptions = 0)
     {
-        $body = $this->getBody();
-        $body->rewind();
-        $body->write($json = json_encode($data, $encodingOptions));
+        $clone = clone $this;
+        $clone->body = new Body(fopen('php://temp', 'w+'));
```
mathmarques: I think it's better to use `$newResponse = $this->withBody(new Body(fopen('php://temp', 'r+')))` instead of clone and set.

akrabat: Where does `newResponse` come from?

I like clone as it's quick.

mathmarques: I was just saying to use [withBody](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Message.php#L296), cause if we have a method for return a new response with a new body, we don't need to reimplement it.

robertprice: I think @mathmarques makes a good point. [withBody](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Message.php#L296) clones and sets the body, so it's the same code really. It may make more sense to do this instead of duplicating code.

robertprice:
Merging #1978 will fix the php7 failing tests.

akrabat:
Thanks!

PR #1977, opened 2016-09-02
use stdClass as a dummy during testing for non-existent classes.

robertprice:
phpunit 5.5.2 on PHP7 complains if non-existent classes are mocked, giving an undefined index error. I have replaced two mock classes that don't exist with stdClass to mitigate this.


akrabat:
Why did you close this?

akrabat:
Ah - Replaced with #1978

robertprice:
Yes :-)

PR #1979, opened 2016-09-02
Extra tests

robertprice:
Some extra tests to increase code coverage.

### DIFF
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -165,14 +165,17 @@ class RequestTest extends \PHPUnit_Framework_TestCase
     {
         $uri = Uri::createFromString('https://example.com:443/foo/bar?abc=123');
         $headers = new Headers([
-            'Content-Type' => 'application/x-www-form-urlencoded',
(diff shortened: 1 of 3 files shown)


robertprice:
This requires #1978 to be merged to fix the php7 build issues.

tests/Http/RequestTest.php
```diff
@@ -165,14 +165,17 @@ public function testGetMethodWithOverrideParameterFromBodyObject()
     {
         $uri = Uri::createFromString('https://example.com:443/foo/bar?abc=123');
         $headers = new Headers([
-            'Content-Type' => 'application/x-www-form-urlencoded',
+            'Content-Type' => 'application/json',
```
akrabat, line 168: Why did you change this? Practically, _METHOD overrides are only going to be used with HTML forms.

robertprice: I was testing decoding of objects, but I can't remember the reasons now. 

I think it's safe for you to ignore this, and for me to learn to write better commit messages.

Issue #1983, opened 2016-09-07
Throwable exception uncaught in Route.php unlike in App.php


Starli0n:
Hi,

I found a bug while dealing with `phpErrorHandler`. I noticed that `Throwable` exceptions are not caught in `Route.php` on line 324 whereas this case is handle in `App.php`

``` php
    try {
        ob_start();
        $newResponse = $handler($this->callable, $request, $response, $this->arguments);
        $output = ob_get_clean();
    } catch (Exception $e) {
        ob_end_clean();
        throw $e;
    }
```

For `Throwable` exceptions, `ob_end_clean()` is never called.

It should be replace by:

``` php
    try {
        ob_start();
        $newResponse = $handler($this->callable, $request, $response, $this->arguments);
        $output = ob_get_clean();
    } catch (Exception $e) {
        ob_end_clean();
        throw $e;
    } catch (Throwable $e) {
        ob_end_clean();
        throw $e;
    }
```

Thanks

rrpadilla:
Following the documentation http://php.net/manual/en/language.errors.php7.php#119652, we should "add a catch block for Exception AFTER catching Throwable first".
Just switch catch blocks:

try {
        ob_start();
        $newResponse = $handler($this->callable, $request, $response, $this->arguments);
        $output = ob_get_clean();
    } catch (Throwable $e) {
        ob_end_clean();
        throw $e;
    } catch (Exception $e) {
        ob_end_clean();
        throw $e;
    }

Starli0n:
I put `Throwable` first.

Thanks

akrabat:
Thank you !

PR #1984, opened 2016-09-07
Fix Issue #1983: Throwable exception uncaught in Route.php

Starli0n:
Hi,

Could you please perform this pull request to fix the issue #1983.

Thanks

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -9,6 +9,7 @@
 namespace Slim;
 
 use Exception;
+use Throwable;
 use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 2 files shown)


dopesong:
Coverage decreased (-0.2%) to 97.003% - you should provide tests for this :)

Starli0n:
@dopesong 

I added a test coverage for PHP7 only.

I do not know how to do better as far as `Throwable` appeared in PHP7.

dopesong:
Check it here: https://github.com/slimphp/Slim/blob/3.x/tests/Handlers/PhpErrorTest.php#L88

Starli0n:
As far as the coverage is performed in PHP5 and following the documentation http://php.net/manual/en/language.errors.php7.php#119652, the `catch (Throwable $t)` code is never reached in PHP5

``` php
try
{
   // Code that may throw an Exception or Error.
}
catch (Throwable $t)
{
   // Executed only in PHP 7, will not match in PHP 5
}
catch (Exception $e)
{
   // Executed only in PHP 5, will not be reached in PHP 7
}
```

Therefore I added `@codeCoverageIgnore` statement

akrabat:
Thank you!

PR #2000, opened 2016-09-27
argument should be rawurldecode, not urldecode. 

luzluna:
in rfc2396.

>    The plus "+", dollar "$", and comma "," characters have been added to
>   those in the "reserved" set, since they are treated as reserved
>   within the query component.
> So reserved chars should be bypassed to users, if in argument field.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -515,7 +515,7 @@ class App
         if ($routeInfo[0] === Dispatcher::FOUND) {
             $routeArguments = [];
             foreach ($routeInfo[2] as $k => $v) {
-                $routeArguments[$k] = urldecode($v);
+                $routeArguments[$k] = rawurldecode($v);
             }
 
             $route = $router->lookupRoute($routeInfo[1]);


akrabat:
What are the BC implications?

luzluna:
if route defined like this
`$app->get('/hello/{name}', function (Request $request, Response $response) {
    $name = $request->getAttribute('name');`
and request is '/hello/+apple'

now
$name is ' apple'

if apply this patch
$name is '+apple'

johnhunt:
I'd definitely say this is a BC break. It might be a better idea to add a second argument to getAttribute() to return the value using rawurldecode():

`$app->get('/hello/{name}', function (Request $request, Response $response) { $name = $request->getAttribute('name', true);`

However, this might be a bit too uglifying..

akrabat:
`getAttribute()` is part of [PSR-7](https://github.com/php-fig/http-message/blob/master/src/ServerRequestInterface.php#L212-L227)...

johnhunt:
I feel as though this belongs in slim 4, the implications of putting it in the 3.x branch could be pretty severe. Does anyone else agree?

geggleto:
👍

geggleto:
@akrabat can we close this one and ask the author to target 4.x?

akrabat:
I hadn't noticed! @luzluna, please rebase against `4.x` and edit the PR to be against `4.x` too.

akrabat:
Can we raise a new PR for [Slim-Http](https://github.com/slimphp/Slim-Http) please?

PR #2034, opened 2016-11-14
Reuse getParsedBodyParam for determining method override in POST body.

sander-bol:
Makes `getMethod()` internally use `getParsedBodyParam()` for determining the value of `_METHOD` in the request body, for purposes of overriding the body.

Question though: wouldn't this type of method-juggling be more appropriately implemented as a middleware rather than living in the core Request object logic, so it becomes an opt-in behaviour?


akrabat:
Yes. `_METHOD` will certainly be moved to middleware for 4.0

akrabat:
Thanks.

sander-bol:
You're welcome, and thanks for accepting my little contribution!

PR #2037, opened 2016-11-16
Add HEAD shortcut route

### LINKED ISSUE #2033: Why are we not supporting HEAD in routes?
Hello

I was implementing a route like this: 

```
$app->group('/pattern', function () {

    $this->get('', '\MyClass:getMethodOnController');
    $this->head('', '\MyClass:headMethodOnController');

});
```

And Slim just crashes because apparently it doesn't support HEAD for routing patterns. Why? Is this an oversight or is there a good reason for it? I find it odd because the `$request` object has the HEAD method available like `$request->isHead();`

I was encouraged to open an issue on Slack.

chadicus:
Fixes #2033

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -142,6 +142,19 @@ class App
     }
 
     /**
+     * Add HEAD route
+     *
+     * @param string $pattern  The route URI pattern
+     * @param callable|string  $callable The route callback routine
+     *
+     * @return \Slim\Interfaces\RouteInterface
+     */
+    public function head($pattern, $callable)
+    {
+        return $this->map(['HEAD'], $pattern, $callable);
+    }
+
+    /**
      * Add POST route
      *
      * @param  string $pattern  The route URI pattern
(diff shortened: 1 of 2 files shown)


codeguy:
Is there a reason you want to define routes that respond ONLY to head requests? You should already be able to issue HEAD requests  against any GET route, you just won't  get back the response body, just the header.

codeguy:
See https://github.com/nikic/FastRoute#a-note-on-head-requests (which Slim uses under the hood)

chadicus:
@codeguy I just saw the issue discussion and thought I'd lend a hand by adding the method. I have no use case for HEAD requests. Feel free to reject the PR.

geggleto:
From FastRoute

> Finally, note that applications MAY always specify their own HEAD method route for a given resource to bypass this behavior entirely.

👍 for me.

akrabat:
Not needed as no one has a use case for responding to HEAD differently to GET.

garnold:
@akrabat https://github.com/slimphp/Slim/issues/2033#issuecomment-278341137

PR #2038, opened 2016-11-17
Remove codeCoverageIgnore annotations

amad:
As they are considered as bad habit!

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -343,11 +343,9 @@ class Route extends Routable implements RouteInterface
                 ob_start();
                 $newResponse = $handler($this->callable, $request, $response, $this->arguments);
                 $output = ob_get_clean();
-            // @codeCoverageIgnoreStart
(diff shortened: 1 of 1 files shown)


akrabat:
They are there for a reason: the code coverage is run in PHP5.6, so that code is never run…

amad:
I take your point about why added. But is it still the case? Because coverage report generated.

akrabat:
Yes - see that the coverage dropped? That's because of this: https://coveralls.io/builds/8878310/source?filename=Slim%2FRoute.php#L347

amad:
Fair enough :+1:
IMO I don't see difference between ignoring and leaving. either way is not green on this report.

But of course it's not big deal at all and I just send it because I didn't realize the reason behind it in the first place.
Thanks

akrabat:
Thanks for looking at it.

The difference for me is that we've noted that it's intentional.

PR #2066, opened 2016-11-20
Improve request parsing with invalid json

moritz-h:
Slim throws a runtime exception if json Request sends "semi valid" json code.

Here you expect json parsing to be an object, array or null:
[https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L1015](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L1015)

If i send invalid json everything is fine, because the result ist null. If i send "semi valid" json like `"foobar"`
the runtime exception gets triggered.
The Problem is that the php function json_decode ([called here](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L197)) decodes a little more than valid json as mentioned here:
[http://php.net/manual/en/function.json-decode.php](http://php.net/manual/en/function.json-decode.php)
So `"foobar"` gets decoded as string, which is not an object, array or null.

I think ether you should accept "semi valid" json as result like php does or you should handle it like invalid json (i think better). But a RuntimeException looks totally wrong here.

I don't know how this should be fixed. The pull request just has tests for what i would expect.

### DIFF
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -751,6 +751,36 @@ class RequestTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(['foo' => 'bar'], $request->getParsedBody());
     }
 
+    public function testGetParsedBodyInvalidJson()
+    {
+        $method = 'GET';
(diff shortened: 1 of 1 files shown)


akrabat:
Thanks. I've also fixed the bug.

PR #2067, opened 2016-11-22
Remove App::VERSION

### LINKED ISSUE #2058: Remove app version constant

codeguy:
Removes VERSION constant which was never updated.

Also fixes a Error Handler unit test that assumed unix line returns, which was not true when executed on Windows systems.

Fixes #2058

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -46,13 +46,6 @@ class App
     use MiddlewareAwareTrait;
 
     /**
-     * Current version
-     *
-     * @var string
-     */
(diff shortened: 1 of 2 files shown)


vlakoff:
This has been undone later, in #2472.

Thus, there is a line [in the changelog](https://github.com/slimphp/Slim/blob/4.x/CHANGELOG.md#removed) that should be removed.

vlakoff:
(by the way, that's an example it's not good to mix different changes in the same PR :)

PR #2075, opened 2016-11-23
2.x - Allow for mcrypt polyfils

RichardHeelin:
The php extension mcrypt has been deprectated in php 7.1, and as such users may be using a polyfil instead of the extension. In those cases the constants will exist but the extension won't be loaded, resulting in the constant attempting to be redefined and a Notice been thrown.


tuupola:
There is also this workaround: https://packagist.org/packages/tuupola/slim-fake-mcrypt

akrabat:
Thanks.

PR #2082, opened 2016-11-29
Send Exception also into NotFoundHandler

ghost:
**Motivation:** To be able to pass some aditional information into ErrorHandler.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -592,7 +592,7 @@ class App
             $params = [$e->getRequest(), $e->getResponse(), $e->getAllowedMethods()];
         } elseif ($e instanceof NotFoundException) {
             $handler = 'notFoundHandler';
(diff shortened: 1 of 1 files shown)


akrabat:
Can you explain the use-case a little more please as you didn't update the `NotFound` handler.

ghost:
When I raise some exception (for example: `PostNotFoundException`, `UserDoesNotExistsException`), it's handy (especially for inner APIs) to provide aditional info about error into response body.

With this you can inhert your own NotFoundException (with meta info - message text for example) and in your presenters repack your domain exceptions into this one.

Api 404 responses now can be richer with some body info.

ghost:
Also - it's expects you to have your own NotFoundHandler.

ghost:
A can update default one to take this exception in declaration but as it wont be used - im not sure its the right thing to do.

Issue #1955, opened 2016-08-12
Trouble mocking the POST body request


dmelo:
I saw a lot of references to use 'slim.input' to represent the body of a POST request, while testing. I searched the source code and saw that the code no longer handles 'slim.input'. How can I mock a POST request passing data through the body?

geggleto:
With something like the following...

``` php
    public function makeRequest($method, $url = '/') {
        $env = Environment::mock();
        $uri = Uri::createFromString('http://example.com'.$url);

        $headers = Headers::createFromEnvironment($env);
        $serverParams = $env->all();

        $body = new RequestBody();
        $uploadedFiles = UploadedFile::createFromEnvironment($env);
        $request = new Request($method, $uri, $headers, [], $serverParams, $body, $uploadedFiles);
        return $request;
    }


$request = $request->withParsedBody([
    //Your Parsed Body
]);
```

dmelo:
It worked. Thank you very much

PR #2086, opened 2016-12-02
adding MOCK_POST_DATA variable to `Environment`

famoser:
I need to mock the POST request body (for unit testing). Relevant stackoverflow question: http://stackoverflow.com/questions/40867377/mock-slim-endpoint-post-requests-with-phpunit
I'm aware of https://github.com/slimphp/Slim/issues/1955, but the solution provided there can not be used to test endpoints (as the `Request` is manually contructed).

It seems as this is impossible with Slim 3 so far (as the RequestBody directly reads from php://input which is read only). I therefore made this pull request to enable this behaviour.

What I've done:
 - I've added a new variable `MOCK_POST_DATA` to the environment
 - I'm passing the `Environment` to the constructor of the `RequestBody`
 - The `RequestBody` constructor checks if `MOCK_POST_DATA` is defined, and uses its value if true
 - I've added a unit test to confirm it works as expected
 - I tried it successfully in an actual project.

Introduced overhead: 
 - The `RequestBody` constructor executes additional logic with each request, even if no testing is done

If there is a better way to archive this functionality I'm very open for suggestions, and will improve the code if necessary.

I'm choosing the 3.x branch for merging as there is currently no development branch, and this change will not break compatibility.

### DIFF
--- a/Slim/Http/Environment.php
+++ b/Slim/Http/Environment.php
@@ -45,6 +45,7 @@ class Environment extends Collection implements EnvironmentInterface
             'REMOTE_ADDR'          => '127.0.0.1',
             'REQUEST_TIME'         => time(),
             'REQUEST_TIME_FLOAT'   => microtime(true),
+            'MOCK_POST_DATA'       => null
         ], $userData);
 
         return new static($data);
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -146,7 +146,8 @@ class Request extends Message implements ServerRequestInterface
         $request = new static($method, $uri, $headers, $cookies, $serverParams, $body, $uploadedFiles);
 
         if ($method === 'POST' &&
-            in_array($request->getMediaType(), ['application/x-www-form-urlencoded', 'multipart/form-data'])
+            in_array($request->getMediaType(), ['application/x-www-form-urlencoded', 'multipart/form-data']) &&
+            !$environment["MOCK_POST_DATA"]
         ) {
             // parsed body must be $_POST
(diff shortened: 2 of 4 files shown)


famoser:
I had to add another exception for `MOCK_POST_DATA` in `Request`. I am not quite satisfied with the result now, it seems to me there should be a more elegant way to do this.
I'd be very helpful if someone could give some feedback; if this feature is even supposed to be inside Slim, and if there's a better to do what I tried to accomplish.

famoser:
There is already a way to archive this; some nice guy over at http://discourse.slimframework.com/t/mock-slim-endpoint-post-requests-with-phpunit/973/7 was able to help me. The code to archive what I needed to:

``` php
//setup environment vals to create request
$env = Environment::mock();
$uri = Uri::createFromString('/1.0/testnode/');
$headers = Headers::createFromEnvironment($env);
$cookies = [];
$serverParams = $env->all();
$body = new RequestBody();
$uploadedFiles = UploadedFile::createFromEnvironment($env);
$request = new Request('POST', $uri, $headers, $cookies, $serverParams, $body, $uploadedFiles);

//write request data
$request->write(json_encode([ 'key' => 'val' ]));
$request->getBody()->rewind();
//set method & content type
$request = $request->withHeader('Content-Type', 'application/json');
$request = $request->withMethod('POST');

//execute request
$app = new App();
$resOut = $app($request, new Response());
$resOut->getBody()->rewind();

$this->assertEquals('full response text', $resOut->getBody()->getContents());
```

PR #2087, opened 2016-12-04
Remove application base url concept. Fixes #2054.

### LINKED ISSUE #2054: Remove base path concept
Just remove it entirely. Let web server configuration handle edge case base path scenarios.

codeguy:
This PR removes the concept of an app-wide "base path". I did leave the Router's base path stuff intact so that one could manually set a base path on the Router object... seemed like the logical thing to do,  but I do encourage feedback and careful review on this PR before merge.

Thanks!

/cc @akrabat @silentworks

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -310,9 +310,6 @@ class App
     {
         // Ensure basePath is set
         $router = $this->container->get('router');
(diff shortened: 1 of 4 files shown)


codeguy:
@akrabat I also left the `Uri::getBaseUrl()` method since that was a separate, unrelated helper method. I assume you'll deal with that when you tackle the PSR-7 class extraction stuff.

PR #2088, opened 2016-12-06
Fixed return type in phpdoc

daviddallet:
Hi,

I just noticed it in my IDE, the phpdoc return types seem wrong in the class Request for the methods getParsedBodyParam and getQueryParam (it says it returns "null" type).

I fixed it by using "mixed" instead.

Regards,

David Dallet


akrabat:
Thanks.

PR #2094, opened 2016-12-07
Fix issue #2091 - PSR-7 contract violation

### LINKED ISSUE #2091: [Http] PSR-7 contract violation
### spec
```php
// MessageInterface.php

/**
 * Retrieves all message header values.
 *
 * The keys represent the header name as it will be sent over the wire, and
 * each value is an array of strings associated with the header.
 *
 *     // Represent the headers as a string
 *     foreach ($message->getHeaders() as $name => $values) {
 *         echo $name . ": " . implode(", ", $values);
 *     }
 *
 *     // Emit headers iteratively:
 *     foreach ($message->getHeaders() as $name => $values) {
 *         foreach ($values as $value) {
 *             header(sprintf('%s: %s', $name, $value), false);
 *         }
 *     }
 *
 * While header names are not case-sensitive, getHeaders() will preserve the
 * exact case in which headers were originally specified.
 *
 * @return string[][] Returns an associative array of the message's headers. Each
 *     key MUST be a header name, and each value MUST be an array of strings
 *     for that header.
 */
public function getHeaders();
```

### problem
Slim [initializes its environment](https://github.com/slimphp/Slim/blob/50029c8a625bd3aaf9d6889da4570f9423b8e61a/Slim/DefaultServicesProvider.php#L47) by feeding it the raw $_SERVER superglobal. Later in the request lifecycle, this environment is used to [create the headers array for the request object](https://github.com/slimphp/Slim/blob/50029c8a625bd3aaf9d6889da4570f9423b8e61a/Slim/Http/Request.php#L140). Unfortunately, when running under a FastCGI execution environment the headers sent by the client are in the format of 'HTTP_FOO_BAR' instead of the original 'Foo-Bar', which violates the PSR-7 contract.

debug headers

### ferfuxsake why do you care?
I am the author of an [HMAC HTTP authentication library](https://github.com/1ma/Psr7Hmac) that aims to work out of the box with any PSR-7 implementation, so in my use case the whole spec is a very hard dependency. I found out this problem while developing and testing a Slim middleware for Psr7Hmac.

1ma:
Fixes #2091

As far as I know in PHP it is impossible to get a hold of the HTTP header names exactly as they were sent by the actual client (unless you use the ```getallheaders()``` function, which is only available in an Apache-based setup). So I had to settle with a plausible reconstruction of the name.

Cheers

### DIFF
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -56,7 +56,7 @@ class Headers extends Collection implements HeadersInterface
             $key = strtoupper($key);
             if (isset(static::$special[$key]) || strpos($key, 'HTTP_') === 0) {
                 if ($key !== 'HTTP_CONTENT_LENGTH') {
-                    $data[$key] =  $value;
+                    $data[static::reconstructOriginalKey($key)] =  $value;
                 }
             }
         }
@@ -218,5 +218,27 @@ class Headers extends Collection implements HeadersInterface
         }
 
         return $key;
+    }
+
+    /**
+     * Reconstruct original header name
+     *
+     * This method takes an HTTP header name from the Environment
+     * and returns it as it was probably formatted by the actual client.
+     *
+     * @param string $key An HTTP header key from the $_SERVER global variable
+     *
+     * @return string The reconstructed key
+     *
+     * @example CONTENT_TYPE => Content-Type
+     * @example HTTP_USER_AGENT => User-Agent
+     */
+    private static function reconstructOriginalKey($key)
+    {
+        if (strpos($key, 'HTTP_') === 0) {
+            $key = substr($key, 5);
+        }
+
+        return strtr(ucwords(strtr(strtolower($key), '_', ' ')), ' ', '-');
     }
 }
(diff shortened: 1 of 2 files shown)


Slim/Http/Headers.php
```diff
@@ -56,7 +56,7 @@ public static function createFromEnvironment(Environment $environment)
             $key = strtoupper($key);
             if (isset(static::$special[$key]) || strpos($key, 'HTTP_') === 0) {
                 if ($key !== 'HTTP_CONTENT_LENGTH') {
-                    $data[$key] =  $value;
+                    $data[self::reconstructOriginalKey($key)] = $value;
```
akrabat, line 59: Could we include both the current `$key` and the reconstructed one?

1ma: I don't think so, because if we put two different copies of the same header in `$data` both will be returned by the `getHeaders()` method, which would also violate PSR-7.

However I can close this PR and submit it again against the 4.x branch, which has no stable release yet and can handle breaking changes wrt 3.x. I think that might be the best course of action.

akrabat: Yeah - it's going to have to be that way. We'll do it in Slim-Psr7

PR #2095, opened 2016-12-07
Let CallableResolver work without container. Fixes #2049.

### LINKED ISSUE #2049: Let router resolve `MyController::method` callable with and without container
Because the DIC is optional, should we let router continue to resolve `MyController::method` route callables even if no DIC container is specified? Or should the DIC be required to use this callable format?

/cc @akrabat

codeguy:
With updated unit tests.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -24,9 +24,9 @@ final class CallableResolver implements CallableResolverInterface
     private $container;
 
     /**
-     * @param ContainerInterface $container
+     * @param ContainerInterface|null $container
      */
-    public function __construct(ContainerInterface $container)
+    public function __construct(ContainerInterface $container = null)
     {
         $this->container = $container;
     }
@@ -55,7 +55,8 @@ final class CallableResolver implements CallableResolverInterface
                 $class = $matches[1];
                 $method = $matches[2];
 
-                if ($this->container->has($class)) {
+                if ($this->container instanceof ContainerInterface &&
+                    $this->container->has($class)) {
                     $resolved = [$this->container->get($class), $method];
                 } else {
                     if (!class_exists($class)) {
@@ -67,7 +68,8 @@ final class CallableResolver implements CallableResolverInterface
                 // check if string is something in the DIC that's callable or is a class name which
                 // has an __invoke() method
                 $class = $toResolve;
-                if ($this->container->has($class)) {
+                if ($this->container instanceof ContainerInterface &&
+                    $this->container->has($class)) {
                     $resolved = $this->container->get($class);
                 } else {
                     if (!class_exists($class)) {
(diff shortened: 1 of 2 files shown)


codeguy:
Also, what is the point of `CallableResolverAwareTrait` exactly? The trait appears to _require_ a container in order to use our own callable resolver. Should I append this PR by adding a `setCallableResolver()` method to `Slim\App` and not rely on the presence of a container _internal_ to the app like so?

```
$container = new \My\Container();
$app = new \Slim\App();
$app->setCallableResolver($container->get('myResolver'));
```

Thoughts?

codeguy:
And how will this work should the user provide a 3rd-party container? If the end-user invokes `$app->setContainer($otherContainer)`, is that container set as the new app resolver?

codeguy:
<del>If my previous comment is true, then we should only add `$app->setContainer($container)` rather than adding that _in addition to_ a new `$app->setCallableResolver($something)` method.</del>

EDIT: The more I think of it, these _should_ be separate things.

&lt;/end-stream-of-consciousness&gt;

codeguy:
<del>What if we do away entirely with the concept of a `CallableResolverInterface` and just use our own. Is anyone using a different resolver anyway? Do we really need to support more than an actual callable or a container-interop container to resolve things?</del>

EDIT: Apparently folks are. That's fine.

mnapoli:
> What if we do away entirely with the concept of a CallableResolverInterface and just use our own. Is anyone using a different resolver anyway? Do we really need to support more than an actual callable or a container-interop container to resolve things?

Woops yes :) In the PHP-DI bridge: https://github.com/PHP-DI/Slim-Bridge/blob/master/src/CallableResolver.php

The reason for that is that with PHP-DI there is a concept of "lazy callable" common to PHP-DI and all its framework integrations. Here is how you can define a controller:

```php
$app->get('/', [MyController::class, 'myAction']); 

$app->get('/', [MyController::class]); 
```

See https://gist.github.com/mnapoli/f8afef0f556010c144e7 which explains in more details. Of course you can use classic PHP callables on top of that.

codeguy:
@mnapoli Thanks! Good to know.

geggleto:
This looks fine.

In answer to your trait question:
The trait simply by-passes the requirement of proper Dependency Injection in order to allow our internal classes to resolve a callable signature.

In the future, we should be using proper DI to add resolution to our internal classes. Using the trait was probably a better choice at the time to preserve the already existing internal API.

However since we can now "Break All The Things", we should be injecting the CallableResolver class instead of using a trait.

codeguy:
Leaving this PR as is. I've opened a new issue (assigned to me) to refactor more CallableResolver stuff #2096

akrabat (approved):

PR #2098, opened 2016-12-10
Refactor callable resolver, router, route, and route group

codeguy:
1. Remove `CallableResolverTrait`
2. Inject resolver into `DeferredCallable` instances instead of container
3. Set default callable resolver when appropriate if custom resolver not provided
4. Refactor `Route`, `Router`, `RouteGroup` to use new callable resolver
5. Update older service provider to inject resolver into router during creation
6. Add router setter/getter methods to `App`
7. Add callable resolver setter/getter methods to `App`

Please review carefully before merging. This is a big change. I didn't touch some places in `App` that still rely on the presence of a container, such as where we need app settings, or response handlers (Found, NotFound, etc.) and so on. Those are on my list. I've also sprinkled `TODO` notes throughout.


Slim/App.php
```diff
@@ -118,6 +131,88 @@ public function __call($method, $args)
...
+     *
+     * @return CallableResolver|null
+     */
+    public function getCallableResolver()
+    {
+        if (! $this->callableResolver instanceof CallableResolverInterface) {
+            $resolver = null;
+
+            // Fetch resolver from container if container is set
+            if ($this->container instanceof ContainerInterface) {
```
akrabat: Should we be using the container at all here? i.e. if you want to change the callable resolver, then you call `setCallableResolver()`.

codeguy: I think my intention was  to retain temporary BC with our current container (to prefer what it provides). But I agree with you. I just pushed a new commit to remove it from this method.

Slim/App.php
```diff
@@ -118,6 +131,76 @@ public function __call($method, $args)
...
+     *
+     * @return RouterInterface
+     */
+    public function getRouter()
+    {
+        if (! $this->router instanceof RouterInterface) {
+            $router = null;
+
+            // Fetch router from container if container is set
+            if ($this->container instanceof ContainerInterface) {
```
akrabat: Same here, I think?

codeguy: Just pushed a new commit. Luckily, this brought several affected unit tests to my attention.

akrabat:
https://github.com/akrabat/slim-bookshelf breaks when I use this branch with it. I get:

> An exception has been thrown during the rendering of a template ("Named route does not exist for name: list-authors") in "layout.twig" at line 35.

codeguy:
@akrabat See:

https://github.com/akrabat/slim-bookshelf/blob/master/app/src/dependencies.php#L17

You'll need the app's router with `$app->getRouter()`,  else inject your container's router with `$app->setRouter($c['router'])`.

codeguy:
@akrabat Should we have a method, such as, `$app->loadFromContainer($container)` that populates major objects from a third-party container to help with upgrading from v3?

codeguy:
Example code:

```
public function loadFromContainer(ContainerInterface $c)
{
    if ($c->has('router')) {
        $this->setRouter($c->get('router'));
    }
    // repeat for other objects
}
```

codeguy:
We could even do `importFromContainer($container)` and `exportToContainer($container)` to expedite stuff both ways.

akrabat:
Aha! Yeah - that's an obvious issue...

akrabat:
I'm not sure we need helper methods yet. Maybe wait and see what happens in beta when people try upgrading?

Certainly we should document this in the upgrading documentation as a minimum. Maybe create an `UPGRADING.md` file in root so that we can note while we we work?

For now, I think that the skeleton's `index.php` needs:

```php
$app->getContainer()['router'] = $app->getRouter();
```

codeguy:
Just pushed an `UPGRADING.md` file to document changes.

UPGRADING.md
```diff
@@ -0,0 +1,5 @@
+# How to upgrade
+
+* [2098] - You must inject the container router into app with `$app->setRouter()`
```
akrabat: Technically, you need to add the App's router to the container for a straight upgrade. If you've created your own router factory in the container though, then you need to set it into the $app.

codeguy: Updated language

PR #2102, opened 2016-12-11
Refactor container out of router, route, route group

codeguy:
<del>Work in progress. Do not merge yet.</del> Ready for review.

PR #2104, opened 2016-12-11
Refactor settings from container

codeguy:
Work in progress. Do not merge yet.


codeguy:
@akrabat What's your opinion on injecting settings in the constructor? Right now we accept a container instance in the constructor... but that will be unnecessary in 4.x. The container will be optional, so I think we should not ask for it in the constructor, but instead with a setter method `setContainer($c)`.

We could update the constructor to accept settings though like:

```
$app = new \Slim\App([
    'foo' => 'bar
]);
```

codeguy:
Currently, there are two ways you may use the 3.x constructor:

```
$app = new App([
    'settings' => [
        // settings
    ],
    // other container items
]);
```

The array argument will be used to populate a default Slim container. Else, you can pass your own container-interop container into the constructor:

```
$app = new App($myContainer);
```

Both scenarios would need to change to the following signature, where the argument is optional and if present, will be merged with the default settings:

```
$app = new App([
    // settings
]);
```

codeguy:
@akrabat I had to merge my latest work from #2102 into this branch. So you can either merge #2102 and then merge this PR, or you can simply merge this PR.

Latest changes in this PR make the `App::__construct()` method accept only an array of app settings which are then merged into the default settings array. It adds a public interface to interact with app settings array. And it separates settings from the app container entirely. To do this, I added getter/setter methods for app handlers (e.g. not found,  not allowed, error, php error).

codeguy:
Ping @akrabat. Can you review and merge this soon? I just merged in other latest changes from the primary 4.x branch.

Slim/App.php
```diff
@@ -65,38 +64,75 @@ class App
...
     {
-        if (is_array($container)) {
-            $container = new Container($container);
-        }
-        if (!$container instanceof ContainerInterface) {
-            throw new InvalidArgumentException('Expected a ContainerInterface');
-        }
-        $this->container = $container;
+        $this->addSettings($settings);
+        $this->container = new Container();
```
geggleto, line 112: ?? I thought we were going to remove the container ;)

geggleto (approved):
👍

Slim/App.php
codeguy: One step at a time :)

PR #2110, opened 2016-12-20
Cleanup CallableResolver

ghost:
Hi There,

I was taking a look at the Callable Resolver and I've noticed how the indentation levels were nested and all coupled into one single method. I've also noticed how some checks and logic were 
intrinsically structured.
Moreover the callable pattern, even though is not used anywhere else I think it's better to have it as a constant, since it's never going to change.
I've run the unit tests. Hope it's opened for discussions, glad to have your positive or negative feedback, thanks!

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -3,7 +3,7 @@
  * Slim Framework (https://slimframework.com)
  *
  * @link      https://github.com/slimphp/Slim
- * @copyright Copyright (c) 2011-2016 Josh Lockhart
+ * @copyright Copyright (c) 2011-2017 Josh Lockhart
  * @license   https://github.com/slimphp/Slim/blob/3.x/LICENSE.md (MIT License)
  */
 namespace Slim;
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -18,6 +18,8 @@ use Slim\Interfaces\CallableResolverInterface;
  */
 final class CallableResolver implements CallableResolverInterface
 {
+    const CALLABLE_PATTERN = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
+
     /**
      * @var ContainerInterface
      */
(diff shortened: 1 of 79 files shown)


ghost:
gotta add full test coverage

akrabat:
You can simply add to this branch and as you push, it'll re-run the CI tests

ghost:
@akrabat yes thanks I know. Just wanted to keep the PR list clean, as it will take some time, before I can work on it. I will reopen it soon

ghost:
The un-covered line here it seems to be like a dead code, and moreover it's not appearing on my local phpunit 4.8.33 running with PHP7 7.0.14, where I've 100% code coverage for the Slim Callable Resolver. Any ideas ?

Slim/CallableResolver.php
```diff
@@ -46,45 +48,63 @@ public function __construct(ContainerInterface $container)
...
+     * @param string $class
+     * @param string $method
+     * @return callable
+     *
+     * @throws \RuntimeException if the callable does not exist
+     */
+    protected function resolveCallable($class, $method = '__invoke')
+    {
+        if ($this->container->has($class)) {
+            return [$this->container->get($class), $method];
```
akrabat, line 86: This changes the behaviour from the current code. 

At the moment, when there isn't a method specified, we return the class object and let PHP resolve to `__invoke`. This code explicitly returns an array with the second parameter as `"__invoke"`.

Are there any implications as a result of this change?

akrabat:
> The un-covered line here it seems to be like a dead code, and moreover it's not appearing on my local phpunit 4.8.33 running with PHP7 7.0.14, where I've 100% code coverage for the Slim Callable Resolver. Any ideas ?

I can't see why that closing `}` is shown as not covered.

Slim/CallableResolver.php
ghost: Hey,
I don’t think it’s going to affect the behaviour, since as we must return a valid callback, when a method is not provided, the class must be invokable.
Given that, we return a callable defined as array which has at index 0 the class and index 1 the method to be invoked, which we explicitly define to be `__invoke’` https://3v4l.org/Uvj29

I’m going to test it deeply, but I’d say wouldn’t be an issue, isn’t it? :)

akrabat: Double check performance too.

ghost: do you have any benchmark suite ?

akrabat: I've asked around and general view is that it won't make any difference.

akrabat:
Thanks.
 It may be worth looking at the equivalent in `4.x` and seeing if that can also be improved.

ghost:
@akrabat someone already made some refactoring https://github.com/slimphp/Slim/pull/2157

Issue #1794, opened 2016-03-03
Slash is appers in the begining of Uri/path on one server, and not appears in another


xorik:
Hi! I found some issue in the Uri parser.

I have different results on different servers `test2` and `/test2` for this code:

```
$app->get('/test2', function()
{
    var_dump($this->request->getUri()->getPath());
});
```

On the first server (dev) index.php is not in the HTTP root (in subdirectory), and on the second (prod) index.php is in the HTTP root.

I have slim-3.2.1 on both servers.

Dev server:

```
  'PATH_INFO' => string '' (length=0)
  'SCRIPT_FILENAME' => string '/var/www/hot/index.php' (length=22)
  'QUERY_STRING' => string 't=test2' (length=7)
  'REQUEST_METHOD' => string 'GET' (length=3)
  'CONTENT_TYPE' => string '' (length=0)
  'CONTENT_LENGTH' => string '' (length=0)
  'SCRIPT_NAME' => string '/hot/index.php' (length=14)
  'REQUEST_URI' => string '/hot/test2' (length=10)
  'DOCUMENT_URI' => string '/hot/index.php' (length=14)
  'DOCUMENT_ROOT' => string '/var/www' (length=8)
  'SERVER_PROTOCOL' => string 'HTTP/1.1' (length=8)
 .....
  'PHP_SELF' => string '/hot/index.php' (length=14)
```

Prod server:

```
  ["DOCUMENT_ROOT"]=>
  string(29) "/var/www/0/71917/varmtarbeid/"
  ["REQUEST_SCHEME"]=>
  string(4) "http"
  ["CONTEXT_PREFIX"]=>
  string(0) ""
  ["CONTEXT_DOCUMENT_ROOT"]=>
  string(29) "/var/www/0/71917/varmtarbeid/"
  ["SCRIPT_FILENAME"]=>
  string(38) "/var/www/0/71917/varmtarbeid/index.php"
  ["REDIRECT_URL"]=>
  string(6) "/test2"
  ["REQUEST_METHOD"]=>
  string(3) "GET"
  ["QUERY_STRING"]=>
  string(0) ""
  ["REQUEST_URI"]=>
  string(6) "/test2"
  ["SCRIPT_NAME"]=>
  string(10) "/index.php"
  ["PHP_SELF"]=>
  string(10) "/index.php"
```

akrabat:
This is related to `getBasePath()`

Does this code give the same output on both servers?

``` php
$uri = $this->request->getUri();
[...]

Issue #1724, opened 2016-01-07
Slim 3 getPath in subdirectory missing left most slash


geggleto:
Example:
https://gist.github.com/geggleto/6cc2d792305fefc2bf64

http://dev/slim3/develop-testbed/public/yes/no => yes/no
http://develop-testbed.slim3.dev/yes/no => /yes/no

getPath is getting ltrim'd somewhere.

akrabat:
This is intentional. If the leading `/` is missing, then the base path is set.

tuupola:
Same as #1554.

Even though intentional it still is confusing for me that behaviour is different depending on whether Slim is installed in subfolder or not. I guess there will be lots of bug reports and support requests about this. In the end I just ended up doing something like following:

``` php
$path = "/" . $request->getUri()->getPath();
$path = str_replace("//", "/", $path);
```

alexweissman:
> If the leading / is missing, then the base path is set.

Do you mean "base path is set" -> "leading `/` will be stripped"?

Also, what is the motivation behind this behavior?  Is this part of PSR-7?

PR #2115, opened 2017-01-01
Strip leading slash from getPath() output when index.php is in webroot

### LINKED ISSUE #1554: Leading slash missing when Slim 3 installed in subfolder
With Slim 3 both rc1 and 3.x-dev the code below returns `/foo/bar` when installed in document root. However when installed into subfolder it returns `foo/bar` without the leading slash. It does not matter whether `RewriteBase` is set or not.

``` php
require_once "./vendor/autoload.php";

$app = new \Slim\App;
$app->get("/foo/bar", function ($request, $response, $args) {
    return $response->write($request->getUri()->getPath() . "\n");
});

$app->run();
```

misha354:
Previously, the leading slash in getPath() output was significant: its presence indicated that the base path is not set.  This caused confusion, because the slash would appear when index.php is in webroot, but would not appear when it's in a subdirectory. The base path concept was removed in #2054, so now '/' no longer needs to be treated as a special case.

Fixes #1554, #1724, #1794

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -201,7 +201,7 @@ class Uri implements UriInterface
         $virtualPath = $requestUri;
         if (stripos($requestUri, $requestScriptName) === 0) {
             $basePath = $requestScriptName;
-        } elseif ($requestScriptDir !== '/' && stripos($requestUri, $requestScriptDir) === 0) {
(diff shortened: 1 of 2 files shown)


akrabat:
The HTTP component has been move to [Slim-Http](https://github.com/slimphp/Slim-Http) where this issue has been handled.

PR #2119, opened 2017-01-08
Update docblocks for route arguments

### LINKED ISSUE #1868: Inconsistent types for route argument setting vs getting
According to it's docblock, `RouteInterface::setArgument()`'s $value argument needs to a string, whereas `RouteInterface::getArgument()` returns `mixed`, and accepts `mixed` type for its $default parameter.

Will be happy to provide a PR to resolve the inconsistency, but need to know which solution you prefer: should route arguments always be strings, or should we accept `mixed` for the value? The latter makes more sense to me - in my own case, I was trying to set a boolean flag on the route as argument (worked perfectly, except that static analysis complains about violating the parameter type).

akrabat:
Ensure that the getters and setters for arguments in Route and
RouteInferface are consistent.

Fixes #1868

Please forward port the 4.x branch too.

### DIFF
--- a/Slim/Interfaces/RouteInterface.php
+++ b/Slim/Interfaces/RouteInterface.php
@@ -25,16 +25,16 @@ interface RouteInterface
      * Retrieve a specific route argument
      *
      * @param string $name
-     * @param mixed $default
+     * @param string|null $default
      *
-     * @return mixed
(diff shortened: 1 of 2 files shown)


codeguy (approved):
Looks good to me!

PR #2121, opened 2017-01-08
Support rendering Exceptions with previous Errors

### LINKED ISSUE #1943: Slim 3.5 PHP errors cause fatal errors in the error handler
In Slim 3.5 on php 7 I am encountering this php error when Slim attempts to handle php errors:

> `Fatal error: Uncaught TypeError: Argument 1 passed to Slim\Handlers\Error::renderHtmlException() must be an instance of Exception, instance of Error given, called in path/vendor/slim/slim/Slim/Handlers/Error.php on line 84 and defined in path/vendor/slim/slim/Slim/Handlers/Error.php:110 Stack trace: #0 path/vendor/slim/slim/Slim/Handlers/Error.php(84): Slim\Handlers\Error->renderHtmlException(Object(Error)) #1 path/vendor/slim/slim/Slim/Handlers/Error.php(48): Slim\Handlers\Error->renderHtmlErrorMessage(Object(Error)) #2 [internal function]: Slim\Handlers\Error->__invoke(Object(Slim\Http\Request), Object(Slim\Http\Response), Object(PDOException)) #3 path/vendor/slim/slim/Slim/App.php(613): call_user_func_array(Object(Slim\Handlers\Error), Array) #4 path/vendor/slim/slim/Slim/App.php(339): Slim\App->handleException(Object(PDOExc in path/vendor/slim/slim/Slim/Handlers/Error.php on line 110`

(adjusted to remove paths)

This appears to be this issue: https://akrabat.com/improved-error-handling-in-slim-3-2-0/ (Specfically, the PHP 7 heading)

As I say, I am on Slim 3.5, which I have confirmed through composer and updated just in case. Has this been reverted by accident? 

Here is the method as it appears in my current codebase:

```
protected function renderHtmlException(\Exception $exception)
    {
        $title = 'Slim Application Error';
//more exciting code
```

And the exact composer version: slim/slim 3.5.0 

If it is relevant, I think the error is originating in a PDO class fetch, but its a little difficult to tell.

akrabat:
If an Exception is thrown with an Error set as it's previous then we
need to render this correctly in the error handler. This simply requires
removing the type hint, however we can't do that for BC, so I've created
the method `renderHtmlExceptionOrError` which is now used.

Fixes #1943

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -81,7 +81,7 @@ class Error extends AbstractError
 
             while ($exception = $exception->getPrevious()) {
                 $html .= '<h2>Previous exception</h2>';
-                $html .= $this->renderHtmlException($exception);
+                $html .= $this->renderHtmlExceptionOrError($exception);
             }
         } else {
             $html = '<p>A website error has occurred. Sorry for the temporary inconvenience.</p>';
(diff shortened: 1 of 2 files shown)


Slim/Handlers/Error.php
```diff
@@ -103,12 +103,30 @@ protected function renderHtmlErrorMessage(\Exception $exception)
...
+    }
+
+    /**
+     * Render exception or error as HTML.
+     *
+     * @param \Exception|\Error $exception
+     *
+     * @return string
+     */
+    protected function renderHtmlExceptionOrError($exception)
```
chrisemerson, line 124: You could simplify this by hinting for Throwable instead - both Error and Exception implement this interface

akrabat: Throwable doesn't exist in PHP5, so it's better to be specific.

chrisemerson: Ah of course, I had forgotten about the support requirements!

codeguy (approved):
Looks good. Merging momentarily.

PR #2122, opened 2017-01-08
Handle invalid XML better

### LINKED ISSUE #2106: RuntimeException thrown when submitting invalid XML
Slim seems to throw a RuntimeException when submitting invalid XML data.

The xml media parser in Request.php, which uses simplexml_load_string to parse the request body, returns false when it's unable to validate the xml, and thereupon, an exception is thrown if the media parser doesn't return an array, an object or a null value.

Is this behaviour by design, or is it a bug? It seems a bit strange to me to stop the processing abruptly at this early point, before reaching a route. Wouldn't it be better to return a null value, and to let the routes - or middleware - decide what to do no valid data is passed to them?

May be related to issue #1807.

akrabat:
If the request's body contains invalid XML, don't throw an exception or
PHP warning. Instead, just return NULL from the medica parser.
This matches the way we handle JSON.

Fixes #2106

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -207,15 +207,27 @@ class Request extends Message implements ServerRequestInterface
 
         $this->registerMediaTypeParser('application/xml', function ($input) {
             $backup = libxml_disable_entity_loader(true);
+            $backup_errors = libxml_use_internal_errors(TRUE);
             $result = simplexml_load_string($input);
             libxml_disable_entity_loader($backup);
+            libxml_clear_errors();
+            libxml_use_internal_errors($backup_errors);
+            if ($result === false) {
+                return null;
+            }
             return $result;
         });
 
         $this->registerMediaTypeParser('text/xml', function ($input) {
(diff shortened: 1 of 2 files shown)


silentworks:
Is this not a BC break, since we are now suppressing a PHP error?

akrabat:
I think it's a bug fix. Will go with majority opinion though.

silentworks (approved):

codeguy:
@akrabat Forward port to 4.x too?

akrabat:
> @akrabat Forward port to 4.x too?

Yes.

codeguy:
Do we want a test for this?

akrabat:
> Do we want a test for this?

Almost certainly :)

codeguy:
Will merge once you get a test added.

akrabat:
Rebased and added unit tests.

akrabat:
Success, @codeguy  :)

akrabat:
Can this go into 4.x too, please?

Issue #2055, opened 2016-11-19
Remove `Slim\Exception\SlimException` class


codeguy:

geggleto:
What are we going to replace them with ? :) Are we going to just extend from `\Exception` or `\Throwable`

codeguy:
Updated title. This only affects the `Slim\Exception\SlimException` class.

mrailton:
Any objections to me working on this?

akrabat:
Fixed by #2124

PR #2124, opened 2017-01-15
Remove Slim\Exception\SlimException

mrailton:
Per issue #2055 this removes Slim\Exception\SlimException by updating MethodNotAllowedException and NotFoundException to extend the \Exception class and pulling in the request and response themselves.  Also removed redundant test.


mrailton:
Not sure how this has caused the test coverage to drop, tho appears to be such a small amount.

akrabat:
Thanks.

PR #2126, opened 2017-01-17
Ensure Request::withUri preserves host correctly

akrabat:
The rules for dealing with the Host header in `withUri()` when `$preserveHost` is `true` state this:

> If the the Host header is missing or empty, and the new URI contains
> a host component, this method MUST update the Host header in the returned
> request.

This now works - it didn't before as we didn't have a test for it…

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -567,7 +567,7 @@ class Request extends Message implements ServerRequestInterface
                 $clone->headers->set('Host', $uri->getHost());
             }
         } else {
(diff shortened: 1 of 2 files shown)


akrabat:
Discovered that my unit test for `$preserveHost` was wrong as it didn't allow for the case of the Request's Host header being different from the Uri's host property. 

Fixed test and hence fixed `withRequest()`.

codeguy:
@akrabat Is this just for 3.x?

akrabat:
> @akrabat Is this just for 3.x?

Better forward port to 4.x too as I need to resync for slim\http

codeguy (approved):
Looks good. Merging into 3.x and 4.x

PR #2132, opened 2017-01-23
Added pragmatic example in code example into README.md

ghost:
Added pragmatic example in code example into README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,7 @@ require 'vendor/autoload.php';
 $app = new Slim\App();
 
 $app->get('/hello/{name}', function ($request, $response, $args) {
-    $response->write("Hello, " . $args['name']);
-    return $response;
+    return $response->write("Hello, " . $args['name']);
 });
 
 $app->run();


akrabat:
Can you explain the reasoning behind this change?

ghost:
Hi,
I was asking myself, why should we use two statements for doing something that can be done in one? At least in the example.
It’s going to be a lean example to show, on how simple and straightforward is, to use (setting up the Hello World) Slim, isn't it? Compared to other PHP micro frameworks, it looks like we have to do extra work to return a response. Why using more code than what is actually needed?
I understand that in the way you used in principle, it's clear for beginners, that the closure should always returns a valid Response, but that’s also clear from the well done documentation Slim has. I hope it’s not a stupid argument. WDYT? :)

ghost:
@akrabat makes sense? :)

akrabat (approved):

akrabat:
Yes. I understand your logic.

PR #2134, opened 2017-01-24
As per PSR-7, type hint @return static

hannesvdvreken:
### Changed

- Some type hints of `with*` methods should return type static.

---

This helps with situations like this where you're extending the Request class to add typed attribute getters.

```php
class AuthenticatedRequest extends Request
{
     /**
      * @return static
      */
     public function withUser(User $user)
     {
         return $this->withAttribute('user', $user);
     }

     /**
      * @return User
      */
     public function getUser()
     {
         return $this->getAttribute('user');
     }
}
```

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -131,7 +131,7 @@ class Request extends Message implements ServerRequestInterface
      *
      * @param  Environment $environment The Slim application Environment
      *
-     * @return self
+     * @return static
      */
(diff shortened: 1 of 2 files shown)


jdrieghe:
Would love to see this one released. It would improve type hinting in the project I'm working on.

hannesvdvreken:
Thanks, Rob!

Issue #2133, opened 2017-01-24
[QUESTION] code quality


ghost:
Looking at `Slim\MiddlewareAwareTrait` I see `protected $middlewareLock = false;`

Why we need to initialise properties even though it’s not necessary?
Also in the same class we have 
`callMiddlewareStack(ServerRequestInterface $req, ResponseInterface $res)`
But in App class we have
` __invoke(ServerRequestInterface $request, ResponseInterface $response)`
Why don’t we have consistency?

Or looking at handleException method in App class,
```php
// Call the registered handler 
return call_user_func_array($callable, $params);
```

Do we really need all these comments ?

Slim it’s an awesome project, and I’d like to contribute to it, but seeing all these little things let me think there is not space for quality code.
What are your thoughts ? Hope to have a insightful discussion :)

ghost:
@0x13a
You want quality? you can't find it in famous PHP projects as I saw!

ghost:
@sharifzadesina why not?

ghost:
@0x13a
Because I am working on one and I am seeing what the hard is writing a standard framework, see Laravel there are many design problems in it which bring us many limitation, for example you can't add custom dependent rules to validation. for writing my own framework for my own job I have read many source code none of them are standard so don't waste your time ;)

akrabat:
My general opinion is that inconsistencies are inevitable when you have lots of volunteer contributors. I certainly won't reject a PR because someone has named their variable `$res` rather than `$response`.

However, I'll take a PR to tidy up and please feel free to do so!

As far as comments are concerned, I encourage them as, again, lots of different people look at this code and if a comment helps someone then it's worth it.

If you don't think that the code is good enough quality, then create small PRs that you think improve it and we can discuss specifics.

ghost:
Thanks for the feedback @akrabat, I wanted know if I could help. I will submit small PRs from time to time.

akrabat:
General rule: the smaller, the better :)

PR #2136, opened 2017-01-26
updated consistent name request and response in MiddlewareAwareTrait

ghost:
Hi all,
as discussed briefly here https://github.com/slimphp/Slim/issues/2133. I'd like to make more consistent use of the variables names, also and specifically `$request` and `$response`, instead of using sometimes those and sometimes `$req` and `$res`.

### DIFF
--- a/Slim/MiddlewareAwareTrait.php
+++ b/Slim/MiddlewareAwareTrait.php
@@ -66,7 +66,10 @@ trait MiddlewareAwareTrait
         $this->stack[] = function (
             ServerRequestInterface $request,
             ResponseInterface $response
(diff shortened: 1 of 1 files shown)


ghost:
yes, i need to setup the psr2 sniffer.. Anyway ready to merge @akrabat

tuupola:
To check for PSR-2 run either `$ composer phpcs` or `$ vendor/bin/phpcs` in Slim folder.

https://github.com/slimphp/Slim/blob/3.x/composer.json#L55

ghost:
thanks @tuupola

PR #2141, opened 2017-02-04
Call not-found/bad-method handlers on invalid HTTP method

### LINKED ISSUE #2005: HTTP method outside whitelist causes 500
# (picking up from the Slack discussion earlier)

If an incoming HTTP request has a verb other than something in Request::$validMethods (e.g. LINK/UNLINK...or CHECKIN...or CHICKEN), Request::filterMethod() throws an InvalidArgumentException on request create from env (which happens on App::run()). This isn't caught anywhere, so it bubbles up to a 500. This happens both with and without a route declared for that URL, that method (yep, the router will let you match on CHICKEN), or neither/both.

By contrast, a verb in $validMethods, but unassigned for a route, will call the notAllowedHandler, which seems like the more proper way of handling the above. Likewise, if you don't have a route for that verb, probably better to call notFoundHandler.

There are a couple catches here:
1. notAllowedHandler assumes you've matched with a route before calling. Which needs a request. The only way I can think of to handle this properly is to set up an exception that builds the request anyway, then throws something that extends \InvalidArgumentException and includes the request along with it. That gets thrown into the router, then we decide how to treat the error (we'll always be erroring out) based on what the router gives us back. A little convoluted, but should give us the behavior we're looking for.
2. The router doesn't enforce the allowed methods array at all, so if we take the above fix then we'll end up with an even more confusing error message (LINK not allowed, please use one of [GET, PUT, POST, LINK]) if someone matches to an unsupported verb. So we'd want to tweak the router to throw an InvalidArgumentException (which _should_ throw us into a 500) when someone throws a verb in there that doesn't make sense.

Does the above concern make sense to someone other than me, and do the miticgations I'm suggesting also make sense? Maybe fully supporting this corner case at the cost of N operations per verb per route match() call doesn't make sense perf-wise, but I'd argue that having a really easy way to make a Slim app 500 is odd too.

If everything makes sense, thumbs-up this and I'll work on a PR.

iansltx:
Partially resolves #2005. Router map() calls don't have method checking,
and we still don't link the valid methods in the request to valid
methods at the application level (which IMO should be since the router
doesn't enforce any sort of method limitation at this point), but now
you won't get an uncaught exception when someone hits you with an HTTP
message with verb CHICKEN.

Also revised tests to check app run handling rather than request create
handling, since internally to the request a bad method should still be
invalid/throw an exception, so we don't want to weaken that guarantee.
But we do want the application not to 500 at runtime.

Along with that, behavior (as tested) differs between whether we have a
route or not, since in any other case the former would give you a bad
method exception, while the latter would give you a not-found. So these
tweaks match up that behavior for obscure HTTP methods as well.

Finally, RouterInterface now has constants for route info array returns.
They're only used on the exceptional path, but they make it a bit
clearer why we're pulling various numeric keys out of the route, rather
than having a non-obvious contract with FastRoute.

Took extra care here to respect order of execution for the existing code
so anything relying on that order will still work. Likewise,
InvalidMethodException subclasses InvalidArgumentException and returns
the same message and code, so anything checking for exception strings
etc., as well as instanceof checks, will still work.

Credit to @nickdnk for building the first red test for this issue.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -289,15 +289,13 @@ class App
      */
     public function run($silent = false)
     {
-        try {
-            $request = $this->container->get('request');
-        } catch (InvalidMethodException $e) {
-            $request = $e->getRequest();
-        }
-
         $response = $this->container->get('response');
 
-        $response = !isset($e) ? $this->process($request, $response) : $this->processInvalidMethod($request, $response);
+        try {
+            $response = $this->process($this->container->get('request'), $response);
+        } catch (InvalidMethodException $e) {
+            $response = $this->processInvalidMethod($e->getRequest(), $response);
+        }
 
         if (!$silent) {
             $this->respond($response);
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -9,6 +9,8 @@
 namespace Slim;
 
 use Exception;
+use Slim\Exception\InvalidMethodException;
+use Slim\Http\Response;
 use Throwable;
 use Closure;
 use InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


iansltx:
I'll work on improving coverage back to where it should be. No need to merge until I do that.

iansltx:
...and by "improving coverage" I actually mean "fixing a completely broken first implementation", as it turns out. Now we're good to go though :)

Slim/App.php
```diff
@@ -287,10 +289,15 @@ public function group($pattern, $callable)
...
+        try {
+            $request = $this->container->get('request');
+        } catch (InvalidMethodException $e) {
+            $request = $e->getRequest();
+        }
+
         $response = $this->container->get('response');
 
-        $response = $this->process($request, $response);
+        $response = !isset($e) ? $this->process($request, $response) : $this->processInvalidMethod($request, $response);
```
akrabat: Please refactor this ternary into the try/catch

iansltx: I assume that means duplicating `$response = $this->container->get('response');` to maintain existing application flow?

akrabat: I think that it can go before the `try`?

iansltx: Existing flow gets the request before the response so I wanted to maintain that order of execution. But sounds like we're cool with swapping order since the two should never be related anyway. Will push a rev once this talk's over.

akrabat: I don't think it should matter. If it does, I'm not sure what I'll think!

iansltx: Cleaned up! Keep up with the feedback :)

akrabat:
Thanks!

PR #2143, opened 2017-02-06
Fixed HTTP EOL

katjuncker:
The HTTP standard specifies that CR LF ("\r\n") should be used for separating elements in an HTTP request/response.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2

Using PHP_EOL could potentially break some applications as PHP_EOL is not guaranteed to be "\r\n".

I added a class constant of EOL to the Response class and used that constant instead of PHP_EOL when converting the Response to a string and in the unit tests.

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -455,17 +455,18 @@ class Response extends Message implements ResponseInterface
      */
     public function __toString()
     {
+        $eol = "\r\n";
         $output = sprintf(
             'HTTP/%s %s %s',
             $this->getProtocolVersion(),
             $this->getStatusCode(),
             $this->getReasonPhrase()
         );
-        $output .= PHP_EOL;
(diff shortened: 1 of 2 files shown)


Slim/Http/Response.php
```diff
@@ -118,6 +118,13 @@ class Response extends Message implements ResponseInterface
     ];
 
     /**
+     * EOL characters used for HTTP response.
+     *
+     * @var string
+     */
```
piotr-cz, line 124: Maybe you can add a line for future reference here
```
@see https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2
```

vlakoff:
`self::EOL` would be a better pick. Maybe even `static::EOL` (one shouldn't have to override it, but who knows).

PR #2146, opened 2017-02-09
Allowing use of PSR-11 containers

Sam-Burns:
Allows users to swap in a PSR-11 container. This means that containers (e.g. Symfony, Ultra-Lite) that support PSR-11 only, can be used without an adapter.

Note that since v1.2, all Container-Interop containers are automatically PSR-11 compliant, but that the reverse is not true.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -17,7 +17,7 @@ use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 use FastRoute\Dispatcher;
 use Slim\Exception\SlimException;
 use Slim\Exception\MethodNotAllowedException;
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -9,7 +9,7 @@
 namespace Slim;
 
 use RuntimeException;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 use Slim\Interfaces\CallableResolverInterface;
 
 /**
--- a/Slim/CallableResolverAwareTrait.php
+++ b/Slim/CallableResolverAwareTrait.php
@@ -9,7 +9,7 @@
 namespace Slim;
 
 use RuntimeException;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 use Slim\Interfaces\CallableResolverInterface;
 
 /**
--- a/Slim/DeferredCallable.php
+++ b/Slim/DeferredCallable.php
@@ -10,7 +10,7 @@
 namespace Slim;
 
 use Closure;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 
 class DeferredCallable
 {
--- a/Slim/Routable.php
+++ b/Slim/Routable.php
@@ -8,7 +8,7 @@
  */
 namespace Slim;
 
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 
 /**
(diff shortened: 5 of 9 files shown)


Sam-Burns:
- Converts the application's container to a PSR-11 container.
- Allows users to swap in another PSR-11 container.
- Requires "container-interop/container-interop": "^2.0", so it is caught by Composer if a user is trying to swap in a container which only implements container-interop/container-interop version 1.1 or previous.
- Exceptions implement `Interop\Container\Exception\ContainerException` and `Psr\Container\ContainerExceptionInterface`, so user-land code can still catch them as Container-Interop exceptions.

piotr-cz:
BTW, _container-interop_ is now officially deprecated in favor of PSR-11 _php-fig/container_

Sam-Burns:
Is there any news on this?  Does anyone have any feedback or questions?

geggleto:
@Sam-Burns Switching the container is not so simple. 

We also need to switch the Exceptions the container throws [`A call to the get method with a non-existing id MUST throw a Psr\Container\NotFoundExceptionInterface.`] which will cause a BC break. 

Due to the BC break, we cannot merge this into 3.x and this PR should be rebased to target 4.x.

Sam-Burns:
@geggleto Referring to the [Container-Interop standard](https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/Exception/ContainerException.php#L13), `Interop\Container\Exception\ContainerException` extends `Psr\Container\ContainerExceptionInterface`.

I think the Slim container is already PSR-11 compliant.  This PR is intended to allow instances of the parent interface to be used instead.  Put simply, it means that containers like the Symfony Service Container can be used with Slim without an adapter, and this is achieved without breaking backward compatibility.

If I've misunderstood you, please do point me in the right direction.  Thanks for taking the time to review my PR!

geggleto:
I need some more time to digest this. I think we can do it after seeing what interop did, I just want to be 100% sure that we are not going to be breaking people's production apps by changing the type-hint.

My biggest concern is with the types of exceptions. The new interface has new types of exceptions. By changing the type-hint to the parent class of the PSR-Container ... one would expect those exceptions be thrown. I think for 99% of people this will work, it's those 1% whom are not using Pimple or standard containers.

Sam-Burns:
The pull request simply allows additional containers, such as the Symfony container, to be used in Slim without writing a custom adapter.  This is done by changing the typehint of the incoming container to the parent interface.  I don't think there is any need for concern that this will change the type of exceptions thrown by a container already in use.

akrabat (approved):

akrabat:
Thanks @Sam-Burns.

Can you don one for 4.x too?

Sam-Burns:
@akrabat will do :+1:

Sam-Burns:
@akrabat ref https://github.com/slimphp/Slim/pull/2174 .

harikt:
@akrabat This actually is a BC break. I wonder how does this actually happened to be in 3.x .

akrabat:
@harikt Can you explain?

If you have code that checks for `instanceof  Interop\Container\ContainerInterface`, then that code will continue to work. Similarly, you can continue to pass a container that implements `Interop\Container\ContainerInterface` where you could before.

harikt:
@akrabat sorry for the confusion. I have looked more carefully what are the changes made.

It looks you guys are right here as if anyone passing an instance of `Interop\Container\ContainerInterface` will work.

Sorry once again. First look made me wonder about it.

akrabat:
@harikt No problem at all :) Always happy to have additional people look at what we do!

harikt:
though you cannot swap the container easily for there is something like this https://github.com/shameerc/Slim3-Aura.Di-Bridge/blob/24d6fc36aab84a80973deb84cfe938266ab181c9/src/Container.php#L75-L119

akrabat:
Agreed. Hopefully, 4.x will be easier in this regard.

PR #2150, opened 2017-02-17
Fix typo of getParams() phpdoc

dopesong:


akrabat:
I bet that was annoying you! Thanks :)

PR #2152, opened 2017-02-19
conflicts with slim/http

harikt:
Hi, 

I have been working on a test project with `http-interop/http-factory`.

The aim of test project was to showcase using the test-project with any psr-7 framework. But due to slim 3.x is not using the same Request / Response classes this makes composer hard to understand the correct class what I need. So instead of requiring slim/http it will take slim/slim and viseversa. 

I would love if conflict is actually added in the 3.x

Have also initiated one in : https://github.com/http-interop/http-factory-slim/pull/5

Thank you for the consideration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,9 @@
         "squizlabs/php_codesniffer": "^2.5",
         "phpunit/phpunit": "^4.0"
     },
+    "conflict": {
+        "slim/http": "0.*"
+    },
     "provide": {
         "psr/http-message-implementation": "1.0"
     },


geggleto:
Hey @harikt, I believe you have cleared this up on your end. Slim does not offer at this point in time a stable standalone PSR-7 Implementation. Slim/Http is not considered stable at all. 

The conflict check does not make a lot of sense as the Slim/Http branch will be targeted at 4.0 and above.

harikt:
Hey @geggleto ,

Thank you for your response.

akrabat:
We'll solve this for 4.

PR #2153, opened 2017-02-20
Bugfix when handling requests with empty body

dhm80:
If you call a page with e. g. Method ```GET``` and ContentType ```text/xml``` the method ```getParsedBody()``` throws an Exception as ```$this->body```is always set (as it contains a StreamInterface) if the body is empty.

The correct procedure would be to check if the body has content.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1006,7 +1006,7 @@ class Request extends Message implements ServerRequestInterface
             return $this->bodyParsed;
         }
 
-        if (!$this->body) {
+        if (!$this->getBody()) {
             return null;
         }
 


akrabat:
Please create a unit test and supply a curl command that shows the problem.

Thanks.

akrabat:
I can't reproduce with this test:

```
    public function testGetParsedBodyXmlWithNoBody()
    {
        $method = 'GET';
        $uri = new Uri('https', 'example.com', 443);
        $headers = new Headers();
        $headers->set('Content-Type', 'application/xml');
        $cookies = [];
        $serverParams = [];
        $body = new RequestBody();
        $body->write('');
        $request = new Request($method, $uri, $headers, $cookies, $serverParams, $body);

        $this->assertEquals('', $request->getParsedBody());
    }
```


Please write a test that shows the problem and add to this branch and then reopen the PR.

PR #2157, opened 2017-02-26
Simplify callable resolver

StJudeWasHere:
Simplify callable resolver to avoid duplicated code.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -49,34 +49,24 @@ final class CallableResolver implements CallableResolverInterface
         $resolved = $toResolve;
 
         if (!is_callable($toResolve) && is_string($toResolve)) {
+            $class = $toResolve;
+            $method = '__invoke';
+
             // check for slim callable as "class:method"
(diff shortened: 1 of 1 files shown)


Slim/CallableResolver.php
```diff
@@ -49,34 +49,24 @@ public function resolve($toResolve)
         $resolved = $toResolve;
 
         if (!is_callable($toResolve) && is_string($toResolve)) {
+            $class = $toResolve;
+            $method = '__invoke';
```
geggleto, line 53: it might be nice to have this as a configurations setting.

StJudeWasHere: I can't think of any case where having this as a setting could be useful. Anyway, this PR isn't intended to add a new feature but to avoid code duplication.

akrabat: I tend to agree with @mnlg.

akrabat:
Out of interest, is there a reason why this isn't relevant to the `3.x` branch?

akrabat (approved):

StJudeWasHere:
@akrabat I wasn't sure what branch to target, the CONTRIBUTING.md file doesn't seem to be up to date.

PR #2161, opened 2017-02-28
Update Request.php to add LINK/UNLINK HTTP methods

amreladawy:
LINK and UNLINK need to be added to allowed methods.


johnhunt (changes requested):
Please put LINK in alphabetical order in the array to maintain consistency.

akrabat:
Sorry, but a far better solution is to remove the list of validMethods altogether and change `filterMethod()` to check for a regex of the form: 

    /^[[:alnum:]!#$%&\'*\/+\-.^_`|~]+$/i

and only throw the `InvalidMethodException` if the `$method` doesn't match that regex.

PR #2162, opened 2017-02-28
Update docblocks to include missing @throws

StJudeWasHere:
Update docblocks to include missing `@throws`

This PR also adds a missing `@var` dockblock and changes one `@var boolean` to `@var bool` for consistency.


akrabat:
Thanks.

PR #2166, opened 2017-03-08
Output buffer middleware

codeguy:
Work in progress. Opening PR for feedback.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -90,7 +90,6 @@ class App
     protected $settings = [
         'httpVersion' => '1.1',
         'responseChunkSize' => 4096,
-        'outputBuffering' => 'append',
         'determineRouteBeforeAppMiddleware' => false,
         'displayErrorDetails' => false,
         'addContentLengthHeader' => true,
@@ -540,11 +539,6 @@ class App
 
         // Create route
         $route = $this->getRouter()->map($methods, $pattern, $callable);
-
-        // TODO: Do we keep output buffering?
-        if (is_callable([$route, 'setOutputBuffering'])) {
-            $route->setOutputBuffering($this->getSetting('outputBuffering', 'append'));
-        }
 
         return $route;
     }
--- a/Slim/Middleware/OutputBuffering.php
+++ b/Slim/Middleware/OutputBuffering.php
@@ -1,0 +1,63 @@
+<?php
+namespace Slim\Middleware;
+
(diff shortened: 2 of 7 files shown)


codeguy:
Changes so far include:

1. Require all route callbacks to return PSR7 response.
2. Remove all output buffering concerns from `App` and `Route` classes.

akrabat (changes requested):
* ~Need tests for the new `OutputBuffering` middleware~
* Can you add a CHANGELOG entry please?

Update - I can't read!

tests/RouteTest.php
```diff
@@ -296,36 +276,11 @@ public function testInvokeWhenEchoingOutput()
     public function testInvokeWhenReturningAString()
```
akrabat, line 276: I think this test should return a string and expect an Exception to happen?

akrabat (changes requested):

Slim/App.php
```diff
@@ -543,11 +542,6 @@ public function map(array $methods, $pattern, $callable)
         // Create route
         $route = $this->getRouter()->map($methods, $pattern, $callable);
 
-        // TODO: Do we keep output buffering?
-        if (is_callable([$route, 'setOutputBuffering'])) {
-            $route->setOutputBuffering($this->getSetting('outputBuffering', 'append'));
-        }
```
geggleto, line 549: I honestly think that we can remove it.

Dump and Die will still work, which is I feel what a majority of people do anyway.

tests/RouteTest.php
geggleto: @akrabat I think the name of the test should change. This test should be renamed.

`testEchoingOutputIsEmpty`

`testReturningStringThrowsException` ?

geggleto (changes requested):

Slim/Middleware/OutputBuffering.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param  callable                $next Middleware callable
+     * @return \ResponseInterface            PSR7 response
+     */
+    public function __invoke(ServerRequestInterface $req, ResponseInterface $res, callable $next)
+    {
+        try {
+            ob_start();
+            $newResponse = $next($req, $res);
+            $output = ob_get_clean();
+        } catch (\Throwable $e) {
```
danopz, line 46: PHP 5 vs PHP 7 #2175

Issue #1617, opened 2015-11-25
PHPDocs


designermonkey:
Can we get the PHP docs put up somewhere? I can never remember the methods and properties available :(

p3ym4n:
👍

tw2113:
Recommendation after some talk in IRC.

http://www.apigen.org/

As easy as this command from a terminal in the root directory of a checkout:

```
apigen generate --source Slim --destination ../docs --template-theme bootstrap --title "Slim Framework"
```

Change the destination to wherever desired and commit afterwards.

tw2113:
Example output:

![apigen](https://cloud.githubusercontent.com/assets/484075/11453976/3bc1ee44-95e4-11e5-8d5c-30420b07e9cd.png)

geggleto:
@tw2113 oh I likey...

akrabat:
Within Slim-Website, we need a script that creates the api docs and adds them to the Slim-Website repository. This way we can run it after we release a new version and update the website.

tflight:
Automated way to [Generate API [docs] to Github pages via Travis](https://github.com/ApiGen/ApiGen/wiki/Generate-API-to-Github-pages-via-Travis)?

designermonkey:
I have been revisiting this, and stumbled at the first hurdle, as apigen seems to not be installable via composer right now without package version issues.

akrabat:
Thinking about this, the API docs need to be on the Slim-Website repository as we're going to need this for 3.x and 4.x.

Maybe this needs to be a bash/php script that is on Slim-Website that checks out the Slim repo, creates the api docs in the right place on the website and then deletes the Slim repo?

designermonkey:
Good idea @akrabat

designermonkey:
Posted too early...

~~I have a script that does this kind of thing as I used to do this for server deployments of client projects. I'll clean it up and post here if this is how it wants to be done?~~

I can't find it any more :(

akrabat:
So close…

ertogs:
Hi there, 
this can be automated with travis ci quite easy. 
also see https://github.com/slimphp/Slim-Website/pull/214
I will create a new PR with setup instructions in a few minutes.

MarcHagen:
And what about just a simple webhook triggering the build?  
I'm used to have a deploy script that handles this stuff for me.

akrabat:
See https://github.com/slimphp/Slim-Website/issues/213

PR #2171, opened 2017-03-16
Add auto generated API docs using apigen with twitter-bootstrap template

crtl:
Hi,
currently there is no further documentation besides the User Guide to get some insights into the framework so generated an API documentation with [apigen](http://www.apigen.org/) using the twitter-bootstrap template.

```
apigen generate --source Slim --destination docs --template-theme bootstrap
```


geggleto:
I will review this later on today :)

tflight:
@crtl have you seen https://github.com/slimphp/Slim/issues/1617 ?

geggleto:
@akrabat @codeguy @silentworks thought's on merging this into this repo?

I was thinking about how it might be nice to have this on the website too.

akrabat:
The only thing I care about is that it's automated

geggleto:
I took a quick look, it does not appear to be automated. It would need to be run after we tag a new version.

crtl:
@akrabat its not, you would have to generate them manually. But its only one command (the one in my post)

akrabat:
As per other issue: https://github.com/ApiGen/ApiGen/wiki/Generate-API-to-Github-pages-via-Travis

akrabat:
Generally, I think the odds of remembering to do it manually are low

akrabat:
Another thought.  This needs to be on the Slim-Website repository as we're going to need this for 3.x and 4.x.

Maybe this needs to be a bash/php script that is on Slim-Website that checks out the Slim repo, creates the api docs in the right place on the website and then deletes the Slim repo?

Issue #2156, opened 2017-02-25
Allow all valid HTTP methods


amreladawy:
As allowed HTTP methods are defined in the protected `$validMethods`  in the Request class, HTTP methods like `LINK `and `UNLINK `are not supported. 

The proposed solution is to extend the `Request` class to add these methods. Wouldn't it be easier to have a setter function in the `Request` class or a function to append other HTTP methods to the  `$validMethods` ?

#2005

amreladawy:
Here is a pull request to add LINK/UNLINK 

https://github.com/slimphp/Slim/pull/2161

akrabat:
See comment on PR. A closed list of methods doesn't make sense, so we need to fix it properly.

johnhunt:
Rob suggested rewriting validateMethod() to use a regex along these lines: https://regex101.com/r/f80dzN/2

Zegnat:
That regular expression (``/^[[:alnum:]!#$%&\'*\/+\-.^_`|~]+$/i``) seems wrong to me. It allows for `/` within a method, but this doesn’t seem to be allowed as per RFC 7230.

The method in the Request Line is defined in [section 3.1.1.](https://tools.ietf.org/html/rfc7230#section-3.1.1):
```
     method         = token
```

Token is defined in [section 3.2.6.](https://tools.ietf.org/html/rfc7230#section-3.2.6):
```
     token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```
(`DIGIT` and `ALPHA` are defined by RFC 5234, the ABNF syntax.)

The correct regex is: [``/^[[:alnum:]!#$%&'*+\-.^_`|~]+$/i``](https://regex101.com/r/f80dzN/3).

akrabat:
I'm not going to disagree :)

akrabat:
Looking it up, Diactoros uses 

    /^[!#$%&\'*+.^_`\|~0-9a-z-]+$/i

(https://github.com/zendframework/zend-diactoros/blob/master/src/RequestTrait.php#L303)

Zegnat:
@akrabat you will want to use double-backtick to encapsulate inline code that includes a backtick itself.

[...]

Issue #2169, opened 2017-03-10
HTTP methods should not be made uppercase.


Zegnat:
While looking at the code for #2156, I spotted [`strtoupper()`](https://github.com/slimphp/Slim/blob/cc9b7892ac51220e47d31899f19aadcdf555c0bd/Slim/Http/Request.php#L347) being applied to methods. This is wrong. According to [RFC 7230 section 3.1.1.](https://tools.ietf.org/html/rfc7230#section-3.1.1) (emphasis mine):

> The method token indicates the request method to be performed on the target resource.  The request **method is case-sensitive**.

I considered changing it in the PR I am working on, but I feel like it might be out of scope. I can imagine this change being a BC breaking change in case people were expecting methods to be normalised. Filing this as a separate issue so we can discuss when and where this should be fixed.

tuupola:
FWIW this is also one part which fails on the PSR-7 integration tests.

https://gist.github.com/tuupola/5992c962f07068c4f4e3333c29d986fa

ghost:
According to PSR-7 this is also right.

```php
[...]

interface RequestInterface extends MessageInterface
{
    [...]

    /**
     * Return an instance with the provided HTTP method.
     *
     * While HTTP method names are typically all uppercase characters, HTTP
     * method names are case-sensitive and thus implementations SHOULD NOT
     * modify the given string.
     *
     * This method MUST be implemented in such a way as to retain the
     * immutability of the message, and MUST return an instance that has the
     * changed request method.
     *
     * @param string $method Case-sensitive method.
     * @return static
     * @throws \InvalidArgumentException for invalid HTTP methods.
     */
    public function withMethod($method);

    [...]
}

```
Tnx for reporting.

Zegnat:
Thanks for referencing PSR-7, @sharifzadesina. I am more knowledgable of the HTTP spec than PSR-7. Can we treat this as a bug then, and fix it against 3.x, even if it might cause BC breaks?

ghost:
@Zegnat All bug fixes are BC breaks, but slim have huge community so I don't think we can have this in 3.x, slim 4 will come soon, we can have this bug fix in version 4.

Zegnat:
I am trying to establish what is and isn’t worth working on for #2156, for 3.x, and what could better be postponed for 4.x.

akrabat:
The `strtoupper()` should certainly be removed from 4.x

akrabat:
[...]

PR #2181, opened 2017-03-21
Allow all valid HTTP methods

mathmarques:
As suggested on #2156, Slim should allow all valid HTTP Methods instead of a closed list.

Thx to @Zegnat too.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -111,23 +111,6 @@ class Request extends Message implements ServerRequestInterface
     protected $uploadedFiles;
 
     /**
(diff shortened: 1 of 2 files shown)


mathmarques:
Any guess on why Coverage decreased?

akrabat (approved):

akrabat:
> Any guess on why Coverage decreased?

You removed a number of lines of code that were covered by the tests, so the percentage of the codebase tested has gone down slightly.

Zegnat:
As this pull is only against 4.x, it might be worth also implementing #2169?

PR #2192, opened 2017-03-27
Fixing risky test, to move file to stream then test content

Sam-Burns:
'Slim\Tests\Http\UploadedFilesTest::testMoveToStream' was marked as risky, because it didn't include assertions.  This PR checks that moving the file contents to a stream properly retains the data.


geggleto:
But this doesn't actually test the moveTo method ?

mathmarques:
This test is supposed to move the uploaded file to a stream, so you could check if  [`moved === true`](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/UploadedFile.php#L255)

PR #2202, opened 2017-04-16
Change visibility of register method and vars

basuke:
When I want to reuse the Slim's Container class from other project,
registerDefaultServices() registers default service providers unless
overritten by configuration. The method and defaultSettings property
is private which prevent overwritten by sub class.

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -49,7 +49,7 @@ class Container extends PimpleContainer implements ContainerInterface
      *
      * @var array
      */
-    private $defaultSettings = [
+    protected $defaultSettings = [
         'httpVersion' => '1.1',
         'responseChunkSize' => 4096,
         'outputBuffering' => 'append',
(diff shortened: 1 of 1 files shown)


akrabat:
Out of interest, what benefit do you get by using Slim's Container rather than using Pimple directly in your non-Slim project?

basuke:
Well, I am writing a library inspired by Slim very much, so I is very natal to me using Slim\Container other than other ContainerInterface compatible containers 😃 

If I can't use Slim\Container as the super class of my Container, I will extend them from Pimple with the features of:
- Psr\Container\ContainerInterface
- property getter as an alias of get().

The latter is what I think good for separation of concern, on configuration and on runtime. But anyway with those features, I have no idea I can write shorter and cleaner code than current Silm\Container.

akrabat:
Ah!

Well, tbh, you'd be better off copying Slim's Container into your project directly and modifying it as appropriate. No point in bringing in the rest of Slim's code, unless you're using all that too I suppose.

basuke:
Well I don't want to copy them, but I understand the policy. Thanks.

PR #2206, opened 2017-04-17
Adding link(), unlink() shorthand functions

amreladawy:
As LINK and UNLINK have been accepted as allowed method, this PR for shorthand routing functions for them.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -244,7 +244,7 @@ class App
      */
     public function any($pattern, $callable)
     {
-        return $this->map(['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'], $pattern, $callable);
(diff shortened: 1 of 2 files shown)


geggleto (approved):

akrabat:
My gut feeling is that link and unlink aren't commonly used enough.

Thoughts, @silentworks, @codeguy ?

amreladawy:
Updated the unit test to cover the new functions.

amreladawy:
I want to add `LINK` and `UNLINK`to the list of methods in the function `any()` but I am not sure if this would match the approach of avoiding literally listing allowed methods just like PR #2156

geggleto:
@akrabat I was thinking the same thing, but I don't see any harm for the short-cuts.

silentworks:
I don't see any harm in writing it the long way by using `map`, I don't see this becoming a common thing anytime soon and don't feel we need to add a shortcut for it.

Issue #2201, opened 2017-04-15
Uri::getBaseUrl() should escape @ in password


Rudloff:
Hello,

Some user decided to use my app behind an Apache basic auth and some of the libraries I use started throwing errors about URLs. Turns out it was because they had a `@` in their password.

And inded this code:
```php
$url = new \Slim\Http\Uri('https', 'example.com', 443, '/', '', '', 'foo', 'b@r');
$url->getBaseUrl();
```
returns `https://foo:b@r@example.com`.

But according the [URL standard](https://url.spec.whatwg.org/), the user and password should be percent-encoded: `https://foo:b%40r@example.com`.

akrabat:
I'm not sure that the URL Standard is an actual RFC.

However, I can see the argument that we should percent-encode at least the `@` sign if it's in the user or password fields.

This URL Standard suggests percent-encoding all these characters:

    U+0020 SPACE, U+0022 ("), U+0023 (#), U+003C (<), U+003E (>), U+003F (?), U+0060 (`),
    U+007B ({), and U+007D (}), U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
    U+005B ([), U+005C (\), U+005D (]), U+005E (^), and U+007C (|) 

If we did this, will we break BC?

Rudloff:
Note that the W3C HTML validator also complains about this if such an URL is used in a `href` attribute.
`Bad value http://foo:b@ar@example.com for attribute href on element a: User or password contains an at symbol ("@") not percent-encoded.`

geggleto:
@akrabat in theory a password could contain any of those values. I would still think this would be fine as it is a bug-fix...

RyanNerd:
@akrabat @geggleto Is this still an open issue? If so I can work on it (have some time on my hands right now)...

akrabat:
This should be fixed, yes @RyanNerd

akrabat:
Actually, see https://github.com/slimphp/Slim/pull/2207 - this is needed for Slim 3 too.

X-Tender:
I don't know if the output of the `user:password@` values are still necessary or "up to date" because in Chrome you get an Deperacation error and the whole website don't work because Chrome blocks all traffic which goes through the base url.

```
[Deprecation] Subresource requests whose URLs contain embedded credentials (e.g. `https://user:pass@host/`) are blocked. See https://www.chromestatus.com/feature/5669008342777856 for more details.
```

PR #2207, opened 2017-04-17
URL encoding password

Sam-Burns:
Fix https://github.com/slimphp/Slim/issues/2201

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -348,7 +348,8 @@ class Uri implements UriInterface
      */
     public function getUserInfo()
     {
-        return $this->user . ($this->password ? ':' . $this->password : '');
+        $password = urlencode($this->password);
+        return $this->user . ($password ? ':' . $password : '');
     }
 
     /**
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -348,8 +348,9 @@ class Uri implements UriInterface
      */
     public function getUserInfo()
     {
+        $user = urlencode($this->user);
         $password = urlencode($this->password);
-        return $this->user . ($password ? ':' . $password : '');
+        return $user . ($password ? ':' . $password : '');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


geggleto:
Can you URL Encode the user too... for obvious reasons X_X

Rudloff (approved):
I confirm this fixes my usecase.

ghost:
First: You should use `rawurlencode` not `urlencode`.
Second: Take a look at [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.2) you will get that your encoding is wrong.

akrabat (approved):

akrabat:
Should this be against `3.x` ?

akrabat:
@sharifzadesina  if it's changed to rawurlencode, what is then wrong?

akrabat:
@Sam-Burns rebase required.

ghost:
@akrabat 
You must use a function like below one for encoding:
```php
[...]
    /**
     * Sets the URI user information.
     *
     * @param string      $user
     * @param string|null $password
     */
    private function setUserInfo($user, $password = null)
    {
        $this->userInfo = '';

        if ('' !== $user) {
            $this->userInfo = self::encodeUserInfo($user);
            if ('' != $password) {
                $this->userInfo .= ':' . self::encodeUserInfo($password);
            }
        }
    }

    /**
     * Encodes URI user information.
     *
     * @see https://tools.ietf.org/html/rfc3986#section-3.2.1
     *
     * @param string $string User information
     *
     * @return string Encoded user information
     */
    private static function encodeUserInfo($string)
    {
        return preg_replace_callback(
            '#(?:[^' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . ':%]+|%(?![A-Fa-f0-9]{2}))#',
            function ($matches) {
                return rawurlencode($matches[0]);
            },
            $string
        );
    }
[...]
```

akrabat:
Needs to be against Slim-Http now. Also, should we be encoding only on output?

Issue #2204, opened 2017-04-17
request->hasHeader('authorization') has different performance on PHP5.6 and PHP7


hefengxian:
Environment
+ PHP Version 5.6.21 windows & Apache/2.4.20 (Win64)
+ PHP Version 7.0.15-0ubuntu0.16.04.4 & Apache/2.4.18 (Ubuntu)
+ slim version `"slim/slim": "^3.5"`

My request is not set `Authorization` header, and this is my code
```php
$request->hasHeader('authorization')
```
means my expected result is `false`

+ when i run this code with PHP 7 on Ubuntu it return `false`
+ when i run this code with PHP 5.6 on Windows it return `true`

so i see the source code about headers, then i do this test

```php
var_dump($_SERVER['HTTP_AUTHORIZATION']);
``` 
+ when i run this code with PHP 7 on Ubuntu it return `null`
+ when i run this code with PHP 5.6 on Windows it return empty string `string(0) ""`

so if when i use PHP 5.6 on Windows and i want know is header `Authorization` is set or not, i can not use method `$request->hasHeader('authorization')`

and method `hasHeader(key)` inside it use method `array_key_exists(key, data)`, so is this reasonable

RyanNerd:
It appears that this is an edge case. Looking at the code in Slim it is loading the environment via the `$_SERVER` global array. I believe that the reason you are getting a `var_dump` of  `HTTP_AUTHORIZATION` as an empty string in your PHP 5.6 version is because you are using the CGI interface which you can use [`HTTP_AUTHORIZATION` to _trick_ Apache into using the `PHP_AUTH_USER` and `PHP_AUTH_PW`](https://support.tigertech.net/php-http-auth)

I don't have an environment to test this but my guess is that on your Windows based apache server your `.htaccess` file that looks something like this:

```
RewriteEngine on
RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization},last]
```
This tells Apache to use and populate the `HTTP_AUTHORIZATION` environment variable with the encoded username and password for basic auth. I'm guessing that Apache is setting the `HTTP_AUTHORIZATION` environment variable to the empty string if no user name or password is in the header (that there aren't authorization credentials) and that this issue has nothing to do with the PHP version or Slim's handling of `hasHeader()`

tuupola:
[...]

PR #2216, opened 2017-04-30
determineAuthorization(): HTTP_AUTHORIZATION may be null or an empty string

RyanNerd:
* `determineAuthorization()` should check if HTTP_AUTHORIZATION is null *OR* an empty string
* Closes issue #2204


tuupola:
It probably would be cleaner to use `empty()`.

Issue #2170, opened 2017-03-13
Suggestions for error-handling in v4


alexweissman:
I propose that error-handling for specific types of exceptions be decoupled from the main application class in v4.  Specifically referencing issues #1646 and #1337, singling out the `NotFoundException` and `MethodNotAllowedException`s for special treatment in `App` seems kind of arbitrary.  

If you take a look at the [`CoreErrorHandler` class we've developed in UserFrosting](https://github.com/userfrosting/UserFrosting/blob/master/app/sprinkles/core/src/Handler/CoreErrorHandler.php), it allows you to map an `Exception` type to a custom `ExceptionHandler` class.  This allows you to map as many different types of exceptions as you want to designated handlers, and even map child exception types to new handlers.  IMO, this is a much more flexible way of dealing with exceptions.

Right now, we have the `errorHandler` service returning an instance of `CoreErrorHandler`, but the problem is that Slim's custom handlers for `NotFoundException` and `MethodNotAllowedException` make it more difficult for developers to do their own switchyarding for these exceptions.  

It would be far more flexible if the `notFoundHandler` and `notAllowedHandler` services were removed, and everything were passed through the `errorHandler` service.  Perhaps something like my `CoreErrorHandler` could be built into Slim, or developers could do their own type-checking on the exception to determine an appropriate course of action.

l0gicgate:
@alexweissman I also have had those thoughts when looking at the error handling in the app class and thought it was a bit clunky.

I'm going to throw a PR together that implements your concept within the next couple days here and submit it for review.

alexweissman:
@l0gicgate do you have any experience with [whoops](https://github.com/filp/whoops)?  I wonder if Slim could just delegate all error-handling to this library.

l0gicgate:
@geggleto @alexweissman looking further at the source, I see that the `notFoundHandler` and the `notAllowedHandler` give insight into the error itself and that's why those handlers were probably created in the first place.

Do we really need to render these in the specific content type? The handlers have capabilities of rendering in HTML, XML and JSON. I'm going to boldly say that I suggest scratching all of that and decoupling the error rendering from the error handling.

We could give the option to the user to pass in a renderer if need be. We could also include by default HTML, JSON, XML and Plain Text renderers.

I dislike the fact that they are tightly coupled together.

l0gicgate:
@alexweissman whoops seems like a great library. I'm not sure though if @geggleto @akrabat want to depend on that.

[...]

PR #2222, opened 2017-05-10
4.x Error Handling

l0gicgate:
**Description**
As discussed on issue #2170 I feel that the error handling has always been a bit clunky and very tightly coupled to the rendering of the error. This PR eliminates all native error handlers and replaces it with `ErrorHandler` which handles all errors. You can create a handler for any exception in these ways:

```php
/*
 * Via Shortcut Method
 * Notice that the callable also gets passed $displayErrorDetails optionally from the App settings
 * You can either pass a callback or a resolvable callable as the second parameter
 * My implementation uses CallableResolver when 
 */
$handler = function ($req, $res, $exception, $displayErrorDetails) {
    return $response;
};
$app->setErrorHandler(\Your\Namespaced\Exception::class, $handler);

/*
 * Via Named Shortcut Methods Currently NotFound and NotAllowed only
 */
$app->setNotFoundHandler(function ($request, $response, $exception) {
    return $response;
});

/*
 * Via App Settings
 */
$app = new \Slim\App([
     /*
      * The default error handler can be set using this variable
      * or the $app->setDefaultErrorHandler() method
      * All exceptions are caught by this handler by default which is an instance of ErrorHandler
      */
    'defaultErrorHandler' => function ($req, $res, $exception, $displayErrorDetails) {
       return $res->withJson(['Oops...']);
    },

     /*
      * You can also set multiple error handlers via the settings object
      * by passing an associative array with 'Namespace\Exception' => function () {}
      * You can also pass a resolvable callable since this PR implements CallableResolver
      */
    'errorHandlers' => [
       \Slim\Exceptions\HttpNotFoundException::class => function (...) {...},
       \Your\Namespaced\Exception::class => ResolvableCallable::class
    ]
]);
```

**Error Handling/Rendering**
The rendering is finally decoupled from the handling. Everything still works the way it previously did. It will still detect the content-type and render things appropriately with the help of `ErrorRenderers`. The core `ErrorHandler` extends the `AbstractErrorHandler` class which has been completely refactored. By default it will call the appropriate `ErrorRenderer` for the supported content types. Someone can now provide their custom error renderer by extending the `AbstractErrorHandler` class and overloading the protected `renderer` variable from the parent. 

```php
class MyCustomErrorRenderer extends \Slim\Handlers\AbstractErrorRenderer
{
    public function render()
    {
        return 'My awesome format';
    }
}

class MyCustomErrorHandler extends \Slim\Handlers\AbstractErrorHandler
{
    protected $renderer = MyCustomErrorRenderer::class;
}
```

**HTTP Exceptions**
I just think it makes sense to throw HTTP based exceptions within the application. These exceptions works with nicely with the proposed native renderers. They can each have a `description` and `title` attribute as well to provide a bit more insight when the native HTML renderer is invoked. 

The base class `HttpException` extends `Exception` and comes with the following sub classes :
- `HttpBadGatewayException`
- `HttpBadRequestException`
- `HttpForbiddenException`
- `HttpInternalServerErrorException`
- `HttpNotAllowedException`
- `HttpNotFoundException`
- `HttpNotImplementedException`
- `HttpUnauthorizedException`

The developer can extend the `HttpException` class if they need any other response codes that we decide not to provide with the base repository. Example if you wanted a 504 gateway timeout exception that behaves like the native ones you would do the following:
I
```php
class HttpForbiddenException extends HttpException
{
    protected $code = 504;
    protected $message = 'Gateway Timeout.';
    protected $title = '504 Gateway Timeout';
    protected $description = 'Timed out before receiving response from the upstream server.';
}
```

**Status**
- [x] Build Passing
- [x] Unit Tests (Coverage Increased +2.03%)
- [ ] Discussion and Code Review.


akrabat:
Thanks for doing this. I will review and provide feedback at some point, but may not get to it until next week.

geggleto (commented):
There's code smell to me here as Exceptions are being treated as standard objects and not thrown.

We need to make a choice here.

1) If we use exceptions we should throw them

2) If we use standard objects then we must implement and expose an Event Messaging system.

I would prefer option 2 over 1.

l0gicgate:
By using Exceptions as objects you mean like it carrying request/response or extra details like allowed methods for said route?

What are my actionable items here @geggleto?

geggleto (approved):

akrabat:
There's a lot here and it's going to take me a while to go through.

First thoughts. 

1. Please get rid of `PhpException`. This only exists in Slim 3 because we had type hinted the handler to need an `Exception` which meant we wouldn't use the standard handler. 
    I don't want the additional stack trace because we've wrapped the Error into an Exception just so we can type hint `handleException()`.

2. Why is `displayErrorDetails` marked as an optional parameter for the error handler signature?

l0gicgate:
@akrabat

1. Will do
2. It's "optional" because not everybody cares about the parameter $displayErrorDetails. I passed it in just in case people needed to use it for a dev/prod scenario where they would want to handle the errors differently. It still gets passed to the handler always, you as a developer can choose to use it in the signature, or not. That's all I meant by "optional".

geggleto:
@akrabat Do we even need these renderers? ... Why are we even rendering content and not just letting exceptions bubble to where a user can catch and do something with them....

even in production, this could be a bad thing... what if I want to deregister all of the renderers? 

Seems like a pain in the butt to add more to deploy....

It makes more sense to me to have it bubble outside of app and catch it at a global level.
we can then decouple all rendering things and App becomes a middleware runner that returns a response or throws an exception.

akrabat:
Currently we render errors and I think that's part of Slim - otherwise, you may as well just use FastRoute directly. What I'm not sure about is whether we should have middleware to do it rather than having it built into dispatch.

geggleto:
@akrabat I'm not saying we don't render them, it's just I would rather do it in a very decoupled way.

I think Middleware is at least a good first step.

dopesong:
@akrabat +1 for moving error handling to middleware.

akrabat:
I suspect that we're going to do a multi-step error handling refactorying process.

This is a big PR which is why it's taking me a while. It's better than what we have now, but it's where I think we'll end up by 4.0.

Slim/Exception/HttpException.php
```diff
@@ -0,0 +1,139 @@
...
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+abstract class HttpException extends \Exception
+{
+    /**
+     * @var ServerRequestInterface|null
+     */
+    protected $request = null;
```
akrabat, line 12: Why does the exception object have a request?

Slim/App.php
```diff
@@ -750,16 +732,25 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
-            $notAllowedHandler = $this->getNotAllowedHandler();
-            return $notAllowedHandler($request, $response, $routeInfo[1]);
+        $exception = null;
+        switch ($routeInfo[0]) {
+            case Dispatcher::FOUND:
+                $route = $router->lookupRoute($routeInfo[1]);
+                return $route->run($request, $response);
+
+            case Dispatcher::METHOD_NOT_ALLOWED:
+                $exception = new HttpNotAllowedException;
```
akrabat, line 719: Why not just throw a new `HttpNotAllowedException` so that it's caught in `run()` ?

Slim/App.php
```diff
@@ -750,16 +732,25 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+                $exception->setAllowedMethods($routeInfo[1]);
+                $exception->setRequest($request);
+                break;
+
+            case Dispatcher::NOT_FOUND:
+                $exception = new HttpNotFoundException;
```
akrabat: Same here - could just throw it?

Slim/App.php
```diff
@@ -750,16 +732,25 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+                $exception->setRequest($request);
+                break;
```
akrabat: What happens if `$routeInfo[0]` isn't one of the three cases?

Slim/App.php
akrabat: To answer my own question. If we return a response, then middleware could process it.

Slim/Exception/HttpException.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    protected $title = '';
+    /**
+     * @var string
+     */
+    protected $description = '';
+    /**
+     * @var bool
+     */
+    protected $recoverable = true;
```
akrabat: What is the property for?

Slim/Handlers/ErrorRenderers/PlainTextErrorRenderer.php
```diff
@@ -0,0 +1,62 @@
...
+class PlainTextErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @return string
+     */
+    public function render()
+    {
+        $e = $this->exception;
+
+        $text = 'Slim Application Error:' . PHP_EOL;
```
akrabat, line 25: Needs to support case when `displayErrorDetails` is `false`

Slim/App.php
l0gicgate: What are the other cases? Or what should the default be, throw an exception?

Slim/Exception/HttpException.php
l0gicgate: Originally, when an exception was thrown you couldn't access a reference of the request at the stage where the exception was thrown. It would pass you a reference of the request that hadn't been passed through the middleware cycle so if any data or properties were bootstrapped to the request during the middleware cycle, you wouldn't be able to access them. 

I believe that's a problem for developers as it doesn't reflect the state of the request at the time the exception was thrown. 

Hope that makes sense.

Slim/Exception/HttpException.php
l0gicgate: My memory is a bit foggy about that but I remember reading through the code comments and there was something about a "recoverable" flag. For example a 500 is most likely not recoverable while another type of exception may be. 

At this moment I don't believe it is being used and could be taken out.

Slim/Handlers/ErrorRenderers/PlainTextErrorRenderer.php
l0gicgate: From what I recall there is no difference between the output of displayErrorDetails true/false for the plain text renderer. 

If there is a difference in format please enlighten me.

Slim/App.php
akrabat: There shouldn't be any other cases, but assumptions change :)

Make it throw an exception.

Slim/Exception/HttpException.php
akrabat: Please remove.

Slim/Handlers/ErrorRenderers/PlainTextErrorRenderer.php
akrabat: It's actually used for OPTIONS. See https://github.com/slimphp/Slim/blob/3.x/Slim/Handlers/NotAllowed.php#L39 where I renamed the method name to make this clearer.

Arguably, we need a better solution, but for now, we're just replicating how it currently works.

Slim/Exception/HttpException.php
akrabat: Okay.

Slim/App.php
l0gicgate: What do you want the exception message to say? I'm clueless.

I was thinking something like: 'Unknown routing error occurred.'

Slim/Handlers/ErrorRenderers/PlainTextErrorRenderer.php
l0gicgate: Actually, all that logic has been changed and can be found in AbstractErrorHandler.php

An OPTIONS request may demand a different content-type than text/plain which wasn't honored before and I fixed that. You can now request JSON output from an OPTIONS request and it will honor the content type.

See https://github.com/l0gicgate/Slim/blob/4.x/Slim/Handlers/AbstractErrorHandler.php#L118-L146

This change is not required.

ghost:
Is this something that should be moved to a middleware?

akrabat:
I think so.  I like a lot of the concepts in here and would like to explore how to make it middleware.

akrabat:
I think this comment is still relevant: https://github.com/slimphp/Slim/issues/1686#issuecomment-261306557

juliangut:
I've been working on a quite similar solution for Slim3, but in my case I'm relying on exception status code instead of having different classes per HTTP status, i think this gives more flexibility for handling purposes

l0gicgate:
@juliangut I think named exceptions are better because they are explicit. Status codes aren’t.

akrabat:
FWiW, I also think named exceptions are more expressive in the codebase. 

I’m on the fence about shipping exceptions that we don’t need for the dispatcher.

akrabat:
Discussion with @l0gicgate on reworking this:

* Move the error handling to middleware that we apply at the outer edge
* ie. `App` just throws either a NotFound exception or a MethodNotAllowed exception
* Write a new `Slim\Middleware\ErrorMiddleware` which then uses renderers
* Renderers in an appropriate namespace
* Later, we may want to detect that no error middleware has been added to the pipeline and add ours to the outside of the stack in run()

Let's use a new PR for this. Thanks for you work so far and again, I'm really sorry that it's taken me this long.

PR #2225, opened 2017-05-12
Check ini_set

gregory-gueret:
If default_mimetype is already set to empty, ini_set is not executed.
Ini_set is not allowed on my shared server for security reasons.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -572,7 +572,9 @@ class App
     protected function finalize(ResponseInterface $response)
     {
         // stop PHP sending a Content-Type automatically
-        ini_set('default_mimetype', '');
+        if (!empty(ini_get('default_mimetype'))) {
+            ini_set('default_mimetype', '');
+        }
 
         if ($this->isEmptyResponse($response)) {
             return $response->withoutHeader('Content-Type')->withoutHeader('Content-Length');


geggleto:
This code change doesn't do anything....
The if reads as...

if not empty default_mime-type set to empty.

In your hosting environment you will have a default mime-type set.

http://php.net/manual/en/ini.core.php#ini.default-mimetype

gregory-gueret:
thank you for your reply,
But my need is do not run ini_set, but I can set a default default_mimetype in php.ini

akrabat:
What's your use-case?

akrabat:
Right, so the problem is that `ini_set` is disabled on your shared_host. Hmm.

gregory-gueret:
Yes `ini_set` is disabled for internal security reasons and this simple modification will make it possible for me to use Slim

geggleto:
I'm pretty torn on this.... while this feature might enable slim for you, it does impose a penalty on the rest of the community as if blocks do have a material cost.

noogen:
How about a compromise by using the @ symbol?  @ini_set('default_mimetype', '');

akrabat:
The`@` symbol is extremely slow.

noogen:
I did not know that.  Saw it uses in other places for handle the same situation.  Did some reading on it.  Now I know.  Thanks @akrabat

geggleto:
@noogen a good rule of thumb is to never use @ ;)

akrabat:
I don't think we should change this in Slim 3.

Issue #1976, opened 2016-09-02
Route group callable


AydinHassan:
Hi,

I was trying to organise a route group definition into an invokable class to tidy up some files and noticed that I didn't have access to app, so I couldn't define any child routes.

It seems `$app` is only binded to closures. Is there any reason we don't just pass `$app` as a parameter to the route group callable at: https://github.com/slimphp/Slim/blob/3.x/Slim/RouteGroup.php#L45 ?

If not, I would be happy to create a PR :)

geggleto:
the app class gets bound to the group.

``` php
$app->group('/admin', function () {
    //We are now in the context of $app
    $this->get('/users', ''); //Example, is the same as $app->get
});
```

AydinHassan:
Only if it is a closure: https://github.com/slimphp/Slim/blob/3.x/Slim/RouteGroup.php#L41-L43. I was trying to use an invokable class.

geggleto:
oh well that's interesting....

Using DI methods you can inject the router into your class which would allow you to register routes, but I guess it wouldn't be nearly as nice.

danopz:
This should be resolved by #2229

PR #2229, opened 2017-05-17
Pass app into group callable

geekish:
As discussed in #1976, route group callables are bound to App, making the use of closures easy, but App is not passed to groups on invoke so an invokable class will not work.

I chose to remove the closure binding altogether as passing App through on invoke negates the need to do this anymore. The tests weren't even using the feature...

### DIFF
--- a/Slim/RouteGroup.php
+++ b/Slim/RouteGroup.php
@@ -34,7 +34,7 @@ class RouteGroup extends Routable implements RouteGroupInterface
      *
      * @param App $app The App to bind the callable to.
      */
-    public function __invoke(App $app)
+    public function __invoke(App $app = null)
     {
         // Resolve route callable
         $callable = $this->callable;
--- a/Slim/RouteGroup.php
+++ b/Slim/RouteGroup.php
@@ -34,7 +34,7 @@ class RouteGroup extends Routable implements RouteGroupInterface
      *
      * @param App $app The App to bind the callable to.
      */
-    public function __invoke(App $app = null)
+    public function __invoke(App $app)
     {
         // Resolve route callable
         $callable = $this->callable;
@@ -42,11 +42,6 @@ class RouteGroup extends Routable implements RouteGroupInterface
             $callable = $this->callableResolver->resolve($callable);
         }
 
-        // Bind RouteGroup to app instance
-        if ($callable instanceof \Closure && $app !== null) {
-            $callable = $callable->bindTo($app);
-        }
-
-        $callable();
+        $callable($app);
     }
 }
--- a/Slim/RouteGroup.php
+++ b/Slim/RouteGroup.php
@@ -32,7 +32,7 @@ class RouteGroup extends Routable implements RouteGroupInterface
     /**
      * Invoke the group to register any Routable objects within it.
      *
-     * @param App $app The App instance to pass to the group callable
+     * @param App $app The App instance to bind/pass to the group callable
      */
     public function __invoke(App $app = null)
     {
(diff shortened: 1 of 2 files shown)


silentworks:
I am happy with this change getting added into Slim 4, @geekish has also provided example code to show the use of this feature, which current implementation doesn't allow.

https://gist.github.com/geekish/3a1cf866aec0a6ee1935007cbf915d78

geggleto:
I need to think about it more. The idea of passing App around strikes me as a code smell.

I understand the desire to fix the problem with route groups, and registering routes and what not.

I think it would make more sense to only pass through the Router object.

//cc @akrabat

silentworks:
@geggleto we can't pass the `Router` instead, because the `Router` doesn't carry these methods, these are helper methods [get, post, put, delete, map] that are created inside of `App`.

geggleto:
@silentworks maybe they should be on the router and not the app ;)

geekish:
I would definitely prefer that we pass Router through instead and get rid of the helper methods on App, but that would be out of the scope for this PR.

Besides that, I don't see how passing App through is any smellier than binding the closure to it; that kind of magic seems pretty smelly to me.

silentworks (approved):

ghost:
I don't like this PR, there is no reason to add `App` to the route group, routing and application are separate packages.

silentworks:
@sharifzadesina we are already doing this in the form of closure binding. So I am not sure why there is fuss about this PR.

ghost:
@silentworks of course you are, and I know this will solve the sequent injection of the `App` class, but this is not good design, imagine router without `App` class, what will happen? your router will be useless.

silentworks:
@sharifzadesina so the point you are making has nothing to do with this PR. Someone else should send another PR to fix the issue you are concerned about.

ghost:
@silentworks Maybe, but by this PR you are making the bad design bigger and bigger. anyway this is just my idea.

silentworks:
@sharifzadesina no we aren't making it bigger, it still the same magnitude as it was before this PR. This PR however is making the use more consistent.

silentworks:
For past commenters and future commenters on this issue, I want to make it clear that coming with an argument about design decision has no effect on this PR, this PR was created after the design decision was taken, this PR is just making it more clear as to how you create routes in route groups.

I find it distasteful that many have commented about bad design on a PR by a first time contributor that has/had no effect on the bad design decision that was taken more than a year ago. If you wish to fix this bad design decision, raise a new issue or send a PR, please don't make this PR out to be what its not.

geggleto (approved):

geggleto:
@geekish I think we should revisit the groups in a different way. :)

akrabat:
One side effect of removing the binding is that it will be that much harder to upgrade from 3 to 4. Is the fact that you don't like the closure binding feature of PHP a reason to remove it?

silentworks:
I think its more of consistency and my IDE would know how to give me autocomplete hint based on `$app` than currently where I get none based on `$this`.

Also since this is v4 I would expect there to be breaks, I actually don't see this as a major break either (although that depends on how many route group you have already using `$this` inside.

akrabat:
I'm happy to pass in `$app` to the closure / `__invoke()`, I'm just questioning whether we should remove the binding of `$app` to  `$this` for closures. 

I think that smoothing out upgraders from v3 warrants it.

We should probably remove any documentation in v4 about the binding though and maybe mark it deprecated for removal in v5?

akrabat (changes requested):
Please reinstate binding `$app` to `$this`, but add a `trigger_error` in the `if` statement of type `E_USER_DEPRECATED` noting that the binding of $this to the group is deprecated.

akrabat:
Note that we could also do this for `3.x` to provide a forward migration path.

geekish:
I've re-added the closure binding, but didn't add a `trigger_error` - I agree there needs to be some way to notify the user about the deprecation but that would trigger on every route using a closure whether the user utilizes the bind or not.

geggleto:
thank you @geekish  :)

akrabat (approved):

akrabat:
Any reason we can't do this for Slim 3 too?

PR #2233, opened 2017-05-22
Minor optimizations

theodorejb:
* Avoid function call overhead when adding a single group to an array
* Fix documented return type of `Router::createRoute`
* Simplify `Router::popGroup` implementation
    * Avoid `instanceof` check by directly returning result of `array_pop`.
    * The `RouterInterface` method was incorrectly documented as returning bool, when in fact the last group is returned if one exists.
    * It might actually be better for this method to not return anything at all, since Slim never uses the returned value.


akrabat (approved):

akrabat:
Thanks.

PR #2237, opened 2017-05-26
Proposal to use Behat for smoke testing major features

Sam-Burns:
This PR is a proposal to introduce a small number of Behat tests, to provide integration tests for Slim 4.

It provides a smoke tests using a basic 'hello world' app, which makes sure that the front controller and routing infrastructure will stay working as development continues.  It also covers the middleware functionality, with a [scenario](https://github.com/Sam-Burns/Slim/blob/behat/tests/behat/feature_files/middlewares.feature#L6-L10) that discusses middleware closures being added to an application.


giftnuss:
My opinion is this should be an independent package. Behat is nice but I dont't think it should be a dev dependency by default.

akrabat:
Please rebase.

geggleto:
@Sam-Burns //poke

PR #2239, opened 2017-06-06
Refactor router class

ipabz:


Slim/Router.php
```diff
@@ -271,12 +272,15 @@ public function getRoutes()
      */
     public function getNamedRoute($name)
     {
-        foreach ($this->routes as $route) {
-            if ($name == $route->getName()) {
-                return $route;
-            }
+        $routes = array_filter($this->routes, function ($route) use ($name) {
```
geggleto: This should only ever return one route.

geggleto: This is also not an improvement on readability in my opinion.

geggleto (changes requested):

Slim/Router.php
```diff
@@ -233,11 +232,13 @@ protected function createDispatcher()
                 'routeParser' => $this->routeParser,
                 'cacheFile' => $this->cacheFile,
             ]);
-        } else {
-            $this->dispatcher = \FastRoute\simpleDispatcher($routeDefinitionCallback, [
-                'routeParser' => $this->routeParser,
-            ]);
-        }
+
+            return $this->dispatcher;
```
danopz, line 236: Why do you want multiple return statements?

Slim/Router.php
```diff
@@ -272,10 +273,11 @@ public function getRoutes()
     public function getNamedRoute($name)
     {
         foreach ($this->routes as $route) {
-            if ($name == $route->getName()) {
+            if ($name === $route->getName()) {
```
danopz, line 276: Could be a potential BC break?

Slim/Router.php
silentworks: I think its the idea of return early, I am however not sure how much of a speed improvement this gives.

danopz: Should be nearly nothing as there is no code between the if and return.

Slim/Router.php
geggleto: Yes, it would.

silentworks:
As I don't think there are any benefits to these changes, I will be rejecting this PR.

PR #2242, opened 2017-06-12
Traits should only know their own scope

ghost:
Getting myself back into Slim :)

In an effort to contain the scope knowledge of the `CallableResolverAwareTrait` with regard to `$this->container`, I have abstracted out container awareness to another trait, and used that in places where `£this->container` was being used throughout the code.

This does add a public method `getContainer()` in few places where there wasn't one before, but I feel this is a small tradeoff for looser coupling.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -45,6 +45,7 @@ use Slim\Interfaces\RouterInterface;
  */
 class App
 {
+    use ContainerAwareTrait;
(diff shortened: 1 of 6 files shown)


akrabat:
There's a good chance this needs to be against the `4.x` branch

ghost:
Okidoke. I didn't realise there was one.

PR #2244, opened 2017-06-14
Adding plain not found response for OPTIONS requests

piotr-cz:
OPTIONS requests are being sent by browsers with the `Accept: */*` header (as opposed to this [MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS)).

This is causing AbstractHandler::determineContentType method to evaluate response format as `text/html` and so NotFound handler creates a HTML response.

Such HTML response is not necessary for simple preflight request and my slow down communication (both on client side by downloading large payload and on server side by generating it; for example I'm rendering a 4Kb html page with stylesheets and suff).

So I've added a check similar to the [one in NotAllowed handler](https://github.com/slimphp/Slim/blob/3.8.1/Slim/Handlers/NotAllowed.php#L36-L39) and corresponding `renderPlainNotFoundOutput` method. NotFound response body for OPTIONS requests will now be just `Not found`.

This is very similar to https://github.com/slimphp/Slim/commit/a2cc9d9801c037f886b9253f936ef5f0994c5f1a when such behavior has been added to `NotAllowed` handler.

### DIFF
--- a/Slim/Handlers/NotFound.php
+++ b/Slim/Handlers/NotFound.php
@@ -32,23 +32,28 @@ class NotFound extends AbstractHandler
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response)
     {
-        $contentType = $this->determineContentType($request);
-        switch ($contentType) {
-            case 'application/json':
-                $output = $this->renderJsonNotFoundOutput();
-                break;
+        if ($request->getMethod() === 'OPTIONS') {
+            $contentType = 'text/plain';
+            $output = $this->renderPlainNotFoundOutput();
+        } else {
+            $contentType = $this->determineContentType($request);
+            switch ($contentType) {
+                case 'application/json':
+                    $output = $this->renderJsonNotFoundOutput();
+                    break;
 
-            case 'text/xml':
-            case 'application/xml':
-                $output = $this->renderXmlNotFoundOutput();
-                break;
+                case 'text/xml':
+                case 'application/xml':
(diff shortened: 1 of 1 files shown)


piotr-cz:
Travis failed on PHP:hhvm build:
```
HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.
```

Sam-Burns:
@piotr-cz We're aware of the build problem.  It is addressed by https://github.com/slimphp/Slim/pull/2245.

akrabat:
@piotr-cz can you rebase and then we can check that travis passes

akrabat (changes requested):
Please rebase for CI tests

piotr-cz:
@akrabat I've rebased, see https://github.com/piotr-cz/Slim/commits/29d517cfb98df8a173088d1ed69b955bf2b29799/.travis.yml

No idea why it doesn't show up, just as a notice after https://github.com/slimphp/Slim/pull/2244#issuecomment-308683978

akrabat (changes requested):
Needs a unit test too. Sorry that I didn't mention this earlier.

piotr-cz:
I need a hand here:
What is the proper way to create an OPTIONS mock request?
The `getRequest` method [may receive a method](https://github.com/slimphp/Slim/blob/3.8.1/tests/Handlers/NotFoundTest.php#L62), but it doesn't use it.

akrabat:
Thanks.

PR #2250, opened 2017-06-18
Add `only` restriction to getParams()

Grafikart:
## Proposal

Enhanced `Request::getParams()` method with an extra argument to filter keys.

```
$request->getParams(['name', 'content']);
```

This will filter parameters exporting only the specified keys (defaulting to null if the key doesn't exist). I am unsure for the "defaulting to null" so feel free to comment or discard the whole PR if you think this feature should not be included by Slim.

## Possible changes 

- Do not default to null if a key doesn't exist (let PHP throw an error or create a custom exception ?)
- Replace the array argument with a spread operator or `func_get_args` so we can write `$request->getParams('name', 'content');`

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1202,14 +1202,22 @@ class Request extends Message implements ServerRequestInterface
      *
      * Note: This method is not part of the PSR-7 standard.
      *
-     * @return array
+     * @param array|null $only list the keys to retrieve.
+     * @return array|null
      */
-    public function getParams()
+    public function getParams($only = null)
     {
         $params = $this->getQueryParams();
         $postParams = $this->getParsedBody();
         if ($postParams) {
             $params = array_merge($params, (array)$postParams);
+        }
+        if ($only) {
+            $onlyParams = [];
+            foreach ($only as $key) {
+                $onlyParams[$key] = isset($params[$key]) ? $params[$key] : null;
+            }
+            return $onlyParams;
         }
 
         return $params;
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1205,7 +1205,7 @@ class Request extends Message implements ServerRequestInterface
      * @param array|null $only list the keys to retrieve.
      * @return array|null
      */
-    public function getParams($only = null)
+    public function getParams(array $only = null)
     {
         $params = $this->getQueryParams();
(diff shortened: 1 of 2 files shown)


Grafikart:
I don't get the change in coveralls % since my changes are covered by new tests :(

Slim/Http/Request.php
```diff
@@ -1202,15 +1202,23 @@ public function getQueryParam($key, $default = null)
      *
      * Note: This method is not part of the PSR-7 standard.
      *
-     * @return array
+     * @param array|null $only list the keys to retrieve.
+     * @return array|null
      */
-    public function getParams()
+    public function getParams($only = null)
```
pachico: Regardless of if this feature will be implemented or not, please correct to:
```php
public function getParams(array $only = null)
```
Your unit tests should have probably tested also non expected inputs.

Cheers

Grafikart: Fixed

akrabat:
I'm on the fence about this one as I'm not sure it adds enough that it should be part of Slim or whether it should be in user-land.

danopz:
Should be user-land IMHO.

silentworks:
This is not that easy to implement in user-land due to our current design and currently it won't break anything in core and its not doing any harm adding it in.

geggleto:
I would like to see this PR implement the 2nd possible change.

Grafikart:
I can implement the 2nd possible change but do I keep the "first argument as array" ? or only the spread operator ?

akrabat:
This is interesting :)

If the key doesn't exist then it shouldn't be in the returned array as otherwise `getParams()` and `getParams(['not_there'])` will return different results which is inconsistent.

I disagree with @geggleto on point 2 as I think that `getParams('foo', 'bar')` isn't as clear that you're selecting a subset from the parameters array.


Finally, I should note that the user land implementation is a one liner:

```php
$some = array_filter($request->getParams(), function ($k) use ($only) { return in_array($k, $only); }, ARRAY_FILTER_USE_KEY);
```

geggleto:
@akrabat that is not a one-liner ;)

I am torn on the 2nd point; I could go either way with it.

akrabat:
Merged after changing to not provide a default of null for '$only` keys that don't exist in the params list.

akrabat:
Thanks @Grafikart for your work on this.

PR #2254, opened 2017-06-29
Replace addContentLengthHeader with ContentLength Middleware

akrabat:
This feature isn't needed in core and is better as optional middleware.

### DIFF
--- a/Slim/Middleware/ContentLength.php
+++ b/Slim/Middleware/ContentLength.php
@@ -1,0 +1,30 @@
+<?php
+namespace Slim\Middleware;
+
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Slim\Http\Body;
+
+class ContentLength
+{
+    /**
+     * Invoke
+     *
+     * @param  ServerRequestInterface $request   PSR7 server request
+     * @param  ResponseInterface      $response  PSR7 response
+     * @param  callable               $next      Middleware callable
+     * @return ResponseInterface                 PSR7 response
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $response = $next($request, $response);
+
+        // Add Content-Length header if not already added
+        $size = $response->getBody()->getSize();
+        if ($size !== null && !$response->hasHeader('Content-Length')) {
+            $response = $response->withHeader('Content-Length', (string) $size);
+        }
+
+        return $response;
+    }
+}
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,9 +1,11 @@
 # How to upgrade
 
+* [2254] - You need to add the `Middleware\ContentLength` middleware if you want Slim to add this automatically.
 * [2166] - You need to add the `Middleware\OutputBufferingMiddleware` middleware to capture echo'd or var_dump'd output from your code.
 * [2098] - You need to add the App's router to the container for a straight upgrade. If you've created your own router factory in the container though, then you need to set it into the $app.
 * [2102] - You must inject custom route invocation strategy with `$app->getRouter()->setDefaultInvocationStrategy($myStrategy)`
 
+[2254]: https://github.com/slimphp/Slim/pull/2254
 [2166]: https://github.com/slimphp/Slim/pull/2166
 [2098]: https://github.com/slimphp/Slim/pull/2098
 [2102]: https://github.com/slimphp/Slim/pull/2102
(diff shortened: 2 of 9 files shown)


damianopetrungaro:
@akrabat 
As discussed on Slack for the AppTest:



    public function testAddSettings()
    {
        $app = new App();
        $app->addSettings(['foo' => 'bar']);
        $this->assertAttributeContains('bar', 'settings', $app);
    }

    public function testAddSetting()
    {
        $app = new App();
        $app->addSetting('foo', 'bar');
        $this->assertAttributeContains('bar', 'settings', $app);
    }


with this it works 😄

damianopetrungaro:
Or even better:

    public function testAddSettings()
    {
        $app = new App();
        $reflection = new ReflectionProperty(App::class, 'settings');
        $reflection->setAccessible(true);
        $reflection->setValue($app, []);
        $app->addSettings(['foo' => 'bar']);
        $this->assertSame(['foo' => 'bar'], $reflection->getValue($app));
    }

    public function testAddSetting()
    {
        $app = new App();
        $reflection = new ReflectionProperty(App::class, 'settings');
        $reflection->setAccessible(true);
        $reflection->setValue($app, []);
        $app->addSetting('foo', 'bar');
        $this->assertSame(['foo' => 'bar'], $reflection->getValue($app));
    }

akrabat:
Ping @geggleto

Slim/Middleware/ContentLength.php
```diff
@@ -0,0 +1,30 @@
...
+     * @return ResponseInterface                 PSR7 response
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $response = $next($request, $response);
+
+        // Add Content-Length header if not already added
+        $size = $response->getBody()->getSize();
+        if ($size !== null && !$response->hasHeader('Content-Length')) {
+            $response = $response->withHeader('Content-Length', (string) $size);
```
geggleto: .... why are we casting it to a string? ... my gut says that is a really bad idea.

tests/Middleware/ContentLengthTest.php
```diff
@@ -0,0 +1,44 @@
...
+        $response = new Response();
+
+        $next = function (ServerRequestInterface $req, ResponseInterface $res) {
+            $res->write('Body');
+            return $res;
+        };
+
+        $newResponse = $mw($request, $response, $next);
+
+        $this->assertEquals(4, $newResponse->getHeaderLine('Content-Length'));
```
geggleto: This test should also assert that the body's content is actually 'Body'

@akrabat what about multi-byte character sets?

geggleto (changes requested):

Slim/Middleware/ContentLength.php
akrabat: [`withHeader()`](https://github.com/php-fig/http-message/blob/master/src/MessageInterface.php#L128) is defined as taking a string as the second parameter.

tests/Middleware/ContentLengthTest.php
akrabat: Why? The code under test doesn't touch the body and ensuring the body is the correct text is covered by tests on `Response` and `App`.

akrabat: Size of multi-byte character sets is an issue for `Slim\Http\Stream()` and should be tested there I think?

i.e. I'm testing that the correct header is added as that's what this Middleware does.

akrabat: There's an argument that the test should be:

```php
$this->assertEquals($response->getBody()->getSize(), $newResponse->getHeaderLine('Content-Length'));
```

or even:

```php
$this->assertTrue($newResponse->hasHeader('Content-Length'));
```

akrabat:
Renamed to `ContentLengthMiddleware` as per https://github.com/slimphp/Slim/pull/2288#discussion_r133005546

UPGRADING.md
```diff
@@ -1,9 +1,11 @@
 # How to upgrade
 
+* [2254] - You need to add the `Middleware\ContentLength` middleware if you want Slim to add this automatically.
```
campersau: `ContentLengthMiddleware`?
Maybe the `OutputBuffering` should also be renamed?

akrabat: Good catch. Thanks.

Yes, 'OutputBuffering' will need renaming too, but that's a separate PR if you fancy doing it?

geggleto (approved):

akrabat:
Rebased to fix conflicts

PR #2261, opened 2017-07-13
Upgraded tests to use the latest PHPUnit (6.2 at this writing)

RyanNerd:
* The latest PHPUnit offers a more clean output and details on failures than previous versions.

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -25,8 +25,9 @@ use Slim\Http\Response;
 use Slim\Http\Uri;
 use Slim\Router;
 use Slim\Tests\Mocks\MockAction;
+use PHPUnit\Framework\TestCase;
 
-class AppTest extends \PHPUnit_Framework_TestCase
+class AppTest extends TestCase
 {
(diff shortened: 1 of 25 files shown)


tests/AppTest.php
```diff
@@ -41,7 +42,7 @@ public static function tearDownAfterClass()
 
     public function testContainerInterfaceException()
     {
-        $this->setExpectedException('InvalidArgumentException', 'Expected a ContainerInterface');
+        $this->expectException(\InvalidArgumentException::class);
```
campersau, line 45: Why not adding `$this->expectExceptionMessage('Expected a ContainerInterface');`?

RyanNerd:
I just looked at the CI failures. I sincerely did not know that the latest PHPUnit requires PHP 7.x. 
This PR may be better to target Slim 4.x (if the decision to require PHP 7 for Slim 4.x hasn't been made).

akrabat:
At the moment Slim 4 is targeting 5.6. If that changes, then we'll update PHPUnit.

PR #2263, opened 2017-07-17
Minor perf improvement during error handling

Sam-Burns:
Makes the error handling in the front controller run very slightly faster, by avoiding `call_user_func_array()`.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -865,8 +865,7 @@ class App
             $handler = $this->getErrorHandler();
             $params = [$request, $response, $e];
         }
-
-        return call_user_func_array($handler, $params);
+        return $handler(...$params);
     }
 
     /**
@@ -882,8 +881,6 @@ class App
     protected function handlePhpError(Throwable $e, ServerRequestInterface $request, ResponseInterface $response)
     {
         $handler = $this->getPhpErrorHandler();
-        $params = [$request, $response, $e];
-
-        return call_user_func_array($handler, $params);
+        return $handler($request, $response, $e);
     }
 }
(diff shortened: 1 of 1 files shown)


Slim/App.php
```diff
@@ -865,8 +865,7 @@ protected function handleException(Exception $e, ServerRequestInterface $request
             $handler = $this->getErrorHandler();
             $params = [$request, $response, $e];
         }
-
-        return call_user_func_array($handler, $params);
+        return $handler(...$params);
```
silentworks: Is this a PHP 7 feature?

chadicus: [PHP 5.6](http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list)

Sam-Burns: The Travis build's passing in PHP 5.6-7.1 and HHVM - think we're all good on that front.

RyanNerd: Why are you using `return $handler(...$params);` for the `handleException()` method, but in the `handlePhpError()` method you implicitly list the parameters: `return $handler($request, $response, $e);`  ?

akrabat: Maybe refactor and replace the `$params` assignments with the `return $handler` call?

That way we can also lose the final `else`.

akrabat (changes requested):

RyanNerd:
A couple of suggestions (thanks for your work so far):

Refactoring handleException to perhaps use the `switch` construct.

It also looks like you are missing the `elseif ($e instanceof SlimException)` logic in this PR:
The `SlimException` is implicitly thrown in the unit tests.

akrabat:
> It also looks like you are missing the `elseif ($e instanceof SlimException)` logic in this PR:
The `SlimException` is implicitly thrown in the unit tests.

In 4.x, there's no `SlimException`.

PR #2265, opened 2017-07-18
Build process

Sam-Burns:
Several build process changes.

These are as follows:

- Add `satooshi/php-coveralls` as a dev dependency, as `squizlabs/php_codesniffer` is, rather than installing it indirectly in the `.travis.yml` process.
- Add a 'canary build' with Travis' `php: nightly`.  It won't break the build if this one fails, but it will tell us if there are any incompatibilities with PHP 7.2 that we need to be aware of.
- Use PHP 7.1 to compute code coverage and run the 'phpcs' static analyser.  This makes it run faster than using PHP 5.6.
- Add a 'prefer lowest' test with Composer.  This checks that we are telling the truth about the minimum necessary versions of dependencies in our `composer.json` file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,29 @@ language: php
 
 dist: trusty
 
-php:
-  - 5.6
-  - 7.0
-  - 7.1
-  - hhvm
+matrix:
+  include:
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
+      env: ANALYSIS='true'
+    - php: hhvm
+    - php: 5.6
+      env: COMPOSER_ARGS='--prefer-lowest'
+    - php: nightly
+  allow_failures:
+    - php: nightly
 
 before_script:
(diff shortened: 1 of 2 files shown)


akrabat (approved):

akrabat:
@Sam-Burns I added this, but am a little concerned that `satooshi/php-coveralls` has pulled in guzzle v3. Is this going to cause us issues?

Sam-Burns:
@akrabat I'm pretty sure it uses an older version of Guzzle, but that the v4-6 versions are differently namespaced.  Do let me know if you notice any issues, but I don't think it's a problem.

akrabat:
@Sam-Burns Okay. The Composer warning isn't nice though, so lets hope they update.

Sam-Burns:
@akrabat Sounds like they're working on tagging a stable release without it.  In the meantime, as it's a dev dependency, you only get the warning when you clone Slim.  It doesn't happen if Slim is a dependency in your Composer file.

PR #2266, opened 2017-07-18
remove container-interop polyfill

1ma:
I propose to stop depending `container-interop/container-interop` now that `pimple/pimple` ^3.1 already pulls down `psr/container`.


Slim/Exception/ContainerException.php
```diff
@@ -9,12 +9,12 @@
...
 
 use InvalidArgumentException;
-use Interop\Container\Exception\ContainerException as InteropContainerException;
+use Psr\Container\ContainerExceptionInterface as PsrContainerException;
 
 /**
  * Container Exception
  */
-class ContainerException extends InvalidArgumentException implements InteropContainerException
+class ContainerException extends InvalidArgumentException implements PsrContainerException
```
Sam-Burns, line 17: Things like this unfortunately cannot be done in Slim 3, as it would break userland code catching `Interop\Container\Exception\ContainerException`.  It has already been done in Slim 4.

Sam-Burns (changes requested):
In general, bumping to Pimple 3.1 (or 3.2) would be a good idea, and having the unit tests expect PSR-11 exceptions wouldn't be a bad thing.  However, with regards to making the framework throw exceptions which are no longer implementations of `Interop\Container\Exception\ContainerException`, this would result in userland code catching exceptions ceasing to work as expected.  These changes have already been made for Slim 4, but I'm not sure we can safely do that in Slim 3.

Slim/Exception/ContainerException.php
1ma: Right, it didn't occur to me that someone might be catching these exceptions.

1ma:
I'm cool if these changes are already in the 4.x branch (didn't know, here's the commit for reference: https://github.com/slimphp/Slim/commit/f742f0f7f2194f00cc22318579dfe3f7450b1242 btw thanks :sweat_smile: )

Maybe we could still bump `pimple/pimple` to ^3.1 or ^3.2 and remove the explicit require of `psr/container` in 4.x?

Sam-Burns:
@1ma Working on the 4.x branch w.r.t. the container now.  Will bear that in mind.  I'll add you as a reviewer when there's a PR.

PR #2270, opened 2017-07-28
Error handlers should respect `outputBuffering` setting

### LINKED ISSUE #2208: Error handlers should respect `outputBuffering` setting
Given this code,

```php
$app->get("/", function ($request, $response, $args) {
    var_dump($request);
    throw new \Exception("oops");
});
```

I should see the var_dump'd information displayed if `outputBufferring` is set to `append` or `prepend`.

This requires changes to:

* `Route::__invoke()` to remove the `ob_end_clean()` calls.
* Pass the `outputBuffering` setting into `Slim\Handlers\AbstracError`'s constructor and store as a protected property
* Update `Slim\Hanlders\Error` to use `ob_get_clean()` and add it to the response in the correct place as per the `outputBuffering` setting.

RyanNerd:
* `Route::__invoke()` no longer calls `ob_get_clean()`
* slim\Handlers\AbstractError `__constructor` has output buffering as an argument which is saved to a protected property.
* `slim\Handlers\Error now handles the output buffering setting correctly.
* `DefaultServicesProvider` is updated to pass in the output buffering setting for the error service object.
* Two unit tests are added to test exception handling with output buffering on (append) and off.

Closes #2208

### DIFF
--- a/Slim/DefaultServicesProvider.php
+++ b/Slim/DefaultServicesProvider.php
@@ -152,7 +152,10 @@ class DefaultServicesProvider
              * @return callable
              */
             $container['errorHandler'] = function ($container) {
-                return new Error($container->get('settings')['displayErrorDetails']);
+                return new Error(
+                    $container->get('settings')['displayErrorDetails'],
+                    $container->get('settings')['outputBuffering']
+                );
             };
         }
 
--- a/Slim/Handlers/AbstractError.php
+++ b/Slim/Handlers/AbstractError.php
@@ -19,13 +19,19 @@ abstract class AbstractError extends AbstractHandler
     protected $displayErrorDetails;
 
     /**
+     * @var bool|string
+     */
+    protected $outputBuffering;
+
+    /**
      * Constructor
      *
      * @param bool $displayErrorDetails Set to true to display full details
      */
-    public function __construct($displayErrorDetails = false)
+    public function __construct($displayErrorDetails = false, $outputBuffering = false)
     {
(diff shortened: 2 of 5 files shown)


geggleto:
Hmm... only broken on HHVM ...

RyanNerd:
@geggleto At first I thought that HHVM doesn't support `var_dump()` (it does) but I suspect `var_dump()` may differ slightly in output. I'm going to change the tests to just do a simple `echo` and see if that fixes things.

geggleto:
🤕

RyanNerd:
Hmmmm... I wonder if HHVM output buffering is handled differently than PHP. (`ob_start()` must  be implicit in HHVM or something like that).
I installed HHVM side by side with PHP on my system at one time. I may try that again just so I can trace through the code to figure things out. This is a bit of a head scratcher.

RyanNerd:
I have tried a number of methods to get this to work with HHVM and I think this is actually [a bug in HHVM](https://github.com/facebook/hhvm/issues/7803)

RyanNerd:
How do I get coverage to pass? I thought that PHPUnit would do the trick. What am I missing?

akrabat:
The Coveralls stats are related to the whole codebase, so changing the amount of code in it affects the percentage covered.

Do your tests cover every path through the code you've changed? If so, that's good enough.

RyanNerd:
@akrabat Thanks. Yes the unit tests are very thorough.

akrabat:
Thanks.

PR #2280, opened 2017-08-04
Create absolute URL when using pathFor()

aimeos:
Absolute URLs are required for redirecting forth and back to payment providers for example. Only the path isn't enough


akrabat:
I think this is a BC break.

akrabat:
Needs an approach that doesn't break BC. 

We also need to consider what to do for 4.x

PR #2286, opened 2017-08-13
Track only the stack tip in MiddlewareAwareTrait

bnf:
As the stack is stored recursively in
the closure of the stack tip,
there is no need to hold an explicit reference
to all middlewares on the stack.

The \SplStack data structure was actually unused
in runtime code, it was only used to fetch the
tip of the stack and to check stack interity in tests.

That means those tests have to change, as the
middlewares now need to be executed, to be able to check whether
the stack has been built correctly.
An advantage of these changes is that the tests
now actually test the recursive closure stack structure, not
the unusued linked-list structure of \SplStack.

TODO: variable naming: Is "tip" too generic? But stackTip/stackTop sounds
clumsy..

### DIFF
--- a/Slim/MiddlewareAwareTrait.php
+++ b/Slim/MiddlewareAwareTrait.php
@@ -9,8 +9,6 @@
 namespace Slim;
 
 use RuntimeException;
-use SplStack;
-use SplDoublyLinkedList;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use UnexpectedValueException;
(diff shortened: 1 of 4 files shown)


akrabat:
This is an interesting catch. `$tip` is fine with me.

akrabat:
Any reason that this can't be applied to `3.x` ?

bnf:
I actually tested it with `3.x` as well (can be cherry-picked without conflicts) and it run's fine. But I think it could be considered as a BC break, for applications that use `MiddlewareAwareTrait`'s `$stack` property somehow. That's why I didn't even ask for merging into`3.x`.

akrabat:
Given this comment at the top of `MiddlwareAwareTrait`:

```
 * This is an internal class that enables concentric middleware layers. This
 * class is an implementation detail and is used only inside of the Slim
 * application; it is not visible to—and should not be used by—end users.
```

I'm inclined to put it into `3.x` too. Thoughts, @geggleto, @silentworks ?

geggleto:
I would like this to be cherry picked and put against 3.x

This is a great PR.

👍

akrabat:
Cherry picked to 3.x here: https://github.com/slimphp/Slim/commit/a08b779d2aa97621fb676409d76f0b414f7f616c

akrabat:
Thanks @bnf

PR #2287, opened 2017-08-13
Make OPTIONS rendering clearer

akrabat:
We render the OPTIONS response via Not Found error handling, so to make this clearer, rename the render method to `renderPlainOptionsMessage`.


geggleto:
Not to be a ... stickler but I see that there is no test for this method ;)

geggleto (approved):

akrabat:
@geggleto Test is [here](https://github.com/slimphp/Slim/blob/3.x/tests/Handlers/NotAllowedTest.php#L47-L57)

This is just a rename of the method to make it clearer what it does.

PR #2288, opened 2017-08-13
Separate routing from dispatching

akrabat:
Move the routing to middleware so that the user can add it wherever they want it to be within the middleware stack. This gives us the new capability to allow for some middleware to run before routing and some after.

If the `Routing` middleware isn't added by the user, then automatically call it in `App::__invoke()` as a last resort.

Also, remove `determineRouteBeforeAppMiddleware` setting as it's no longer necessary.

### DIFF
--- a/Slim/Middleware/Routing.php
+++ b/Slim/Middleware/Routing.php
@@ -1,0 +1,62 @@
+<?php
+namespace Slim\Middleware;
+
+use FastRoute\Dispatcher;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Slim\Interfaces\RouterInterface;
+
+/**
+ * Perform routing and store matched route to the request's attributes
+ */
+class Routing
+{
+    protected $router;
+
+    public function __construct(RouterInterface $router)
+    {
+        $this->router = $router;
+    }
+
+    /**
+     * Invoke
+     *
+     * @param  ServerRequestInterface $request   PSR7 server request
+     * @param  ResponseInterface      $response  PSR7 response
+     * @param  callable               $next      Middleware callable
+     * @return ResponseInterface                 PSR7 response
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $request = $this->doRouting($request);
+        return $next($request, $response);
+    }
+
+    /**
+     * Perform routing
+     *
+     * @param  ServerRequestInterface $request   PSR7 server request
+     * @return ServerRequestInterface
+     */
+    public function doRouting(ServerRequestInterface $request)
+    {
+        $routeInfo = $this->router->dispatch($request);
+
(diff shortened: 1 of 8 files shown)


Slim/Middleware/Routing.php
```diff
@@ -0,0 +1,62 @@
...
+
+use FastRoute\Dispatcher;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Slim\Interfaces\RouterInterface;
+
+/**
+ * Perform routing and store matched route to the request's attributes
+ */
+class Routing
```
geggleto: We need a new name for this.

Slim/Middleware/Routing.php
```diff
@@ -0,0 +1,62 @@
...
+        return $next($request, $response);
+    }
+
+    /**
+     * Perform routing
+     *
+     * @param  ServerRequestInterface $request   PSR7 server request
+     * @return ServerRequestInterface
+     */
+    public function doRouting(ServerRequestInterface $request)
```
geggleto: .............. 👎 you can... doBetter

geggleto (changes requested):
Also will this not only affect Slim currently when you use the invoke method. We would need to make more changes to affect run?

Slim/Middleware/Routing.php
silentworks: Whats wrong with this?

geggleto: doRouting ... surely we can come up with a better name :P

Slim/Middleware/Routing.php
akrabat: Any preference? `RoutingMiddleware` ?

Though if we postfix with `Middleware`, we'll have to change the other Middleware classes too.

Slim/Middleware/Routing.php
akrabat: `performRouting()` ?

Slim/Middleware/Routing.php
geggleto: I would prefer to have everything suffixed ...

Slim/Middleware/Routing.php
geggleto: `dispatchRequest` ?

akrabat: Can't be `dispatchRequest` as this code doesn't dispatch - it only performs the routing work and then we dispatch later.

Slim/Middleware/Routing.php
akrabat: Renamed to `RoutingMiddleware`

Slim/Middleware/Routing.php
silentworks: Maybe `prepare` or `perform` then?

geggleto (approved):

akrabat:
Rebased

akrabat:
Renamed `doRouting()` to `performRouting()`

PR #2289, opened 2017-08-13
Simplify container usage

Sam-Burns:
Pimple exceptions are now PSR-11 compliant.  This means that we don't need to wrap its exceptions in our own, to ensure we throw exceptions which implement this relevant interfaces.  This PR removes code which did that, while bumping the Pimple version.


akrabat:
Sorry, but this becomes moot with https://github.com/slimphp/Slim/pull/2290 …

PR #2291, opened 2017-08-15
Rename OutputBuffering to OutputBufferingMiddleware

campersau:
As suggested in https://github.com/slimphp/Slim/pull/2254#discussion_r133139725


akrabat:
Thanks.

PR #2298, opened 2017-08-24
Support HTTP "2" as a valid protocol version (i.e no minor version)

### LINKED ISSUE #2296: $validProtocolVersions should include "2" (without minor version)
See: https://http2.github.io/faq/#is-it-http20-or-http2

https://github.com/slimphp/Slim/blob/a641082cdc08769c6fe195180ed73a63ca828b73/Slim/Http/Message.php#L39

RyanNerd:
`ValidProtocolVersions` for HTTP "2" no longer include minor version.
@see https://http2.github.io/faq/#is-it-http20-or-http2
Closes #2296

### DIFF
--- a/Slim/Http/Message.php
+++ b/Slim/Http/Message.php
@@ -39,7 +39,7 @@ abstract class Message implements MessageInterface
     protected static $validProtocolVersions = [
         '1.0' => true,
         '1.1' => true,
-        '2.0' => true,
+        '2' => true,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


Slim/Http/Message.php
```diff
@@ -39,7 +39,7 @@
     protected static $validProtocolVersions = [
         '1.0' => true,
         '1.1' => true,
-        '2.0' => true,
```
danopz, line 42: This is a BC break as people may already use 2.0 as HTTP version, I would just have added 2 instead of removing the .0 part.

akrabat (dismissed):
Ad @danopz notes, we can't remove `'2.0'`, so please add `'2'` as an addition to list.

PR #2309, opened 2017-09-28
Send correct status code regardless of location header

### LINKED ISSUE #1730: Can't create 200 OK response with Location header
I'd like to respond `200 OK` with `Location` HTTP header, but what I'm getting is `302 Found` instead:

```
    return $response
            ->withHeader('Location', '/upload/' .basename($filePath))
            ->withStatus(200)
            ;
```

The same happens when I try to use `204 No Content` response.

Surprisingly `201 Created` works as expected. Can we make it work for 200 as well for clients that do not know how to handle 201, please?

kelunik:
PHP thinks it's clever, but it isn't. Fixes #1730.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -388,6 +388,16 @@ class App
     {
         // Send response
         if (!headers_sent()) {
+            // Headers
+            foreach ($response->getHeaders() as $name => $values) {
+                foreach ($values as $value) {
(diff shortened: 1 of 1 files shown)


akrabat:
Also merged to 4.x

akrabat:
Thanks @kelunik

Issue #2318, opened 2017-10-23
Problem mocking https requests (wrong port)


aurmil:
hello,

I don't know if I do something wrong or docs is wrong or bug

code:

```php
<?php

require 'vendor/autoload.php';

$baseUrl = parse_url('https://example.com');
$envOptions = [
    'REQUEST_METHOD' => 'GET',
    'SCRIPT_NAME' => '/index.php',
    'REQUEST_URI' => '/test',
    'HTTP_HOST' => $baseUrl['host']
];
if ('https' === strtolower($baseUrl['scheme'])) {
    $envOptions['HTTPS'] = 'on';
    // $envOptions['SERVER_PORT'] = '443';
}
$env = Slim\Http\Environment::mock($envOptions);

$app = new Slim\App(['environment' => $env]);

$app->get('/test', function ($request, $response) {
    $baseUrl = $request->getUri()->getBaseUrl();
    return $response->write($baseUrl);
});

$app->run();
```

output: `https://example.com:80`

I have to uncomment line `$envOptions['SERVER_PORT'] = '443';` to get it right: `https://example.com`

[docs](https://www.slimframework.com/docs/cookbook/environment.html):

> SERVER_PORT
The port on the server machine being used by the web server for communication. For default setups, this will be ‘80’; using SSL, for instance, will change this to whatever your defined secure HTTP port is.
HTTPS
Set to a non-empty value if the script was queried through the HTTPS protocol.

tflight:
When the environment is mocked, the `SERVER_PORT` [defaults to 80](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Environment.php#L39) and then array_merge is run with the mocked environment options. So if `SERVER_PORT` is **not** set when mocking the environment, port 80 will always be set regardless of scheme/protocol.

Looks like the documentation is incorrect, or at least misleading. The documentation can be fixed, or we could change the way mock works to check if `HTTPS` is not null and `SERVER_PORT` is null... and if so default to port 443 instead of 80... or something like that.

aurmil:
thanks, did not say more as I did not know if code or docs was to be changed

PR #2323, opened 2017-10-31
Add 443 as SERVER_PORT(default) when request is https

juanpagfe:
Hi.  I solved the #2318 issue. Just added a validation wich changes the default port when request is HTTPS to 443.


Slim/Http/Environment.php
```diff
@@ -29,14 +29,19 @@ class Environment extends Collection implements EnvironmentInterface
      */
     public static function mock(array $userData = [])
     {
+        //Validates if default protocol is HTTPS to set default port 443
+        $defport = (isset($userData["HTTPS"]) ||
+                    (isset($userData["SERVER_PROTOCOL"]) && substr($userData["SERVER_PROTOCOL"], 0, 5) === "HTTPS"))
+                      ? 443 : 80;
```
akrabat: Please use an `if` statement.

Slim/Http/Environment.php
```diff
@@ -29,14 +29,19 @@ class Environment extends Collection implements EnvironmentInterface
      */
     public static function mock(array $userData = [])
     {
+        //Validates if default protocol is HTTPS to set default port 443
+        $defport = (isset($userData["HTTPS"]) ||
```
akrabat: `["HTTPS"]` may contain the string `off`, which should use port 80.

Slim/Http/Environment.php
```diff
@@ -29,14 +29,19 @@ class Environment extends Collection implements EnvironmentInterface
...
+                    (isset($userData["SERVER_PROTOCOL"]) && substr($userData["SERVER_PROTOCOL"], 0, 5) === "HTTPS"))
```
akrabat: [`SERVER_PROTOCOL`] will contain `HTTP/1.0`, `HTTP/1.1` or `HTTP/2`. It will never contain `HTTPS` and so please do not check this key.

Slim/Http/Environment.php
akrabat: Use single quoted strings if there's no variable interpolation.

akrabat (changes requested):
Some changes required as noted.

Also, please supply a unit test so that we don't regress your fix.

mahagr:
Why not use `REQUEST_SCHEME` which is either `http` or `https`?

akrabat:
> Why not use `REQUEST_SCHEME` which is either `http` or `https`?

No reason that it also couldn't be checked. I hadn't heard of it, so didn't mention it.

Slim/Http/Environment.php
```diff
@@ -29,14 +29,22 @@ class Environment extends Collection implements EnvironmentInterface
...
+        if (isset($userData['HTTPS']) && $userData['HTTPS'] !== 'off') {
+            $defport = 443;
+        } else {
+            $defport = 80;
+        }
+
         $data = array_merge([
             'SERVER_PROTOCOL'      => 'HTTP/1.1',
             'REQUEST_METHOD'       => 'GET',
+            'REQUEST_SCHEME'       => 'http',
```
akrabat: If you are going to put `REQUEST_SCHEME` here, then it needs to be either `http` or `https` … 

Similarly, if you're going to support it here, you should test for it in $userData for setting the port.

phpunit.xml.dist
```diff
@@ -12,7 +12,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
+         stopOnFailure="true"
```
akrabat: Please remove this change.

akrabat (changes requested):

akrabat (approved):

akrabat:
Thank you @juanpagfe  :)

akrabat:
@juanpagfe Fancy creating a similar PR for https://github.com/slimphp/Slim-Http ?

mahagr:
@akrabat Any change making the same change into Uri class:

https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Uri.php#L169-L170

Mostly because of HTTPS isn't always available.

akrabat:
@mahagr Which web server doesn’t set `HTTPS` when the connection is over TLS?

mahagr:
See discussion in:

https://github.com/getgrav/grav/commit/a9c82710a6f33b636a84cc42e4d0b1964d7c1483
https://github.com/getgrav/grav/issues/1698

juanpagfe:
@akrabat Of course, I'll create a PR in a minute.

Issue #2324, opened 2017-11-02
Add a code of conduct


akrabat:
Personally, I like and have used the [Contributor Covenant](https://www.contributor-covenant.org) on other projects.

Does anyone disagree or have another preference?

akrabat:
The main advantage I've seen has been clarity of expectations. Anecdotally, on other projects, we've had devs say to us privately that they only considered submitting a PR because the CoC gave them the confidence that they wouldn't be attacked in the comments.

geggleto:
Do we need it? I have felt that our community has been incredibly good in this aspect.

geggleto:
Also; do we as maintainers have the resources to monitor and enforce a CoC?

juanpagfe:
@geggleto I think it is always needed.  Even if it has not been needed.  Better to have it and don't need it than need it and don't have it.

akrabat:
Discussion on Slack suggests that basing on https://www.ruby-lang.org/en/conduct/ / https://github.com/laravel/framework/blob/5.5/CODE_OF_CONDUCT.md may work for our community.

i.e. something like:

> This document provides community guidelines for a safe, respectful, productive, and collaborative place for any person who is willing to contribute to the Slim community. It applies to all “collaborative space”, which is defined as community communications channels (such as mailing lists, submitted patches, commit comments, etc.). Any violations of the code of conduct may be reported by contacting one or more of the project maintainers.
>
> * Participants will be tolerant of opposing views.
> * Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.
> * When interpreting the words and actions of others, participants should always assume good intentions.
> * Behaviour which can be reasonably considered harassment will not be tolerated.

chris48s:
> Do we need it? I have felt that our community has been incredibly good in this aspect.

@geggleto One of the findings from GitHub's [open source survey](http://opensourcesurvey.org/2017/#insights) was that

> "Documentation that clearly explains a project's processes, such as contributing guides and codes of conduct, is valued more by groups that are underrepresented in open source, like women."

While it is amazing that the existing Slim community is fortunate enough not to have issues with poor behaviour which need to be addressed, adopting a code of conduct may remove a barrier to some potential contributors which you may not be aware is a problem for them.

damianopetrungaro:
@akrabat 

[...]

PR #2326, opened 2017-11-03
Add a Code of Conduct

akrabat:
See #2324

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,6 +1,6 @@
 # Code of Conduct
 
-The Slim Framework code of conduct is derived from the Ruby code of conduct. This document provides community guidelines for a safe, respectful, productive, and collaborative place for any person who is willing to contribute to the Slim Framework community. It applies to all “collaborative space”, which is defined as community communications channels (such as mailing lists, IRC, Slack, forums, submitted patches, commit comments, etc.). Any violations of the code of conduct may be reported by contacting one or more of the project maintainers.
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,8 @@
+# Code of Conduct
+
+The Slim Framework code of conduct is derived from the Ruby code of conduct. This document provides community guidelines for a safe, respectful, productive, and collaborative place for any person who is willing to contribute to the Slim Framework community. It applies to all “collaborative space”, which is defined as community communications channels (such as mailing lists, IRC, Slack, forums, submitted patches, commit comments, etc.). Any violations of the code of conduct may be reported by contacting one or more of the project maintainers.
```
geggleto: I think we should probably be a bit more specific about how this is possible.

What about adding all of us to the Authors list in composer.json?

akrabat: Let's get a generic email address set up.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,8 @@
...
+
+* Participants will be tolerant of opposing views.
+* Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.
+* When interpreting the words and actions of others, participants should always assume good intentions.
+* Behaviour which can be reasonably considered harassment will not be tolerated.
```
codeguy: What does "reasonably" mean? That's really the only thing I see that will likely prompt push back. Otherwise this looks great.

geggleto: that's a great question; I think reasonably will end up being what we (maintainers) believe is the correct thing under the context...

Minimally, I think what this document is saying is: `Be Nice, Courteous and Understanding`

geggleto (approved):

silentworks (approved):

CODE_OF_CONDUCT.md
silentworks: I think what @geggleto suggested is probably better to use as `reasonably` is subjective.

akrabat:
Updated wording of last point. Is that better @silentworks, @geggleto, @codeguy?

geggleto:
👍 LGTM @akrabat

PR #2327, opened 2017-11-03
Encode user info

### LINKED ISSUE #2201: Uri::getBaseUrl() should escape @ in password

akrabat:
If the username or password includes an `@`, `:` or other reserved
characters, they need to be encoded.

Fixes #2201

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -378,10 +378,29 @@ class Uri implements UriInterface
     public function withUserInfo($user, $password = null)
     {
         $clone = clone $this;
-        $clone->user = $user;
-        $clone->password = $password ? $password : '';
+        $clone->user = $this->filterUserInfo($user);
+        if ($clone->user) {
+            $clone->password = $password ? $this->filterUserInfo($password) : '';
+        }
 
         return $clone;
+    }
+
+    /**
+     * Filters the user info string.
+     *
+     * @param string $query The raw uri query string.
+     * @return string The percent-encoded query string.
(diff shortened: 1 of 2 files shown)


silentworks (approved):

mahagr:
Just a note: `filterUserInfo()` has `/u` modifier, but `filterPath()` and `filterQuery()` do not. Should they all have the modifier?

akrabat:
@mahagr Now that is a very good question :)

I think that they should all have `/u` as I think unicode is allowed in URIs. (e.g http://☃.net)

mahagr:
Also another thing... Should `__construct()` also escape user and password?

mahagr:
And from your example: hostname isn't filtered either...

akrabat:
Yes - I would appreciate a PR :)

akrabat:
Re hostname: is there any reason to filter it?

mahagr:
You can use Uri class outside of request to build a new URL to be shown your page.

PR #2328, opened 2017-11-04
* is missed

ares333:
I checked rfc3986

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -640,7 +640,7 @@ class Uri implements UriInterface
     protected function filterPath($path)
     {
         return preg_replace_callback(
-            '/(?:[^a-zA-Z0-9_\-\.~:@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/',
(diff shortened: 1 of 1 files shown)


vlakoff:
https://www.ietf.org/rfc/rfc3986.txt [Page 53]:

> The mark characters that are typically unsafe to decode,
> including the exclamation mark ("!"), asterisk ("*"), single-quote
> ("'"), and open and close parentheses ("(" and ")"), have been moved
> to the reserved set in order to clarify the distinction between
> reserved and unreserved and, hopefully, to answer the most common
> question of scheme designers.

akrabat:
Why did you close?

ares333:
I closed it because just one character was included in the patch and I didn't check the rfc very carefully.So I am not sure about it.😓

vlakoff:
Note it's a negated character class. The characters it contains are *not* escaped.

Because the other characters – `!` `'` `(` `)` – from my quote above are also omitted, I guess the omission of `*` is on purpose.

akrabat:
Thanks for noticing @ares333 & thanks @vlakoff for checking the regex. Seems that we're all good.

PR #2332, opened 2017-11-06
Fixed regression, removing user info with Uri::withUserInfo('') is not clearing password anymore

mahagr:
Unfortunately, Slim 3.9 broke redirects with user authority as the password isn't cleared when `Uri::withUserInfo('')`  is called.

Url: `user:password@domain.com/path` became: `Location: //:password@/path`.


akrabat:
Good catch. Can you provide a test too?

geggleto:
@mahagr we need a test so we can merge this plz :)

mahagr:
Sorry, I've been super busy lately. Thanks @splitbrain for the test!

PR #2333, opened 2017-11-07
Avoid deleting output buffer to fix 14 tests that PHPUnit have considered risky

dmelo:
When my projects updated from 3.8.1 to 3.9.0, many of my phpunit tests started resulting into risky tests. It turned out that reverting back to 3.8.1 caused phpunit to stop classifying the tests as risky. Running phpunit on Slim, I also saw that 3.8.1 had only one risky test while 3.9.0 have 18 risky tests.

The 17 new risky tests have the message: "Test code or tested code did not (only) close its own output buffers".

This commit uses `ob_get_contents` and `ob_clean` instead of `ob_get_clean` to avoid destroying the current output buffer.

All the tests are still passing but now, only 3 of the 17 tests are classified as Risky due to  "Test code or tested code did not (only) close its own output buffers".

Here is the lisf of risky tests on 3.9.0

```
dmelo@merov2:~/proj2/Slim$ git checkout 3.9.0
Previous HEAD position was 53853027... Uppdate App:VERSION to 3.8.1
HEAD is now at 575a8b53... Version 3.9.0
dmelo@merov2:~/proj2/Slim$ ./vendor/bin/phpunit --stderr --verbose
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

Runtime:	PHP 7.1.9 with Xdebug 2.5.1
Configuration:	/home/dmelo/proj2/Slim/phpunit.xml.dist

...............................................................  63 / 588 ( 10%)
..................R.........R.............SR................... 126 / 588 ( 21%)
.............................RRRRRRRRRRRR.R.................... 189 / 588 ( 32%)
.........SSSSSSSSSSSSS......................................... 252 / 588 ( 42%)
............................................................... 315 / 588 ( 53%)
............................................................... 378 / 588 ( 64%)
............................................................... 441 / 588 ( 75%)
............................R.................................. 504 / 588 ( 85%)
........................................................R...... 567 / 588 ( 96%)
.....................

Time: 973 ms, Memory: 16.00MB

There were 18 risky tests:

1) Slim\Tests\AppTest::testExceptionErrorHandlerDoesNotDisplayErrorDetails
Test code or tested code did not (only) close its own output buffers

/home/dmelo/proj2/Slim/vendor/phpunit/phpunit/phpunit:52

2) Slim\Tests\AppTest::testExceptionErrorHandlerDisplaysErrorDetails
Test code or tested code did not (only) close its own output buffers

[...]

PR #2337, opened 2017-11-09
Fix output buffering issue that was creating a risky test in PHPUnit.

Rezouce:
Same problem as here but at the Route level:
https://github.com/slimphp/Slim/pull/2333

It fixes the risky test for `RouteTest::testInvokeWithException`.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -292,10 +292,28 @@ class App
         $response = $this->container->get('response');
 
         try {
+            ob_start();
             $response = $this->process($this->container->get('request'), $response);
         } catch (InvalidMethodException $e) {
             $response = $this->processInvalidMethod($e->getRequest(), $response);
+        } finally {
+            $output = ob_get_clean();
         }
+
+        if (!empty($output) && $response->getBody()->isWritable()) {
+            $outputBuffering = $this->container->get('settings')['outputBuffering'];
+            if ($outputBuffering === 'prepend') {
+                // prepend output buffer content
+                $body = new Http\Body(fopen('php://temp', 'r+'));
+                $body->write($output . $response->getBody());
+                $response = $response->withBody($body);
+            } elseif ($outputBuffering === 'append') {
+                // append output buffer content
+                $response->getBody()->write($output);
+            }
+        }
+
+        $response = $this->finalize($response);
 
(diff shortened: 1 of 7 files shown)


moritz-h:
@Rezouce Would it make sense to also use `ob_get_contents()` instead of `ob_get_clean()` here: https://github.com/slimphp/Slim/blob/3.9.0/Slim/Handlers/Error.php#L61-L64
and remove it here: https://github.com/slimphp/Slim/blob/3.9.0/Slim/Handlers/Error.php#L68
Otherwise I think in the finally block the buffer would be destroyed a second time in exception case.

akrabat:
Travis test fails.

Rezouce:
@moritz-h you are right, I didn't see that the output buffering management was duplicated in the error handler.
Instead of duplicating the logic in both the route and the error handler, I chose to move the output buffering management at the app level instead.

akrabat:
I'm wondering if anyone depends on output buffering being at route level.

silentworks (approved):

RyanNerd (changes requested):
In `Slim/Slim/Handlers/AbstractError.php` in the constructor the second argument is [`$outputBuffering`](https://github.com/slimphp/Slim/blob/c9a768c5a062c5f1aaa0a588d7bb90e8ce18bfd6/Slim/Handlers/AbstractError.php#L31) since this is being moved to the App level this class should no longer need this argument.
Also`PhpError.php` which extends `AbstractError` should be looked at to see if it is using the `$outputBuffering` setting.

akrabat:
Do we break anything if we remove a parameter to a constructor?

moritz-h:
The `$outputBuffering` constructor parameter was just recently added with Slim 3.9.0. All usages within the handler are also removed with this PR.
But on calling the constructor the parameter should then also be removed:
https://github.com/slimphp/Slim/blob/3.9.0/Slim/DefaultServicesProvider.php#L157

Rezouce:
I removed the unnecessary output buffering from the AbstractError, thank you @RyanNerd and @moritz-h for your inputs.
@akrabat I don't think removing this extra parameter will provoke any breaking change. Some people may continue to provide it but that shouldn't be a problem.

PR #2341, opened 2017-11-15
Add code to allow the CallableResolver to resolve constructor dependencies.

TomWright:
This PR makes it nice and easy to ensure a Controller receives all of it's required dependencies without having to interact with the container itself.

See how `tests/Mocks/ResolvingCallableDependencies/MultipleDependencyTest.php` interacts with the `MockDependency` in the modified files below.
Without this change the following would have been required to get an instance of that dependency.

    public function __construct(ContainerInterface $container = null)
    {
        $this->dependency = $container->get(MockDependency::class);
    }

This functionality means the controllers will no longer need direct access to the Container instance, which in itself isn't a problem, it can become annoying when your dependencies start building up.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -77,8 +77,6 @@ final class CallableResolver implements CallableResolverInterface
      * @param string $class
      * @param string $method
      * @return callable
-     *
-     * @throws \RuntimeException if the callable does not exist
      */
     protected function resolveCallable($class, $method = '__invoke')
     {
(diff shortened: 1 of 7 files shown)


geggleto:
Hey Tom, While I appreciate the amount of work that you have put into this Pull Request, this feature is a non-starter. We couldn't merge this PR as it would be a breaking change. Furthermore we are removing the requirement of having a container for the next major release of the framework.

geggleto:
Furthermore you can use PHP-DI's container if you wish to have auto-wiring!

TomWright:
Thanks for the quick response.
I had only ran phpunit locally so wasn't aware of the phpcs formatting errors yet. Other than those though I had attempted to make it backwards compatible.

I appreciate that you can get the auto-wiring functionality by using PHP-DI but thought it may be useful to have similar functionality out of the box. If you don't agree then that's fine also and I'll stick to using PHP-DI!

Thanks for your time.

PR #2342, opened 2017-11-15
Regex simplification in CallableResolver

### LINKED ISSUE #1960: Regex simplification
Just because...

In the callable resolver, there is a regex that matches a callable string. This is based on that which is in the PHP.net documentation.

The aim of the regex is to validate the latter part of a callable string and split it into class and method parts.

As the callable resolver is not required to cover validation of a class/method name (as PHP won't allow invalid class/method names and would error long before the regex is processed), I propose we simplify the regex to remove the superfluous computation. I also propose we use correct start and end qualifiers for a string, as we are only interested in single line strings, and not multiline as defined by our use of `^$`.

```
'!\A([^\:]+)\:([^\Z]+)\Z!'
```

Another reason to choose this, is that by using the current regex, we should have implemented tests to prove it's validity by providing class and method names to test all of it's possible outcomes. This is a silly waste of time as we don't need to test validity with the regex.

I'm open to comment on this, and discussion.

NB: This regex still passes all current tests.

juanpagfe:
Fix #1960 issue.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -18,7 +18,7 @@ use Slim\Interfaces\CallableResolverInterface;
  */
 final class CallableResolver implements CallableResolverInterface
 {
-    const CALLABLE_PATTERN = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
+    const CALLABLE_PATTERN = '!\A([^\:]+)\:([^\Z]+)\Z!';
 
     /**
      * @var ContainerInterface


vlakoff:
I tickled on the `[^\Z]` and indeed, "Anchors cannot be used inside character classes".

vlakoff:
Also,
* there is no need to escape ":"
* `\A` `\z` don't seem useful (btw, rather `\z` than `\Z`)
* I haven't checked thoroughly, but your changes match on two colons (e.g. "Foo::bar") so maybe it broke regular callbacks, which are handled just after in the code.

To summarize, this could do the job: `^([^:]+):([^:]+)$`

vlakoff:
@akrabat What about the issues I pointed above? Namely, the incorrect class `[^\Z]` and the false positive on "Foo::bar".

akrabat:
@vlakoff: Doh! Somehow I missed it :(

akrabat:
Reverted in 3.9.2

PR #2345, opened 2017-11-18
Set status to 302 if Location header set

### LINKED ISSUE #2344: Redirects no longer work in 3.9
Hello, I am able to successfully use this code in 3.8.1:

```
$app->get('/', function (Request $request, Response $response) {
	if (UserState::isLoggedIn()) {
		return $response->withHeader('Location', '/dashboard');
	}

	return $response->withHeader('Location', '/login');
})->setName('home');
```

However in 3.9 a white screen appears with no output (I do have displayErrorDetails set to true). There are no errors in my PHP error log and no errors in Apache, either.

The only way I can get a redirect to work is to replace the return lines above with: `header("Location: /dashboard"); exit;` but I much prefer to use the tools provided by the framework.

Thank you for reading!

Apologies if this has already been raised; I did not see it in the issue list.

akrabat:
If the Location header is set, then set to the status code to 302 if
it is 200. This ensures that code like this still works:

    return $response->withHeader('Location', '/login');

However if the status code is set to something other than 200, when we
assume intent by the user, so do not override their code. e.g. this
code works does not change the status code from 201:

    return $response->withStatus(201)->withHeader('Location', '/item/1');

Closes #2344

### DIFF
--- a/Slim/Http/Response.php
+++ b/Slim/Http/Response.php
@@ -250,6 +250,34 @@ class Response extends Message implements ResponseInterface
     }
 
     /*******************************************************************************
+     * Headers
+     ******************************************************************************/
+
+    /**
+     * Return an instance with the provided value replacing the specified header.
+     *
+     * If a Location header is set and the status code is 200, then set the status
+     * code to 302 to mimic what PHP does. See https://github.com/slimphp/Slim/issues/1730
+     *
+     * @param string $name Case-insensitive header field name.
+     * @param string|string[] $value Header value(s).
+     * @return static
+     * @throws \InvalidArgumentException for invalid header names or values.
+     */
+    public function withHeader($name, $value)
+    {
+        $clone = clone $this;
+        $clone->headers->set($name, $value);
+
+        if (200 == $clone->getStatusCode() && 'location' == strtolower($name)) {
+            $clone = $clone->withStatus(302);
+        }
+
+        return $clone;
+    }
+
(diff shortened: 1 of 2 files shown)


nickdnk:
May I suggest using === for comparison:

```
if (200 === $clone->getStatusCode() && 'location' === strtolower($name)) {
    $clone = $clone->withStatus(302);
}
```

akrabat:
> May I suggest using === for comparison:

Done.

Slim/Http/Response.php
```diff
@@ -249,6 +249,34 @@ public function getReasonPhrase()
...
+     * @param string|string[] $value Header value(s).
+     * @return static
+     * @throws \InvalidArgumentException for invalid header names or values.
+     */
+    public function withHeader($name, $value)
+    {
+        $clone = clone $this;
+        $clone->headers->set($name, $value);
+
+        if (200 === $clone->getStatusCode() && 'location' === strtolower($name)) {
```
silentworks: Please don't Yoda the codebase. Lets keep it consistent with everywhere else in the codebase.

silentworks (changes requested):

Slim/Http/Response.php
akrabat: Done. We should put in a phpcs sniff to prevent it!

silentworks (approved):

PR #2357, opened 2017-11-30
Fix minor issue in SmallChunksStream mock [Dev]

### LINKED ISSUE #2281: tests/Mocks/SmallChunksStream write() function returns the wrong type
https://github.com/slimphp/Slim/blob/3.x/tests/Mocks/SmallChunksStream.php#L104

This function is echoing back the string argument as a string when it should probably be returning an int of `strlen($string);`

See: https://github.com/php-fig/http-message/blob/master/src/StreamInterface.php#L112

RyanNerd:
Closes #2281

### DIFF
--- a/tests/Mocks/SmallChunksStream.php
+++ b/tests/Mocks/SmallChunksStream.php
@@ -103,6 +103,6 @@ class SmallChunksStream implements StreamInterface
 
     public function write($string)
     {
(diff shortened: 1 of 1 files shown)


akrabat:
Thanks. Do we need this for Slim 4.x too?

PR #2360, opened 2017-12-02
Fix example in README

mnapoli:
Am I missing something or is the example wrong? I think it's a PSR-7 response so the `write()` method exists on the body only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ require 'vendor/autoload.php';
 $app = new Slim\App();
 
 $app->get('/hello/{name}', function ($request, $response, $args) {
-    return $response->write("Hello, " . $args['name']);
+    return $response->getBody()->write("Hello, " . $args['name']);
 });
 
 $app->run();


tflight:
Slim's response object does have a convenience method for `write`, [seen here](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Response.php#L280). So the example does work, but doesn't conform to PSR-7. Therefore you couldn't swap out Slim's response with another PSR-7 response with the existing example and your example would be needed.

mnapoli:
Mmh OK, I was confused since this is also not consistent with https://www.slimframework.com/

Anyway thanks!

tflight:
I don't disagree with you though. We probably should be using PSR-7 compatible examples whenever possible and at least those two examples should be consistent.

tflight:
@mnapoli reading my response again, it might not be clear what i was trying to say. I think your PR is valid and is better than the existing example.

mnapoli:
Reopening in case you want to merge it, feel free to close it again if it's not necessary :)

akrabat:
Thanks @mnapoli

PR #2366, opened 2017-12-08
Suppress deprecation warnings for mcrypt on PHP 7.1

### LINKED ISSUE #2362: mcrypt_module_open is deprecated - PHP 7.1.11
Hi, I just upgraded my PHP version from 5.6 to 7.1.11, and slim throws an error:

`Function mcrypt_module_open() is deprecated in D:\Project\vendor\slim\slim\Slim\Http\Util.php:106`

I updated slim to 2.6.3 version, based on this link (https://www.slimframework.com/2017/01/07/slim-2.6.3.html), it says:

> Slim 2.6.3 handles the deprecation of mcrypt in PHP 7.1

But didn't worked either. I also tried to install mcrypt_compat (https://github.com/phpseclib/mcrypt_compat) but error is still there.

Any ideas how to fix it?

Thank you.

rechl:
Fixes #2362

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -106,21 +106,21 @@ class Util
         $module = @mcrypt_module_open($settings['algorithm'], '', $settings['mode'], '');
 
         //Validate IV
-        $ivSize = mcrypt_enc_get_iv_size($module);
+        $ivSize = @mcrypt_enc_get_iv_size($module);
         if (strlen($iv) > $ivSize) {
             $iv = substr($iv, 0, $ivSize);
         }
 
         //Validate key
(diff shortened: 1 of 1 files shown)


geggleto:
I don't think using `@` is an acceptable solution. This allows things to fail silently which is bad.

vlakoff:
On a related note, I love how the function just returns the unencrypted data if extension mcrypt isn't loaded. This could be pretty fun if one moves its application to a different server.

akrabat:
Silence operator is the wrong solution here.

We should probably put an end of life note on Slim 2.x's README too.

PR #2369, opened 2017-12-14
Test against PHP 7.2 [Dev]

danielmadu:


geggleto:
before I merge this; I want to discuss dropping hhvm with the other maintainers.

PR #2379, opened 2018-01-10
Add support for PSR-15 middlewares

### LINKED ISSUE #2050: Update middleware signature for PSR-15

bnf:
PSR-15 middlewares can be registered using the existing API:
`App::add()` and `Routable::add()`
Those methods are extended to accept objects that implement
the MiddlewareInterface (or class names that refer to classes
which implement the MiddlewareInterface).

We do not add new entry points for middleware registration, as the type
of the middleware can be detecting through the MiddlewareInterface
which all PSR-15 middlewares implement.

Both, classes that do not implement MiddlewareInterface, and closures will
be handled as single pass middleware – as before.
That means there's no support for closures with a PSR-15 single-pass
signature.

The detection of the middleware type is implemented
in the callable resolver. While that might look like
something which is out of scope for the CallableResolver, reasons are:

1. The internal slim middleware stack stays at is – it handles callables only.
    (and in cannot be changed anyway as long double-pass middlewares
    need to be supported. Double-pass middlewares expect a Response object
    that needs to be passed along the middleware stack. Therefore we'd need to
    wrap PSR-15 middleware's in any case – which is what this change provides)
2. the CallableResolver resolves (in terms for transforms) a callable
    from PSR-15 style middleware.

TODO: Add more tests for:

- [ ] Adapter\SinglePassMiddleware
- [ ] Routable with PSR-15 Middlewares

Resolves #2050

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -54,7 +54,8 @@ final class CallableResolver implements CallableResolverInterface
 
         if (!is_callable($toResolve) && is_string($toResolve)) {
             $class = $toResolve;
-            $method = '__invoke';
+            $instance = null;
+            $method = null;
 
             // check for slim callable as "class:method"
             $callablePattern = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
@@ -64,18 +65,21 @@ final class CallableResolver implements CallableResolverInterface
             }
 
             if ($this->container instanceof ContainerInterface && $this->container->has($class)) {
-                $resolved = [$this->container->get($class), $method];
+                $instance = $this->container->get($class);
             } else {
                 if (!class_exists($class)) {
                     throw new RuntimeException(sprintf('Callable %s does not exist', $class));
                 }
-                $resolved = [new $class($this->container), $method];
+                $instance = new $class($this->container);
             }
 
             // For a class that implements RequestHandlerInterface, we will call handle()
-            if ($resolved[0] instanceof RequestHandlerInterface) {
-                $resolved[1] = 'handle';
+            // if no method has been specified
+            if ($instance instanceof RequestHandlerInterface && $method === null) {
+                $method = 'handle';
             }
+
+            $resolved = [$instance, $method ?? '__invoke'];
         }
 
         if ($resolved instanceof RequestHandlerInterface) {
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -36,7 +36,7 @@ final class CallableResolver implements CallableResolverInterface
     }
 
     /**
-     * Resolve toResolve into a closure that that the router can dispatch.
+     * Resolve toResolve into a callable that that the router can dispatch.
      *
      * If toResolve is of the format 'class:method', then try to extract 'class'
      * from the container otherwise instantiate it and then dispatch 'method'.
@@ -50,9 +50,19 @@ final class CallableResolver implements CallableResolverInterface
      */
     public function resolve($toResolve): callable
     {
-        $resolved = $toResolve;
+        if ($toResolve instanceof RequestHandlerInterface) {
+            return [$toResolve, 'handle'];
+        }
 
-        if (!is_callable($toResolve) && is_string($toResolve)) {
+        if (is_callable($toResolve)) {
+            if ($toResolve instanceof \Closure && $this->container instanceof ContainerInterface) {
+                return $toResolve->bindTo($this->container);
+            }
+            return $toResolve;
+        }
+
+        $resolved = $toResolve;
+        if (is_string($toResolve)) {
             $class = $toResolve;
             $instance = null;
             $method = null;
@@ -76,14 +86,10 @@ final class CallableResolver implements CallableResolverInterface
             // For a class that implements RequestHandlerInterface, we will call handle()
             // if no method has been specified
             if ($instance instanceof RequestHandlerInterface && $method === null) {
-                $method = 'handle';
+                return [$instance, 'handle'];
             }
 
             $resolved = [$instance, $method ?? '__invoke'];
-        }
-
-        if ($resolved instanceof RequestHandlerInterface) {
-            $resolved = [$resolved, 'handle'];
         }
 
         if (!is_callable($resolved)) {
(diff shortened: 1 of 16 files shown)


geggleto:
Thanks :) I will take a good look at this today or tomorrow, I know akrabat is at a conference this week so he might be able to get around to it next.

Slim/Adapter/SinglePassMiddleware.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->next = $next;
+
+        /* Call the PSR-15 middleware and let it return to our handle()
+         * method by passing `$this` as RequestHandler. */
+        return $this->middleware->process($request, $this);
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        return call_user_func($this->next, $request, $this->response);
```
weierophinney: Since `$this->next` is `callable, and this PR requires PHP 7 and up, that means you can simplify the above to:

```php
return ($this->next)($request, $this->response);
```

My understanding is that `call_user_func` is slower than PHP's native dereferencing in PHP 7, so the above will also be more performant.

Slim/CallableResolver.php
```diff
@@ -8,8 +8,10 @@
  */
 namespace Slim;
 
+use Interop\Http\Server\MiddlewareInterface;
```
weierophinney: Just a heads-up: PSR-15 passed today, though we're waiting to announce it until the packages are on Packagist — though that should be within the next few hours after I write this. As such, you'll be able to require `psr/http-server-middleware` instead of `http-interop/http-server-middleware`, and the above will become `Psr\Http\Server\MiddlewareInterface`.

Slim/Adapter/SinglePassMiddleware.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This class wraps a PSR-15 style single pass middleware,
+ * into an invokable double pass middleware.
+ *
+ * This is an internal class. This class is an implementation detail
+ * and is used only inside of the Slim  application; it is not visible
+ * to—and should not be used by—end users.
+ */
+final class SinglePassMiddleware implements RequestHandlerInterface
```
weierophinney: Instead of `SinglePassMiddleware`, may I suggest `StandardsBasedMiddleware` or `PsrMiddleware` here?

I've noted in communicating with ZF folks that many are not clear what is meant by "single-pass" and "double-pass", despite us having documented it. Having a name that indicates the type of middleware in terms of where you might find a definition helps.

Also, I'd add an `@see` annotation with a link to the accepted standard (we'll have that on the fig-standards repo shortly).

Slim/Routable.php
```diff
@@ -87,13 +87,13 @@ public function getCallableResolver()
     /**
      * Prepend middleware to the middleware collection
      *
-     * @param callable|string $callable The callback routine
+     * @param MiddlewareInterface|callable|string $middleware The middleware routine
```
weierophinney, line 94: Have you imported this interface within this class file?

Slim/Adapter/SinglePassMiddleware.php
bnf: Well, I personally prefer a name that describes the functionality. With `PsrMiddleware` it's not that clear whether the middleware is implementing PSR 15 or (as in our case) adapts a PSR-15 middleware.
Anyway, I've no strong objections against `PsrMiddleware` so i'll push a patch that renames that one (consistent naming across frameworks is probably not that bad).

Slim/Adapter/SinglePassMiddleware.php
bnf: I used `call_user_func` since that was used throughout the Slim codebase. I think those occurrences should be changed in one go. Will push a patch for that.

Slim/Routable.php
bnf: I missed that one, will fix.

Slim/CallableResolver.php
bnf: Yes, I pushed a patch for the switch to `psr/http-server-middleware`.

.travis.yml
```diff
@@ -6,18 +6,20 @@ dist: trusty
...
     - php: 7.0
     - php: 7.1
       env: ANALYSIS='true'
     - php: hhvm
-    - php: 5.6
+    - php: 7.0
       env: COMPOSER_ARGS='--prefer-lowest'
     - php: nightly
   allow_failures:
+    - php: hhvm
```
akrabat: Let's just take hhvm out - we're not supporting it in 4.x

akrabat:
I've been thinking about this a bit. While I appreciate that it's easier to keep the internals the same as Slim 3, I'd like to explore if we can make PSR-15 the default internally and "wrap" the current signatures. Going forward, this makes more sense to me as we probably won't support the current signatures in Slim 5.

Thoughts?

l0gicgate:
@bnf are you moving forward with this PR? I'm curious as I will tackle it after my next couple of PRs if you aren't.

bnf:
Hi @akrabat, hi @l0gicgate,

Sorry, I missed your comments/thoughts. Therefore now my (very late) reaction:

I still think this is the right approach and actually considered it ready. Therefore I didn't do more work on this, because I was waiting for it to be merged. It's actually used in production as a backport for slim v3: https://packagist.org/packages/bnf/slim3-psr15
The main motivation for sticking with the old middleware dispatcher is the `$response` object. It needs to be handed over all the way from the first double-pass style middleware to the very last one.

> I'd like to explore if we can make PSR-15 the default internally and "wrap" the current signatures

Going with a pure PSR-15 solution and adding a wrapper for double pass middlewares (as in #2555) results in (hard to detect) breaking behaviour, as the wrapper has to create a new `$response` object for every double pass middleware. That means upgrading to Slim v4 will be a bit of a pain. (IMO)
To illustrate that: The following code will no longer work with #2555, as the `$response` will not be passed, but rather silently dropped:

```
$response = $response->withStatus(403);
```

Existing middleware code that expect to be able to pass a modified `$response` to the `$next` middleware will fail with that change. I know that it's a technical limitation due to the changes in PSR-15.
For exactly that reason I decided to keep the existing dispatcher instead of creating a new pure PSR-15 one.
If you find that limitation of dropping `$response` to be ok, than please go ahead with #2555 otherwise I'd ask you to reconsider this approach.

Thanks, Benjamin

l0gicgate:
Closing this as we are moving forward with #2555. Thank you for your contribution @bnf

akrabat:
> Closing this as we are moving forward with #2555. Thank you for your contribution @bnf

To this point, we only need one PR to get PSR-15 support into Slim 4. 

However, the BC problems that @bnf highlights (and @l0gicgate also highlighted in Slack at some point in the past to me) do need to be explicitly discussed I think.

My initial reaction as @l0gicgate noted is that passing the response down the chain is bad, so my natural reaction is that I don't mind breaking it in Slim 4. However, is that the right call? I could definitely be wrong.

PR #2385, opened 2018-01-22
Add public getter for the routes pattern

mewejo:
Simple PR to add a public getter method to the Route class so it can be used elsewhere in the app.


mewejo:
Appears it's already there 😆 closing.

geggleto:
yes :)

PR #2388, opened 2018-01-29
Allow SameSite flag to be used in cookies

### LINKED ISSUE #2387: Not possible to use SameSite in cookies
It is currently not possible to use the SameSite flag with lax or strict.

This flag is fairly new (inception in middle 2016) and intended to mitigate CSRF attacks. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

This is due to how cookies are built in [Http/Cookies.php](Slim/Slim/Http/Cookies.php), it skips unknown flags.

Adding this flag should be rather simple, but I've never contributed. I will try to make a proper PR for it (code standard / tests etc).

sg3s:
This should resolve #2387 if accepted.

(btw, I think you should use trailing commas for array lists, but of course didn't add those to stay inline with the rest)

### DIFF
--- a/Slim/Http/Cookies.php
+++ b/Slim/Http/Cookies.php
@@ -42,7 +42,8 @@ class Cookies implements CookiesInterface
         'path' => null,
         'expires' => null,
         'secure' => false,
-        'httponly' => false
+        'httponly' => false,
+        'samesite' => null
     ];
 
     /**
@@ -148,6 +149,11 @@ class Cookies implements CookiesInterface
 
         if (isset($properties['httponly']) && $properties['httponly']) {
             $result .= '; HttpOnly';
+        }
+
+        if (isset($properties['samesite']) && in_array(strtolower($properties['samesite']), ['lax', 'strict'], true)) {
+            // While strtolower is needed for correct comparison, the RFC doesn't care about case
+            $result .= '; SameSite=' . $properties['samesite'];
         }
 
         return $result;
--- a/tests/Http/CookiesTest.php
+++ b/tests/Http/CookiesTest.php
@@ -35,7 +35,8 @@ class CookiesTest extends \PHPUnit_Framework_TestCase
             'path' => null,
             'expires' => null,
             'secure' => true,
-            'httponly' => true
+            'httponly' => true,
+            'samesite' => null
         ];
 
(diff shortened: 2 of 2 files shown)


sg3s:
I do not know why Travis failed on hhvm, not familiar with either and the errors look vague to me.

If it is something I can fix I would love to hear suggestions.

akrabat:
This looks great. Would you mind raising a PR in https://github.com/slimphp/Slim-Website to document it?

Also, Slim 4.x will use https://github.com/slimphp/Slim-Http, so could you port this PR to that respository too, so we have it for the future?

akrabat:
If you rebase against `3.x` and then force push to your branch on your clone, then Travis should pass now.

sg3s:
Thank you for the feedback. 

I can rebase & push quite easily, but looking back on my code I see that I have mistaken 'non-strict comparison' for 'case-insensitive comparison'. Which is a dumb mistake. To compare I need to normalize the input properly.

1. Use strtolower or ucfirst (or something else?) on the input? (the used [case doesn't actually matter](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1) for setting the cookie flags)
2. Want a specific test for that case?

Cookies don't seem to be documented in v3 (I'm guessing the class shouldn't be present, it is mentioned in the [v3 upgrade guide](https://www.slimframework.com/docs/v3/start/upgrade.html#cookies) that it was removed) so where should I put this?

I will cherry-pick the patch to Slim-Http/v4 after figuring out the above.

akrabat:
> 1. Use strtolower or ucfirst (or something else?) on the input? (the used case doesn't actually matter for setting the cookie flags)

Use strtolower() or stripos().

> 2. Want a specific test for that case?

Yes.

akrabat:
@sg3s are you likely to be able to work on this?

sg3s:
@akrabat My apologies, I think I've missed the previous comment and I have been quite busy so I wasn't checking up on it either. 

Updated, added test & rebased. I think I've done that correctly, feedback is welcome :)

akrabat:
Thanks @sg3s! Fancy putting the same patch into https://github.com/slimphp/Slim-Http ?

PR #2395, opened 2018-02-10
removed unused imports

sergiy-petrov:
I've removed all unused imports in project.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -8,8 +8,6 @@
  */
 namespace Slim;
 
-use Exception;
-use Throwable;
 use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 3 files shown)


Slim/Route.php
```diff
@@ -8,8 +8,6 @@
  */
 namespace Slim;
 
-use Exception;
```
damianopetrungaro, line 11: This is used in https://github.com/slimphp/Slim/pull/2395/files#diff-5cbcc564b9520431aac614f483da529cR312
But use the backslash, so you may remove it or use the backslash here.

akrabat: Please do not use `\Exception` inline. import it with `use`.

akrabat:
Needs rebasing and removal of the use of `\Exception`

damianopetrungaro:
@akrabat btw we may add a cs fixer and automate all those processes.

akrabat:
> @akrabat btw we may add a cs fixer and automate all those processes.

Good idea.

damianopetrungaro:
I can work on it tomorrow if is it ok for you.

akrabat:
Closing as requested changes haven't been completed and we may go a different route.

Issue #2393, opened 2018-02-09
Difference between displayErrorDetails and printing to error log


edudobay:
Hi! I was configuring error reporting for an application and started to wonder about the `displayErrorDetails` setting: why is displaying errors to the client coupled to *not showing them on the log*? There are use cases during development where I might want to display the errors to the client but also have them at the log (e.g. if the client may be a mobile app or another application which might not give me a chance to look at the full response body).

Overall, it seems to me that switching displaying of error details and printing to the log are separate concepts and should be independent settings. Does this make sense to implement? Of course Slim is not a logging framework and the implementation for this should be kept simple. I would be happy to contribute a PR for this!

odan:
Hi @edudobay. I'm not sure whether I get your question right. Have you tried to implement a [custom slim error handler](https://www.slimframework.com/docs/handlers/error.html#custom-error-handler)? There you have the possibility to decide which (and when) error details are displayed to the client and can log all errors in detail using a logger (e. g. Monolog).

edudobay:
Hi! Yes, I had no problem implementing a custom error handler, but to keep things simple I wanted to reuse the logging behavior that Slim already provides out of the box. The problem is that I can't reuse *just* the logging part because it's coupled to the rendering part. If I turn on error rendering, logging gets turned off and that’s what I’d like to change.

My current solution is an ugly subclass that “fakes” the displayErrorDetails setting when delegating to the log writer. Of course this is coupled to the Slim implementation; the proposal for Slim 4.x at #2222 ([AbstractErrorHandler](https://github.com/slimphp/Slim/pull/2222/files#diff-d83bb89ad1f138cb4c7ed85ce0dbcfe9)) would break this, for example.
```php
class ErrorHandlerWithLogging extends \Slim\Handlers\Error
{

    protected function writeToErrorLog($throwable)
    {
        $oldValue = $this->displayErrorDetails;
        $this->displayErrorDetails = false;
        parent::writeToErrorLog($throwable);
        $this->displayErrorDetails = $oldValue;
    }
}
```

The most obvious “pretty” solution for me would be for the constructor to take two flags, say `displayErrorDetails` and `writeToErrorLog` - and the latter could default to the opposite of the former when not provided, to guarantee backwards-compatible behavior. Something like this:

```php
class Error extends AbstractError
{
    public function __construct($displayErrorDetails = false, $writeToErrorLog = null)
    {
        if ($writeToErrorLog === null) {
[...]

PR #2398, opened 2018-02-14
4.x Error Middleware

l0gicgate:
### Preface
The original PR #2222 was closed and some of the concepts and code will be implemented in this PR. 

### Description
This PR eliminates all native error handlers and replaces it with `ErrorMiddleware` which handles all errors and provides multiple different renderers (HTML, XML, JSON, Plain Text) to enhance error display.

### Implementation
```php
use Slim\App;
use Slim\Middleware\ErrorMiddleware;
use Slim\Middleware\RoutingMiddleware;

$app = new App();

/*
 * The routing middleware should be added earlier than the ErrorMiddleware
 * Otherwise exceptions thrown from it will not be handled by the middleware
 */
$routingMiddleware = new RoutingMiddleware($app->getRouter());
$app->add($routingMiddleware);

/*
 * The constructor of `ErrorMiddleware` takes in 4 parameters
 * @param CallableResolverInterface $callableResolver -> Callable Resolver Interface of your choice
 * @param bool $displayErrorDetails -> Should be set to false in production
 * @param bool $logErrors -> Parameter is passed to the default ErrorHandler
 * @param bool $logErrorDetails -> Display error details in error log
 * which can be replaced by a callable of your choice.
 * Note: This middleware should be added last. It will not handle any exceptions/errors
 * for middleware added after it.
 */
$callableResolver = $app->getCallableResolver();
$errorMiddleware = new ErrorMiddleware($callableResolver, true, true, true);
$app->add($errorMiddleware);

...
$app->run();
```

**Adding custom handlers for different named exceptions**
```php
...
$callableResolver = $app->getCallableResolver();
$errorMiddleware = new ErrorMiddleware($callableResolver, true, true, true);

$handler = function ($req, $res, $exception, $displayErrorDetails) {
    return $res->withJson(['error' => 'Caught MyNamedException']);
}
$errorMiddleware->setErrorHandler(MyNamedException::class, $handler);
$app->add($errorMiddleware);
```

### Error Logging
If you would like to pipe in custom error logging to the default `ErrorHandler` that ships with Slim you can simply extend it and stub the `logError()` method.

```php
namespace MyApp\Handlers;

use Slim\Handlers\ErrorHandler;

class MyErrorHandler extends ErrorHandler {
    public function logError($error)
    {
        // Insert custom error logging function.
    }
}
```

```php
use MyApp\Handlers\MyErrorHandler;
use Slim\App;
use Slim\Middleware\ErrorMiddleware;

$app = new App();
$callableResolver = $app->getCallableResolver();

$myErrorHandler = new MyErrorHandler(true); // Constructor parameter is $logErrors (bool)
$errorMiddleware = new ErrorMiddleware($callableResolver, true, true, true);
$errorMiddleware->setDefaultErrorHandler($myErrorHandler);
$app->add($errorMiddleware);

...
$app->run();
```

### Error Handling/Rendering
The rendering is finally decoupled from the handling. Everything still works the way it previously did. It will still detect the content-type and render things appropriately with the help of `ErrorRenderers`. The core `ErrorHandler` extends the `AbstractErrorHandler` class which has been completely refactored. By default it will call the appropriate `ErrorRenderer` for the supported content types. Someone can now provide their custom error renderer by extending the `AbstractErrorHandler` class and overloading the protected `renderer` variable from the parent. 

```php
class MyCustomErrorRenderer extends \Slim\Handlers\AbstractErrorRenderer
{
    public function render()
    {
        return 'My awesome format';
    }
}

class MyCustomErrorHandler extends \Slim\Handlers\ErrorHandler
{
    protected $renderer = MyCustomErrorRenderer::class;
}
```

### HTTP Exceptions
I just think it makes sense to throw HTTP based exceptions within the application. These exceptions works with nicely with the proposed native renderers. They can each have a `description` and `title` attribute as well to provide a bit more insight when the native HTML renderer is invoked. 

The base class `HttpSpecializedException` extends `Exception` and comes with the following sub classes :
- `HttpBadRequestException`
- `HttpForbiddenException`
- `HttpInternalServerErrorException`
- `HttpNotAllowedException`
- `HttpNotFoundException`
- `HttpNotImplementedException`
- `HttpUnauthorizedException`

The developer can extend the `HttpSpecializedException` class if they need any other response codes that we decide not to provide with the base repository. Example if you wanted a 504 gateway timeout exception that behaves like the native ones you would do the following:
I
```php
class HttpForbiddenException extends HttpSpecializedException
{
    protected $code = 504;
    protected $message = 'Gateway Timeout.';
    protected $title = '504 Gateway Timeout';
    protected $description = 'Timed out before receiving response from the upstream server.';
}
```

### Status
Work in progress
- [X] Remove Existing Error Handling
- [X] Implement Error Middleware
- [X] Implement supporting tests
- [X ] Discussion and Code Review.


Slim/App.php
```diff
@@ -555,51 +434,29 @@ public function group($pattern, $callable)
     public function run()
     {
         // create request
-        $request = Request::createFromGlobals($_SERVER);
+        if (is_null($this->request)) {
```
asheliahut: `null === $this->request` is better here no overhead of function call.

Slim/App.php
```diff
@@ -682,22 +541,14 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
         $router = $this->getRouter();
 
         // If routing hasn't been done, then do it now so we can dispatch
-        if (null === $routeInfo) {
+        if (is_null($routeInfo)) {
```
asheliahut: should stay ===

Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -0,0 +1,99 @@
...
+    public function render()
+    {
+        $e = $this->exception;
+        $title = 'Slim Application Error';
+
+        if ($this->displayErrorDetails) {
+            $html = '<p>The application could not run because of the following error:</p>';
+            $html .= '<h2>Details</h2>';
+            $html .= $this->renderExceptionFragment($e);
+        } else {
```
asheliahut: you can just return before the else and have this html set as a second return lowering cyclomatic complexity by 1.

Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @param Exception $exception
+     * @return string
+     */
+    private function renderExceptionFragment($exception)
+    {
+        $html = sprintf('<div><strong>Type:</strong> %s</div>', get_class($exception));
+
+        if (($code = $exception->getCode())) {
```
asheliahut: if someone passes an exception code of 0 this will never execute.

Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -0,0 +1,99 @@
...
+            $html .= sprintf('<div><strong>Code:</strong> %s</div>', $code);
+        }
+
+        if (($message = $exception->getMessage())) {
```
asheliahut: if someone sets a message of '0' it also will fail to run.

Slim/Error/Renderers/PlainTextErrorRenderer.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @param \Exception|\Throwable $e
+     * @return string
+     */
+    private function formatExceptionFragment($e)
+    {
+        $text = sprintf('Type: %s' . PHP_EOL, get_class($e));
+
+        if ($code = $e->getCode()) {
```
asheliahut: what about 0 here too

Slim/Error/Renderers/PlainTextErrorRenderer.php
```diff
@@ -0,0 +1,62 @@
...
+            $text .= sprintf('Code: %s' . PHP_EOL, $code);
+        }
+        if ($message = $e->getMessage()) {
```
asheliahut: and '0' here too

Slim/Error/Renderers/XmlErrorRenderer.php
```diff
@@ -0,0 +1,51 @@
...
+        $e = $this->exception;
+        $xml = "<error>\n  <message>{$e->getMessage()}</message>\n";
+        if ($this->displayErrorDetails) {
+            do {
+                $xml .= "  <exception>\n";
+                $xml .= "    <type>" . get_class($e) . "</type>\n";
+                $xml .= "    <code>" . $e->getCode() . "</code>\n";
+                $xml .= "    <message>" . $this->createCdataSection($e->getMessage()) . "</message>\n";
+                $xml .= "    <file>" . $e->getFile() . "</file>\n";
+                $xml .= "    <line>" . $e->getLine() . "</line>\n";
```
asheliahut: May want to exchange \n for . PHP_EOL for working with windows support.

Slim/Exception/HttpException.php
```diff
@@ -0,0 +1,122 @@
...
+    protected $description = '';
+
+    /**
+     * HttpException constructor.
+     * @param string|array|null $details
+     */
+    public function __construct($details = null)
+    {
+        if (is_string($details)) {
+            parent::__construct($details);
```
asheliahut: you can just return the parent constructor otherwise do $this->details = $details
thus removing the else.

Slim/Exception/HttpNotAllowedException.php
```diff
@@ -0,0 +1,42 @@
...
+    public function getAllowedMethods()
+    {
+        $details = $this->getDetails();
+
+        if (isset($details['allowedMethods'])) {
+            $allowedMethods = $details['allowedMethods'];
+
+            if (is_array($allowedMethods)) {
+                return implode(', ', $allowedMethods);
+            } elseif (is_string($allowedMethods)) {
```
asheliahut: this can be an if and not an else if because of the return will stop it anyways.

Slim/Exception/HttpNotAllowedException.php
```diff
@@ -0,0 +1,42 @@
...
+            return '';
+        }
+    }
+
+    /**
+     * @param string|array $methods
+     */
+    public function setAllowedMethods($methods)
+    {
+        if (is_null($this->details)) {
```
asheliahut: null === $this->details

Slim/Exception/HttpNotFoundException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace Slim\Exception;
+
+class HttpNotFoundException extends HttpException
+{
+    protected $code = 404;
+    protected $message = 'Not found.';
+    protected $title = '404 Not Found';
+    protected $description = 'The requested resource could not be found. Please verify URI and try again.';
```
asheliahut: Please verify the URI and try again.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,250 @@
...
+use Slim\Exception\HttpNotAllowedException;
+use Slim\Interfaces\ErrorHandlerInterface;
+use Slim\Interfaces\ErrorRendererInterface;
+use Exception;
+use Throwable;
+
+/**
+ * Default Slim application error handler
+ *
+ * It outputs the error message and diagnostic information in either JSON, XML,
```
asheliahut: either -> one of the following

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,250 @@
...
+     *
+     * @return ErrorRendererInterface
+     *
+     * @throws \RuntimeException
+     */
+    protected function determineRenderer()
+    {
+        $renderer = $this->renderer;
+
+        if ((!is_null($renderer) && !class_exists($renderer))
```
asheliahut: null ===  $renderer

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,250 @@
...
+            || (!is_null($renderer) && !in_array('Slim\Interfaces\ErrorRendererInterface', class_implements($renderer)))
+        ) {
+            throw new \RuntimeException(sprintf(
+                'Non compliant error renderer provided (%s). ' .
+                'Renderer must implement the ErrorRendererInterface',
+                $renderer
+            ));
+        }
+
+        if (is_null($renderer)) {
```
asheliahut: null === $renderer

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * @return int
+     */
+    protected function determineStatusCode()
+    {
+        if ($this->method === 'OPTIONS') {
+            return 200;
+        } elseif ($this->exception instanceof HttpException) {
```
asheliahut: change to if because of return

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Wraps the error_log function so that this can be easily tested
+     *
+     * @param string $error
+     */
+    protected function logError($error)
+    {
+        error_log($error);
```
asheliahut: I feel like this error log should also be able to take in an outside error logging, as in be able to take in a logger middleware of some kind.

Slim/Handlers/ErrorHandler.php
```diff
@@ -0,0 +1,19 @@
...
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2017 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/3.x/LICENSE.md (MIT License)
+ */
+namespace Slim\Handlers;
+
+/**
+ * Default Slim application error handler
+ *
+ * It outputs the error message and diagnostic information in either
```
asheliahut: either-> one of the following

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+     * @return ResponseInterface
+     */
+    public function handleException($exception)
+    {
+        /**
+         * Retrieve request object from exception and replace current request object if not null
+         */
+        if (method_exists($exception, 'getRequest')) {
+            $request = $exception->getRequest();
+            if (!is_null($request)) {
```
asheliahut: === null

Slim/Middleware/RoutingMiddleware.php
```diff
@@ -60,6 +65,21 @@ public function performRouting(ServerRequestInterface $request)
...
+            $exception = new HttpNotAllowedException();
+            $exception->setAllowedMethods($routeInfo[1]);
+        }
+
+        /**
+         * We pass the Request object to the exception so it reflects the
+         * exact state of the request at the time of the exception can be
+         * accessed by the end user if necessary
+         */
+        if (!is_null($exception)) {
```
asheliahut: !=== null

Slim/Error/AbstractErrorRenderer.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+namespace Slim\Error;
+
+use Slim\Http\Body;
+use Slim\Interfaces\ErrorRendererInterface;
+
+/**
+ * Default Slim application error renderer
+ *
+ * It outputs the error message and diagnostic information in either JSON, XML,
```
asheliahut: either should be "one of the following"

asheliahut (changes requested):
Just some thoughts.

l0gicgate:
Thank you @asheliahut for the code review. I'm aware this is a large PR and I really appreciate you going through everything. I have made the suggested changes besides the suggestion you made for the `AbstractErrorHandler::logError()` method, If people want to implement logging they can simply extend the class and replace the `::logError()` method. That's the main reason why I've left it abstract.

Slim/Error/Renderers/XmlErrorRenderer.php
odan: I'm developing under windows and linux and if I would run my api integration test under windows the tests would fail because the response is not the same like under linux. So please never use PHP_EOL because it will complicate thinks, especially in combination with windows and integration tests.

Slim/Error/Renderers/PlainTextErrorRenderer.php
```diff
@@ -0,0 +1,71 @@
...
+class PlainTextErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @return string
+     */
+    public function render()
+    {
+        $e = $this->exception;
+
+        $text = 'Slim Application Error:' . PHP_EOL;
```
odan: Use `\n` instead of `PHP_EOL` to make integration tests compatible with windows.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,259 @@
...
+        return 'text/html';
+    }
+
+    /**
+     * Determine which renderer to use based on content type
+     * Overloaded $renderer from calling class takes precedence over all
+     *
+     * @return ErrorRendererInterface
+     *
+     * @throws \RuntimeException
```
odan: Add `use RuntimeException` on top to remove the backslash.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,259 @@
...
+     */
+    protected function determineRenderer()
+    {
+        $renderer = $this->renderer;
+
+        if ((
```
odan: There'd be plenty of room to put this statement in one line.
`if (($renderer !== null && !class_exists($renderer)) {`

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,259 @@
...
+
+        return 500;
+    }
+
+    /**
+     * @return ResponseInterface
+     */
+    protected function formatResponse()
+    {
+        $e = $this->exception;
```
odan: This extra variable ($e) is not really necessary.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+        $params = [$this->request, $this->response, $exception, $this->displayErrorDetails];
+
+        return call_user_func_array($handler, $params);
+    }
+
+    /**
+     * Get callable to handle scenarios where an error
+     * occurs when processing the current request.
+     *
+     * @param string $type
```
odan: What for values can be passed here?

odan:
Hi! 

I have some questions / ideas about this PR and the error handling in general.

1) If Slim 4 will provide a new Error middleware like in this PR, why do we still need the `setErrorHandler()` and `setDefaultErrorHandler` methods in the App class? There should be only one style how to add middleware to the stack and not multiple different methods. This is important, because I want to decide in which order I place my middleware handlers. 

2) I don't like the PHP_EOL magic constant because the response will be different on windows/linux. The Response should be **exactly** the same for all platforms.

odan:
The name of the middleware `ErrorMiddleware` implies that it handles PHP errors. But in PHP there is a small difference between Exceptions and Errors. If this middleware handles "only" Exceptions and no "Errors" then the it should be renamed to "ExceptionMiddleware".

The question is how to handle PHP [errors](http://php.net/manual/de/class.error.php) (extended from Throwable)? Are you planning a new separate Middleware (e.g. ErrorMiddleware) or a generic "ThrowableMiddleware" for Exceptions  **and** Errors together?

Slim/Interfaces/ErrorRendererInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * @package Slim
+ * @since   4.0.0
+ */
+interface ErrorRendererInterface
+{
+    /**
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     */
+    public function __construct($exception, $displayErrorDetails);
```
designermonkey: Interfaces should not have a constructor as you are limiting the ability to define implementation detail vs defining expected behaviour.

Instead, your methods `render` and `renderWithBody` should accept the exception instance.

l0gicgate:
@odan 

- I am going to replace all instances of `PHP_EOL` to `\n`. Thanks for pointing that out.
- The `App` class does not contain `setErrorHandler()` and `setDefaultErrorHandler()` anymore. The first commit on this PR removes all that.
- Definitely open for proper naming suggestion for the middleware's name. It does handle both `Exception` and `Throwble` by default.

Thank you for reviewing this PR!

Slim/Interfaces/ErrorRendererInterface.php
l0gicgate: Thank you for pointing that out, I will refactor the interface accordingly!

Slim/Middleware/ErrorMiddleware.php
l0gicgate: I don't understand your question. Can you clarify?

edudobay:
Hi! As I started to point out at #2393, I believe that there should be a separate `writeToErrorLog` (or similar) setting, that doesn’t need to subordinate to `displayErrorDetails`. A couple reasons for that:

* Newcomers might be confused when they turn of display of error details and things stop going to the error log.
* Anyways we might need to keep a log of errors even though they were displayed to the client (maybe the client was an API so we couldn’t really see the output).
* Writing to a log and rendering the error to the client seem like separate concerns to me. These could even be dealt with by separate classes; but anyway it seems to me that it would be more logical to have separate settings for separate (independent) behaviors.

If you agree with these concerns, you can refer to #2393 for some remarks I’ve already made and I’ll be happy to help with the implementation. (I already have a draft but it is written for Slim 3.)

designermonkey:
I think this is an awesome piece of work, and definitely the direction Slim should be going.

I left a comment, but I have a couple of points to raise; not complaints, more architectural ideas.

Looking at the `AbstractRenderer` and it's constructor, I see there is a boolean passed as an argument. Generally when passing a boolean, it's a sure sign that there are in fact two concerns being addressed by the class/method, and therefore a refactor in the making.

Maybe we could look at renderers like `RendersHtmlErrorsWithDetails` and `RendersHtmlErrorsWithoutDetails`? This is a strategy in effect, and can be handled by the container based on a more global config option, rather than passing the boolean `displayErrorDetails` around.

l0gicgate:
@edudobay Thank you for your suggestions. I will refactor `AbstractErrorHandler` accordingly.

`ErrorMiddleware` will take 2 parameters for the constructor `$displayErrorDetails` and `$logErrors` which will both be passed down to the default `ErrorHandler` that Slim ships with.

I will also implement a way to pass a custom callable to `ErrorHandler` so it stubs the default `error_log()` usage.

l0gicgate:
@edudobay as per the commits after our discussion. The error logging has been decoupled from `displayErrorDetails`. I updated the PR documentation to reflect how custom logging can be done as well.

l0gicgate:
@designermonkey the whole reason for having standalone error renderers is so they can be used by the end user as well if they decide to implement their own error handlers. Using global values from the containers would bind those renderers to the container which is not something I want to do. Am I misunderstanding what you're proposing?

designermonkey:
@l0gicgate I'm just generally stating that boolean values as constructor and method arguments is a code smell, and should be refactored in some way. The suggestion of container use was only an idea of how. :)

odan:
Maybe mixing error handling and error logging violentes SRP. What do you think about splitting this into 2 Middleware classes? (Just an idea)

For example. 

1. `new ExceptionMiddleware(bool $displayErrorDetails)`
2. `new LogExceptionMiddleware($logger)`

Log all exceptions from your middlewares. When your middleware throw exception this middleware catch it, log it and throw again. You will newer miss any exception. [Example](https://github.com/php-middleware/log-exceptions/blob/master/src/LogExceptionMiddleware.php)

designermonkey:
@odan I would think that logging is a form of exception handling in itself. Recently I've been using messaging patterns more, and message handling can have more than one handler, so maybe a direction to go? Exception handling could indeed be a kind of middleware pipe allowing multiple handlers?

odan:
@designermonkey Well, it's still possible to log and handle the error in the same callback. Then we could everything there. For example:

```php
$errorMiddleware = new ErrorMiddleware(false);

$handler = function ($req, $res, $exception, $displayErrorDetails) {
   // get the logger and log something.
    $logger->error(...);
    return $res->withJson(['error' => 'Caught MyNamedException']);
}
$errorMiddleware->setErrorHandler(MyNamedException::class, $handler);
$app->add($errorMiddleware);
```

l0gicgate:
@designermonkey I refactored the `ErrorRendererInterface` after our discussion in this [commit](https://github.com/slimphp/Slim/pull/2398/commits/595f9819cb137d5725473aa898b134872e62dd2a) I however did not split the methods for displaying/not displaying error details.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Write to the error log if displayErrorDetails is false
+     *
+     * @return void
+     */
+    protected function writeToErrorLog()
+    {
+        $renderer = new PlainTextErrorRenderer();
+        $error = $renderer->render($this->exception, $this->displayErrorDetails);
```
edudobay: Maybe the details should *always* be displayed in the error log? To me it doesn’t make sense to log a generic "Application error" without details.

(Maybe we are starting to cross boundaries and violating SRP as @odan pointed out?)

l0gicgate: You do have a point. I can modify this so it’s always displaying the error details when writing to log if everyone is in favor of this option!

@odan @designermonkey @akrabat @asheliahut @geggleto

akrabat:
```
/*
 * The routing middleware should always be added first
 * before the error handling middleware is added
 */
```

Why? I want the error handling middleware to be the outermost middleware as there will be other middleware before routing that I want the error handler to pick up.

Slim/Handlers/AbstractErrorHandler.php
odan: On a production server I have to log all errors (into a file), but for security reasons I am not allowed to show the users any error details. Therefore, it is still necessary to be able to configure both options (displayErrorDetails and logging) independently of each other.

akrabat: For more advanced requirements, I expect that the user should be able to easily replace Slim's error handling with their own.

I don't want this component to turn into a massive thing as we're _Slim_! However, we should be reasonably flexible.

l0gicgate:
@akrabat what I meant by that is `ErrorMiddleware` should be the last one to be added. In the example `RoutingMiddleware` should be added first, but I don't mean that it should be added first in your whole application.

Slim/Handlers/AbstractErrorHandler.php
l0gicgate: So how about I decouple `$displayErrorDetails` from the logging and add another parameter `$logErrorDetails` which can be enabled or disabled. So for instance you may not want to display the error details to the end user but you may want to have them displayed in your log.

odan: @l0gicgate Sounds good

l0gicgate: @odan I've implemented `$logErrorDetails` in this [commit](https://github.com/slimphp/Slim/pull/2398/commits/897958f4e6efb6d297b098ada9d9cdc01013fe92)

Slim/App.php
```diff
@@ -106,6 +87,7 @@ public function __construct(array $settings = [], ContainerInterface $container
      * Get container
      *
      * @return ContainerInterface|null
+     * @codeCoverageIgnore
```
akrabat: Why are we ignoring code coverage for methods in App?

Slim/App.php
```diff
@@ -58,32 +49,21 @@ class App
     protected $router;
 
     /**
-     * @var callable
+     * @var ServerRequestInterface|null
      */
-    protected $notFoundHandler;
+    protected $request;
```
akrabat: I do not want $request or $response as properties on App. They were intentionally removed from Slim 3's DIC because they are immutable and people use them as if they aren't.

This also means that I don't want `set/getRequest()` and `set/getResponse()`.

Also, `run()` and `process()` are separate for a reason: it allows someone to inject their own request and response when running Silm.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+use Slim\Interfaces\ErrorHandlerInterface;
+use Exception;
+use RuntimeException;
+use Throwable;
+
+class ErrorMiddleware
+{
+    /**
+     * @var bool
+     */
```
akrabat: Add a gap between each property. It just looks nicer. This applies to all files.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+     * @param ServerRequestInterface $request   The most recent Request object
+     * @param ResponseInterface      $response  The most recent Response object
+     * @param callable $next
+     *
+     * @return ResponseInterface
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $this->request = $request;
+        $this->response = $response;
```
akrabat: Do not store response. Each renderer should return a response that they have created.

Slim/Middleware/RoutingMiddleware.php
```diff
@@ -60,6 +65,21 @@ public function performRouting(ServerRequestInterface $request)
 
             // add route to the request's attributes
             $request = $request->withAttribute('route', $route);
+        } elseif ($routeInfo[0] === Dispatcher::NOT_FOUND) {
+            $exception = new HttpNotFoundException();
```
akrabat: Call ` $exception->setRequest($request);` and then `throw` directly here.

Slim/Middleware/RoutingMiddleware.php
```diff
@@ -60,6 +65,21 @@ public function performRouting(ServerRequestInterface $request)
...
+        } elseif ($routeInfo[0] === Dispatcher::METHOD_NOT_ALLOWED) {
+            $exception = new HttpNotAllowedException();
+            $exception->setAllowedMethods($routeInfo[1]);
```
akrabat: Call ` $exception->setRequest($request);` and then `throw` directly here.

I know it's a duplication from above, but will be clearer to read and remove the test for not null below.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+     * Resolve custom error handler from container or use default ErrorHandler
+     * @param Exception|Throwable $exception
+     * @return ResponseInterface
+     */
+    public function handleException($exception)
+    {
+        /**
+         * Retrieve request object from exception and replace current request object if not null
+         */
+        if (method_exists($exception, 'getRequest')) {
```
akrabat: Test for an instance of `HttpException`. That way we can be sure that the Request object in getRequest() is a PSR7 Request object.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+            $request = $exception->getRequest();
+            if ($request !== null) {
```
akrabat: Can we make `HttpException`'s constructor require a `Request` object? Then we don't need the null test here.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+        $count = count($selectedContentTypes);
+
+        if ($count) {
+            $current = current($selectedContentTypes);
+
+            /**
+             * Ensure other supported content types take precedence over text/plain
+             * when multiple content types are provided via Accept header.
+             */
+            if ($current === 'text/plain' && $count > 1) {
```
akrabat: I'm not sure about this. Conceptually, if the client has set `text/plain` first, then it should win. Is there a major browser which sets `text/plain` first or something?

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+     * @param ServerRequestInterface $request   The most recent Request object
+     * @param ResponseInterface      $response  The most recent Response object
+     * @param Exception|Throwable    $exception The caught Exception object
+     * @param bool $displayErrorDetails Whether or not to display the error details
+     *
+     * @return ResponseInterface
+     */
+    public function __invoke(
+        ServerRequestInterface $request,
+        ResponseInterface $response,
```
akrabat: Don't need a Response here. The renderer should create its own.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * ErrorMiddleware constructor.
+     * @param bool $displayErrorDetails
+     * @param bool $logErrors
+     * @param bool $logErrorDetails
+     */
+    public function __construct($displayErrorDetails, $logErrors, $logErrorDetails)
+    {
+        $this->displayErrorDetails = $displayErrorDetails;
+        $this->defaultErrorHandler = new ErrorHandler($logErrors, $logErrorDetails);
```
akrabat: I don't think that we should instantiate an ErrorHandler here. More often than not, we won't need it.

akrabat: Why are `$logErrors` and `$logErrorDetails` passed directly to the ErrorHandler directly and not passed in via `__invoke` like `$displayErrorDetails` is?

I can imagine that if I register my own ErrorHandler, I'd like the option to implement logging of errors as per the user's configuration of the error middleware.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Get callable to handle scenarios where an error
+     * occurs when processing the current request.
+     *
+     * @param string $type Exception/Throwable name. ie: RuntimeException::class
+     * @return callable|ErrorHandler
+     */
+    public function getErrorHandler($type)
+    {
+        if (isset($this->handlers[$type]) && is_callable($this->handlers[$type])) {
```
akrabat: Ideally, we should integrate the container and collect from there, so that we can lazily instantiate ErrorHandler classes.

Slim/Middleware/ErrorMiddleware.php
```diff
@@ -0,0 +1,173 @@
...
+            return $this->handlers[$type];
+        }
+        return $this->defaultErrorHandler;
```
akrabat: Related, would be nice to collect from the container rather than having to have it instantiated every request.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @return ErrorRendererInterface
+     *
+     * @throws RuntimeException
+     */
+    protected function determineRenderer()
+    {
+        $renderer = $this->renderer;
+
+        if ($renderer !== null
```
akrabat: Don't just test here. If the `$renderer` is not null, then we can instantiate it and return it within this `if()` statement. That way we can test using `instancepf ErrorRendererInterface` and we don't have to check for `$renderer === null` later in the method.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+
+        return new $renderer();
+    }
+
+    /**
+     * @return int
+     */
+    protected function determineStatusCode()
+    {
+        if ($this->method === 'OPTIONS') {
```
akrabat: Probably not this PR, but I think we should handle Options in its own middleware or in App or Route directly.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+        return 500;
+    }
+
+    /**
+     * @return ResponseInterface
+     */
+    protected function formatResponse()
+    {
+        $response = $this->response;
+        $body = $this->renderer->renderWithBody($this->exception, $this->displayErrorDetails);
```
akrabat: The renderer should return a response that's configured with the content-type correct header. Just add the status and we're good to go.

Slim/Handlers/AbstractErrorHandler.php
akrabat: In general, I agree. A separate PR after this could allow for handling a PSR3 logger.

Slim/Error/AbstractErrorRenderer.php
```diff
@@ -0,0 +1,34 @@
...
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     * @return \Slim\Http\Body
+     */
+    public function renderWithBody($exception, $displayErrorDetails)
+    {
+        $output = $this->render($exception, $displayErrorDetails);
+        $body = new Body(fopen('php://temp', 'r+'));
+        $body->write($output);
+        return $body;
```
akrabat, line 34: Renderers should return a Response. As such, I'm not sure what an abstract implementation looks like, so this can probably be removed.

Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -0,0 +1,105 @@
...
+ * Default Slim application HTML Error Renderer
+ */
+class HtmlErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     * @return string
+     */
+    public function render($exception, $displayErrorDetails)
```
akrabat, line 24: Return a Response.

Slim/Error/Renderers/JsonErrorRenderer.php
```diff
@@ -0,0 +1,51 @@
...
+ * Default Slim application JSON Error Renderer
+ */
+class JsonErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     * @return string
+     */
+    public function render($exception, $displayErrorDetails)
```
akrabat, line 23: Return a Response.

Slim/Error/Renderers/JsonErrorRenderer.php
```diff
@@ -0,0 +1,51 @@
...
+        $error = ['message' => $exception->getMessage()];
+
+        if ($displayErrorDetails) {
+            $error['exception'] = [];
+            do {
+                $error['exception'][] = $this->formatExceptionFragment($exception);
+            } while ($exception = $exception->getPrevious());
+        }
+
+        return json_encode($error, JSON_PRETTY_PRINT);
```
akrabat: Can we use `JSON_PRETTY_PRINT| JSON_UNESCAPED_SLASHES` please?

Slim/Error/Renderers/PlainTextErrorRenderer.php
```diff
@@ -0,0 +1,71 @@
...
+ * Default Slim application Plain Text Error Renderer
+ */
+class PlainTextErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     * @return string
+     */
+    public function render($exception, $displayErrorDetails)
```
akrabat, line 23: Return a Response.

Slim/Error/Renderers/XmlErrorRenderer.php
```diff
@@ -0,0 +1,54 @@
...
+ * Default Slim application XML Error Renderer
+ */
+class XmlErrorRenderer extends AbstractErrorRenderer
+{
+    /**
+     * @param \Exception|\Throwable $exception
+     * @param bool $displayErrorDetails
+     * @return string
+     */
+    public function render($exception, $displayErrorDetails)
```
akrabat, line 23: Return a Response.

Slim/Exception/HttpBadGatewayException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace Slim\Exception;
+
+class HttpBadGatewayException extends HttpException
```
akrabat: Please remove this Exception. It's not a common enough use-case to my mind.

Slim/Exception/HttpException.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * @var string
+     */
+    protected $description = '';
+
+    /**
+     * HttpException constructor.
+     * @param string|array|null $details
+     */
+    public function __construct($details = null)
```
akrabat: Can we make Request the first constructor parameter (required)? I can't think of a downside.

Slim/Exception/HttpException.php
```diff
@@ -0,0 +1,122 @@
...
+abstract class HttpException extends \Exception
+{
+    /**
+     * @var ServerRequestInterface|null
+     */
+    protected $request = null;
+    /**
+     * @var ResponseInterface|null
+     */
+    protected $response = null;
```
akrabat: No need for `$response`. Create one if you need one.

Slim/Exception/HttpException.php
akrabat: What's the difference between details, title and description?

Slim/Exception/HttpNotAllowedException.php
```diff
@@ -0,0 +1,44 @@
+<?php
+namespace Slim\Exception;
+
+class HttpNotAllowedException extends HttpException
```
akrabat: Rename to `HttpMethodNotAllowedException`.

Slim/Handlers/AbstractErrorHandler.php
```diff
@@ -0,0 +1,274 @@
...
+
+        if ($this->exception instanceof HttpNotAllowedException) {
```
akrabat: As the exception should return a response, this can go into the exception itself.

Slim/Handlers/ErrorHandler.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+namespace Slim\Handlers;
+
+/**
+ * Default Slim application error handler
+ *
+ * It outputs the error message and diagnostic information in one of the following formats:
+ * JSON, XML, Plain Text or HTML based on the Accept header.
+ */
+class ErrorHandler extends AbstractErrorHandler
```
akrabat: Would rather have an interface and rename `AbstractErrorHandler` to `ErrorHandler`

Slim/Interfaces/ErrorHandlerInterface.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @param Exception|Throwable $exception
+     * @param $displayErrorDetails
+     * @return ResponseInterface
+     */
+    public function __invoke(
+        ServerRequestInterface $request,
+        ResponseInterface $response,
```
akrabat: Response not required.

Slim/Interfaces/ErrorRendererInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+namespace Slim\Interfaces;
+
+/**
+ * ErrorRendererInterface
+ *
+ * @package Slim
+ * @since   4.0.0
+ */
+interface ErrorRendererInterface
```
akrabat, line 17: I think we just need `render` which returns a response.

akrabat (changes requested):
A number of bits and bobs in here to look at @l0gicgate.

If you'd like me to raise a PR against this one addressing them, let me know.

Slim/App.php
l0gicgate: Removed and added tests for those methods.

Slim/App.php
l0gicgate: Removed and modified run() signature to take in an optional $request parameter for test purposes.

Slim/Middleware/ErrorMiddleware.php
l0gicgate: Done.

Slim/Middleware/ErrorMiddleware.php
l0gicgate: Done.

Slim/Middleware/RoutingMiddleware.php
l0gicgate: Done.

Slim/Middleware/RoutingMiddleware.php
l0gicgate: Done.

Slim/Middleware/ErrorMiddleware.php
l0gicgate: Done.

Slim/Middleware/ErrorMiddleware.php
l0gicgate: I don't think that's a good idea. Users might be using these exceptions within their applications and having to keep request in scope all the time when throwing one of these named exceptions is annoying. I use these exceptions a lot in the API that I am building at the moment and I would be hindered by having to pass $request every time.

akrabat: Okay. Accepted for now. I'm mildly concerned that you're throwing HTTP exceptions from somewhere where you don't have a request object, but it's your architecture :)

Slim/Handlers/AbstractErrorHandler.php
l0gicgate: Done.

Slim/Middleware/ErrorMiddleware.php
l0gicgate: Refactored.

l0gicgate: Absolutely right. Refactored.

Slim/Error/Renderers/JsonErrorRenderer.php
l0gicgate: Done.

Slim/Exception/HttpBadGatewayException.php
l0gicgate: Removed.

Slim/Exception/HttpException.php
l0gicgate: Refactored as discussed.

Slim/Exception/HttpException.php
l0gicgate: Done.

Slim/Exception/HttpNotAllowedException.php
l0gicgate: Done.

Slim/Handlers/ErrorHandler.php
l0gicgate: Done.

Slim/Interfaces/ErrorHandlerInterface.php
l0gicgate: Refactored.

Slim/Exception/HttpException.php
l0gicgate: This may be a bit too fancy. I could concatenate `$title` and `$description` and concatenate it into `$message`. I really outdid myself here 😂 

As for `$details`, I thought it would give a user the chance to add onto the exception. Right now it also stores allowed methods when `HttpMethodNotAllowed` is being thrown. But it can be used to store any data the end user wants.

```php
class HttpForbiddenException extends HttpSpecializedException
{
    protected $code = 403;
    protected $message = 'Forbidden.';
    protected $title = '403 Forbidden';
    protected $description = 'You are not permitted to perform the requested operation.';
}
```

Can we settle on:
```php
class HttpForbiddenException extends HttpSpecializedException
{
    protected $code = 403;
    protected $message = 'Forbidden. You are not permitted to perform the requested operation.';
}
```

Thoughts?

Slim/Interfaces/ErrorHandlerInterface.php
l0gicgate: Actually, what if during the middleware progression a user did attach headers to the response, example if a CORS middleware appended headers, wouldn't we want to have the response object that reflects the correct state at the time the time the exception was thrown?

akrabat: Where does it come from given that the error handler is called from the catch() of the Middleware and that we won't have $response passed into the middleware in PSR-15?

Issue #2399, opened 2018-02-16
Custom Dispatcher Implementation for 3.x/4.x


l0gicgate:
As discussed on Slack, there are a few issues with the way `FastRoute\Dispatcher` returns information after a route has been dispatched.

**Issues**
- `FastRoute\Dispatcher::dispatch()` returns an array with an inconsistent structure
- It only appends allowed methods to the return value when no allowed methods have been found on the dispatched route.

**Solution**
As discussed, I proposed that we implement our own dispatcher based on the logic found in  `FastRoute\Dispatcher\GroupCountBased`. Our dispatcher would return a `DispatcherResults` object which would have the following methods:
- `::getStatus()` which would return `Dispatcher::FOUND`, `Dispatcher::NOT_ALLOWED` or `Dispatcher::NOT_FOUND`
- `::getAllowedMethods()` which would return the allowed methods for dispatched route. This would be done dynamically as it has to loop through all the routes in order to compute the allowed methods. This is the exact reason why at the moment it is only done when the `Dispatcher::NOT_ALLOWED` status is being returned. This way, we offset the overhead on the end-user when they call this method.
- `::getHandler()` when the route is found, this is the way we get the handler (callable) associated with the found route.

**Implementation**
@akrabat mentioned that we could implement this in `3.x` and append the `DispatcherResults` object generated to a request attribute `dispatcherResults` or `dispatchResult` and consequently phase out `routeInfo` on `4.x` which is currently being appended while routing is done.

**Questions**
- Are you in favor of this implementation?
- Should we implement it in `3.x` and phase out `routeInfo` in `4.x`
- Should the new attribute be called `dispatchResult` or `dispatcherResults`
- Should the object returned by our dispatcher be `DispatcherResults` or `DispatchResults` or something else.

**Status**
- Proof of concept of custom Dispatcher can be seen [here](https://gist.github.com/l0gicgate/b47f084f7180ca119e06cc4fc55ee99b)
- Need to finalize discussion

RyanNerd:
* I am in favor of this implementation. Including the _work-around_ for `3.x`.
* Absolutely phase out `routeInfo` in `4.x`
* I am in favor of the plural for the attribute: `dispatcherResults`
* I am in favor of `DispatcherResults`  (if the attribute is also plural). I'd also be okay with `DispatcherInfo`

l0gicgate:
[...]

PR #2405, opened 2018-02-23
4.x New Dispatcher & Routing Results

l0gicgate:
As per #2399 

As discussed on Slack, there are a few issues with the way `FastRoute\Dispatcher` returns information after a route has been dispatched.

**Issues**
- `FastRoute\Dispatcher::dispatch()` returns an array with an inconsistent structure
- It only appends allowed methods to the return value when no allowed methods have been found on the dispatched route.

**Changes**
- `routeInfo` now deprecated and replaced with `routingResults`
- `RoutingResults` is an immutable object instantiated by the Slim Dispatcher which now offers a way to get all a route's allowed methods which was not accessible before via `FastRoute\GroupCountBased\Dispatcher`

**Usage**
```php
use Slim\App;

$app = new App();
$app->get('/foo', function (Request $request, Response $response) {
    $routingResults = $request->getAttribute('routingResults');
    $uri = $routingResults->getUri();
    $method = $routingResults->getMethod();
    $routeArguments = $routingResults->getRouteArguments();
    $allowedMethods = $routingResults->getAllowedMethods();
});
```

**Status**
- [X] Implement new Dispatcher & Routing Results
- [X] Implement FastRoute test suite for Dispatcher
- [x] Code Review & Discussion

### DIFF
--- a/Slim/Dispatcher.php
+++ b/Slim/Dispatcher.php
@@ -23,9 +23,9 @@ class Dispatcher extends GroupCountBased
     private $uri;
 
     /**
-     * @var array|null
+     * @var array
      */
-    private $allowedMethods;
+    private $allowedMethods = [];
 
     /**
      * @param string $httpMethod
@@ -70,8 +70,7 @@ class Dispatcher extends GroupCountBased
             return $this->dispatcherResultsFromVariableRouteResults($result);
         }
 
-        $this->allowedMethods = $this->getAllowedMethods($httpMethod, $uri);
-        if (count($this->allowedMethods)) {
+        if (count($this->getAllowedMethods($httpMethod, $uri))) {
             return $this->dispatcherResults(self::METHOD_NOT_ALLOWED);
         }
 
@@ -108,14 +107,14 @@ class Dispatcher extends GroupCountBased
      */
     public function getAllowedMethods($httpMethod, $uri)
     {
-        if ($this->allowedMethods !== null) {
-            return $this->allowedMethods;
+        if (isset($this->allowedMethods[$uri])) {
+            return $this->allowedMethods[$uri];
         }
 
-        $this->allowedMethods = [];
+        $this->allowedMethods[$uri] = [];
         foreach ($this->staticRouteMap as $method => $uriMap) {
             if ($method !== $httpMethod && isset($uriMap[$uri])) {
-                $this->allowedMethods[] = $method;
+                $this->allowedMethods[$uri][] = $method;
             }
         }
 
@@ -123,10 +122,10 @@ class Dispatcher extends GroupCountBased
         foreach ($varRouteData as $method => $routeData) {
             $result = $this->dispatchVariableRoute($routeData, $uri);
             if ($result[0] === self::FOUND) {
-                $this->allowedMethods[] = $method;
+                $this->allowedMethods[$uri][] = $method;
             }
         }
 
-        return $this->allowedMethods;
+        return $this->allowedMethods[$uri];
     }
 }
--- a/Slim/Dispatcher.php
+++ b/Slim/Dispatcher.php
@@ -1,0 +1,124 @@
+<?php
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2017 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+namespace Slim;
+
+use FastRoute\Dispatcher\GroupCountBased;
+
+class Dispatcher extends GroupCountBased
+{
+    /**
+     * @var string
+     */
+    private $httpMethod;
+    /**
+     * @var string
+     */
+    private $uri;
+
+    /**
+     * @param string $httpMethod
+     * @param string $uri
+     * @return DispatcherResults
+     */
+    public function dispatch($httpMethod, $uri)
+    {
+        $this->httpMethod = $httpMethod;
+        $this->uri = $uri;
+
+        if (isset($this->staticRouteMap[$httpMethod][$uri])) {
+            return $this->dispatcherResults(self::FOUND, $this->staticRouteMap[$httpMethod][$uri]);
+        }
+
+        $varRouteData = $this->variableRouteData;
+        if (isset($varRouteData[$httpMethod])) {
+            $result = $this->dispatchVariableRoute($varRouteData[$httpMethod], $uri);
+            return $this->dispatcherResultsFromVariableRouteResults($result);
+        }
+
+        // For HEAD requests, attempt fallback to GET
+        if ($httpMethod === 'HEAD') {
+            if (isset($this->staticRouteMap['GET'][$uri])) {
+                return $this->dispatcherResults(self::FOUND, $this->staticRouteMap['GET'][$uri]);
+            }
+            if (isset($varRouteData['GET'])) {
+                $result = $this->dispatchVariableRoute($varRouteData['GET'], $uri);
+                return $this->dispatcherResultsFromVariableRouteResults($result);
+            }
+        }
+
+        // If nothing else matches, try fallback routes
+        if (isset($this->staticRouteMap['*'][$uri])) {
+            return $this->dispatcherResults(self::FOUND, $this->staticRouteMap['*'][$uri]);
+        }
+        if (isset($varRouteData['*'])) {
+            $result = $this->dispatchVariableRoute($varRouteData['*'], $uri);
+            return $this->dispatcherResultsFromVariableRouteResults($result);
+        }
+
+        if (count($this->getAllowedMethods($httpMethod, $uri))) {
+            return $this->dispatcherResults(self::METHOD_NOT_ALLOWED);
+        }
+
+        return $this->dispatcherResults(self::NOT_FOUND);
+    }
+
+    /**
+     * @param $status
+     * @param null $handler
+     * @param array $arguments
+     * @return DispatcherResults
+     */
+    protected function dispatcherResults($status, $handler = null, $arguments = [])
+    {
+        return new DispatcherResults($this, $this->httpMethod, $this->uri, $status, $handler, $arguments);
+    }
+
+    /**
+     * @param array $result
+     * @return DispatcherResults
+     */
+    protected function dispatcherResultsFromVariableRouteResults($result)
+    {
+        if ($result[0] === self::FOUND) {
+            return $this->dispatcherResults(self::FOUND, $result[1], $result[2]);
+        }
+        return $this->dispatcherResults(self::NOT_FOUND);
+    }
+
+    /**
+     * @param string $httpMethod
+     * @param string $uri
+     * @param bool $includeRequestedMethod
+     * @return array
+     */
+    public function getAllowedMethods($httpMethod, $uri, $includeRequestedMethod = true)
+    {
+        $httpMethod = $includeRequestedMethod ? 'ArtificialMethod' : $httpMethod;
+
+        $allowedMethods = [];
+        foreach ($this->staticRouteMap as $method => $uriMap) {
+            if ($method !== $httpMethod && isset($uriMap[$uri])) {
+                $allowedMethods[] = $method;
+            }
+        }
+
+        $varRouteData = $this->variableRouteData;
+        foreach ($varRouteData as $method => $routeData) {
+            if ($method === $httpMethod) {
+                continue;
+            }
+            $result = $this->dispatchVariableRoute($routeData, $uri);
+            if ($result[0] === self::FOUND) {
+                $allowedMethods[] = $method;
+            }
+        }
+
+        return $allowedMethods;
+    }
+}
--- a/Slim/Dispatcher.php
+++ b/Slim/Dispatcher.php
@@ -30,7 +30,7 @@ class Dispatcher extends GroupCountBased
     /**
      * @param string $httpMethod
      * @param string $uri
-     * @return DispatcherResults
+     * @return RoutingResults
      */
     public function dispatch($httpMethod, $uri)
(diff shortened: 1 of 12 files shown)


Slim/Middleware/RoutingMiddleware.php
```diff
@@ -53,35 +56,32 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+        $dispatcherResults = $this->router->dispatch($request);
+        $routeStatus = $dispatcherResults->getRouteStatus();
 
-        if ($routeInfo[0] === Dispatcher::FOUND) {
-            $routeArguments = [];
-            foreach ($routeInfo[2] as $k => $v) {
-                $routeArguments[$k] = urldecode($v);
-            }
+        switch ($routeStatus) {
+            default:
```
akrabat: `default` should be last in a switch statement. Arguably, it should throw an Exception as it should never happen.

Slim/Dispatcher.php
```diff
@@ -0,0 +1,128 @@
...
+            if ($method === $httpMethod) {
+                continue;
+            }
+            $result = $this->dispatchVariableRoute($routeData, $uri);
+            if ($result[0] === self::FOUND) {
+                $allowedMethods[] = $method;
+            }
+        }
+
+        return $allowedMethods;
```
akrabat: Should we be caching the list of allowed methods in an instance variable keyed by the function's parameters? It appears that we currently have to do this same work twice when we're in a method not allowed scenario.

Slim/DispatcherResults.php
```diff
@@ -0,0 +1,137 @@
...
+    public function getRouteHandler()
+    {
+        return $this->routeHandler;
+    }
+
+    /**
+     * @param bool $urlDecode
+     * @return array
+     */
+    public function getRouteArguments($urlDecode = true)
```
akrabat: Why would we not want to url decode the route arguments?

Slim/Router.php
```diff
@@ -266,11 +263,13 @@ protected function createDispatcher()
 
         if ($this->cacheFile) {
             $this->dispatcher = \FastRoute\cachedDispatcher($routeDefinitionCallback, [
+                'dispatcher' => '\\Slim\\Dispatcher',
```
geggleto: We should use the class constant here

geggleto: This however makes the app not extendable and a user cannot replace the dispatcher

geggleto (changes requested):

Slim/DispatcherResults.php
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @var string
+     */
+    protected $uri;
+
+    /**
+     * @var int
+     *
+     * Not Found = 0
```
damianopetrungaro: What do you think about using an Enum for this `$routeStatus` or if it's an overhead use constants instead of comments for valid statuses?

Slim/DispatcherResults.php
```diff
@@ -0,0 +1,137 @@
...
+        return $this->routeHandler;
+    }
+
+    /**
+     * @param bool $urlDecode
+     * @return array
+     */
+    public function getRouteArguments($urlDecode = true)
+    {
+        if ($urlDecode) {
```
damianopetrungaro: Avoid nested statement:
```
if (!$urlDecode) {
	return $this->routeArguments;
}

$routeArguments = [];
foreach ($this->routeArguments as $key => $value) {
    $routeArguments[$key] = urldecode($value);
}

return $routeArguments;
```

Slim/DispatcherResults.php
geggleto: It probably should be a class constant

Slim/Dispatcher.php
```diff
@@ -0,0 +1,128 @@
...
+        if (isset($varRouteData[$httpMethod])) {
+            $result = $this->dispatchVariableRoute($varRouteData[$httpMethod], $uri);
+            $dispatcherResults = $this->dispatcherResultsFromVariableRouteResults($result);
+            if ($dispatcherResults->getRouteStatus() === Dispatcher::FOUND) {
+                return $dispatcherResults;
+            }
+        }
+
+        // For HEAD requests, attempt fallback to GET
+        if ($httpMethod === 'HEAD') {
```
damianopetrungaro, line 54: Wdyt about adding `https://github.com/php-fig/http-message-util` to avoid hardcoded HTTP methods?

geggleto: HTTP Methods don't change. I dont think we need to include a package for constants, we aren't javascript.

Slim/Router.php
l0gicgate: Actually the Router can be extended and this method overridden. When instantiating the RoutingMiddleware you can pass a router to it which could be the router of your choice. An alternative would be to pass settings to the Router constructor.

Slim/Dispatcher.php
damianopetrungaro: I always include it in for descriptive HTTP status codes and it contains have HTTP methods too.
Seen that is a FIG package I'm ok with it.
But I get your point of view so it's not a big deal 😄

Slim/DispatcherResults.php
l0gicgate: The routeStatus is what comes from Fastroute. We can convert those to constants. Should the constants be on the Dispatcher or Dispatcher results and what should they be named? FOUND, NOT_FOUND, NOT_ALLOWED?

damianopetrungaro: Yup, class constant or Enum will be more explicit so for me both are good.

damianopetrungaro: 1: `FOUND`, `NOT_FOUND`, `NOT_ALLOWED` sounds good to me.
2: You use them in `DispatcherResults` so it may be the right place (imho use an Enum and inject it is the best approach)

l0gicgate: Actually, no need to make changes. This PR extends FastRoute\Dispatcher\GroupCountBased class which implements  FastRoute\Dispatcher interface which has those constants. They can be used as Slim\Dispatcher::FOUND, Slim\Dispatcher::NOT_FOUND, Slim\Dispatcher::NOT_ALLOWED

I will amend the comment in DispatcherResults to reference where those integers come from.

Slim/Dispatcher.php
l0gicgate: If anywhere, this should be on the Slim\Http package.

tests/RouterTest.php
```diff
@@ -205,19 +207,19 @@ public function testCreateDispatcher()
...
+        $this->assertInstanceOf('\Slim\Dispatcher', $method->invoke($this->router));
     }
 
     public function testSetDispatcher()
     {
         $this->router->setDispatcher(\FastRoute\simpleDispatcher(function ($r) {
             $r->addRoute('GET', '/', function () {
             });
-        }));
+        }, ['dispatcher' => '\\Slim\\Dispatcher']));
```
damianopetrungaro: Use `Dispatcher::class` instead of hardcoded namespace, if it will change we have to change it in a lot of places

tests/RouterTest.php
```diff
@@ -362,7 +364,7 @@ public function testRouteCacheFileCanBeDispatched()
         $method->setAccessible(true);
 
         $dispatcher = $method->invoke($this->router);
-        $this->assertInstanceOf('\FastRoute\Dispatcher', $dispatcher);
+        $this->assertInstanceOf('\Slim\Dispatcher', $dispatcher);
```
damianopetrungaro: @l0gicgate  Here namespace too 😄

tests/RouterTest.php
```diff
@@ -205,19 +207,19 @@ public function testCreateDispatcher()
...
         $this->router->setDispatcher(\FastRoute\simpleDispatcher(function ($r) {
             $r->addRoute('GET', '/', function () {
             });
-        }));
+        }, ['dispatcher' => Dispatcher::class]));
         $class = new \ReflectionClass($this->router);
         $prop = $class->getProperty('dispatcher');
         $prop->setAccessible(true);
-        $this->assertInstanceOf('\FastRoute\Dispatcher', $prop->getValue($this->router));
+        $this->assertInstanceOf('\Slim\Dispatcher', $prop->getValue($this->router));
```
damianopetrungaro: @l0gicgate  Here namespace too 😄

tests/RouterTest.php
```diff
@@ -205,19 +207,19 @@ public function testCreateDispatcher()
         $class = new \ReflectionClass($this->router);
         $method = $class->getMethod('createDispatcher');
         $method->setAccessible(true);
-        $this->assertInstanceOf('\FastRoute\Dispatcher', $method->invoke($this->router));
+        $this->assertInstanceOf('\Slim\Dispatcher', $method->invoke($this->router));
```
damianopetrungaro: @l0gicgate  Here namespace too 😄

tests/DispatcherTest.php
```diff
@@ -0,0 +1,625 @@
...
+class DispatcherTest extends TestCase
+{
+    protected function getDispatcherClass()
+    {
+        return 'Slim\\Dispatcher';
+    }
+
+    protected function getDataGeneratorClass()
+    {
+        return 'FastRoute\\DataGenerator\\GroupCountBased';
```
damianopetrungaro: Are we including it as dependency?
We may use namespace here too instead of hardcoded 😄

tests/DispatcherTest.php
```diff
@@ -0,0 +1,625 @@
...
+
+/**
+ * Class DispatcherTest
+ * @package Slim\Tests
+ */
+class DispatcherTest extends TestCase
+{
+    protected function getDispatcherClass()
+    {
+        return 'Slim\\Dispatcher';
```
damianopetrungaro: @l0gicgate  Here namespace too 😄

Slim/DispatcherResults.php
l0gicgate: This is what is currently being done https://github.com/slimphp/Slim/blob/4.x/Slim/Middleware/RoutingMiddleware.php#L66

Slim/Dispatcher.php
l0gicgate: As per one of our previous discussions, this avoids the initial performance hit of looping through every single route. It is being returned on demand instead of initially doing it every time we dispatch. This is also the way Fastroute does it.

Slim/Middleware/RoutingMiddleware.php
l0gicgate: What kind of exception would you like to be thrown here? Just a regular exception or RuntimeException

Slim/DispatcherResults.php
akrabat: What I mean is that currently in Slim 3 it's not optional. Why is the choice a method parameter here?

Slim/Middleware/RoutingMiddleware.php
akrabat: Runtime is fine as it should never happen :)

Slim/Dispatcher.php
akrabat: I understand that. 

Given `getAllowedMethods()` has been called, why don't we keep a copy of the calculated `$allowedMethods` in a local member variable, so that we don' have to do all the work again if `getAllowedMethods()` is called again with exactly the same parameters?

From what I can tell if we're in a Method Not Allowed situation, we call `getAllowedMethods()` on line 68 of `Dispatcher.php` and then again in line 83 of RoutingMiddleware.php, so it appears to me that we're doing the same work twice.

Slim/Router.php
```diff
@@ -213,9 +212,7 @@ public function map($methods, $pattern, $handler)
      *
      * @param  ServerRequestInterface $request The current HTTP request object
      *
-     * @return array
-     *
-     * @link   https://github.com/nikic/FastRoute/blob/master/src/Dispatcher.php
+     * @return DispatcherResults
```
akrabat: We need to add

    $uri = rawurldecode($uri);

before calling the dispatcher's `dispatch()` in this method. Not sure if it should be a separate PR or not.

Slim/DispatcherResults.php
l0gicgate: As per our discussion on slack, we're leaving this as is.

Slim/Router.php
l0gicgate: This commit cc0c0b3 applies the requested change.

damianopetrungaro:
@l0gicgate btw before merging could you please squash all your commits?

l0gicgate:
@damianopetrungaro I'm not the one merging, @akrabat will take care of that!

damianopetrungaro:
You can do it also on your local branch:
`git rebase -i HEAD~{number of commit to squash}` and clean a little bit the history for this PR 😄

akrabat:
FWiW, I am not a fan of rebasing to clean history. Rebasing to remove a commit that's a typo fix is fine, but I personally prefer to see the route that was taken to get to where we are now.

Slim/Dispatcher.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * @param string $httpMethod
+     * @param string $uri
+     * @return array
+     */
+    public function getAllowedMethods($httpMethod, $uri)
+    {
+        if ($this->allowedMethods !== null) {
```
akrabat: This will give the wrong answer if I do this:

    $methods1 = $o->getAllowedMethods('POST', '/foo');
    $methods2 = $o->getAllowedMethods('GET', '/bar');

At least, I think that `$methods2` will be the allowed methods for `/foo`

l0gicgate: You are correct, I made a mistake. I fixed this logic in latest commit. Good find!

akrabat (approved):

damianopetrungaro:
Great work @l0gicgate !

PR #2407, opened 2018-02-24
Allow echo if output buffering disabled

fatcat386:
Hi!
Please consider this PR. After upgrading from 3.9.0 to 3.9.1 i face problem of rendering non-HTML content with some PDF libraries which set headers with header() and then echo() themselves, so i not return $response from controller. But default $response exists and overwriting Content-Type header until i disable outputBuffering. It all works until 3.9.1 where ob_start is used in App::run() constantly. I don't want to catch PDF-output and put it to $response as it is complicate things and also overhead mem. Keep in mind other libraries.
Thank you!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -290,26 +290,34 @@ class App
     public function run($silent = false)
     {
         $response = $this->container->get('response');
+        $outputBuffering = $this->container->get('settings')['outputBuffering'];
 
-        try {
-            ob_start();
-            $response = $this->process($this->container->get('request'), $response);
-        } catch (InvalidMethodException $e) {
-            $response = $this->processInvalidMethod($e->getRequest(), $response);
-        } finally {
-            $output = ob_get_clean();
-        }
+        if ($outputBuffering) {
+            try {
+                ob_start();
+                $response = $this->process($this->container->get('request'), $response);
+            } catch (InvalidMethodException $e) {
+                $response = $this->processInvalidMethod($e->getRequest(), $response);
+            } finally {
+                $output = ob_get_clean();
+            }
 
-        if (!empty($output) && $response->getBody()->isWritable()) {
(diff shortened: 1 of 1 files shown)


odan:
The unit tests are failing here:

```
1) Slim\Tests\AppTest::testExceptionOutputBufferingOff

RuntimeException: Unexpected data in output buffer. Maybe you have characters before an opening <?php tag?
/home/travis/build/slimphp/Slim/Slim/App.php:612
/home/travis/build/slimphp/Slim/Slim/App.php:324
/home/travis/build/slimphp/Slim/tests/AppTest.php:2245
/home/travis/build/slimphp/Slim/vendor/phpunit/phpunit/phpunit:52
```

akrabat:
The change to output buffering came about because of https://github.com/slimphp/Slim/issues/2208 which was fixed with https://github.com/slimphp/Slim/pull/2270. Your PR breaks this fix.

As your PDF library is setting headers and echo'ing directly, you should probably just `exit` after calling it.

Another possible solution would be to modify `Route::__invoke()` to use `headers_list()`to add the headers set by the library to the response if `$newResponse` is a string.

Can you open an issue with an example showing how to reproduce the problem?

fatcat386:
Thanks for your answer!

To reproduce problem we need just call this from controller:
`header('Content-Type: application/pdf');`
`echo SOME_PDF_CONTENT;`
After that we can get two possible results:
1. page with raw pdf code if outputBuffering is 'append' or 'prepend' and additional header from default response which break application/pdf.
2. a blank page if outputBuffering is false, and additional header from default response which break application/pdf. We lose all echos here starting from 3.9.1, earlier it works and headers was correct as default response can't mess them after echo.

But nevermind, `exit` helps me, shame on me)
Maybe it must be in documentation somewhere)

PR #2409, opened 2018-02-25
Add support of routes, that contain national characters

SmiSoft:
Add support of routes, that contain national (for me - Russian) characters.

Preface

Some years ago, most browsers began to support national characters in URI, for example:
http://somesite.ru/новости
("новости" is "news" in Russian). This URIs are percent-encoded https://en.wikipedia.org/wiki/Percent-encoding and they are more user-friendly.

Problem

Slim cannot use such uris in routes, because "raw" uri will appears as:
/%D0%BD%D0%BE%D0%B2%D0%BE%D1%81%D1%82%D0%B8
so the plain pattern:
$app->get('/новости', function ($request, $response, $args) {
  // some code
});
won't work.

Solutions

One of the possible solutions is to url-encode all patterns, but in such case, some special for patterns characters, like curly braces will be encoded too early and the pattern wouldn't work.
I found a balanced solution: encode only national characters and don't touch symbols from lower half of ANSI table, this will guarantee, that pattern, that didn't use national characters will work as before, but most routes with national characters (and national characters in combination with safe characters, like latin characters, numbers, underscores etc.) will work perfect too.

Advantage

Routes with national characters will work.

Disadvantage

- A source-file, that defines routes (index.php by default) MUST be UTF-8 encoded. In most cases, this is true, so it's tiny flaw.
- Some complex routes like /новости{1} won't work, but I never saw like uris on real internet sites, so I don't think, that it's a great flaw.

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -76,6 +76,33 @@ class Route extends Routable implements RouteInterface
     protected $callable;
 
     /**
+     * Convert hexadecimal number (range 0..15) into uppercased hex digit (0..9, A..F)
+     * @param $byte number in range 0..15
+     * @return uppercased hex digit (0..9, A..F) or underscore (in case of error)
+     */
+    public static function digit2hex($byte){
+        if(($byte>=0) && ($byte<=9))
+            return chr($byte+ord('0'));
+        if(($byte>=10) && ($byte<=15))
+            return chr($byte+ord('A')-10);
+        return '_';
+    }
+
+    /**
+     * Convert symbols from higher half of ANSI table into it's percent-encoded presentation
+     * @param $str UTF-8 encoded string
+     * @return string with higher half of ANSI table percent-encoded
+     */
+    public static function patternEncode($str){
+        $arr=str_split($str);
+        foreach ($arr as &$value) {
+            if(ord($value)>127)
+                $value='%' . self::digit2hex(ord($value) >> 4) . self::digit2hex(ord($value) & 0xF);
+        }
+        return implode($arr);
+    }
+
+    /**
      * Create new route
      *
      * @param string|string[]   $methods The route HTTP methods
@@ -87,7 +114,7 @@ class Route extends Routable implements RouteInterface
     public function __construct($methods, $pattern, $callable, $groups = [], $identifier = 0)
     {
         $this->methods  = is_string($methods) ? [$methods] : $methods;
-        $this->pattern  = $pattern;
+        $this->pattern  = self::patternEncode($pattern);
         $this->callable = $callable;
         $this->groups   = $groups;
         $this->identifier = 'route' . $identifier;
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -80,11 +80,14 @@ class Route extends Routable implements RouteInterface
      * @param $byte number in range 0..15
      * @return uppercased hex digit (0..9, A..F) or underscore (in case of error)
      */
-    public static function digit2hex($byte){
-        if(($byte>=0) && ($byte<=9))
+    public static function digit2hex($byte)
+    {
+        if (($byte >= 0) && ($byte <= 9)) {
             return chr($byte+ord('0'));
-        if(($byte>=10) && ($byte<=15))
+        }
+        if (($byte >= 10) && ($byte <= 15)) {
             return chr($byte+ord('A')-10);
+        }
         return '_';
     }
 
@@ -96,8 +99,9 @@ class Route extends Routable implements RouteInterface
     public static function patternEncode($str){
         $arr=str_split($str);
         foreach ($arr as &$value) {
-            if(ord($value)>127)
+            if (ord($value) > 127) {
                 $value='%' . self::digit2hex(ord($value) >> 4) . self::digit2hex(ord($value) & 0xF);
+            }
         }
         return implode($arr);
     }


odan:
This issue is a nice example of how the router should be implemented as middleware in the future. Then you would have the possibility to add some kind of "InternationalRouteMiddleware" one step ahead of the core router into the stack. As a result, the core framework could be kept "slim" and the custom routing middleware could be placed where and when necessary.

akrabat:
> This issue is a nice example of how the router should be implemented as middleware in the future. Then you would have the possibility to add some kind of "InternationalRouteMiddleware" one step ahead of the core router into the stack. As a result, the core framework could be kept "slim" and the custom routing middleware could be placed where and when necessary.

Slim 4: has RoutingMiddleware: https://github.com/slimphp/Slim/blob/4.x/Slim/Middleware/RoutingMiddleware.php

akrabat:
At first glance, this need to be opt-in. Ideally, it would be a piece of middleware that runs over all the added patterns and converts them. I'm not sure that's possible with the current access in the router though, but @l0gicgate may know.

SmiSoft:
Sorry, but I think, you are wrong. For example, authentification module should be external module (not included in kernel), because there exists site, that don't require authentification. The database support module may be external, there are not many sites, that don't use databases, but they may exist. The same is ORM, template engine, captha generator, logger and many other useful modules. But as for me, UTF-8 router must be included in kernel unconditionally, because most browsers (or we won't count IE as a browser?) encode URL in UTF-8 and it is a de-facto standart of the internet.

odan:
Hi @SmiSoft The slim router uses [nikic/FastRoute](https://github.com/nikic/FastRoute) and my first thought would have been to solve the problem right there. But then I found this issue [154](https://github.com/nikic/FastRoute/issues/154). 

Nikic is unfortunately not willing to solve this problem within his library. Here are the reasons:

> FastRoute expects a decoded URL. As shown in the README, this can be achieved using a call to rawurldecode().

Then I tried this litte workaround and it works :-)

```php
$app->get('/' . rawurlencode('новости'), function ($request, $response) {
    // it works
});
```

Unfortunately, this is not mentioned in the documentation of FastRoute. Apparently in Slim, the regex pattern, is not automatically urlencoded. See [Route.php](https://github.com/slimphp/Slim/blob/3.x/Slim/Route.php#L90), which could be considered a bug. 

But I also believe that such a major change in the encoding of Slim could lead to compatibility problems with existing projects. The problem is that suddenly something doesn't work anymore seems to me very big.

Here is a shorter function to encode the pattern:

```php
function encodeUrlPattern($url)
{
    return preg_replace_callback('~' . '[\x7F-\xFF]' . '~', function ($match) {
        return urlencode($match[0]);
    }, $url);
}
```

Edit: Maybe this should be documented in Slim too?

akrabat:
> Sorry, but I think, you are wrong.

Fair points.

However, how do we do it that it's not a performance hit for people using ASCII only routes?

l0gicgate:
I'm not in favor of the proposed fix as I believe there would be a performance hit having to iterate over the route patterns and replace characters versus using a regex tree like **Symfony 4 Router** does.

odan:
The Symfony 4 RouterCompiler has a option for 'utf8' encoded routes.

https://github.com/symfony/routing/blob/master/RouteCompiler.php#L103

Maybe Slim could implement such a new option too?

SmiSoft:
@odan: thanks for **constructive** comment. Certainly, FastRoute shouldn't be affected by this change. URLEncoding is not an option, I told in my first comment to pull request:
> One of the possible solutions is to url-encode all patterns, but in such case, some special for patterns characters, like curly braces will be encoded too early and the pattern wouldn't work.
So, I offer my solution. Rather controversial, but this problem should be raised.
> Here is a shorter function to encode the pattern
A great solution. I test it here: http://pastebin.ru/X2AQrqcs Results:

Input data: nothing to encode
encodeUrlPattern=0.15625 225000
patternEncode=2.0625 225000
urlencode=0.078125 245000

Input data: very many chars to encode
encodeUrlPattern=6.3125 825000
patternEncode=19.546875 825000
urlencode=0.078125 955000

So, your function is 3...10 times faster, than mine. But it cannot be compared to standart encoder, that is more than 280 times faster. Tests was performed on my ten years old notebook, but the absolute value was not the target, only relative values.

>performance hit for people using ASCII only routes
6.3 seconds for 5000 routes (for @odan solution, which is faster)... Reasonable. Maybe, it's really worth to enable this optionally? But router (where this option logically required) has no access to DI container with 'settings'. The only visible way is to move patternEncode to App\map method. Maybe, I missed something?

After some time, I found an intresting thing. In Uri.php file I found the function filterPath:
https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Uri.php#L642
I add a rawurldecode to the call:
>return rawurldecode(preg_replace_callback(
Perfect! Russian UTF-8 route works perfect at this point. To complete change, it seems, that this foreach should be commented:
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L569
but I'm not sure. More, I don't found pitfall of this solution, they should be, as my ass sense :-)

I think, it's worth to discuss this kind of solution (maybe, it even worth to close this pull request and switch it to Issues branch?).

SmiSoft:
@odan: thanks for **constructive** comment. Certainly, FastRoute shouldn't be affected by this change. URLEncoding is not an option, I told in my first comment to pull request:
> One of the possible solutions is to url-encode all patterns, but in such case, some special for patterns characters, like curly braces will be encoded too early and the pattern wouldn't work.

So, I offer my solution. Rather controversial, but this problem should be raised.
> Here is a shorter function to encode the pattern

A great solution. I test it here: http://pastebin.ru/X2AQrqcs Results:

Input data: nothing to encode
encodeUrlPattern=0.15625 225000
patternEncode=2.0625 225000
urlencode=0.078125 245000

Input data: very many chars to encode
encodeUrlPattern=6.3125 825000
patternEncode=19.546875 825000
urlencode=0.078125 955000

So, your function is 3...10 times faster, than mine. But it cannot be compared to standart encoder, that is more than 280 times faster. Tests was performed on my ten years old notebook, but the absolute value was not the target, only relative values.

>performance hit for people using ASCII only routes

6.3 seconds for 5000 routes (for @odan solution, which is faster)... Reasonable. Maybe, it's really worth to enable this optionally? But router (where this option logically required) has no access to DI container with 'settings'. The only visible way is to move patternEncode to App\map method. Maybe, I missed something?

After some time, I found an intresting thing. In Uri.php file I found the function filterPath:
https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Uri.php#L642
I add a rawurldecode to the call:
>return rawurldecode(preg_replace_callback(

Perfect! UTF-8 route works perfect at this point. To complete change, it seems, that this foreach should be commented:
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L569
but I'm not sure. More, I don't found pitfall of this solution, they should be, as my ass sense :-)

I think, it's worth to discuss this kind of solution (maybe, it even worth to close this pull request and switch it to Issues branch?).

akrabat:
> I think, it's worth to discuss this kind of solution (maybe, it even worth to close this pull request and switch it to Issues branch?).

Please yes. I'd love to know if there's any BC issues with the rawurldecode proposal.

Issue #2391, opened 2018-02-04
Slim always removes port from Host: header


garnold:
It appears that Slim is always overwriting the `Host:` header and dropping the port. [I believe the `||` should be an `&&` here](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L200).

geggleto:
Hmm. I actually can't think of a reason why the first condition of that if would ever be false.

garnold:
@geggleto Right, but it's an `||`. The first condition is likely to be `false`, but the second condition is likely to be `true`.

PR #2411, opened 2018-02-28
Bug fix: Stop Slim overwriting Host header if it's defined (#2391)

imjoehaines:
See #2391

The previous condition skipped checking if the Host header was already set and would overwrite it with the value from the given URI.

This fixes it to only add the Host header if it's missing and also appends the port if it's non-standard.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -197,8 +197,10 @@ class Request extends Message implements ServerRequestInterface
             $this->protocolVersion = str_replace('HTTP/', '', $serverParams['SERVER_PROTOCOL']);
         }
 
-        if (!$this->headers->has('Host') || $this->uri->getHost() !== '') {
-            $this->headers->set('Host', $this->uri->getHost());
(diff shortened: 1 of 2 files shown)


Slim/Http/Request.php
```diff
@@ -197,8 +197,12 @@ public function __construct(
...
         }
 
-        if (!$this->headers->has('Host') || $this->uri->getHost() !== '') {
-            $this->headers->set('Host', $this->uri->getHost());
+        if (!$this->headers->has('Host') && $this->uri->getHost() !== '') {
+            $port = $this->uri->getPort() ? ":{$this->uri->getPort()}" : '';
+
+            $this->headers->set('Host', $this->uri->getHost() . $port);
+
+            unset($port);
```
geggleto: This isn't required

imjoehaines: Removed

akrabat:
Thanks @imjoehaines

PR #2414, opened 2018-03-02
Use rawurldecode instead of urldecode for route argument

juherr:
[urldecode](https://secure.php.net/manual/en/function.urldecode.php) is used in order to decode part of the route arguments but according to the php documentation:
> Plus symbols ('+') are decoded to a space character. 

According to the RFC-1866, the '+' to space conversion is only valid after the '?' part of the url: https://stackoverflow.com/a/40292260/4234729

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -567,7 +567,7 @@ class App
         if ($routeInfo[0] === Dispatcher::FOUND) {
             $routeArguments = [];
             foreach ($routeInfo[2] as $k => $v) {
-                $routeArguments[$k] = urldecode($v);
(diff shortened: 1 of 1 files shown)


juherr:
@slimphp Hi, can I have any feedback on my PR? 
Currently, I am converting space to '+' which is not really logic.

Travis is failing on php7.1 without any reason.

danopz:
Duplicate + BC break for Slim 3: #2000

juherr:
@danopz Oh! I didn't find the duplication, sorry!

PR #2415, opened 2018-03-02
Allow easier extensions to the main App class

zelding:
This seems/feels like a miss-type

Either the property should be protected, or the class should be final.
If the `$container` is private, most of the public base class methods would fail, because there the container hasn't been set

This way people will be able to easily extend the Slim\App class.

If this was a deliberate design-choice than, I'm sorry, but I don't see the point.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -59,7 +59,7 @@ class App
      *
      * @var ContainerInterface
      */
-    private $container;
+    protected $container;
 
     /********************************************************************************
      * Constructor


geggleto:
It was not a mistype. 
You can still get the container if you wish via `$this->getContainer()`. 

The reason being it ensures that apps that extend Slim can't inadvertently break the container without a lot of effort. The framework itself does a lot of bootstrapping like setting framework dependencies in the container.

zelding:
Thank you for the info, I'll do it the _proper_ way than ×)

Issue #2175, opened 2017-03-18
Slim 4 and PHP versions


Sam-Burns:
More of a question than an issue, but what versions of PHP are to be supported by Slim 4?  Has dropping PHP 5 support in the new major version been considered?

akrabat:
The last time we discussed it, the plan was for PHP 5.6+

Sam-Burns:
Ok, well if I can't persuade the team to go PHP 7 only, perhaps [this PR](https://github.com/slimphp/Slim/pull/2178) might be of interest.

RyanNerd:
Please, in the name of sanity make PHP 7+ a requirement for Slim 4.0

akrabat:
Why? What do you gain as a user of the framework?

tabuna:
Require support for PHP 7 reasonably

RyanNerd:
I'm a user of and a contributor to Slim. The benefits of strict typing alone are self-evident plus the other benefits of version 7 are many. 
As a user of the framework, I've had to step through the framework code on multiple occasions to make sure what I was having issues with was my own code and not a flaw in the framework. As an example, PHP 7 has type declarations so expected value types are returned from a function and this helps with the concept of least astonishment. If a function argument is expecting a bool type and a string with the word 'guacamole' is passed instead the issue is immediately obvious -- prior versions instead waste my time forcing me to chase duck typing issues.

If you go even further and insist on 7.1+ you get:
- Scoped constants (which perform better than static declarations)
- Nullable return types
- Negative string offsets
- The void return type
- more

And for any PHP dev that may have been living in a cave for the past two years here are the new features of 7.0:
- Scalar type hints
- Return type declarations
- Anonymous classes
- The Closure::call() method
- Generator delegation
- Generator return expressions
- The null coalesce operator
- The space ship operator
- Throwables
- Level support for the dirname() function
- The Integer division function
- Uniform variable syntax

Sam-Burns:
With basic security support for PHP 5.6 and 7.0 ending next year, and bug-fixing for PHP 5 already gone, it would be nice to see Slim users encourage to upgrade.  For those who can't, maybe a good period of support for Slim 3 might be an option.  Perhaps this could be an alternative to developing the latest version of the Slim framework in an old language version?

RyanNerd:
[...]

PR #2421, opened 2018-04-03
[Slim 4] Add PHP 7.0 type definitions

danopz:
According to #2175 I started adding argument  and response type definitions where the type is clear right now. There are still some cases where we can not define some type right now but shouldn't block this one.

~This PR is build upon #2404.~

### DIFF
--- a/Slim/RoutingResults.php
+++ b/Slim/RoutingResults.php
@@ -6,6 +6,9 @@
  * @copyright Copyright (c) 2011-2018 Josh Lockhart
  * @license   https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
  */
+
+declare(strict_types=1);
+
 namespace Slim;
 
 class RoutingResults
@@ -37,7 +40,7 @@ class RoutingResults
     protected $routeStatus;
 
     /**
-     * @var callable|null
+     * @var null|string
      */
     protected $routeHandler;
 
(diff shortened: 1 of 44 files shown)


Slim/RoutingResults.php
```diff
@@ -49,19 +52,19 @@ class RoutingResults
     /**
      * RoutingResults constructor.
      * @param Dispatcher $dispatcher
-     * @param $httpMethod
-     * @param $uri
+     * @param string $httpMethod
+     * @param string $uri
      * @param int $routeStatus
-     * @param callable|null $routeHandler
+     * @param string|null $routeHandler
```
akrabat, line 58: I'm not sure that the `$routeHandler` is a string.

akrabat: Then again, maybe `$routeHandler` is actually the route's `$identifier`. Can you confirm @l0gicgate ?

l0gicgate: `$routeHandler` refers to the route's identifier that's correct. The variable is in fact a string.

akrabat: Maybe we should rename the variable?

l0gicgate: `$routeIdentifier` sounds pretty descriptive, what do you think?

danopz: Should probably done in an extra PR as this one is just for PHP7

akrabat: > Should probably done in an extra PR as this one is just for PHP7

Yes, that was my assumption and gentle hint to @l0gicgate  :)

PR #2423, opened 2018-04-06
3.x middleware fifo support

sunchuo:
middleware fifo support
add defaultSetting:
'middlewareFifo' => false

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -57,6 +57,7 @@ class Container extends PimpleContainer implements ContainerInterface
         'displayErrorDetails' => false,
         'addContentLengthHeader' => true,
         'routerCacheFile' => false,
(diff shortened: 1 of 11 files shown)


RyanNerd (commented):
There are some files that were changed (minimally) but my question is why change these?
- Slim/Exception/ContainerException.php
- Slim/Exception/ContainerValueNotFoundException.php
- Slim/Exception/NotFoundException.php
- Slim/Http/Body.php

I hope to have some time Sunday to see if I can get your changes to my laptop and do some smoke tests. My initial going through your changes is that the code looks good.

akrabat:
We won't provide a choice unless Slim 4 goes to FIFO and we want to provide a migration path.

PR #2425, opened 2018-04-13
Add a redirect route shortcut

ellotheth:
Per a short conversation with @akrabat.

(I'm blanket overwriting whatever status code the `ResponseInterface` has because I wasn't sure the flexibility in e.g. `Slim\Http\Response::withRedirect` is necessary at this level, but maybe it's worth keeping the behavior the same for consistency?)

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -12,6 +12,7 @@ use Psr\Container\ContainerInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\UriInterface;
 use Slim\Http\Headers;
 use Slim\Http\Request;
 use Slim\Http\Response;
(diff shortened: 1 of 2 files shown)


akrabat (changes requested):
Thanks @ellotheth!

Please add a test that ensures that `UriInterface` as `$destination` also works.

akrabat:
> (I'm blanket overwriting whatever status code the ResponseInterface has because I wasn't sure the flexibility in e.g. Slim\Http\Response::withRedirect is necessary at this level, but maybe it's worth keeping the behavior the same for consistency?)

This is correct. You should only change the response object on the way back anyway.

ellotheth:
@akrabat Righto, done!

akrabat:
Can you also do this against the `3.x` branch?

ellotheth:
Yup, and I'll add the docs. Should be be done by tomorrow night.

akrabat (approved):

PR #2429, opened 2018-04-22
Added StatusCode class

Dmitry-Kucher:
Added status codes enum class to remove hardcoded status code values from Response class.
It's increases readability of code.
Ex.: use `StatusCodes::HTTP_BAD_REQUEST` instead of `400` code and so on.


odan:
Thanks for the PR. :+1: 

`StatusCodes` sounds very (too) generic.
Would it be possible to give it a more specific (and singular) name like: `HttpStatusCode` ?

akrabat:
> `StatusCodes` sounds very (too) generic.
Would it be possible to give it a more specific (and singular) name like: `HttpStatusCode` ?

While I'm not totally convinced on the merit of this yet, the class name is fine as it's in the `Slim\Http` namespace.

pachico:
I would also delete `418 => 'I\'m a teapot'` since wasn't more than an April fools which is not expected to be implemented.

Dmitry-Kucher:
@pachico It was in original [Response file](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Response.php#L93) so to keep consistency I added it to my class. Deletion maybe isn't a good idea if someone wants to use it for Easter egg in API.

odan:
@akrabat I agree, but in the context of a controller / action method the `Http` prefix is not redundant because it's not clear which status code you mean. You first have to check the use statements on top of the PHP file to know what a `StatusCode` is.

Example: Here it is clear:

```php
public function __invoke(Request $request, Response $response) 
{
    return $response->withStatus(HttpStatusCode::NOT_FOUND);
}
```

And here it is not clear:

```php
public function __invoke(Request $request, Response $response) 
{
    return $response->withStatus(StatusCode::NOT_FOUND);
}
```

But anyway, I'd be happy to call the class at least singular. :-)

Dmitry-Kucher:
@odan agreed with you. 100% about singular naming, but not absolutely about HTTP prefix for class name.
In my case code would be different:

```php
public function __invoke(Request $request, Response $response) 
{
    return $response->withStatus(StatusCode::HTTP_NOT_FOUND);
}
```

`HTTP` already presented in constant name. If everyone will agree I can remove HTTP prefix from constant and move it to classname

akrabat:
FWiW, We definitely keep `418`.

akrabat:
This example proves my point:

```
public function __invoke(Request $request, Response $response) 
{
    return $response->withStatus(StatusCode::NOT_FOUND);
}
```

We don't have `HttpRequest`, `HttpResponse`, `HttpUri`, `HttpHeaders` as the other class names in `Slim\Http`.

danopz:
> I would also delete 418 => 'I\'m a teapot' since wasn't more than an April fools which is not expected to be implemented.

We cannot just remove it as it would be a BC break.
Also http://save418.com/

pachico:
Not for 3.*, I agree, but it could for 4.0. In my opinion, it's just a silly and childish joke that got far more attention than it deserved and made its way to libraries implementations (and it clearly states that should NOT be implemented).

odan:
@Dmitry-Kucher Yes, all the `::HTTP_` constant prefixes are quite redundant, because this class is only responsible for HTTP status codes, right? But yes, this depends on the class name of course.

Another question: If at all, should this class not be part of the [slimphp/Slim-Http](https://github.com/slimphp/Slim-Http) package?

akrabat:
> Another question: If at all, should this class not be part of the slimphp/Slim-Http package?

If we take this for Slim 3.x, then we'll need it in Slim-Http too for Slim 4.x.

Dmitry-Kucher:
@odan Am I right that this package is for Slim 4.x, but my PR is for 3.x. I also can make a PR for slimphp/Slim-Http package if needed

Dmitry-Kucher:
> Yes, all the ::HTTP_ constant prefixes are quite redundant, because this class is only responsible for HTTP status codes, right? 

Agreed, but it's not true for all constants. [CONTINUE constant](http://sandbox.onlinephpfunctions.com/code/2ee35ff49c9d117f2f5a234608cddaa720ddbd60) doesn't work for `php < 7.0`

So, maybe `HTTP` is redundant for constant contains more than one word in their names

odan:
> So, maybe HTTP is redundant for constant contains more than one word in their names

@Dmitry-Kucher I would agree with `HTTP_CONTINUE`, but only for technical reasons (PHP < 7.0). The rest should be without `HTTP_` to make it more consistent.

Dmitry-Kucher:
Not sure that it's related only to technical reasons. Statuses `OK`, `FOUND`, `ACCEPTED` and so on looks not so informative.

Even:

```php
HttpStatusCode::OK
```

Looks not so good, because it means something positive, but `FOUND` also means positive and it's not clear what's difference.

```php
StatusCode::HTTP_OK
```

More clear that it's about `200` status.
Maybe it's related only to personal taste

danopz:
> @Dmitry-Kucher I would agree with HTTP_CONTINUE, but only for technical reasons (PHP < 7.0). The rest should be without HTTP_ to make it more consistent.

No, all should be either with or without the HTTP prefix, do not mix them.

clue:
@Dmitry-Kucher I appreciate the effort (both for Slim and for @ReactPHP where we're considering similar suggestions) :+1: 

Maybe this has been discussed before and I'm missing something, but I wonder what's the motivation for moving this to a separate class `StatusCode` (or `HttpStatusCode`)? Has defining them as part of the `Response` class been considered?

It's my understanding that HTTP *response* status codes are likely used in combination with a *response* object, so a separate reference and import seem to be a bit odd to me. Also, if you're looking at sites such as https://httpstatuses.com/200, you can find some references that other libs/languages use. In particular, Symfony happens to use the same `Response::HTTP_OK` naming convention. Personally, I'm not a Symfony user and I'm not saying compatibility is a must, but it's certainly a nice property and probably something worth considering.

Dmitry-Kucher:
@clue I saw Symfony implementation before suggested my PR. I made the separate class for `StatusCode` because it should be a Enum, maybe more suitable would be to use something like [php-enum](https://github.com/myclabs/php-enum), but include huge amount of status codes into response isn't elegant way, that's why my PR looks as it looks.

Another reason why I'm not using [Symfony http-foundation](https://github.com/symfony/http-foundation/blob/master/Response.php#L126) is double standards 😄 Not sure why they are add constants for status codes and not using them in messages

carc1n0gen:
@odan 

```
use Slim\Http\StatusCode as HttpStatusCode
```
😃

silentworks:
Having it in a separate class makes it available to user land code which I think is a good idea. So now in my responses I can make use of it, instead of trying to remember the codes too.

PR #2431, opened 2018-04-27
Bug fix: Reset Route arguments for each call

### LINKED ISSUE #2430: Reset Route's attributes on prepare()
I'm writing some functional tests and I noticed that if I call twice a route like that:
`/view[/{code:[\w-]+}]`

`Route::prapare()` method doesn't clean `Route::attributes`.
Then if the first call has `code` attribute (`/view/hello-world`), `prepare()` method set it in `attributes` field.
If the second call doesn't have `code` attribute (`/view`) when `App` call `Route::prepare()` method with empty `$arguments`, the `code` still remain `hello-world`, breaking my request handler.

Maybe in `Route::prepare()` we could reset its attributes?

mathmarques:
Route arguments weren't reset for each call of `App::process` and if the route had optional arguments on the url this could cause unexpected behaviors.

This Fix #2430

### DIFF
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -69,6 +69,13 @@ class Route extends Routable implements RouteInterface
     protected $arguments = [];
 
     /**
+     * Route arguments parameters
+     *
+     * @var null|array
+     */
+    protected $savedArguments;
+
+    /**
      * The callable payload
      *
      * @var callable
@@ -286,7 +293,15 @@ class Route extends Routable implements RouteInterface
      */
     public function prepare(ServerRequestInterface $request, array $arguments)
     {
-        // Add the arguments
+        // Save the arguments added on routes
+        if (!isset($this->savedArguments)) {
+            $this->savedArguments = $this->getArguments();
+        }
+
+        // Reset the arguments
+        $this->setArguments($this->savedArguments);
+
+        // Add the route arguments
         foreach ($arguments as $k => $v) {
             $this->setArgument($k, $v);
         }
--- a/Slim/Route.php
+++ b/Slim/Route.php
@@ -73,7 +73,7 @@ class Route extends Routable implements RouteInterface
      *
      * @var null|array
      */
-    protected $savedArguments;
+    protected $savedArguments = [];
 
     /**
      * The callable payload
@@ -233,11 +233,15 @@ class Route extends Routable implements RouteInterface
      *
      * @param string $name
      * @param string $value
+     * @param bool $temp
      *
      * @return self
      */
-    public function setArgument($name, $value)
+    public function setArgument($name, $value, $temp = false)
     {
+        if (!$temp) {
+            $this->savedArguments[$name] = $value;
+        }
         $this->arguments[$name] = $value;
         return $this;
     }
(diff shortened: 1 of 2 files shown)


akrabat:
Why is process() called more than once for a given route?

mathmarques:
@akrabat  It's not more than once for a given route, but process can be invoked more than once for a single Slim's instance. And if one route is matched more than once, this problem will occur.
An use case: https://github.com/slimphp/Slim/pull/1688

geggleto (approved):

fiskWasTaken:
This is indeed a problem you will run into if you try to use Slim under PPM/ReactPHP, so this would be appreciated

However spellcheck `Proccess` in your unit test method names, or consider shorter names (`testOptionalArgTransience`, `testOptionalArgTransienceWithSetArguments`)

akrabat (changes requested):
If I call `setArguments()`, then `$savedArguments` needs resetting I think?

mathmarques:
I don't think so @akrabat .
After app initialization the only way to call `setArguments()` is after the route is matched, and I think if you use `setArguments()` at this time it's only for that one request.

akrabat:
This test fails:

```php
    public function testInvokeSequentialProccessAfterAddingAnotherRouteArgument()
    {
        $app = new App();
        $route = $app->get('/foo[/{bar}]', function ($req, $res, $args) {
            return $res->write(count($args));
        })->setArgument('baz', 'quux');

        // Prepare request and response objects
        $env = Environment::mock([
            'SCRIPT_NAME' => '/index.php',
            'REQUEST_URI' => '/foo/bar',
            'REQUEST_METHOD' => 'GET',
        ]);
        $uri = Uri::createFromEnvironment($env);
        $headers = Headers::createFromEnvironment($env);
        $cookies = [];
        $serverParams = $env->all();
        $body = new RequestBody();
        $req = new Request('GET', $uri, $headers, $cookies, $serverParams, $body);
        $res = new Response();

        // Invoke process with optional arg
        $resOut = $app->process($req, $res);

        $this->assertInstanceOf('\Psr\Http\Message\ResponseInterface', $resOut);
        $this->assertEquals('2', (string)$resOut->getBody());

        // Prepare request and response objects
        $env = Environment::mock([
            'SCRIPT_NAME' => '/index.php',
            'REQUEST_URI' => '/foo/bar',
            'REQUEST_METHOD' => 'GET',
        ]);
        $uri = Uri::createFromEnvironment($env);
        $headers = Headers::createFromEnvironment($env);
        $cookies = [];
        $serverParams = $env->all();
        $body = new RequestBody();
        $req = new Request('GET', $uri, $headers, $cookies, $serverParams, $body);
        $res = new Response();

        // add another argument
        $route->setArgument('one', '1');

        // Invoke process with optional arg
        $resOut2 = $app->process($req, $res);

        $this->assertInstanceOf('\Psr\Http\Message\ResponseInterface', $resOut2);
        $this->assertEquals('3', (string)$resOut2->getBody());
    }
```

Arguably, it's unlikely to happen, but this test works with `3.x` today.

mathmarques:
This change should fix it.
Couldn't think another way to solve it. Any better solution?

akrabat:
> Couldn't think another way to solve it. Any better solution?

Seems reasonable to me. I've renamed the variable to make it a bit clearer though.

mathmarques:
@akrabat Should I patch this on 4.x? Or maybe raise an issue for review every change on 3.x tests since 4.x was created and add them to 4.x and fix it?

akrabat:
@mathmarques Good point - this needs to go to 4.x too.

A proportion of them have been ported forward, but it'd be good to go through and check.

PR #2434, opened 2018-05-08
308 is also a redirect code

sjinks:
Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location,

> Status of responses including a Location header: 201, 301, 302, 303, 307, 308

I am not sure 201 is appropriate here but 308 was missing from the list.


silentworks (approved):

geggleto (approved):

akrabat:
Thanks.

Are you able to forward-port to [Slim-Http](https://github.com/slimphp/Slim).

PR #2439, opened 2018-05-15
Update docbock for Request::getParam() second param's type

hhovakimyan:
`getParam` method `$default` argument type is marked as string, but actually any type of parameters can be passed for `$default` argument and everything will work. That's why I changed it's type to `mixed`


akrabat (approved):

akrabat:
Thanks.

PR #2441, opened 2018-05-16
Bug fix: Replace header when sending first one of each header

### LINKED ISSUE #2246: Cannot remove Cache-Control: no-store and Pragma: no-cache
Hi,
I'm trying to serve files by using PHP (Only logged in users can download them).

I'm trying to make caching to work, but Slim 3 seems to force some headers over my own. The two headers are Cache-Control and Pragma. I'm setting my own Cache-Control  a no-cache header + cache header is sent. I tried removing both by using withoutHeader('Pragma') and  withoutHeader('Cache-Control') but this is completely ignored.

I installed  https://www.slimframework.com/docs/features/caching.html, and I can define the new headers by using this middleware, but the browser is still not caching as it is still receiving the double headers saying it should not cache anything.

I'm not using any other middlewares than the ones that come by default with Slim 3.

How can I disable Slim 3 to prevent caching??

Greetings and thanks.

### LINKED ISSUE #2282: Unable to overwrite default HTTP headers in response
### The problem

I try to make a response cached by a browser. To do it I run:

```php
$app->get('/test', function ($request, $response, $arguments) {
    return $response
        ->withHeader('Cache-Control', 'public, max-age=86400')
        ->write('Hello');
});
```

When I send a request to the route I get a response with such headers:

```
Cache-Control: no-store, no-cache, must-revalidate
Cache-Control: max-age=86400, public
```

As a result, the browser doesn't cache the response.

### Investigation

I've found that Slim sends headers using this code (`Slim\App::respond`):

```php
// Headers
foreach ($response->getHeaders() as $name => $values) {
    foreach ($values as $value) {
        header(sprintf('%s: %s', $name, $value), false);
    }
}
```

The second argument of the `header` function stands for «do not remove previous and default headers» being equal `false`. That's why browser gets the default headers in the response.

To workaround this I had to add `header_remove('Cache-Control')` to the route code. I believe this is a dirty hack.

### Suggested solution

Call the `header` function with the second argument equals `true` for the first value of the header and equals `false` for the rest:

```php
// Headers
foreach ($response->getHeaders() as $name => $values) {
    foreach ($values as $index => $value) {
        header(sprintf('%s: %s', $name, $value), $index === 0);
    }
}
```

I would make a pull request if you agree.

P.S. There is a similar issue here: https://github.com/slimphp/Slim/issues/2246. There is no solution in that issue.

akrabat:
When we call `header()` to send the headers in the Response, set the
`replace` parameter to true for the first of that header name and then
set it to false.

Don't do this if the header is Set-Cookie so we don't break sessions.

Fixes #2282
Fixes #2246

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -26,10 +26,31 @@ use Slim\Http\RequestBody;
 use Slim\Http\Response;
 use Slim\Http\Uri;
 use Slim\Router;
+use Slim\HeaderStackTestAsset;
 use Slim\Tests\Mocks\MockAction;
+
+/**
+ * Emit a header, without creating actual output artifacts
+ *
+ * @param string $value
+ */
+function header($value, $replace = true)
+{
+    \Slim\header($value, $replace);
+}
 
 class AppTest extends \PHPUnit_Framework_TestCase
 {
+    public function setUp()
+    {
+        HeaderStackTestAsset::reset();
+    }
+
+    public function tearDown()
+    {
+        HeaderStackTestAsset::reset();
+    }
+
     public static function setupBeforeClass()
     {
         // ini_set('log_errors', 0);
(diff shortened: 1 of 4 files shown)


ncou:
Hi,

just my 2 cents opinion, but i believe this modification could break some applications who use multiple header. I think for example to the header Content-Security-Policy you could use multiple headers with cumulative policies. With this modification only the first header will be sent. But this case is probably rare...

Keep up the good work.

tests/AppTest.php
```diff
@@ -1742,6 +1763,80 @@ public function testResponseWithStreamReadYieldingLessBytesThanAsked()
         $this->expectOutputString(str_repeat('.', Mocks\SmallChunksStream::SIZE));
     }
 
+    public function testResponseReplacesPreviouslySetHeaders()
```
damianopetrungaro, line 1766: wdyt about using a `dataProvider` and trying to pass the test using "strange" values too?

akrabat:
@ncou That's not what this patch does. It will allow multiple `Content-Security-Policy` headers if you add them to the `Request` object. 

This patch replaces headers set by `header()` if the same header name is used in a header in the `Request` object.

ncou:
hummm i think i am starting to understand (or perhaps not :)). In fact this patch will replace (one time) the header if it's already present in the header() list. To sumup it will overwrite the possible headers added by the server (like for example the "Cache-Control" header). And for the Set-Cookie it will do an exception and not replace the first occurrence, because it could be added by the server (i suppose it's to handle the case with the session cookies).
Am i right ?

akrabat:
@ncou pretty much, yes.

ncou:
thank you it's more clear. :+1:

Issue #2065, opened 2016-11-20
Update Twig-view / PHP-View


geggleto:
- Need to update Twig/PHP -view for 4.x

dopesong:
No description provided. ?

1f7:
Suggest add array to items collection
https://github.com/slimphp/Twig-View/pull/66
need to collect items from different classes that not have Response
not important as it will be called `offsetAdd` or `addGlobal` or like on another

pine3ree:
Hello @geggleto, @codeguy 
It would be nice to have a View template renderer interface and a different signature for the `render` method:
```php
public function render($template, array $data = []);
```
I believe that a template renderer should not be aware of nor interact with `Psr\Http\Message\ResponseInterface`.
The response can be written to directly inside middleware using the response interface methods. Something like this:
```php
return $response->write(
    $this->view->render('/my/template/file.phtml', [
        //..key=>value pairs here
    ]);
);

//or

return new HtmlResponse($this->view->render('/my/template/file.phtml', [/*...*/]);
```

kind regards

akrabat:
Issues need to be opened on relevant repositories for Slim 4 updates.

PR #2445, opened 2018-05-21
Slim4 RendererInterface 

jonathankowalski:
As proposed here https://github.com/slimphp/Slim/issues/2065 Slim4 should provides a renderer interface which is not aware of PSR Response.
This is not the renderer's responsability to update Response.


akrabat:
Can you explain why you  are proposing this please?

jonathankowalski:
Sure ! Due to this issue : https://github.com/slimphp/Slim/issues/2065

designermonkey:
It's a little bit of a bad design to add an interface to a system that has no implementations, and no place of use; it just sits there with no real explanation of what it's for.

As this is purely related to views, and that views are not part of the core package, I feel this would be better placed in view packages.

While I appreciate the reasoning behind it, I'm against this going into the core.

PR #2447, opened 2018-05-25
array_merge was re-indexing arrays with numerical index

yolcuiskender:
turning array keys of post requests like.

Array replace fixes this bug.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1211,7 +1211,7 @@ class Request extends Message implements ServerRequestInterface
         $params = $this->getQueryParams();
         $postParams = $this->getParsedBody();
         if ($postParams) {
-            $params = array_merge($params, (array)$postParams);
+            $params = array_replace($params, (array)$postParams);
         }
 
         if ($only) {
--- a/tests/Http/RequestTest.php
+++ b/tests/Http/RequestTest.php
@@ -213,6 +213,26 @@ class RequestTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(['foo' => 'bar'], $request->getParsedBody());
     }
 
+    public function testCreateFromEnvironmentWithMultipartAndQueryParams()
+    {
+        $_POST['123'] = 'bar';
+
+        $env = Environment::mock([
+            'SCRIPT_NAME' => '/index.php',
+            'REQUEST_URI' => '/foo',
+            'REQUEST_METHOD' => 'POST',
+            'QUERY_STRING' => '123=zar',
+            'HTTP_CONTENT_TYPE' => 'multipart/form-data; boundary=---foo'
+        ]);
+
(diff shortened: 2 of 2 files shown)


ghost:
Hello! Thanks for the PR. We need a test reproducing the error as you have witnessed it and then we can assess whether the fix is right.

The only issue I see with the proposed fix is what would happen if the query params had a key that was the same as the body params? Merging would of course merge, but replace would favour body over query.

We would have to decide what was the expected and favoured outcome.

yolcuiskender:
Hello John!
Thank you very much for the extended comments on my request, I truly appreciate it.

I'm testing it with following code::
routes.php

    $app->post('/test/{foo}', function (Request $request, Response $response, array $args) {
    
        $data_post = $request->getParams();
        $data_query = $request->getQueryParams();
    
        echo '<pre>';
    
        echo '<h2>getParams</h2>';
        print_r($data_post);
        echo '<h2>getQueryParams</h2>';
        print_r($data_query);
    
        echo '<h2>$_POST</h2>';
        print_r($_POST);
        echo '<h2>$_GET</h2>';
        print_r($_GET);
    
        echo '</pre>';
    
        die();
    
    });

Here are the results:
**Current:**
current-merge

**Proposed solution:**
solution-replace

The current solution already mixes the wrong parameters into getParams. As you can see there are the keys 0 and 1 inside getParams which in reality are 123 from query params as 0 and  and 123 from  body as 123 which have lost their numerical indexes.

The proposed solution currently favours the post array but I don't think this should be an issue in this case as it wasn't functioning properly in the previous version.

I'm using slim for a long time but obviously you guys are the ones with more experience in this project so I always consider that I might be wrong. 

Nevertheless I'll be more than happy to help!

Thank you

ghost:
I just realised I'm commenting with the wrong account. Will switch to @designermonkey next time ;)

designermonkey:
Right, I understand the issue now. This is a bit of an edge case.

@akrabat what are your thoughts here? Should we replace, or cater for an edge case?

silentworks:
The proposed solution look like the way to go. I'm just wondering if there are any side-effects to making that change.

akrabat:
This feels like a bug and I'm surprised no one has ever noticed this before. I suppose it's because most people don't use purely numeric key names.

Personally, I think we should fix this.

akrabat (changes requested):
Please provide tests

akrabat:
@yolcuiskender Are you likely to get back to this?

yolcuiskender:
@akrabat yes sir, I'm just having problems writing the test. I didn't want to disturb. I'll be writing the correct test once I've figured out.

yolcuiskender:
I've added the tests sir.

ghost:
Thanks @yolcuiskender.

yolcuiskender:
Thank you too sir, sorry for keep you waiting.

PR #2462, opened 2018-07-06
Add createFromPsr7Request method

MilesChou:
so that we can use `Slim\Request` simply on the other framework.

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -157,6 +157,26 @@ class Request extends Message implements ServerRequestInterface
     }
 
     /**
+     * Create new HTTP request form PSR-7 request
+     *
+     * @param  ServerRequestInterface $request The PSR-7 request
+     *
+     * @return static
+     */
+    public static function createFromPsr7Request(ServerRequestInterface $request)
+    {
+        return new static(
+            $request->getMethod(),
+            $request->getUri(),
+            new Headers($request->getHeaders()),
+            $request->getCookieParams(),
+            $request->getServerParams(),
+            $request->getBody(),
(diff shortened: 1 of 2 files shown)


akrabat:
Can you explain more about the use-case for this change please?

MilesChou:
In my case,  I want to refactoring from Slim, and using the following code to transfer Laravel request to Slim:

```php
// Laravel 5.6
Route::any('/{any}', function (SlimApp $app, ServerRequestInterface $request) {
    $slimRequest = Request::createFromPsr7Request($request);

    return $app($slimRequest, new Response());
})->where('any', '.*')->name('slim');
```

Because the legacy code is:

```php
public function __invoke(Request $request, Response $response)
{
    // \Slim\Request
    $params = $request->getParams();

    // ...
}
```

danopz:
For me there are the interface referenced, so you should be able to use existing PSR7 requests.

https://github.com/slimphp/Slim/blob/dcfa982165d6583305f83b7688e37cee5e34a8bc/Slim/App.php#L497

Or is it some custom class?

MilesChou:
I used custom class:

```php
class MyHandler
{
    public function __invoke(Request $request, Response $response)
    {
        // \Slim\Request
        $params = $request->getParams();

        // ...
    }
}

$slimApp->get('/foo', new MyHandler());

// will see the fatal error: "method getParams not defined"
$slimApp(new StandardPsr7Request(), new StandardPsr7Response());
```

akrabat:
I don't think that there's a wide-enough use-case for this. If I'm wrong, it'll come up again.

PR #2463, opened 2018-07-06
Update composer.json

ghost:
removed unnecessary marking

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,10 +49,8 @@
     },
     "scripts": {
         "test": [
-            "@phpunit",
-            "@phpcs"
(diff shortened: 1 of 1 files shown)


akrabat:
I don't think this change works when phpunit is not in the path

ghost:
Please see this. @akrabat 
[https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands](url)

Note(Official) in getcomposer.org;
In this example no matter if the phpunit binary is actually in vendor/bin/phpunit or bin/phpunit it will be found and executed.

PR #2464, opened 2018-07-07
Add text/plain to known content types

filips123:
Add `text/plain` to known content types in AbstractHandler.php.


akrabat:
What is the use-case that this change is for?

filips123:
It could be for some text based or CLI websites and applications.

akrabat:
Do you have an actual app in production that needs this change? I'm struggling to imagine a text/plain website and haven't seen one in the wild.

filips123:
No. This is just an idea, but it may be usable for someone.

akrabat:
I think we'll wait for someone to have a production use-case for this one.

PR #2466, opened 2018-07-09
Bug fix: Replace incorrect sprintf placeholders with default placeholders

### LINKED ISSUE #2433: Possible format string issue in Slim\Http\UploadedFile::moveTo()
```php
throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $targetPath));
throw new RuntimeException(sprintf('Error removing uploaded file %1s', $this->name));
throw new RuntimeException(sprintf('%1s is not a valid uploaded file', $this->file));
throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $targetPath));
throw new RuntimeException(sprintf('Error moving uploaded file %1s to %2s', $this->name, $targetPath));
```

According to the [documentation](http://php.net/sprintf), `1` and `2` in `%1s` and `%2s` are *width specifiers* (that is, the minimum length of the converted string).

That is, %1s will give the string with at least one character, %2s with at least to characters, that is:

```
> echo sprintf("Error moving uploaded file %1s to %2s", "a", "b");
Error moving uploaded file a to  b
```

But other than with empty or single letter file names, this is probably meaningless.

As far as I understand, the original intent was to facilitate internationalization to make sure that the first parameter is always the name of the source file, and the second one is the destination. If so, the correct syntax will be

```
sprintf('Error moving uploaded file %1$s to %2$s', $this->name, $targetPath)
```

E.g.,

```
> echo sprintf("Error moving uploaded file %1\$s to %2\$s\n", "file1", "file2");
Error moving uploaded file file1 to file2
> echo sprintf("Error moving uploaded file %2\$s to %1\$s\n", "file1", "file2");
Error moving uploaded file file2 to file1
```

llvdl:
Replaces incorrect sprintf placeholders with default placeholders. Fixes #2433.

### DIFF
--- a/Slim/Http/UploadedFile.php
+++ b/Slim/Http/UploadedFile.php
@@ -177,7 +177,7 @@ class UploadedFile implements UploadedFileInterface
     public function getStream()
     {
         if ($this->moved) {
-            throw new \RuntimeException(sprintf('Uploaded file %1s has already been moved', $this->name));
+            throw new \RuntimeException(sprintf('Uploaded file %s has already been moved', $this->name));
         }
         if ($this->stream === null) {
             $this->stream = new Stream(fopen($this->file, 'r'));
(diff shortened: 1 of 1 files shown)


akrabat:
Thanks.

Is this needed on [Slim-Http](https://github.com/slimphp/Slim-Http) too?

llvdl:
Good call. Slim-Http needs to be updated as well. I will create a PR for Slim-Http after work.

PR #2468, opened 2018-07-13
4.x HttpSpecializedException Constructor Fix

l0gicgate:
We didn't catch this during the code review for the `4.x-ErrorMiddleware` PR #2398.

The local variable setting logic for the `$message` variable was omitted in the `HttpSpecializedException` constructor logic.

```php
public function __construct(ServerRequestInterface $request, $message = null, $previous = null) 
{
    parent::__construct($request, $message, $this->code, $previous);
}
```

When you extend `HttpSpecializedException` and set the `$message` variable locally it gets nulled because of the missing parent constructor logic.
```php
class HttpMethodNotAllowedException extends HttpSpecializedException
{
    $message = 'My locally set message.';
}
```

### DIFF
--- a/Slim/Exception/HttpSpecializedException.php
+++ b/Slim/Exception/HttpSpecializedException.php
@@ -26,6 +26,10 @@ abstract class HttpSpecializedException extends HttpException
      */
     public function __construct(ServerRequestInterface $request, $message = null, $previous = null)
     {
-        parent::__construct($request, $message, $this->code, $previous);
+        if ($message !== null) {
+            $this->message = $message;
+        }
+
+        parent::__construct($request, $this->message, $this->code, $previous);
     }
 }


Slim/Exception/HttpSpecializedException.php
```diff
@@ -26,6 +26,10 @@
      */
     public function __construct(ServerRequestInterface $request, $message = null, $previous = null)
     {
-        parent::__construct($request, $message, $this->code, $previous);
+        if ($message !== null) {
```
akrabat, line 29: Why does the `if` statement exist? i.e. why doesn't setting `$message` to `null` when constructing the object not set `$this->message` to null?

l0gicgate: This is an abstract class, other exceptions such as `HttpMethodNotAllowedException` extend this class and have preset `$message` properties. Without that if statement in the constructor that message gets nulled and rendered useless.

PR #2482, opened 2018-08-09
Rename  variable to $callable

StJudeWasHere:
The $handler variable is called $callable everywhere else, so it seems to make sense to rename it for clarity.

### DIFF
--- a/Slim/Interfaces/RouterInterface.php
+++ b/Slim/Interfaces/RouterInterface.php
@@ -30,11 +30,11 @@ interface RouterInterface
      *
      * @param string[] $methods Array of HTTP methods
      * @param string   $pattern The route pattern
-     * @param callable $handler The route callable
+     * @param callable $callable The route callable
      *
      * @return RouteInterface
      */
-    public function map($methods, $pattern, $handler);
+    public function map($methods, $pattern, $callable);
 
     /**
      * Dispatch router for HTTP request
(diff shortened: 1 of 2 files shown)


akrabat:
This is interesting. PSR-15 calls them handlers and as we're going to support PSR-15 in Slim 4, maybe we should use their terminology.

Thoughts @silentworks, @geggleto & everyone else?

auroraeosrose:
Do the handlers have a specific call signature that is important (types, parameters, return types) that makes them more than any generic PHP callable?

That's generally my personal naming reason for naming a callable anything but callable

I'll often phpdoc mark the special callable's signature somewhere and then use that unique variable name everywhere a specific callable type is used.  If only php had delegates (callable definitions)

unguul:
Maybe take advantage of that type hinting and say "handler of type callable".
```
map(...  callable $handler){...}
```
Consistency with the PSR is kept. We know it's a callable because of its type and we know it handles that route because it's called "handler".

l0gicgate:
@unguul we can't use the `callable` type hinting as we are still supporting Slim 3 style `Class:method` callables.

unguul:
@l0gicgate Good point. Thank you!

This makes `$handler` an even better choice then as `$callable` suggests the parameter is expected to be a php `callable` when that's not the case.

I suggest the following: Route definitions have handlers which can be php `callable`s or `Class::method` strings, container keys, etc. 

We stick to the PSR and we remove some potential for confusion. Win win.

l0gicgate:
This PR is no longer relevant, please re-open with updated interfaces if you still believe this issue should be addressed.

PR #2483, opened 2018-08-09
Make groups use the Router's callable resolver

StJudeWasHere:
This PR makes the RouteGroup's use the Routers callable resolver instead of using the one in the App class.

This would avoid situations where setting a new callable resolver in the App class after the router has been created, would cause to have different resolvers for routes and groups.


akrabat:
Can you add tests please.

l0gicgate:
#2555 implements this and test coverage has been implemented there as well. Follow issue there.

PR #2485, opened 2018-08-10
Add tests for the return value of Route::setArgument and Route::setArguments methods

piotr-cz:
Similar in how it's tested in [Route::testSetName](https://github.com/slimphp/Slim/blob/3.10.0/tests/RouteTest.php#L167)


piotr-cz:
Travis job failed for php 5.5 because of

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

akrabat:
The actual PHP 5.5 build error is `[Composer\Downloader\TransportException] Peer fingerprint did not match` - we need to solve this.

Getting it here https://travis-ci.org/slimphp/Slim/jobs/414476047

Issue #2486, opened 2018-08-13
If server is on port != 80 but incoming request is on port 80, the wrong port number appears in the uri


lornajane:
I have this setup:
 - PHP running on port 8080 via `php -S localhost:8080`
 - Ngrok running a tunnel pointing to port 8080. The URL for requests is therefore http://somethingsomething.ngrok.io (on port 80)

## What I expected

I expected that this code (in a route):
```php
$uri = $request->getUri();
echo $uri->getPort();
```
Would return either nothing/false or port 80, since the request came in on that port and I'm building a link to a route inside the same application.

## What actually happened

The code returns "8080", and therefore my application returns data that looks like: http://somethingsomething.ngrok.io:8080 which doesn't work!

## Further notes

I can create a PR if someone can help me work out what the right thing to do here is.  Looking at https://github.com/slimphp/Slim/blame/3.x/Slim/Http/Uri.php#L184 I think the default to `SERVER_PORT` is incorrect. Because in this case, there isn't a port number in the URL (since it's port 80), we never pick up the port number from the host data.

What I'm not clear on: in which situation would `SERVER_PORT` be the correct default over whatever is in the `HTTP_HOST`? Because I can't think of a use case for this but I don't want to remove it and break things for someone else.

l0gicgate:
`SERVER_PORT` will return the port that's being accessed when making requests to the PHP application.

In this case, it should return `8080` since your PHP server is listening on `8080`. There's no way for the server to know that it's being tunneled to unfortunately. 

And https://github.com/slimphp/Slim/blame/3.x/Slim/Http/Uri.php#L184 is the correct behavior. `Environment::get()` get method takes two parameters `get($key, $default)` as shown here https://github.com/slimphp/Slim/blob/3.x/Slim/Collection.php#L63 

If the `$key` you are trying to get is nonexistent it will return the `$default` parameter.

carc1n0gen:
People usually get around this type of thing by having the proxy/tunnel add a header to the request containing the port, then on the slim side you can retrieve that port number from the header.  I don't have an example for Ngrok because I am not familiar with it.

akrabat:
> People usually get around this type of thing by having the proxy/tunnel add a header to the request containing the port, then on the slim side you can retrieve that port number from the header. I don't have an example for Ngrok because I am not familiar with it.

[...]

PR #2487, opened 2018-08-13
Bug Fix: Pick port defaults depending if the host was in HTTP_HOST or SERVER

lornajane:
This change relates to #2486 where a local server on port 8080 receiving a request from ngrok that is on port 80 ends up with a URI containing the ngrok host but the local server port number. This change fixes it for me BUT it causes a test to fail (`UriTest::testCreateEnvironmentWithIPv6HostNoPort`) because the test expects the response to be port 8080, but I'm not sure I agree that's the right behaviour. I can fix the test if we can agree that NULL is the expected response to `getPort()` called on a request without a port number?

### DIFF
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -177,7 +177,7 @@ class Uri implements UriInterface
         if ($env->has('HTTP_HOST')) {
             $host = $env->get('HTTP_HOST');
             // set a port default
-            $port = 80;
+            $port = null;
         } else {
             $host = $env->get('SERVER_NAME');
             // set a port default
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -173,15 +173,17 @@ class Uri implements UriInterface
         $username = $env->get('PHP_AUTH_USER', '');
         $password = $env->get('PHP_AUTH_PW', '');
 
-        // Authority: Host
+        // Authority: Host and Port
         if ($env->has('HTTP_HOST')) {
             $host = $env->get('HTTP_HOST');
+            // set a port default
+            $port = 80;
         } else {
             $host = $env->get('SERVER_NAME');
+            // set a port default
+            $port = (int)$env->get('SERVER_PORT', 80);
         }
 
-        // Authority: Port
-        $port = (int)$env->get('SERVER_PORT', 80);
         if (preg_match('/^(\[[a-fA-F0-9:.]+\])(:\d+)?\z/', $host, $matches)) {
             $host = $matches[1];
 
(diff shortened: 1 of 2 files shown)


silentworks:
In a case that someone should have an application (intranet apps) running on a different port, I don't think this change would work. e.g. `http://myintranet.net:3200`

sananguliyev (approved):

akrabat:
> In a case that someone should have an application (intranet apps) running on a different port, I don't think this change would work. e.g. `http://myintranet.net:3200`

Actually, I think that this scenario would continue to work as the `HTTP_HOST` value includes the port number when it's non-standard.

i.e. when I run `php -S 0.0.0.0:8888 -t public/`,  then `$_SERVER['HTTP_HOST']` is set to ` localhost:8888` as that's the URL that I navigate to to get to the website. We have [this code](https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Uri.php#L188-L190) which then kicks in to extract the port number.

akrabat:
I think it's reasonable to default to port 80 if HTTP_HOST is set. We need to consider SSL though as that needs the port to be set to 443, so may be we need to check `SERVER_PORT` for 443 explicitly before setting the default to 80?

lornajane:
Hmm, this is a really good point. Because the default should be 443 or 80, depending on the protocol, but I'm not sure we always know the protocol of the incoming request, only what the local server is using. If I use https:// for my ngrok tunnel, I still don't get that (default 443) port number in the `HTTP_HOST`

tuupola:
For tunneled requests these could be derived from `X-Forwarded-Port` and `X-Forwarded-Proto`. Not sure if it is the best idea to trust them by default, nor if ngrok sets these. I do know that atleast Cloudflare sets them them when using SSL proxying.

akrabat:
I've looked it up. According to the [spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23), if the  `Host` header exists and doesn't have a port, then it is the default for the protocol. 

i.e. if the port is missing from `Host` header, we should check `getScheme()` and set 443 or 80 as appropriate.

lornajane:
At the risk of thickening the plot, we don't have scheme either. Reading the git history the `X-Forwarded*` fields were deliberately removed since they can't always be trusted and that's the only place I can see the scheme. Luckily @akrabat has a middleware that reads those headers and applies the correct scheme to the request object which helped part of my problem but not the port number part.

akrabat:
Yes, you’d still need rka-scheme-and-host-middleware (now renamed to [proxy-detection-middleware](https://github.com/akrabat/proxy-detection-middleware)) to pick up the scheme. That middleware should probably handle the port better too.

Further thinking about the `Uri` part, it might be enough to set `$port` to null when the host header exists.

lornajane:
This works perfectly for me (using the middleware as well) both locally with a custom port number and with the ngrok tunnel in front of it.

akrabat:
Thanks.

Issue #1316, opened 2015-06-26
urlFor() is a misnomer; proposal for pathFor() and/or an actual urlFor() in 3.x


rszrama:
I'm taking the time to learn Slim 3.x after learning about it from Josh, and I came across this function in the Router. It's very helpful, but it's a misnomer in that it doesn't actually provide URLs for named routes. It provides paths to those routes with argument replacement. It lacks the protocol or domain name that would be required for the return value to be considered a URL.

To remedy this, I propose at least renaming the existing function to pathFor(), which is still very useful i its own right. If you wanted, you could then also provide a urlFor() function that includes the protocol and domain in its return value. (I wouldn't recommend doing what Drupal has done by overloading url() to return either a path by default or a full URL if 'absolute' => TRUE.)

I'm happy to submit a pull request in this direction but figured it best to get feedback before doing so. :smile:

codeguy:
I have no objections. @rszrama Also got your email. Will get back to you tomorrow when I have some time to write a decent response :)

rszrama:
:+1:

geggleto:
I am okay with renaming it to pathFor.

Melbournite:
In general, I am okay with it. It probably makes sense. But should it wait for a later release?

I know it's not yet entered beta or any RC stage and this is when we're meant to change stuff, but the fact is people have found it stable enough to start using, myself included. Not only this, there's already a few tutorials, videos etc. floating around with the urlFor() function used. Renaming it now would essentially break them... Easy to fix I realize, but at the end of the day we're only changing a word.

On the other hand, the inconvenience is only for early users, rather than potentially many more later down the path.

rszrama:
Carpe diem! https://github.com/slimphp/Slim/pull/1320

I hear ya on the "it will bother the people using the framework now", so it seems the best thing to do here will be to just rename the function and not add an actual urlFor(). This allows anyone currently affected to do a simple search / replace and everyone in the future to tighten up their language.

(fwiw, I've maintained a few projects with a lot of early adopters whose livelihoods I've held in my git repos, so I'm not ignoring the inconvenience; it just seems like now is still the appropriate time for a change like this to be made, and the developers affected will be the most competent to address it.)

Melbournite:
If changed, Twig-View is going to have to be updated too, which relies on urlFor. Not sure if other service providers do also.

rszrama:
[...]

PR #2493, opened 2018-08-22
Rewrite Router->urlFor() to generate complete Urls

Jmp00000001:
Hello!

Like @splitbrain mentioned in https://github.com/slimphp/Twig-View/issues/101 there is actually no way to generate a complete url while slim runs in a sub directory.

This pull request refactors https://github.com/slimphp/Twig-View/pull/102 TwigExtension->urlFor() to Router->urlFor()

Is there any chance you can merge this?
Let me know if this pull request need improvement.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -486,9 +486,9 @@ class Router implements RouterInterface
      */
     public function fullUrlFor($name, array $data = [], array $queryParams = [])
     {
-        // Check if container['request'] is initialized.
-        if (!$this->container->has('request')) {
-            throw new RuntimeException('Request object is not initialized yet.');
+        // Check if container is set.
+        if (empty($this->container)) {
+            throw new RuntimeException('Container not set.');
         }
 
         $path = $this->pathFor($name, $data, $queryParams);
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -446,7 +446,7 @@ class Router implements RouterInterface
      * @throws \Psr\Container\ContainerExceptionInterface
      * @throws RuntimeException If named route does not exist or Request not initialized
      */
-    public function urlFor($name, array $data = [], array $queryParams = [])
+    public function fullUrlFor($name, array $data = [], array $queryParams = [])
     {
         // Check if container['request'] is initialized.
         if (!$this->container->has('request')) {
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -444,7 +444,6 @@ class Router implements RouterInterface
      * @return string
      *
      * @throws \Psr\Container\ContainerExceptionInterface
-     * @throws \Psr\Container\NotFoundExceptionInterface
      * @throws RuntimeException If named route does not exist or Request not initialized
      */
     public function urlFor($name, array $data = [], array $queryParams = [])
@@ -461,6 +460,7 @@ class Router implements RouterInterface
         if (is_string($uri)) {
             $uri = Uri::createFromString($uri);
         }
+
         $scheme = $uri->getScheme();
         $authority = $uri->getAuthority();
         $host = ($scheme ? $scheme . ':' : '') . ($authority ? '//' . $authority : '');
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -425,13 +425,7 @@ class Router implements RouterInterface
      */
     public function pathFor($name, array $data = [], array $queryParams = [])
     {
-        $url = $this->relativePathFor($name, $data, $queryParams);
-
-        if ($this->basePath) {
-            $url = $this->basePath . $url;
-        }
-
-        return $url;
+        return $this->urlFor($name, $data, $queryParams);
     }
 
     /**
@@ -450,7 +444,13 @@ class Router implements RouterInterface
      */
     public function urlFor($name, array $data = [], array $queryParams = [])
     {
-        return $this->pathFor($name, $data, $queryParams);
+        $url = $this->relativePathFor($name, $data, $queryParams);
+
+        if ($this->basePath) {
+            $url = $this->basePath . $url;
+        }
+
+        return $url;
     }
 
     /**
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -433,6 +433,8 @@ class Router implements RouterInterface
     /**
      * Build the path for a named route including the base path
      *
+     * This method will be deprecated in Slim 4. Use urlFor() from now on.
+     *
      * @param string $name        Route name
      * @param array  $data        Named argument replacement data
      * @param array  $queryParams Optional query string parameters
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -11,12 +11,12 @@ namespace Slim;
 use FastRoute\Dispatcher;
 use Psr\Container\ContainerInterface;
 use InvalidArgumentException;
+use Psr\Http\Message\UriInterface;
 use RuntimeException;
 use Psr\Http\Message\ServerRequestInterface;
 use FastRoute\RouteCollector;
 use FastRoute\RouteParser;
 use FastRoute\RouteParser\Std as StdParser;
-use Slim\Http\Uri;
 use Slim\Interfaces\RouteGroupInterface;
 use Slim\Interfaces\RouterInterface;
 use Slim\Interfaces\RouteInterface;
@@ -475,31 +475,21 @@ class Router implements RouterInterface
     /**
      * Get fully qualified URL for named route
      *
-     * @param string $name Route name
+     * @param UriInterface $uri
+     * @param string $routeName
      * @param array $data Named argument replacement data
      * @param array $queryParams Optional query string parameters
      *
      * @return string
      *
-     * @throws \Psr\Container\ContainerExceptionInterface
-     * @throws RuntimeException If named route does not exist or Request not initialized
      */
-    public function fullUrlFor($name, array $data = [], array $queryParams = [])
+    public function fullUrlFor(UriInterface $uri, string $routeName, array $data = [], array $queryParams = []): string
     {
-        // Check if container is set.
-        if (empty($this->container)) {
-            throw new RuntimeException('Container not set.');
-        }
-
-        $path = $this->pathFor($name, $data, $queryParams);
-
-        /** @var Uri $uri */
-        $uri = $this->container->get('request')->getUri();
-
+        $path = $this->urlFor($routeName, $data, $queryParams);
         $scheme = $uri->getScheme();
         $authority = $uri->getAuthority();
-        $host = ($scheme ? $scheme . ':' : '') . ($authority ? '//' . $authority : '');
+        $protocol = ($scheme ? $scheme . ':' : '') . ($authority ? '//' . $authority : '');
 
-        return $host . $path;
+        return $protocol . $path;
     }
 }
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -450,8 +450,6 @@ class Router implements RouterInterface
     /**
      * Build the path for a named route.
      *
-     * This method is deprecated. Use pathFor() from now on.
-     *
      * @param string $name        Route name
      * @param array  $data        Named argument replacement data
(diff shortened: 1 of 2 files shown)


Slim/Router.php
```diff
@@ -434,22 +435,36 @@ public function pathFor($name, array $data = [], array $queryParams = [])
...
+     * @throws \Psr\Container\ContainerExceptionInterface
+     * @throws RuntimeException If named route does not exist or Request not initialized
      */
     public function urlFor($name, array $data = [], array $queryParams = [])
     {
-        trigger_error('urlFor() is deprecated. Use pathFor() instead.', E_USER_DEPRECATED);
-        return $this->pathFor($name, $data, $queryParams);
+        // Check if container['request'] is initialized.
+        if (!$this->container->has('request')) {
+            throw new RuntimeException('Request object is not initialized yet.');
```
splitbrain: For one of the use cases (creating email templates) it might very well happen that this mechanism is called without a request context when emails are generated from a command line script or from a task queue.

There is no way to automatically build the base URL without a Request object. So instead of just throwing an Exception here, it would probably be helpful to be able to configure the base URL from a config setting and fall back to it if no Request is available.

akrabat:
There's a potential problem here in that this is not backwards compatible.

vlakoff:
Refs #1316.

Actually, the name `urlFor` was not that wrong, as an URL can be relative (i.e. the scheme and autority are [optional](https://en.wikipedia.org/wiki/URL#Syntax)). Meanwhile, the "path" is a very specific part of the URL, which doesn't include the query string and the anchor.

I'd suggest:
* adding a method `fullUrlFor` (easy to remember I think),
* and replacing again `pathFor` with `urlFor` (mainly because "url" is by far the most used name in other PHP frameworks). In Slim 3, allow both `pathFor` and `urlFor` (removing the deprecation notice, to ease migration), and in Slim 4, deprecate `pathFor`.

As a result, in Slim 4 we would have `urlFor` and `fullUrlFor`, and we could use `fullUrlFor` already in Slim 3.

akrabat:
> I'd suggest:
> 
> * adding a method `fullUrlFor` (easy to remember I think),
> * and replacing again `pathFor` with `urlFor` (mainly because "url" is by far the most used name in other PHP frameworks). In Slim 3, allow both `pathFor` and `urlFor` (removing the deprecation notice, to ease migration), and in Slim 4, deprecate `pathFor`.
> 
> As a result, in Slim 4 we would have `urlFor` and `fullUrlFor`, and we could use `fullUrlFor` already in Slim 3.

I agree. Let's call this `fullUrlFor` and make sure we also create a PR for Slim 4 too.

l0gicgate:
@akrabat I can add this PR as part of the router refactoring process going on at the moment for the Slim 4 branch. It’ll probably be last though.

akrabat:
> @akrabat I can add this PR as part of the router refactoring process going on at the moment for the Slim 4 branch. It’ll probably be last though.

Okay. Separate PR :)

akrabat:
Next step for this one is for @Jmp00000001 to update this PR.

Jmp00000001:
Hello,

I will have a look after work.

This should be updated?
> I'd suggest:
> 
> adding a method fullUrlFor (easy to remember I think),
> and replacing again pathFor with urlFor (mainly because "url" is by far the most used name in other PHP frameworks). In Slim 3, allow both pathFor and urlFor (removing the deprecation notice, to ease migration), and in Slim 4, deprecate pathFor.
> As a result, in Slim 4 we would have urlFor and fullUrlFor, and we could use fullUrlFor already in Slim 3.

Jmp00000001:
I have
* refactored my implementation of `Router::urlFor()` to `Router:fullUrlFor()`
* restored the old `Router::urlFor()`
* removed the deprication warning in `Router::urlFor()`

Let me know if I missed something :)

vlakoff:
Just a nitpick: we could also reverse the Slim 3 BC alias, i.e. make `pathFor()` call `urlFor()`, as the latter should be the preferred/canonical one.

Jmp00000001:
Here you go :)

Slim/Router.php
akrabat: Let's treat this as a separate issue for a subsequent PR.

akrabat:
> Here you go :)

Please remove the `This method is deprecated. Use pathFor() from now on.` docblock comment.

Slim/Router.php
```diff
@@ -453,22 +473,36 @@ public function pathFor($name, array $data = [], array $queryParams = [])
...
+     * @throws RuntimeException If named route does not exist or Request not initialized
      */
-    public function urlFor($name, array $data = [], array $queryParams = [])
+    public function fullUrlFor($name, array $data = [], array $queryParams = [])
     {
-        trigger_error('urlFor() is deprecated. Use pathFor() instead.', E_USER_DEPRECATED);
-        return $this->pathFor($name, $data, $queryParams);
+        // Check if container['request'] is initialized.
+        if (!$this->container->has('request')) {
+            throw new RuntimeException('Request object is not initialized yet.');
```
l0gicgate: Can you add a test case for this condition please?

Slim/Router.php
```diff
@@ -453,22 +473,36 @@ public function pathFor($name, array $data = [], array $queryParams = [])
...
+        }
+
+        $path = $this->pathFor($name, $data, $queryParams);
+
+        /** @var Uri $uri */
+        $uri = $this->container->get('request')->getUri();
+        if (is_string($uri)) {
```
l0gicgate: Can you add a test case for this condition as well please?

l0gicgate (changes requested):
Please add test coverage for the to aforementioned conditions and I'll merge. Thanks!

Jmp00000001:
I will have a look today/tomorrow.
Just got back from vacation :)

Slim/Router.php
Jmp00000001: I think the check that `$uri` is a string is not required at all.
There is no way `Request::getUri()` returns something different than a `UriInterface` object:
- Only the Request constructor sets the `$uri` and expects an `UriInterface`. The constructor will throw otherwise (at latest when calling class methods like `Uri::getPort()`)
- The docblock of `Request::getUri()` also mentions: `This method MUST return a UriInterface instance`

No clue what I was thinking..
I simply remove that check.

Slim/Router.php
Jmp00000001: The constructor of `Slim\Container` initializes the request object via `DefaultServicesProvider::register($container)`.
So when the `$container` is set, the `$container['request']` is also set.
This will throw everything but a `RuntimeException`..

I changed it to `if (empty($this->container)) {...}` instead.
(`ContainerTest::testGetRequest()` already covers the test if Container has the request object of type `Slim\Request`)

Slim/Router.php
l0gicgate: So this brings me to question the method signature in the first place. Look at the way I implemented it in Slim 4:
https://github.com/slimphp/Slim/blob/4.x/Slim/Routing/RouteParser.php#L129-L136

Since this is a new method, we should force a `UriInterface` being passed in by the user instead of trying to do gymnastics to get to it.

Slim/Router.php
Jmp00000001: These should be obsolete now, since the `$uri` gets passed as argument and `$request` ist not needed anymore

vlakoff:
I'd suggest renaming `relativePathFor` to `relativeUrlFor` as well.

* For the record, refs #1521 and #1552 which implemented this method.

<br> I'd also like to reference this SO answer: [Absolute vs relative URLs](https://stackoverflow.com/questions/2005079/absolute-vs-relative-urls/39545949#39545949).

And here are the correspondences:

| Slim | SO answer |
| --- | --- |
| `fullUrlFor()` | 1. Absolute URL |
| `urlFor()` | 4. Root-relative URL |
| `relativeUrlFor()` | 2. Relative URL |

l0gicgate:
@vlakoff we cannot change `relativePathFor` in the `3.x` branch as it would be a BC change.

However, I'm open to that for Slim 4 since we're not even in alpha yet.

vlakoff:
Same as the urlFor() change, keeping the old name as a BC alias in Slim 3.

l0gicgate:
@vlakoff I'm going to PR that for Slim 4 on the `RouteParser`. I think using the same terminology everywhere makes sense. We'll deprecate `relativePathFor()` and `pathFor()` and use as you described:
* RouteParser::fullUrlFor() - Absolute URL
* RouteParser::urlFor() - Root Relative URl
* RouteParser::relativeUrlFor() - Relative URL

I also want to point out that there's not going to be any deprecation warning. Since `Router` was split into `RouteCollector` and `RouteParser` those are entirely new interfaces and there's no point in including deprecated methods/warnings since the user will have to re-learn the new interfaces anyway.

As for Slim 3, I don't think any further changes are warranted since we're so deep in the release cycle. The functionality is there and is usable, that's what matters most.

Jmp00000001:
Thanks for merging!

vlakoff:
Considering that `urlFor()` and `relativeUrlFor()` don't produce URLs strictly speaking, as they don't contain the _protocol_ and _server_, but only the _path_ + _optional query string_, I'd like to make sure everyone is comfortable with these new names.

I chose these names trying to apply the best possible the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). But on second thought, an other acceptable naming scheme could be "`relativePathFor()` / `pathFor()` / `urlFor()`".

This is really worrying me and it's still time to change if necessary, please provide your feedback on this matter!

l0gicgate:
Wouldn’t the term “Uri” instead of path apply here then @vlakoff?

- Uri - Uniform Resource Identifier
- Url - Uniform Resource Locator

A URL should be with protocol. A URI should be part of a url much like a path is. So `relativeUriFor()`, `uriFor()` and `fullUrlFor()` would make sense.

vlakoff:
That's quite a headache!

I have done some reading, and it (probably arguably) seems that what we are dealing with are really URLs, as they are resources' locations; for the value to be useful, the scheme and domain have to be known, i.e. specified but could also be deduced. In our case, the browser does the latter, by completing the provided URL using the current location.

Further reading: [What is the difference between a URI, a URL and a URN?](https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn)

Also worth noting:
> **But doesn't the W3C now say that URLs and URIs are the same thing?**
> Yes. The W3C realized that there is a ton of confusion about this. They issued a [URI clarification document](http://www.w3.org/TR/uri-clarification/) that says that it is now OK to use the terms URL and URI interchangeably (to mean URI). It is no longer useful to strictly segment URIs into different types such as URL, URN, and URC.

Finally, using "uri" surely would be very confusing for the users!

l0gicgate:
>  it is now OK to use the terms URL and URI interchangeably (to mean URI). It is no longer useful to strictly segment URIs into different types such as URL, URN, and URC.

By that logic it's totally fine to use `url` as a prefix/suffix for these method even when referring to a path. I think the key here is in the documentation.

vlakoff:
Agree with the above.

So, personally I'd bend staying on `urlFor()`/`fullUrlFor()`/`relativeUrlFor()`.
(BTW, `relativeUrlFor()` should only be used for very specific cases)

To summarize:
* `urlFor()`: the regular one, e.g. for the `href` attributes
* `fullUrlFor()`: when you need… the full url, with scheme and domain
* `relativeUrlFor()`: plz use this one as little as possible, relative paths cause so much troubles!

So, shall we settle on this?

l0gicgate:
@vlakoff

We are going to leave the terminology as in in Slim 3.

But for Slim 4 the `RouteParserInterface` is as you described:
* RouteParser::fullUrlFor()
* RouteParser::urlFor()
* RouteParser::relativeUrlFor()

So I think this is settled! Thanks for your input.

vlakoff:
I apologize for the extended chat… I really wanted to be sure the new naming convention was good. I would have felt so bad if I had introduced a bad naming scheme in Slim 4!

PR #2502, opened 2018-09-17
style: Add code style fixer

damianopetrungaro:
@akrabat before I run the fixer, please review the rules I used and point me out which branch do we want to merge it.

I'll rebase on the right branch later on.


akrabat:
Given that we have phpcs/phpcbf installed, I wonder if we should look at doctrine-coding-standards?

https://www.doctrine-project.org/projects/doctrine-coding-standard/en/latest/reference/index.html#introduction

This would apply to the `4.x` branch, btw.

l0gicgate:
I ran the Doctrine coding standards that @akrabat mentioned on the **4.x** branch of my latest PR and these are the [errors that were found](https://gist.github.com/l0gicgate/18c1c706cc71d3f1ddeb8c5c904b0a5f)

A few things I dislike:
- They want you to drop suffixes like `Exception` and `Interface`
- Comments require full stops, question marks or exclamation
- Strict type hinting is enforced on everything, which may interfere with some existing extension of classes such as `GroupCountBased` from the FastRoute package.
- It puts comments on single lines for object properties which I think is less readable
- It aligns all defining statements `=` operator. Somehow doesn't do it for constructors or doc blocks tho. Something to get used to.

I do like:
- Strict type hinting is enforced on everything. This is both a pro and a con, mostly a pro tho.
- It removes author and copyright information from doc blocks, which I think we should do since they're just painful to maintain. We should simply have a copyright file with all that information, [that's how Doctrine does it](https://github.com/doctrine/dbal/blob/master/LICENSE)

I could probably get over both of those after some time. Good news is, not that many things to fix.

damianopetrungaro:
@l0gicgate 
I do like the most of the thing that you dislike, but apart from it (we may discuss it on slack if you want), I don't want to use that specific code style fixes because it will introduce a lot of BC in the codebase that are not needed.

I didn't have more time to investigate on it, if you want to help me you're welcome and we can communicate using Slack if you want to 👍

akrabat:
I also dislike the loss of suffixes and lined up `=` and `=>` operator. I don't mind comments being complete sentences, but really don't care enough to enforce it.

l0gicgate:
We will not be implementing a code style fixer for the moment. I will be putting a code style documents together in the future.

PR #2503, opened 2018-09-19
Add docker for local development

meadsteve:
I created this to help me develop locally (I don't have php setup correctly on my machine). I thought it might be useful for other people.


akrabat:
I think this is too specific for the general case. However, I wouldn't mind a cookbook entry in the [docs](https://github.com/slimphp/Slim-Website/) about running a Slim application in Docker.

Issue #2481, opened 2018-08-06
[Proposal] Support no output buffering at all


1ma:
### Use case

Adding the ability to completely disable output buffering would empower the developer to push content to the user agent while still processing an HTTP request (i.e. while still inside the callback or Action class). This would allow, for instance, to give real-time feedback at every iteration of a long-running loop.

The following is a minimal demonstration in pure PHP that can be served with the builtin webserver to see the effect firsthand:
```php
// index.php

header('Content-Type: text/plain');

foreach (range('A', 'F') as $char) {
  sleep(1);
  echo "$char\n";

  ob_flush();
  flush();
}
```
![demo](https://i.imgur.com/jTfzDWj.gif)

### Current status

Slim 3 has an `outputBuffering`  setting whose legal values are `false`, `'append'` and `'prepend'` (default `'append'`). However, `false` seems to [simply instruct the framework to neither prepend nor append](https://github.com/slimphp/Slim/blob/0618c56a3a2c6f6e621a11d2863712969ddc0670/Slim/App.php#L322-L331) stray `echo`'s to the PSR7 response.

As far as I understand, Slim does not currently support this behaviour.

### Proposed API

I believe that the cleanest way to add this feature would be through a new  specialized implementation of the `StreamInterface` that would handle all the details under the hood in its `write` method. The former sample could be then rewritten like this:

```php
namespace Demo;

use Slim\Http;

final class NonBufferedAction
{
    public function __invoke(Http\Request $request, Http\Response $response): Http\Response
    {
        $response = $response
            ->withBody(new NonBufferedBody())
            ->withHeader('Content-Type', 'text/plain');

        foreach (range('A', 'F') as $char) {
            sleep(1);
            $response->getBody()->write("$char\n");
        }

        return $response;
    }
}
```

What are yer thoughts?

llvdl:
I think that would break the ability for middleware to replace or alter the response.

I think a generator as a body would be a good fit here. If Slim could be altered to allow a generator as well as a `Psr\Http\Message\StreamInterface` instance, then the following code could be used:

```php
class NonBufferedAction
{
  public function __invoke(Request $request, Response $response): Response
  {
    return $response
      ->withBody(function (): Generator  { // won't work now, as Slim expects a StreamInterface instance
        foreach (['alpha', 'beta', 'gamma', 'delta'] as $data) {
          sleep(1);
          yield $data;
        }
      });
  }
}
```

akrabat:
[...]

PR #2504, opened 2018-09-19
WiP: Non buffered body

1ma:
This PR implements the proposal made in #2481 as a new subtype of `StreamInterface`.

## Code

The implementation of most methods is trivial because the interface was designed for classes wrapping an internal stream, and in this case there simply isn't one.

`write($string)` essentially destroys all active Output Buffering levels (but saving their contents), then prepends this to `$string` and flushes everything to the webserver. Successive calls to `NonBufferedBody::write()` won't destroy anything else (unless the programmer runs `ob_start()` and `echo` in-between calls for any reason) and flush `$string` right away.

Any previously buffered data is displayed in a FIFO order. In other words, data echoed at the output buffer level 1 will be printed before data echoed at level 2.

Additionally, I had to modify `Response::withHeader($name, $value)` to send the headers eagerly when the body is of type `NonBufferedBody`, otherwise the framework won't attempt to send the headers until the end stages of `App::run()`. This change is quite delicate but I believe it cannot break any previous code because `NonBufferedBody` simply did not exist before.

## Working example

```php
<?php

use Slim\Http;

require_once __DIR__ . '/../vendor/autoload.php';

$app = new \Slim\App;

$app->get('/', function (Http\Request $request, Http\Response $response) {
    $response = $response
        ->withHeader('Content-Type', 'text/plain');

    foreach (range('A', 'F') as $char) {
        sleep(1);
        $response->getBody()->write("$char\n");
    }
});

$app->add(function (Http\Request $request, Http\Response $response, callable $next) {
    $nonBufferedResponse = $response
        ->withBody(new Http\NonBufferedBody)
        ->withHeader('X-Accel-Buffering', 'no');

    return $next($request, $nonBufferedResponse);
});

$app->run();
```

### HTTP headers must be sent before the content

When pushing data to the webserver while still running PHP code, one must send the headers first. Calling `Response::withHeader($name, $value)` **after** any call to `NonBufferedBody::write($string)` will trigger an `Cannot modify header information - headers already sent` PHP error. Obviously, after sending the first chunks of data you cannot push more headers in the middle of the body.

### NonBufferedBody must be set on the response before setting any HTTP header

Since `Response` now looks at the type of the `$body` attribute to determine whether to send HTTP headers eagerly or not, it is imperative to replace the underlying `StreamInterface` of the response with  `Response::withBody(new NonBufferedBody)` **before doing any call to** `Response::withHeader($name, $value)`.

I believe that this issue is best avoided with a "non-buffered middleware" that can be attached to the routes that need to behave this way, like in the example.

### Some middlewares may break

As [noted by @llvdl](https://github.com/slimphp/Slim/issues/2481#issuecomment-414155203) in the original issue, outgoing middlewares won't be able to modify the body of unbuffered responses nor add any HTTP headers. Only appending new data to the body will be possible in this scenario, which is generally not terribly useful.

### Webservers buffer output by default

Even after calling PHP's `flush()` function webservers usually buffer the output on their own.

#### Nginx + php-fpm

Nginx has the `fastcgi_buffering off;` directive to explicitly disable output buffering. Additionally, you can send a special `X-Accel-Buffering` header set to `no` to do it dynamically from PHP (see example).

#### Apache

??? TBD

## TODO

- [ ] Tests
- [ ] Documentation
- [ ] Better class naming?

### DIFF
--- a/Slim/Http/NonBufferedBody.php
+++ b/Slim/Http/NonBufferedBody.php
@@ -1,0 +1,144 @@
+<?php
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2017 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/3.x/LICENSE.md (MIT License)
+ */
+namespace Slim\Http;
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Represents a non-readable stream that whenever it is written pushes
+ * the data back to the browser immediately.
+ */
+class NonBufferedBody implements StreamInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString()
+    {
+        return '';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function detach()
+    {
+        return null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
+    {
+        return null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tell()
+    {
+        return 0;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function eof()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isSeekable()
+    {
+        return false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function rewind()
+    {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isWritable()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write($string)
+    {
+        $buffered = '';
+        while (0 < \ob_get_level()) {
+            $buffered = \ob_get_clean() . $buffered;
+        }
+
+        echo $buffered . $string;
+
+        \flush();
+
+        return \strlen($string) + \strlen($buffered);
+    }
+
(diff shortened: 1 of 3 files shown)


akrabat:
Is it worth doing this in Slim-Psr7 for use with Slim 4?

1ma:
I believe it is, yes.

akrabat:
Looks good to me. We need a docs PR for it on [Slim-Website](https://github.com/slimphp/Slim-Website).

l0gicgate:
I ran the Doctrine coding standards that @akrabat mentioned on the **4.x** branch of my latest PR and these are the [errors that were found](https://gist.github.com/l0gicgate/18c1c706cc71d3f1ddeb8c5c904b0a5f)

A few things I dislike:
- They want you to drop suffixes like `Exception` and `Interface`
- Comments require full stops, question marks or exclamation

I could probably get over both of those after some time. Good news is, not that many things to fix.

l0gicgate:
Where are you at with this @1ma do you need more feedback?

Just a couple things from my end:
- Test coverage is incomplete for `NonBufferedBody`
- I can do the docs PR if you you can summarize the functionality in detail
- Class naming is adequate in my opinion

Thanks!

1ma:
My problem is that I don't see a satisfactory way to test the `NonBufferedBody::write()` method, as it has an `echo`.

For this kind of thing I'd usually wrap the system under test in a `ob_start()/ob_get_clean()` pair and make assertions on the returned string, but this write() method needs precisely to end any active output buffers, so the echoed data always goes straight to STDOUT. Do you have any ideas?

l0gicgate:
I understand @1ma. After some thought I think this PR is fine, it's an opt-in use with some gotchas described above. Could you make a PR on the `Slim-Website` repo documenting the feature? I'll merge after. Thanks

1ma:
Great, will do.

l0gicgate:
Let's port this to Slim-Psr7 as well @1ma. I'll try and do the docs for it actually this weekend, I'll have you review them.

1ma:
Works for me. I don't intend to shy away from the work, but writing in English is not really my forte :) Thank you.

PR #2507, opened 2018-09-25
Remove HTTP method case-insensitivity from Router

JeanPaiva:
Now `$router->map(['get'], '/foo', 'foo:bar');` will return `Slim\Exception\HttpNotFoundException`

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -187,9 +187,6 @@ class Router implements RouterInterface
             $pattern = $this->processGroups() . $pattern;
         }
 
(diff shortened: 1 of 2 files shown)


JeanPaiva:
What would be this the correct way to test this PR?
https://github.com/slimphp/Slim/pull/2507/commits/d54b5bac81a5d6874c79ba9ff24f5a8392403208

akrabat:
Rebased

akrabat:
Thank you.

PR #2510, opened 2018-09-30
getRoutePattern method for retrieving all route patterns

doganoo:
For a project I'm working on, I need to know all the registered route patterns (with optional filtering). I want to share the code I create because I think there are others who might need the same.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -713,4 +713,31 @@ class App
         // No handlers found, so just throw the exception
         throw $e;
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


akrabat:
Thanks for this. I think it's too specific to be in Slim itself, but I'd love a cook book entry on the [Slim-Website](https://github.com/slimphp/Slim-Website/) docs for this.

doganoo:
Hey, I can understand :)

I will try to add if i find time for that. Thank you!

PR #2515, opened 2018-10-03
Remove else conditions

omarkdev:
This PR has the objective of removing `else` conditions in the code, either by applying the return early pattern or by adding ternary operators, this helps to reduce the cyclomatic complexity of the code

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -128,9 +128,9 @@ class Container extends PimpleContainer implements ContainerInterface
                     null,
                     $exception
                 );
-            } else {
-                throw $exception;
             }
+
+            throw $exception;
         }
     }
 
--- a/Slim/Http/Cookies.php
+++ b/Slim/Http/Cookies.php
@@ -129,11 +129,10 @@ class Cookies implements CookiesInterface
         }
 
         if (isset($properties['expires'])) {
-            if (is_string($properties['expires'])) {
-                $timestamp = strtotime($properties['expires']);
-            } else {
-                $timestamp = (int)$properties['expires'];
-            }
+            $timestamp = is_string($properties['expires'])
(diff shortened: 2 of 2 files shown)


vlakoff:
People, please stop with this "no else" trend. The current code is fine.

Jeffrey Way should make an official statement to urge people to stop with this...

akrabat:
I'm mildly confused why you think someone who is involved with a completely different project should make such a statement rather than one of the leadership of this project :)

Slim/Container.php
```diff
@@ -128,9 +128,9 @@ public function get($id)
                     null,
                     $exception
                 );
-            } else {
-                throw $exception;
             }
+
+            throw $exception;
```
akrabat, line 133: 👍

Slim/Http/Cookies.php
```diff
@@ -129,11 +129,10 @@ protected function toHeader($name, array $properties)
...
         if (isset($properties['expires'])) {
-            if (is_string($properties['expires'])) {
-                $timestamp = strtotime($properties['expires']);
-            } else {
-                $timestamp = (int)$properties['expires'];
-            }
+            $timestamp = is_string($properties['expires'])
+                ? strtotime($properties['expires'])
+                : (int)$properties['expires'];
+
```
akrabat, line 135: Please revert as this ternary statement is harder to read.

akrabat (changes requested):

Slim/Http/Cookies.php
l0gicgate: This would be a bit cleaner imo
```php
$expires = $properties['expires'];
$timestamp = is_string($expires) ? strtotime($expires) : (int) $expires;
```

PR #2516, opened 2018-10-05
Remove PHPUnit 5.x

mathmarques:
Remove unnecessary dependency on PHPUnit 5.x (for php5.6) and remove travis test with `--prefer-lowest`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
     - php: nightly
 
 before_script:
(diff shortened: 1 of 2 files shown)


akrabat:
Do we need a `--prefer-lowest` for 7.1 and 7.2?

mathmarques:
I don't think we need `--prefer-lowest`, we've used it for 5.6 only.

PR #2519, opened 2018-10-11
Added getBasePath method to Router

### LINKED ISSUE #1964: since there is a method named setbasepath in router ,why not have a getter method ?
quote [https://github.com/slimphp/Slim/issues/1608]: "getBasePath is on the URI object: $basePath = $request->getUri()->getBasePath()"
1. why not have a getter method in router class?
2 when and where to set BasePath?  
like this:    before  "$app->getContainer()[router]->setBasePath(..., ...); $app->run()"
why not add a Configuration item named BasePath

chantron:
* Added docblock for new method
* Added tests for new method

fixes #1964

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -112,6 +112,16 @@ class Router implements RouterInterface
     }
 
     /**
+     * Get the base path used in pathFor()
+     *
+     * @return string
+     */
+    public function getBasePath()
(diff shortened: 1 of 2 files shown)


BoGnY:
👍

akrabat:
Thank you.

PR #2520, opened 2018-10-16
Put HeaderStackTestAsset into Slim\Tests\Asset namespace

akrabat:
This allows us to minimise what we put into the Slim namespace to just
the header() and headers_sent() functions.

We also rename HeaderStackTestAsset to HeaderStack as the namespace is
descriptive.


l0gicgate:
Reviewed, looks good to me!

PR #2522, opened 2018-10-17
Fix some static code analyse errors with PHPStan level 1

marcelthole:
Hi,

i've run PHPStan over the 4.x branch and found a lot of wrong Interface usages. 
For the first step i fixed some simple errors

PR #2505, opened 2018-09-20
Add PHPStan (config) and some fixes

tersmitten:

PR #2523, opened 2018-10-17
Add phpstan as static code analysing tool as dev dependency

marcelthole:
And here is the dev dependency for phpstan.
I split it, because i didn't know if you would add a new package in your repository
/see #2522


marcelthole:
oh sorry. There is already a phpstan PR #2505 - my fault

akrabat:
PHPStan support has landed via a different PR. Thanks for helping Slim though!

PR #2528, opened 2018-10-28
Update route group $callable parameter type

ugokoli:
Added @param callable|Closure $callable. This is important for advanced IDEs not to show warning error colouring when closure is used.


akrabat:
Thank you.

Issue #2537, opened 2018-11-08
Cannot register media type parser containing '+' character


PaulKiddle:
I want to register a handler for `application/activity+json` and return a custom class, however the request is just treated as JSON and returned as an array.

Looks like herein lies the bug https://github.com/slimphp/Slim/blob/3.x/Slim/Http/Request.php#L1034

I can understand why it's useful to do this if there's no specific handler registered but I think the method of execution is a bit too aggressive for situations like mine.

akrabat:
Fixed by #2538

PR #2538, opened 2018-11-12
Allow registration of media type parser containing `+` character

PaulKiddle:
Fix for #2537

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1031,18 +1031,15 @@ class Request extends Message implements ServerRequestInterface
         $mediaType = $this->getMediaType();
 
         // Check if this specific media type has a parser registered first
-        $mediaTypeHasParser = (isset($this->bodyParsers[$mediaType]) === true);
-
-        if (!$mediaTypeHasParser) {
+        if (!isset($this->bodyParsers[$mediaType])) {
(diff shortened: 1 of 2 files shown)


Slim/Http/Request.php
```diff
@@ -1030,13 +1030,19 @@ public function getParsedBody()
...
         $mediaType = $this->getMediaType();
 
-        // look for a media type with a structured syntax suffix (RFC 6839)
-        $parts = explode('+', $mediaType);
-        if (count($parts) >= 2) {
-            $mediaType = 'application/' . $parts[count($parts)-1];
+        // Check if this specific media type has a parser registered first
+        $mediaTypeHasParser = (isset($this->bodyParsers[$mediaType]) === true);
+
+        if (!$mediaTypeHasParser) {
```
danopz: I feel it would be easier to understand if you would do
```php
if (!isset($this->bodyParsers[$mediaType]) {
    $mediaType = 'doSplitTheType';
}

if (isset($this->bodyParsers[$mediaType]) {
    //handle
}
```

PaulKiddle: Good call. Have updated this.

akrabat:
Thank you. Nice find.

PR #2543, opened 2018-11-22
Ensure PSR-7 stream read() is called with an int argument per spec

### LINKED ISSUE #2511: Stream Read TypeError
Hello,

Slim causes a TypeError when using a stream class with ```declare(strict_types=1)``` that relies on ```fread``` within ```StreamInterface::read()```. I noticed this issue because part of my code uses zend-diactoros, which [now has](https://github.com/zendframework/zend-diactoros/pull/329) ```declare(strict_types=1)```.

Here is the stack trace:

```
PHP Fatal error:  Uncaught TypeError: fread() expects parameter 2 to be integer, string given in /var/www/client-api/vendor/zendframework/zend-diactoros/src/Stream.php:277
Stack trace:
#0 /var/www/client-api/vendor/zendframework/zend-diactoros/src/Stream.php(277): fread(Resource id #17, '112')
#1 /var/www/client-api/vendor/slim/slim/Slim/App.php(464): Zend\\Diactoros\\Stream->read('112')
#2 /var/www/client-api/vendor/slim/slim/Slim/App.php(337): Slim\\App->respond(Object(Zend\\Diactoros\\Response))
#3 /var/www/client-api/public/index.php(11): Slim\\App->run()
#4 {main}
  thrown in /var/www/client-api/vendor/zendframework/zend-diactoros/src/Stream.php on line 277
```

It seems that Slim could pass the string value of the ```Content-Length``` header to ```read()``` [here](https://github.com/slimphp/Slim/blob/6d347477d09ab7910e625cee74c8108b7b85c7f9/Slim/App.php#L464). Since ```StreamInterface::read()``` expects an integer, I think this should be fixed in Slim. Thoughts?

I am using Slim 3.11.0.

Thank you!

soren121:
This fixes #2511.

The PSR-7 spec [requires](https://github.com/php-fig/http-message/blob/f6561bf28d520154e4b0ec72be95418abe6d9363/src/StreamInterface.php#L127-L129) that the length argument of the stream read method be an integer, but Slim retrieves the content length as a string due to its use of $response->getHeaderLine. This causes compatibility issues with PSR-7 implementations that use strict typing, namely [zendframework/zend-diactoros](https://github.com/zendframework/zend-diactoros) 2.0+.

I'm not sure how to write a unit test for this change as it's a type change that doesn't affect output/behavior except with strict PSR-7 implementations.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -461,7 +461,7 @@ class App
             if (isset($contentLength)) {
                 $amountToRead = $contentLength;
                 while ($amountToRead > 0 && !$body->eof()) {
-                    $data = $body->read(min($chunkSize, $amountToRead));
+                    $data = $body->read(min((int)$chunkSize, (int)$amountToRead));
                     echo $data;
 
                     $amountToRead -= strlen($data);


akrabat:
Thank you for this. Good catch.

PR #2549, opened 2018-11-28
Follow Single Responsibility Principle

AliN11:
I made some changes in the Handlers section to obey SRP. Also I think O/C principle could be performed. 
Thanks for reviewing.
Note: Tests need to be updated.


akrabat:
This is a BC break as the `renderXyzErrorMessage()` methods can be overridden. 

Also, the use of a static method on `Render` makes it impossible to test `Error` independently of `Render` and so would not be accepted.

akrabat:
However, thank you for your interest in Slim, we appreciate it! You may want to look at the `4.x` error handling and see if you like the approach used there.

PR #2550, opened 2018-11-29
Make the route parser a service

TiMESPLiNTER:
This is a proposal to make the route parser a service which gets injected into the router. That way one doesn't need to override the whole `router` service definition to use another route parser.

### DIFF
--- a/Slim/DefaultServicesProvider.php
+++ b/Slim/DefaultServicesProvider.php
@@ -8,6 +8,7 @@
  */
 namespace Slim;
 
+use FastRoute\RouteParser\Std as StdParser;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Slim\Handlers\PhpError;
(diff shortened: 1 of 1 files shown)


l0gicgate:
@akrabat I'm fine with this change. It's very minor, provides slightly more flexibility to end user for customization. Let me know if you're fine with merging this as well.

l0gicgate (dismissed):

akrabat:
My concern here is that you can't do this in Slim 4. How many people need to change the parser and how complex is it to extend the router to do so?

l0gicgate:
Yeah that can’t be done in Slim 4 via the container. You can instantiate router with a different parser though no need to extend.

akrabat:
Having looked some more, I'm not totally sure what problem is being solved here. Replacing the router factory doesn't seem that complicated if a different parser is required. What's the use-case?

l0gicgate:
I agree @akrabat actually. You should simply replace the entire `Router` inside of the default service provider instead. I don't think you need to offer this kind of granularity via the default service provider. Replace the `Router` if you need to access those settings.

PR #2553, opened 2018-12-10
Update 4.x example

### LINKED ISSUE #2552: Update the slim4 example please.
Currently the example in the Slim4 branch doesn't work due to 
```
$app = new Slim\App();
```
requires changing to something like:
```
$psrFactory = new Psr17Factory();
$app = new Slim\App(
    $psrFactory
);
```

The run line also needs changing, as:

```
$app->run();
```
requires a request to be passed in. However I can't seem to see an obviously correct way of doing that. Any suggestion please.

akrabat:
Update the example for 4.x so that it works with the current tip of the 4.x branch.

Closes #2552

### DIFF
--- a/example/index.php
+++ b/example/index.php
@@ -31,15 +31,11 @@ $responseFactory = new DecoratedResponseFactory(new ResponseFactory(), new Strea
  */
 $app = new App($responseFactory);
 
-
 /**
  * Add middleware as required. This is a LIFO stack.
  * We recommend adding ErrorMiddleware at least
  */
 $app->add(new ErrorMiddleware($app->getCallableResolver(), $responseFactory, true, true, true));
-
-
-$app = new App();
 
 /**
  * Define the Slim application routes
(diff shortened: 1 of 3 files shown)


example/index.php
```diff
@@ -1,25 +1,48 @@
...
+
+
+/**
+ * Add middleware as required. This is a LIFO stack.
+ * We recommend adding ErrorMiddleware at least
+ */
+$app->add(new ErrorMiddleware($app->getCallableResolver(), $responseFactory, true, true, true));
+
+
+$app = new App();
```
froschdesign: @akrabat 
I do not think so! 😉

froschdesign:
In a future version I would add links to the documentation for each step.
Apart from that: LGTM and it works! 👍

akrabat:
> In a future version I would add links to the documentation for each step.

Agreed.

Thanks for looking over it!

Issue #2408, opened 2018-02-25
Should Slim 4's middleware be LIFO or FIFO?


akrabat:
Slim 3 currently uses a Last In, First Out (LIFO) model for middleware.

That is: 

```php
$app->add('middleware3');
$app->add('middleware2');
$app->add('middleware1');
```

will execute the middleware in the order:  

1. middleware1
2. middleware2
3. middleware3

This works if you imagine middleware as layers of an onion and adding a new middleware adds a new layer to the outside of the onion.

An alternative, would be to use First In, First Out (FIFO) model.

This is, the same PHP code would execute the middleware in this order:

1. middleware3
2. middleware2
3. middleware1

This works if you imagine middleware as a pipleline going from top to bottom where the first one you add is the first one that's executed, then the second one added is executed, etc.

Should Slim 4 change to the FIFO model?

odan:
Definitely **FIFO** ("first in, first out") :-)

The code goes from top to bottom, so the middleware should also run in the same direction. 

Reading the middleware settings would make it easier for the developer to read and maintain, as it corresponds to the mental and natural order of code and logic.

Showing code of a middleware stack in a training or presentation (slide) is better understood by everyone. Because everyone reads the code from above and can therefore follow the sequence better.

FIFO simplifies the writing of tests which have to consider the sequence of the stack. etc...

FIFO has become standard in the PHP community and in all other middleware based frameworks (see Zend Expressive).

tflight:
With FIFO, which order would the middlewares run after `$app`? (It really doesn't matter much to me, I'm just curious.)

bnf:
For technical reasons I would prefer LIFO. With FIFO Slim could not [1] implicitly build the recursive middleware stack when calling `add` (see `MiddlewareAwareTrait`) [2], but would need to cache the list of middlewares first, to create the stack (by reversing the cached middleware list) when the stack is executed.

While FIFO might be more convenient, I think LIFO fits better for Slim – well, because it should be slim.

[1] Switching the middleware implementation to a queue based one (where each middleware's `$next` points back to the Middleware Dispatcher which then calls the next middleware) would allow to keep the implicit building functionality, but requires managing a middleware index (I personally prefer the current implementation)
[2] A middleware needs have a reference to the next middleware. In FIFO the next middleware is not yet known when calling `App:add`.

SadeghPM:
FIFO mode is more natural and Zend expressive also using FIFO mode.

aurmil:
[...]

Issue #2053, opened 2016-11-19
Remove traits if possible, simplify stack trace


codeguy:
Will require more investigation.

geggleto:
I floated the idea of removing Callable Resolver to it's own repo. Ideally it can be provided as a static or singleton class inside the DIC. There really isnt a reason for it being a trait.

Moving it to it's own class should stop some of the jumping across the files.

codeguy:
@geggleto I wasn't paying much attention to Slim when this was added, so I'm not 100% familiar with the internals of the current implementation. Can you walk me through it sometime soon on Slack?

geggleto:
@codeguy Yeah, just ping me when you want to go through it.

akrabat:
Maybe the trait should become a class and be composed into `App` and `Route`?

mathmarques:
I think that a class `MiddlewareAware` composed into App and Route would be better than extends an abstract class.
If you want, I can make this change.

akrabat:
> If you want, I can make this change.

Yes. This probably affects into the PSR-15 support that @bnf has started.

l0gicgate:
Closed as per #2555

PR #2555, opened 2018-12-20
PSR-15 Middleware Support

l0gicgate:
This PR is for upgrading current middleware support from PSR-7 to PSR-15. Backward compatibility for PSR-7 middleware is still present. The order of middleware is being executed is still the same as Slim 3 (LIFO or Last In First out)
<br/>

**All the existing middleware helper methods remain unchanged:**
- App::add($middleware)
- Route::add($middleware)
- RouteGroup::add($middleware)

**New typed methods have been added:**
- App::addMiddleware(MiddlewareInterface $middleware)
- Route::addMiddleware(MiddlewareInterface $middleware)
- RouteGroup::addMiddleware(MiddlewareInterface $middleware)

**The compatible types of middleware are as follows:**
- Object implementing `MiddlewareInterface`
- Class name string of a class that implements `MiddlewareInterface`, which will be resolved at the time of execution
- A `function name` or `Closure` implementing this signature `function (ServerRequestInterface $request, RequestHandlerInterface $handler) {...}`

**Deprecated Features**
- Double Pass Middleware with this signature `function ($request, $response, $next) {...}`

**This PR effectively closes:**
- #2050 
- #2053
- #2379
- #2408
- #2483

**Status**
- [x] Code Review & Discussion:


lordrhodos:
thank you @l0gicgate for the work. I have just started playing around a little bit with the `4.x` branch and hit the middleware issue when setting up a first base application. Your PR fixed that 👍 🎉 

After reading through the changes I have some remarks, event though I am not really into the Slim internals:

I would suggest to remove the `addLegacy` functions. From what I can tell Slim aims in its core for a clean and simple code base. Easy to read and understand as a developer with a focus on simplicity and performance. I love those things about it from what I learned so far. IMO adding the `addLegacy` functions just adds duplicate funcionality which is not needed. I do not think convenience support for BC updates justifies these methods. If people need to upgrade to 4.x they should do this manually by wrapping the legacy middleware. How to do this should be well documented and for BC convenience you could provide an update script/task that automates the wrapping.

Just my 0.0002 cents 😜

l0gicgate:
@lordrhodos thank you for reviewing this PR. These methods were put in place as per the pre-PR discussion with @akrabat. I think they need to be left in for the end user to be able to migrate easier, while this is still a BC change, I believe they still need to be there and deprecated on the next major release.

akrabat:
I think that `add()` needs to support DI lookup.

i.e. `$app->add(Slim\Middleware\ContentLengthMiddleware::class);` should work as it could be a key to the DI container which injects dependencies that I only want to resolve if this this middleware actually runs.

l0gicgate:
> I think that `add()` needs to support DI lookup.

Okay so you want the DI lookup to be done right then and there inside `App::add()`? Because we can't do a deferred callable for this since it isn't an invokable.

akrabat:
> > I think that add() needs to support DI lookup.
> Okay so you want the DI lookup to be done right then and there inside App::add()? Because we can't do a deferred callable for this since it isn't an invokable.

We might need a better deferred callable. We don't want to go backwards in terms of ability.

l0gicgate:
@akrabat I added support for deferred PSR-15 middleware instantiation as per your request.

l0gicgate:
@lordrhodos I ended up adding detection logic for PSR-7 middleware and removing the `addLegacy()` methods. You can now basically throw any type of middleware at the `add()` method and it will handle it.

Slim/Middleware/DeferredResolutionMiddlewareWrapper.php
```diff
@@ -0,0 +1,81 @@
...
+    private $container;
+
+    /**
+     * @var string
+     */
+    private $resolvable;
+
+    /**
+     * DeferredResolutionMiddlewareWrapper constructor.
+     * @param ContainerInterface|null $container
```
lordrhodos: I think you've got the order wrong here, or is this the new LIFO phpdoc param style ? 😉

lordrhodos:
> I ended up adding detection logic for PSR-7 middleware and removing the addLegacy() methods.

@l0gicgate this sounds great in terms of providing backwards compatibility!!! I like the decision.  Maybe you should discuss if PSR-7 support should be deprecated for a next major version 5 though, to slim down the code base.

Slim/Middleware/DeferredResolutionMiddlewareWrapper.php
l0gicgate: Good catch!

Slim/Middleware/Psr7MiddlewareWrapper.php
```diff
@@ -0,0 +1,60 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Class LegacyMiddlewareWrapper
+ * @package Slim\Middleware
+ */
+class Psr7MiddlewareWrapper implements MiddlewareInterface
```
froschdesign: I see the word "wrapper" in (class-)names as problematic, because "wrapper" is nameless and without any technical meaning.
Call a spade a spade: If it is a decorator, call it a decorator.

l0gicgate: Well the Psr7MiddlewareWrapper effectively wraps around a callable and mimics a PSR-15 MiddlewareInterface.

I’m definitely open to suggestion on wording, I couldn’t think of another word to describe what was happening.

lordrhodos: @froschdesign Agreed with a `wrapper` could be anything, but I think this is not a decorator more an adapter technically speaking or not?

froschdesign: @lordrhodos 
>  …think this is not a decorator more an adapter…

Sorry, I haven't read the entire code in this pull request, but you are right if the `DeferredResolutionMiddlewareWrapper` and `Psr7MiddlewareWrapper` adds no additional functionality to the _wrapped_ classes.

Slim/Middleware/Psr7MiddlewareWrapper.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @param ServerRequestInterface $request
+     * @param RequestHandlerInterface $handler
+     * @return ResponseInterface
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $this->responseFactory->createResponse();
+        $next = function (ServerRequestInterface $request) use ($handler) {
```
bnf: The is not the signature of the (classic) double pass middlewares, it actually is:

```
function (ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
```

Yes I know, you can drop unused argments, but this kinda hides the fact that `$response` argument (which will be passed as second paramter) is thrown away by this wrapper. The double pass middleware is degraded to be a single pass middleware.
That effectively means that a middleware that uses this `$next` callable will no longer be able to do the following:

```
$response = $response->withStatus(403);
```

This means the double pass middleware is changed to work like a single pass layer. I think it would be better to drop support entirely, than providing double pass middleware support which is broken.
Therefore #2379 suggested to keep the old middleware dispatcher as it allows to pass the `$response` objects through all middlewares (skipping PSR-15 ones, without throwing the `$response` away).
So either we live with this limitation (which is rather sad), or go with the apporach in #2379, or store the `$response` out of band (which is ugly, but would keep compatibility).

Slim/Middleware/Psr7MiddlewareWrapper.php
l0gicgate: Drop the word "Wrapper" from `DeferredResolutionMiddlewareWrapper` and use "Adapter" instead of "Wrapper" for `Psr7MiddlewareWrapper` maybe?

Slim/Middleware/Psr7MiddlewareWrapper.php
l0gicgate: This discussion happened on Slack between @akrabat and I. We are doing partial PSR-7 middleware support for this release, meaning that the `Response` object will not be passed in the middleware pipeline but simply provided to make it easier for people to return a `Response` from their existing PSR-7 middleware implementations. `ServerRequest` should be used to passed data in the middleware pipeline instead anyway. That being said, support for PSR-7 will most likely entirely be dropped in Slim 5.

bnf: > `ServerRequest` should be used to passed data in the middleware pipeline instead anyway.

Yes, I fully agree on that, but that policy is kinda new and arised due to PSR-15. Prior to that initiative, it was common to pass a modified response from outer middlewares to inner ones. 
Again, why do you want to provide a legacy layer which clutters the code and doesn't even bring full backwards support as benifit with it?

Please look how easy the changes in #2379 are. It allows for quick and safe adaption to Slim v4 and then Slim v5 can drop the old dispatcher and can become a pure PSR-15 implementation.

> support for PSR-7 will most likely entirely be dropped in Slim 5

PSR-7 be removed in Slim 5? I doubt ;) I guess you mean double-pass style middlewares.

l0gicgate: I've looked at #2379 and like I said, @akrabat and I have had this discussion on Slack already and that's the way he wanted things to be done for Slim 4. 

And yes, that's what I meant. Double-pass style middleware support will be dropped entirely in Slim 5.

Slim/App.php
```diff
@@ -92,6 +98,9 @@ public function __construct(
         $this->responseFactory = $responseFactory;
         $this->container = $container;
         $this->addSettings($settings);
+
+        $this->middlewareRunner = new MiddlewareRunner();
+        $this->add($this);
```
geggleto: This seems like some sorta anti-pattern to me...

Slim/App.php
```diff
@@ -112,22 +122,27 @@ public function getContainer()
...
-     * This method prepends new middleware to the app's middleware stack.
-     *
-     * @param  callable|string    $callable The callback routine
-     *
-     * @return static
+     * @param MiddlewareInterface|callable|string $middleware
+     * @return self
      */
-    public function add($callable)
+    public function add($middleware)
```
geggleto: I feel like there's likely bugs in here just do to the sheer complexity of this method. It would be preferable to split this out into protected / private functions to handle each case which will give us a better ability to test the different situations.

Slim/App.php
```diff
@@ -475,14 +484,14 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
-     * @param  ResponseInterface      $response The most recent Response object
+     * @param  RequestHandlerInterface $handler
      *
      * @return ResponseInterface
      *
      * @throws HttpNotFoundException
      * @throws HttpMethodNotAllowedException
      */
-    public function __invoke(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
geggleto: This feels broken as process might implement the middleware interface but does not run the middleware stack ?

l0gicgate: It’s always the first middleware in the middleware runner queue as it is added in the constructor where you commented saying it feels like an anti-pattern. 

The app adds itself to the middleware queue as the first middleware (which gets called last because LIFO). 

It’s the exact same thing as what was happening in MiddlewareAwareTrait. I understand your concerns though, but this is a reflection of `App::__invoke()` which also didn't run the middleware stack, it's just part of it. 

I do dislike that we have to expose this as a public method though.

Slim/App.php
l0gicgate: I can split it sure. 

As for bugs, there’s additional tests written for this to cover all the cases shown below. 

If you can think of additional cases, let me know and I’ll write tests for those as well.

Slim/App.php
l0gicgate: As per Slack comments, I ended up separating `App::process()` into its own middleware and removing the interface implementation. It makes much more sense, now `App` does not expose methods to the end user that are solely for internal purposes. Great catch @geggleto.

Slim/App.php
l0gicgate: Fixed with  [dc52640](https://github.com/slimphp/Slim/pull/2555/commits/dc5264064d965af5d1f197723833d6630a0221f2)

akrabat:
@bnf pointed out something important: 

> Going with a pure PSR-15 solution and adding a wrapper for double pass middlewares (as in #2555) results in (hard to detect) breaking behaviour

We have two choices here:

1. pass a shared response object to all the PSR-7 middleware
2. make it explicit to the upgrader that something has changed. e.g. `$app->addPsr7Middleware()`

Anyone have an opinion on which one?

l0gicgate:
> 1. pass a shared response object to all the PSR-7 middleware

I think we're going to have to go with option #2. It is less user-friendly for migration but it increases the chances that they'll go read the docs and understand that PSR-7 is only partially supported.

bnf:
> We have two choices here:

Or maybe we have a third one:

* Drop support for legacy (in this thread sometimes also called PSR-7) middlewares at all

There is no technical need for Slim to bundle the incompatible (as seen from the BC view) implementation, as wrapping a PSR-15 middleware by injecting a *new* `ResponseInterface` for every middleware can be done outside slim for both, normal and lazy middlewares. (Bundling this support is technically really only necessary if we would try to stay compatible, as #2379 did.)

*Advantages:*
 * Slim stays clean (and well, slim)
 * It becmomes very clear that it is not Slim's fault when the legacy and non-recommended pattern (altering the `ResponseInterface` on the way down the middleware stack) now breaks (as the developer is aware of what the wrapping does, when using a wrapper class, as opposed to some magic wrapping that happens behind Slim's doors)
 * Upgrade documentation could provide a simple example of a PSR-15 middleware adapter which creates a new response. This will probably actually result in quicker adoption of the PSR-15 interfaces. Why? I think it's really easy to understand how to rewrite an existing middleware when looking at a simple wrapper/adapter class. (Well, there is one exception: when the response was altered you get a problem, but that's the intention: The developers need to detect this pattern upfront, and then rewrite the code, before getting into production (where is silently fails and becomes hard to debug).

> I think we're going to have to go with option 2

Don't think s;, option 1 is valid: As this pull request is using a queue a stateful dispatcher is needed anyway (`$this->stages`), so adding another state (shared Response) would be possible.

PS: The name "PSR-7 middleware" is quite misleading (as there are no middlewares in PSR-7 and as PSR-15 middlewares *do* use PSR-7 interfaces, this sounds as if the new middlewares are referenced. Yes, naming is hard. Double vs Single pass is quite unknown. But maybe "legacy" or "classical" or "slimv3 middlewares" is better than?

akrabat:
> * Drop support for legacy (in this thread sometimes also called PSR-7) middlewares at all

I'm open to this if we can document what to do easily

l0gicgate:
@bnf ultimately I would like a cleaner codebase and I’m personally fine with dropping Double Pass middleware as well if @akrabat sees this as an option 

I see Slim 4 as a clean slate, at some point we’re going to drop Double Pass middleware support so why not now?

The migration process from 3 to 4 already has a few BC changes, let’s break everything now instead of later and not have to carry the technical debt. 

We’re going to have unhappy users no matter what we do. There’s nothing wrong with sticking with Slim 3.x if the advantages from Slim 4 don’t seem to outweigh the cons of having to refactor your code. I don’t think adoption will suffer all that much. 

Also, in this PR I migrated all the core middleware from double pass to single pass and it wasn’t hard at all, it only took about an hour.

akrabat:
If we drop Slim3 middleware format do we support pseudo-PSR-15closures ?

i.e. can a user do this?

```
$app->add(function($request, $handler) {
    return $handler->handle($request):
});
```

l0gicgate:
@akrabat yes that is possible. That function will get wrapped with an adapter though.

bnf:
@akrabat Well, very good question. Pseudo PSR-15 closures would offer the same BC problem, as we "silently" change the signature (though that's easier to detect, as the previously passed third parameter now becomes `null` and results in hard errors when trying to be used as callable).

It's a bit of a pity that sth like the following is a bit clumsy to write:
```
$app->add(new class implements MiddlewareInterface {
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
    {
        return $handler->handle($request);
    }
});
```

But it's something I use rather often, as seen in the README for https://github.com/bnf/slim3-psr15
So when dropping the legacy middlewares I'd probably add a proper type declaration on `App::add` to accept only `MiddlewareInterface`.

Slim/MiddlewareRunner.php
```diff
@@ -0,0 +1,107 @@
...
+    public function add(MiddlewareInterface $middleware)
+    {
+        array_unshift($this->middleware, $middleware);
+    }
+
+    /**
+     * @param ServerRequestInterface $request
+     * @return ResponseInterface
+     */
+    public function run(ServerRequestInterface $request): ResponseInterface
```
bnf, line 60: In order to support subrequests, `run()` needs to create a new runner on every run (that may be new class that is instanciated or a clone of itself).

`run()` currently has side effects: it stores the state (`$this->stages`) of the *current* run in the (singleton) class, that means a second call to `run`  – which might happens during the first run – would result in conflicting states that are tried to be stored in *one* (singleton) instance of this class.

Note, handling this explicitly is (now) necessary as this patch switches to a queue based dispatcher, away from the previous stack based on, where workarounds like that were not needed as the sequence of the middlewares was stored implictly thanks to the closures (`$this->tip = function (…) use ($callable, $next) …`) that connected these middlewares by wrapping them in a recursive call chain.

I know #2078 removed support for `App::subRequest`, but I think the pattern of support subrequests has not been dropped(?)

lordrhodos:
I am not so deeply involved in the topic as you guys are, but I like to add some points from the outside:

* I ususally expect an incremented major version to introduce breaking changes ([semantic versioning](https://semver.org/)) and the need to update the code on my end. If I want to use the fancy features of the next major version I need to get my hands dirty. I do not expect my code to run without any changes. An upgrade guide / documenation is mandatory!
* Slim should stay slim, drop (all) legacy stuff thats tries to ease the upgrade pain. Instead offer first class documenation for a) the new major release and b) the upgrade process. If documentation is not enough, provide tools to ease the upgrade. If you need support for `LegacyMiddlewareAdapter` move this to a dedicated library and document the usage. I think @l0gicgate is right about breaking things now and get rid of the technical debt. Highly appreaciated as I have no Slim3 projects in production 😉 
* I think @bnf is right about the naming of Slim3 middleware format, so I suggest you agree on an *official* term for it. I personally think double vs. single pass is valid as this is referenced in the [PSR-15 meta docs](https://www.php-fig.org/psr/psr-15/meta/), although it may not be known amongst so many developers.

Slim/MiddlewareRunner.php
l0gicgate: Can you give me a concrete example as to why one would run the middleware queue twice?

I don’t even know how that would be possible. I could explicitly lock the run method while queue is in progress and automatically rewind at the end if necessary. 

In any case there was no explanation on #2044 as to why subRequest functionality was dropped and I’m not sure if subRequest support still exists or not.

akrabat:
One thing about Slim is that it's _simple_ to get going with. This is why it's used in so many tutorials for SDKs and libs for other components. We need to not lose that.

```
$app->add(new class implements MiddlewareInterface {
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
    {
        return $handler->handle($request);
    }
});
```

is too cumbersome! Though I wonder how many people do closures for middleware compared to doing closures for route callables?

l0gicgate:
@akrabat I don’t think many people use anonymous functions as middleware a whole lot.

I also think having a typed `add()` method is safer. It adds a roadblock when migrating. Otherwise as @bnf said it may introduce bugs in existing callable usage that expects the 3 parameters.

bnf:
Can you elaborate why the middleware runner/dispatcher uses a queue based approach instead of the previous stack based one, that slim used to use?

Maybe I start with a comparison, to depict the difference of queue and stack based dispatching, so we know what we are talking about here:

**Legend:**
*M* = Middleware
*D* = Dispatcher
*H* = Request Handler
*C* = Intermediate handler – might be a closure (`$this->tip = function` for slim v3, would be an anonymous `RequestHandlerInterface` for a native PSR-15 implementation)

**Stack based (as in slim v3):**
```
D -> C1 -> M1 -> C2 -> M2 -> C3 -> M3 -> C4 -> M4 -> R
```
The stack based dispatcher **D** wraps every middleware **Mn** into an intermediate handler **Cn** which is passed to "previous" middleware (that is the `$next` closure as known from slim v3, would be a PSR-15 RequestHandler for a native approach in v4). It could also be seen as recursive graph:

```
D -> C1 -> (
  M1 -> C2 -> (
    M2 -> C3 -> (
      M3 -> C4 -> (
        M4 -> H
      )
    )
  )
)
```

This reverse way of passing a reference to the "next" middleware (e.g. **M3**) to the "previous" middleware (e.g. **M2**) using a helper closure **C3** is the reason that slim v3 middleware order is LIFO (and not FIFO), as the dispatcher needs to know the last middleware (**M4**) first, to be able to pass a reference to pass the intermediate helper **C4** to **M3**.

**Queue based (as in this pull request)**
```
D (has $stages) -> M1 -> D -> M2 -> D -> M3 -> D -> M4 -> R
```

The queue based dispatcher manages an `array` (or `ObjectStorage`) to store the available middlewares and uses a counter (or modification of that `array`/`ObjectStorage`) to keep track of the position in the dispatch cycle we currently are. The counter is increased (or as in this pull request the next middlewares is removed from the queue) on every call(back) to **D** from a Middleware **Mn**. All middlewares therefore call back to the dispatcher. That means this approach does not need intermediate closures/handlers and looks a bit cleaner from the call stack point of view – but to store the ordering and list of middleware this dispatcher needs to keep track of the middleware in a queue on the heap, which is susceptible for concurrency issues (think of subrequests).
That means for subrequests it's important that the dispatcher creates an independent state (a new Runner instance or a clone of itself), to prevent conflicting states.

I'ld like to understand whether the queue based approach is used for a (good) reason or because it was easier to implement, more obvious, or because you want this implementation to be able to support FIFO?
I personally find the stack based approach looks cleaner (looking with theoretical computer science eyes here) as state is nicely isolated in a intermediate closure/handler and because the call stack does all the job of managing the state for us.
Maybe that is the drawback of that approach, because the call stack may become (too) huge? I don't know… I don't have numbers… Slim used to use the approach of using a clean and simple dispatcher which is hard to "break". If we change that, we should know why. I wouldn't like Slim to switch to the queue based approach only because all the others do it that way.

l0gicgate:
@bnf the main reason is cleaner call stack, the second reason was originally to enable the possibility of multi-mode queue (FIFO and LIFO). 

I also believe that the queue is much more explicit than the stack based, which makes it easier to read and understand as a developer not that this should be the reason to adopt queue over stack though. 

That being said I do understand the drawbacks that you’ve presented.

bnf:
> This is why it's used in so many tutorials for SDKs and libs for other components. We need to not lose that.

Goot point. So talking about simplistic syntax sugar, maybe:

```php
use function Slim\middleware;

$app->add(middleware(function($request, $handler) {
    return $handler->handle($request):
}));
```

froschdesign:
@bnf 
https://github.com/zendframework/zend-stratigility/tree/master/src/functions

akrabat:
I like the function idea and we could do others. Interesting to see that ZF came to the same conclusion.

bnf:
Ah, didn't know the ZF functions.

Minor thing: I don't like that they type declare a return type of `Zend\Stratigility\Middleware\CallableMiddlewareDecorator`, one should really declare to return `\Psr\Http\Server\MiddlewareInterface`, as that's what the user of that function should (only) expect to receive here, every change of that return type would be a BC break, other than that, that's pretty much what I meant.

Still, my proposal (which is not mine as we see now) was for pure syntax sugar, in case of readability I personally prefer explicitly instanciating classes, instead of functions that wrap class instanciation into a shorter syntax. (as functions returning objects are hard to grasp mentally without an IDE, but I know that's not a focus for most)

I find the following more readable:
```php
$app->add(new ClosureMiddleware(function($request, $handler) {
    return $handler->handle($request):
}));
```
But sure, this doesn't look "sexy" in code examples...

l0gicgate:
@bnf I agree on the return type. I also agree on the name `ClosureMiddleware` over the sugar function `middleware()`.

Slim/MiddlewareRunner.php
bnf: The usecase are sub requests. Think of a backend context that renders a preview of the frontend, or whatever. See https://github.com/slimphp/Slim/pull/2581 for a patch that adds the missing test for this case.

> I could explicitly lock the run method while queue…rewind at the end if necessary.
Wouldn't help, as the invocation might happen during the invocation of a middleware.

If we really go with the queue based dispatcher we should create a new runner (cleaner than clone) in `run`. Just saw also zf switched to that approach (they used `clone` before, and I wanted to reference them for that as "bad" example – anyway, good that improved already): https://github.com/zendframework/zend-stratigility/commit/04d418adbfca1a1a769261c8a044be087548aae6#diff-26f56039a32365ad4c575dca7b275c56

l0gicgate: Very good point. I will implement that without using clone. I will also merge the tests from #2581 first and rebase this PR on that to make sure that it passes that test.

akrabat:
Maybe we're overthinking the handling of closures?

Part of me thinks we should just take the closure directly. We can't type hint `add()` because of the DI resolver (`$app->add(MyMiddleware::class)`), so it's probably fine to allow `$app->add(function($request, $handler) {} )` on the grounds that if you pass in a closure that uses 3 parameters, you'll get an exception if you try to return the second parameter as it's not a `Response`.

We also need to support closures for route handlers too as these are really common. i.e. `$app->get('/', function($request, $args) {} );`. Again, if you tried to operate on the second parameter thinking it was a Response object, you'd find out by the PHP fatal error.

akrabat:
Do we need to provide an adapter from doubleness middleware to PSR-15 or do we just document it?

Similarly, do we provide an adapter for Slim 3 route handles to Slim 4 route handlers?

I suppose it's not hard to provide them as optional things you explicitly use.

```php
$app->add(new DoublePassMiddlewareAdapater(MyOldMiddleware::class)):
```

```php
$app->get('/', new Slim3Handler('MyController:action'));
```

akrabat:
I think we're all agreed that Slim 4's `add()` will not take Slim 3 style middleware classes directly.

This means that `add()` will take one of:

1. a class that's an instance of PSR-15's MiddlewareInterface
2. a string to be resolved to an instance of PSR_15's MiddlewareInterface when the middleware is invoked.
3. (TBD) a closure of the format `function($request, $handler)`

Note that we need to agree the way forward on item 3.

akrabat:
I think we're all agreed that the Middleware Runner needs to support sub-requests, so let's get that working too.

bnf:
I think that when we use a `mixed` argument for `add()` anyway (to support lazy instantation), then it's fine to support closures as well. As we "lost" the advantage of a proper type declaration anyway, it's fine to support Closure. I agree that BC is not a big issue here.

On the other hand, an alternative could be the following:
In another project I used two separte signatures for the instanciated and lazy registration case:

```php
interface MiddlewareDispatcher {
    /** Adds a middleware to the stack */
    public function add(MiddlewareInterface $middleware);

    /** Adds a lazily instanciated middleware to the stack */
    public function lazy(string $identifier);
}
```

But I figure that a generic method name `lazy()` would probably be inappropriate for the generic `App` class – and an ugly camel case method name like `lazyAdd()` doesn't sound very "slim".

lordrhodos:
@akrabat 
why not combine your two thoughts into one clean approach?

> Do we need to provide an adapter from doubleness middleware to PSR-15 or do we just document it?
> ...
> ```php
> $app->add(new DoublePassMiddlewareAdapater(MyOldMiddleware::class)):
> ```

use a general adapter, e.g. `MiddlewareAdapter`

> This means that `add()` will take one of:
> 
>    1. a class that's an instance of PSR-15's MiddlewareInterface
>
>    2. a string to be resolved to an instance of PSR_15's MiddlewareInterface when the middleware is invoked.
>
>   3. (TBD) a closure of the format `function($request, $handler)`

let it handle all the internal resolving part and type hint to `MiddlewareInterface` in `App::add` method. Use the mixed argument in the adpaters constructor to keep the flexibility. 

IMO this would lead to the cleanest way, e.g. pass in a PSR-15 middleware
```php
$app->add(new Psr15Middleware());
```

Pass in non PSR-15 middleware using the adapter:

```php
$app->add(new MiddlewareAdapter('LegacyMiddleware:custom'));
```

Or am I missing something here?

bnf:
> Or am I missing something here?

Yes, the adapter example is a bit simplified. It actually would need a reference to the container in most real world cases (to be able inject dependencies into the middleware):

```php
$app->add(new MiddlewareAdapter(MiddlewareClassName::class, $container));
```
This would become very ugly for a lot of lazy middlewares.

l0gicgate:
> This would become very ugly for a lot of lazy middlewares.

This is exactly what [DeferredResolutionMiddleware](https://github.com/slimphp/Slim/blob/39eddaf6c4388e9e51d4a6378fabaa36dadd42ca/Slim/Middleware/DeferredResolutionMiddleware.php) in this PR does. However at the moment the `App::add()` method is in charge of instantiating it and passing the container to it.

To come back to the `App::add()` method I think it needs to be typed because people who migrate from Slim 3 could pass it a Closure which would then fail when they execute because the signature doesn't match. We need to BC the method to force the user to go look at the docs or pass it the correct interface.

**Slim 4 Middleware Closure:**
```php
function (ServerRequestInterface $request, RequestHandlerInterface $handler) {...}
```

**Slim 3 Middleware Closure:**
```php
function (ServerRequestInterface $request, ResponseInterface $response, callable $next) {...}
```

l0gicgate:
> Do we need to provide an adapter from doubleness middleware to PSR-15 or do we just document it?

I'm not sure I understand, I thought we agreed to drop double pass middleware support. People are just going to have to migrate to single pass.

I really strongly believe we need to type `add()`. In a normal code base, how many different middleware will you have to migrate? 10 maybe? That'll take you 30 minutes. It's really not the end of the world.

As for deferred resolution we should force the end user to wrap it in a PSR-15 compatible class like `DeferredResolutionMiddleware` from this PR. We can maybe provide automatic container setting when it gets added via `add()`. All I'll do is detect that it's an instance of `DeferredResolutionMiddleware` then check if the container has already been set on it, if it hasn't then we'll set it automagically so the synthax becomes a bit more sugary

(We can change the name `DeferredResolutionMiddleware` to `LazyMiddleware` or something shorter if need be)
```php
$app->add(new DeferredResolutionMiddleware(MyClass::class));
```

Or as @bnf suggested, we could add a helper method which does the wrapping of `DeferredResolutionMiddleware` for you
```php
$app->addLazy(MyClass::class);
```

akrabat:
I am definitely on the other side when it comes to `add()` which I think should be a convenience  method that takes `mixed` as this the path that looks nicest when writing your code in whatever style you prefer. Slim doesn't take sides in what structure you prefer.

I'm definitely happy with runtime errors when you use the wrong signature for a closure for instance. I'm not okay with impossible-to-detect errors though which @bnf pointed out is what happens if we silently handle DoublePassMiddleware with individual Response objects.

I'm also okay with an `addMiddleware()` method that it is strictly typed, so maybe we have `add()` as the convenient loosely-typed one and `addMiddleware()` for the people that want the strict typing?

This is arguably equivalent to having `route()` as the "strict" way to register an action but also support convenience methods like `get()`, `post()` etc.

Slim/Middleware/RoutingDetectionMiddleware.php
```diff
@@ -0,0 +1,75 @@
...
+use Slim\Exception\HttpNotFoundException;
+use Slim\Interfaces\RouterInterface;
+use Slim\Route;
+use Slim\RoutingResults;
+
+/**
+ * Class RoutingDetectionMiddleware
+ * @package Slim\Middleware
+ */
+class RoutingDetectionMiddleware implements MiddlewareInterface
```
akrabat: This is mis-named. It dispatches the route (and also routes if routing hasn't already been performed).

How about renaming to DispatchMiddleware?

akrabat:
We would massively simplify `App` if we remove `setContainer()`, `setRouter()` and `setCallableResolver()`. Any reason why we can't do this?

Slim/App.php
```diff
@@ -115,19 +132,42 @@ public function setContainer(ContainerInterface $container)
...
-     * @return static
+     * @param MiddlewareInterface|string $middleware
+     * @return self
+     */
+    public function add($middleware): self
+    {
+        if (is_string($middleware)) {
+            $deferredContainerResolver = $this->getDeferredContainerResolver();
+            $middleware = new DeferredResolutionMiddleware($middleware, $deferredContainerResolver);
+        } elseif (!($middleware instanceof MiddlewareInterface)) {
```
akrabat, line 125: I would like `add()` to accept a closure too and wrap it in `ClosureMiddleware`.

l0gicgate: > I would like `add()` to accept a closure too and wrap it in `ClosureMiddleware`.

Done

tests/CallableResolverTest.php
```diff
@@ -40,29 +41,24 @@ public function setUp()
...
-            static $called_count = 0;
-            return $called_count++;
+            return true;
         };
         $resolver = new CallableResolver(); // No container injected
         $callable = $resolver->resolve($test);
         $callable();
-        $this->assertEquals(1, $callable());
+
+        $this->assertEquals(true, $callable());
```
akrabat, line 49: We appear to be calling `$callable()` twice. Is this intentional?

l0gicgate: removed that just now

Slim/Middleware/DispatchMiddleware.php
```diff
@@ -0,0 +1,72 @@
...
+     * @throws HttpNotFoundException
+     * @throws HttpMethodNotAllowedException
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        /** @var RoutingResults|null $routingResults */
+        $routingResults = $request->getAttribute('routingResults');
+
+        // If routing hasn't been done, then do it now so we can dispatch
+        if ($routingResults === null) {
```
akrabat: Put `$request->getAttribute('routingResults')` directly into the `if` statement as we don't need the temporary variable.

Slim/MiddlewareRunner.php
```diff
@@ -0,0 +1,137 @@
...
+        return $this->middleware;
+    }
+
+    /**
+     * @param MiddlewareInterface[] $middleware
+     * @return self
+     */
+    public function setMiddleware(array $middleware): self
+    {
+        $this->middleware = $middleware;
```
akrabat: Rather than just assigning, we could clear `$this->middleware` and then assign each item in `$middleware` using `$this->add()`. This would mean that there'd be a PHP type error on assignment rather than a runtime exception when `run()` is called later.

Slim/MiddlewareRunner.php
```diff
@@ -0,0 +1,137 @@
...
+     */
+    protected $stages;
+
+    /**
+     * MiddlewareRunner constructor.
+     * @param array $middleware
+     */
+    public function __construct(array $middleware = [])
+    {
+        $this->middleware = $middleware;
```
akrabat: Assign `$middleware` using `$this->setMiddleware()` here so that it we use the type-safe code path.

Slim/MiddlewareRunner.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * @return SplObjectStorage
+     */
+    protected function buildStages(): SplObjectStorage
+    {
+        $stages = new SplObjectStorage();
+        foreach ($this->middleware as $middleware) {
+            if (!($middleware instanceof MiddlewareInterface)) {
```
akrabat: We shouldn't need this test if we ensure that the only way to add middleware to this class is via `add()`.

Slim/Route.php
```diff
@@ -81,30 +82,34 @@ class Route extends Routable implements RouteInterface
...
+     * @param InvocationStrategyInterface|null  $invocationStrategy
+     * @param RouteGroup[]                      $groups The parent route groups
+     * @param int                               $identifier The route identifier
      */
-    public function __construct($methods, string $pattern, $callable, array $groups = [], int $identifier = 0)
-    {
-        $this->methods  = is_string($methods) ? [$methods] : $methods;
-        $this->pattern  = $pattern;
+    public function __construct(
+        $methods,
```
akrabat: Should we make `$methods` always be an array like `$groups` is? 

Would save the `is_string() ? :` call.

Slim/MiddlewareRunner.php
l0gicgate: That was in place because of the `setMiddleware()` method. Can be removed if we throw from there. This is also why we need generics.

Slim/Route.php
l0gicgate: Couldn’t we use spread operator here? `...$methods`? We should definitely type that property. Though i’m in favor of that. Nobody instantiates routes manually anyway it’s all internal work done by the router so.

tests/RouteTest.php
```diff
@@ -51,37 +65,89 @@ public function testConstructor()
...
     {
-        $callable = $this->routeFactory()->getCallable();
+        $route = $this->createRoute();
+
+        $this->assertTrue(is_callable($route->getCallable()));
+    }
+
+    public function testGetCallableResolver()
+    {
+        $callable = $callable ?? function (ServerRequestInterface $request, ResponseInterface $response, $args) {
```
akrabat: `$callable` will never exist!

Slim/MiddlewareRunner.php
akrabat: Resolved with 5e8fc67

Slim/MiddlewareRunner.php
akrabat: Resolved with 5e8fc67

Slim/MiddlewareRunner.php
akrabat: Resolved with 5e8fc67

Slim/Route.php
akrabat: Resolved via 4bf9785

l0gicgate (commented):
Looks good. Approved.

l0gicgate (commented):
Looks good. Approved.

l0gicgate (commented):
Looks good. Approved.

tests/RouteTest.php
akrabat: Resolved via 9aa8c6b

akrabat (approved):

PR #2558, opened 2019-01-03
Bug fix: Ensure integer is passed to read()

farpat:


akrabat:
Can you explain the problem you are having and why this fixes it?

farpat:
Firstly sorry for my English, I'm French.

You use Zend\Diactoros\Stream@read where you send the parameter $length being a string (even if this one contains a number like "138") and when fread is called - a few lines later - a fatal error is triggered.

This is why I purpose this fix.

akrabat:
Which `read()` line causes the fatal error?

farpat:
The fatal error is provoked in line ~277 of file " vendor/zendframework/zend-diactoros/src/Stream.php " => `$result = fread($this->resource, $length);` 

`$length` must be an int not a string.

akrabat:
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L277 is a comment just above the definition of `group()` ?

farpat:
Sorry, I fix the the filename in my previous comment => https://github.com/zendframework/zend-diactoros/blob/master/src/Stream.php#L277

akrabat:
Ah!

Which `read()` call in `Slim\App` is causing the problem as there are two.

farpat:
Sorry, I did not investigate enought to say it but I can search this if you want.

akrabat:
My main concern is that 

```php
$data = $body->read(min((int)$chunkSize, (int)$amountToRead));
```

already casts to `int`. However:

```php
echo $body->read($chunkSize);
```

Could possibly be an `int`. Your fix doesn't change this variable, so maybe

```php
$chunkSize      = $settings['responseChunkSize'];
```

needs the cast?

farpat:
This concern surely the line `$chunkSize      = $settings['responseChunkSize'];` as you talk.

Also, you'are right => My previous commit does not fix this but my second commit must be fix the issue normally.

Thanks for your patience.

akrabat (approved):

PR #2574, opened 2019-02-02
 I want to understand twice run slim application.

LeafChage:
I want to understand twice run slim application.

My application occurred error because of it.
I lost a day.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -23,6 +23,7 @@ use FastRoute\Dispatcher;
 use Slim\Exception\SlimException;
 use Slim\Exception\MethodNotAllowedException;
 use Slim\Exception\NotFoundException;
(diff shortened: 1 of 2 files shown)


l0gicgate:
Why are you calling the `App::run()` method twice?

There is no need to add this to the existing code base.

LeafChage:
@l0gicgate 
It is my mistake.
But I was supposed to prevent, if it can not run twice.

l0gicgate:
Okay, well it sounds like that's an issue on your end due to the logic that you're utilizing.

You can always extend the App class and overload the `App::run()` method to use your code or fork the repository and modify the code according to your use case.

I'm closing this PR.

LeafChage:
Thanks

Issue #2386, opened 2018-01-24
FastRoute falls back to GET on HEAD request


garnold:
We were trying to figure out why `HEAD` requests from bots were making their way into our controllers. Then we discovered that [FastRoute falls back to `GET` on `HEAD` requests](https://github.com/nikic/FastRoute/blob/master/src/Dispatcher/RegexBasedAbstract.php#L33).

This is unexpected.

Related, I've noticed a number of issues where Slim users are requesting an `$app->head(...)` shortcut be added to Slim, and the response is often to use `$app->map(['HEAD'], ...)` instead. But given FastRoute's fallback you can just use `$app->get(...)` to handle `HEAD` requests. So my guess is this fallback behavior isn't well known in the Slim community.

geggleto:
Hmm, HEAD should be GET just w/o a response body.

damianopetrungaro:
As @geggleto said: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4

akrabat:
Do we correctly ditch the body when it's a HEAD?

designermonkey:
I'm not sure we do from a quick glance. We have the `isEmpty` method on a Response, but I can't see where we would check the Request to see if it is a `HEAD` request.

akrabat:
**Todo:**

1. make sure that the user can detect a HEAD request inside the route callable that's attached to a GET request
2. Ensure that we don't send the request body if it's a HEAD request.

akrabat:
Addressed in #2547

akrabat:
Re-opening - 3.x needs checking.

lordrhodos:
I did some research for the 3.x code. 

> **Todo:**
>
>    1. make sure that the user can detect a HEAD request inside the route callable that's attached to a GET request

As the PSR-7 request is passed to the callable it is no problem to check manually if this is actually a `HEAD` request, e.g.

```php
$app->get('/', function (Request $request, Response $response, array $args) {
...
    $isHeadMethod = $request->getMethod() === 'HEAD';
...
});
```

> 2. Ensure that we don't send the request body if it's a HEAD request.

This is kind of tricky. The mentioned `Response::isEmpty` method just checks for a response status code that should not contain a body:

https://github.com/slimphp/Slim/blob/715d1306dd221f016b2f3f1d132a0ada3c3ac9a8/Slim/Http/Response.php#L376-L382

[...]

PR #2576, opened 2019-02-06
make sure HEAD requests do not return a body

lordrhodos:
Suggests a fix for issue #2386 

@akrabat  @l0gicgate I need some feedback if this is the right approach

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -633,6 +633,11 @@ class App
             }
         }
 
+        // clear the body if this is a HEAD request
+        if ($this->isHeadRequest($request)) {
+            return $response->withBody(new Body(fopen('php://temp', 'r+')));
+        }
+
         return $response;
     }
 
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -11,7 +11,6 @@ namespace Slim;
 use Exception;
 use Psr\Http\Message\UriInterface;
 use Slim\Exception\InvalidMethodException;
-use Slim\Http\Response;
 use Throwable;
 use Closure;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


Slim/App.php
```diff
@@ -645,6 +647,11 @@ protected function finalize(ResponseInterface $response)
      */
     protected function isEmptyResponse(ResponseInterface $response)
     {
+        $request = $this->container->get('request');
+        if ($request->getMethod() === 'HEAD') {
```
l0gicgate: You will need to make this if case insensitive. This will fail with lower case method 'head'.

Slim/App.php
```diff
@@ -614,7 +614,8 @@ protected function finalize(ResponseInterface $response)
         // stop PHP sending a Content-Type automatically
         ini_set('default_mimetype', '');
 
-        if ($this->isEmptyResponse($response)) {
+        $request = $this->container->get('request');
+        if ($this->isEmptyResponse($response) && $request->getMethod() !== 'HEAD') {
```
l0gicgate: this also needs to be a case insensitive if case.

l0gicgate:
A couple of things:

1. You will need to clear out the existing Response body in `App::finalize()`;
2. Add the test case from the `4.x` branch as seen [here](https://github.com/slimphp/Slim/blob/4.x/tests/AppTest.php#L1454-L1470)

Thank you!

l0gicgate (approved):
Everything looks good to me @lordrhodos. Just need @akrabat to have a glance and we can merge and close #2386.

akrabat:
Im a little twitchy about using the container's Request object, but it seems the best choice here

akrabat (approved):

akrabat:
Do we need this for Slim 4 too?

l0gicgate:
@akrabat I already did this for Slim 4 see #2547

PR #2578, opened 2019-02-12
Fix dispatching of non handle() methods for PSR15 request handlers

bnf:
If a method has been explicitly defined, the callable resolver
should honor that for all classes (even for classes that
implement RequestHandlerInterface) and only default to 'handle'
(for PSR-15 RequestHandlers) if no method has been specified.

An example:
When FooBar implements RequestHandlerInterface and
'FooBar:customMethod' is specified to be dispatched, the
CallableResolver must not resolve to the handle() method
even though FooBar implements the RequestHandlerInterface.
It should only do that, if no method name is given: 'FooBar'

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -54,7 +54,8 @@ final class CallableResolver implements CallableResolverInterface
 
         if (!is_callable($toResolve) && is_string($toResolve)) {
             $class = $toResolve;
-            $method = '__invoke';
+            $instance = null;
+            $method = null;
 
             // check for slim callable as "class:method"
             $callablePattern = '!^([^\:]+)\:([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)$!';
@@ -64,18 +65,21 @@ final class CallableResolver implements CallableResolverInterface
             }
 
             if ($this->container instanceof ContainerInterface && $this->container->has($class)) {
-                $resolved = [$this->container->get($class), $method];
+                $instance = $this->container->get($class);
             } else {
                 if (!class_exists($class)) {
                     throw new RuntimeException(sprintf('Callable %s does not exist', $class));
                 }
-                $resolved = [new $class($this->container), $method];
+                $instance = new $class($this->container);
             }
 
             // For a class that implements RequestHandlerInterface, we will call handle()
-            if ($resolved[0] instanceof RequestHandlerInterface) {
-                $resolved[1] = 'handle';
+            // if no method has been specified explicitly
+            if ($instance instanceof RequestHandlerInterface && $method === null) {
+                $method = 'handle';
             }
+
+            $resolved = [$instance, $method ?? '__invoke'];
         }
 
         if ($resolved instanceof RequestHandlerInterface) {
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -36,7 +36,7 @@ final class CallableResolver implements CallableResolverInterface
     }
 
     /**
-     * Resolve toResolve into a closure that that the router can dispatch.
+     * Resolve toResolve into a callable that the router can dispatch.
      *
      * If toResolve is of the format 'class:method', then try to extract 'class'
      * from the container otherwise instantiate it and then dispatch 'method'.
--- a/Slim/Interfaces/CallableResolverInterface.php
+++ b/Slim/Interfaces/CallableResolverInterface.php
@@ -20,7 +20,7 @@ namespace Slim\Interfaces;
 interface CallableResolverInterface
 {
     /**
-     * Invoke the resolved callable.
+     * Resolve $toResolve into a callable
      *
      * @param mixed $toResolve
      *
--- a/tests/CallableResolverTest.php
+++ b/tests/CallableResolverTest.php
@@ -146,6 +146,16 @@ class CallableResolverTest extends TestCase
         $this->assertEquals("1", RequestHandlerTest::$CalledCount);
     }
 
+    public function testResolutionToAPsrRequestHandlerClassWithCustomMethod()
+    {
+        $request = $this->createServerRequest('/', 'GET');
+        $resolver = new CallableResolver(); // No container injected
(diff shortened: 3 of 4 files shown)


akrabat:
What's the use-case of implementing `RequestHandlerInterface` and then not wanting to use `handle()` ?

bnf:
@akrabat I see three usecases/reasons:

1. The main usecase is backwards compatibility:
When `class:method` dispatched `method` in previous version for classes that implement `RequestHandlerInterface`, it should stay that way (if possible).

The usecase for implementing `RequestHandlerInterface` in that scenario could be the following:
If the route target wraps some PSR-15 middleware internally, it passes itself (`$this`) to the middleware `process` function to achieve, that the wrapped middleware calls back to it's `handle` method. In this case the `handle` method is implemented to perform some internal work, but not meant to be dispatched by slim. The developer explicitly defined that by specifing a method in `class:method`. He really only wants to call `handle` if he specifies `class` as target (or defines the method name to be `handle`).

2. Probably a very bad reason, but I still name it.
Having multiple methods in a controller where `handle()` is the main action, that has some additional action that could be invoked when subsequent requests or so. (As said, this is an ugly usecase and should probably be split into multiple classes)

3. Being compliant to the documentation:
If someone defines a string `class:method` it is expected be honored, instead of being silently ignored (an alternative to silently ignoring `:method` would be an exception, but I'd vote against that).

l0gicgate:
I'm a bit perplexed by the existing `RequestHandlerInterface` logic that we have in `CallableResolver`. It's getting really messy in there, `handle()` shouldn't even be supported to be honest. The logic is so implicit, we need to stop including all this magic to make it easy and force the developers to make decisions and create adapters for existing mechanisms that don't necessarily fit their usage.

These are all gotchas that nobody actually knows about unless they read the code in `CallableResolver` or thoroughly scan our docs. 

In any case, I think that @bnf's logic is the better behavior. However this should be removed altogether from CallableResolver.

bnf:
> It's getting really messy in there

Well, I agree on that. For that reason my next pull request would have been to heavily refactor that class into smaller, type safe functions, that are way easier to grasp:
https://github.com/bnf/Slim/blob/834fd687938d84cea6f7a7f0c6858158f02c870b/Slim/CallableResolver.php
The related branch is here: https://github.com/slimphp/Slim/compare/4.x...bnf:improvements-experiment-v3
So maybe it's not a matter of removing that functionality, but rather cleaning the code up.

l0gicgate:
> https://github.com/bnf/Slim/blob/834fd687938d84cea6f7a7f0c6858158f02c870b/Slim/CallableResolver.php

That looks much better, after #2555 is merged you can open a different PR and refactor the entire `CallableResolver` as in the link you provided or you can use this PR and refactor once the merge happens but you'll have to rebase.

@akrabat we need to hold off on all merging before #2555 is in. I've rebased twice already.

bnf:
> @akrabat we need to hold off on all merging before #2555 is in. I've rebased twice already.

I think that's a bad attitude. #2555 is far from being ready to merge. Rebasing is a regular developer task, nothing to be mentioned as a burden. Blocking all other improvements for that reason is selfish and prevents progress, sorry.

This patch (in combination with the followup as shown in https://github.com/slimphp/Slim/compare/4.x...bnf:improvements-experiment-v3) and #2580 are actually preparatory improvements from the other PSR-15 patchset #2379 that you just closed. I re-added these as separate pull requests, as they would have been lost otherwise. What I want to say: these patches (in combination with the original PSR-15 patch #2379) do already wait way longer than your PSR-15 patch to be merged.

l0gicgate:
@bnf we are aiming to merge #2555 before the end of the week as @akrabat told me via Slack earlier this week. I understand your frustrations with #2379 being open for so long then closed but if I recall you didn't reply until mid January when I asked you back in October if you were going to rebase your PR. 

I really appreciate all your contributions to the project in the recent weeks, I'm pushing hard so we can release Slim 4, please be understanding of my decisions of doing things in a certain order.

Thank you

bnf:
> when I asked you back in October if you were going to rebase your PR

You didn't ask for a rebase. You asked:

> are you moving forward with this PR?

There was nothing to move "forward", it was ready for a very long time. ;)
Again, @akrabat was talking about exploring an alternative approach (native PSR-15 dispatcher). Now you proposed one, but I'm not convinced that it is the right one (I raised those complaints: stack vs queue dispatcher), and I don't see a public final decision on that. Therefore my assessment: It's far from being ready to merge.

> please be understanding of my decisions of doing things in a certain order

A project like this in not just you and *your* current patch and **your decisions**. Please think about that. I saw the same behaviour when your error handling refactorings took long to be reviewed. You became quite impatient. Technical reviews take time and are not about personal feelings, or the invested time, but about raw facts.

l0gicgate:
@bnf actually in October your PR needed rebasing because some changes had been made to the `4.x` branch since you had opened it back in August. That's what I meant by "are you moving forward with this PR".

As for my behavior, I've been almost solo dev'ing the `4.x` branch and tackling all the major core functionality that needed to be reworked. While you may dislike my impatience you need to understand that it is frustrating to have roadblocks constantly holding up further development when the finish line is in sight.

You're more than welcome to be more involved, I need all the help I can get but please don't take offense on my decisions to direct the focus to an area that I believe needs to be prioritized. This isn't about feelings, it's about reaching a milestone in a timely manner.

akrabat:
We merge what we can, when we can. 

I know that I'm the major blocker here as much because I'm slow to respond due to other things in my life. I'm also a slow thinker and need to see things and then consider them, which slows down major changes like the middleware one.

I would like to get 2555 in sooner rather than later, but if there's reservations, then it'll wait until we've resolved them to my satisfaction. I'm really valuing the discussion we've got going on 2555 and think we'll end up with a better solution.

i.e. don't go at each other's throats when you can just band together and blame me!

akrabat:
Re use cases mentioned earlier: 3 on its own is compelling enough to justify fixing this for alpha.

akrabat (approved):

Issue #2363, opened 2017-12-07
Slim Error Handling Issue


jbalbert:
Hi I just want to ask this issue about slim 3's error handling.

I can added custom error handler and exceptions. Like This:

**$container["errorHandler"] = function ($c) {
    return new \Slim\Handlers\ApiErrorHandler($c["logger"]);
};

$container["phpErrorHandler"] = function ($c) {
    return $c["errorHandler"];
};

class ApiErrorHandler extends \Slim\Handlers\Error
{	
	protected $logger;
	
	public function __construct(LoggerInterface $logger)
	{
		$this->logger = $logger;
	}

	public function __invoke(Request $request, Response $response, \Exception $exception){

	$this->logger->critical($exception->getMessage());
        $status = $exception->getCode() ?: 500;
        $message = $exception->getMessage();

        return $response->withJSON($message, $status);

	}

}**

However it does not capture fatal errors. I tried this one:

      **public function test($req, $res, $arg){

		$numer = 1 / 0;
		var_dump($numer);exit;
      }**

But I get response like this :

**<br />
<b>Warning</b>:  Division by zero in
<b>/var/www/html/myProject/backend/api/module/Controllers/UserController.php</b> on line
<b>24</b>
<br />
float(INF)**
 
That test code supposedly results to an exception. But slim's error handler doesn't catch it.

But when i do this : 

**$app->get('/', function () {    
    throw new Exception('Division by zero.');
	});**

It catches the exception.  And I got result like this : "Division by zero." And the error was inserted into my logs file.

Can you please tell where did I go wrong here? or Can please tell me the correct way to use slim's errorhandler?

jbalbert:
My bad it seems I was mistaken for the division by zero. ^_^

PR #2584, opened 2019-02-18
Slim 2.x Mcrypt Deprecation Notice Silencing

l0gicgate:
This is a proposed fix for #2362 

@akrabat I'm not sure if you're okay with using `@mcrypt_` to silence the warnings since of the behavior in 7.1 (7.2 is fine).

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -103,24 +103,24 @@ class Util
         $settings = array_merge($defaults, $settings);
 
         //Get module
-        $module = mcrypt_module_open($settings['algorithm'], '', $settings['mode'], '');
+        $module = @mcrypt_module_open($settings['algorithm'], '', $settings['mode'], '');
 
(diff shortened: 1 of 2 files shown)


akrabat:
IIRC `@` has quite a significant performance impact?

l0gicgate:
How often do encrypt/decrypt get called? Isn’t this just a utility file?

The performance hit is probably manageable.

akrabat:
The more I think about this, the more I think it's the wrong approach.

The problem with adding `@` is that if the user upgrades to 7.2 or 7.3, then they'll get a white screen with no error to help them.

I think we recommend that they explicitly disable the warning by turning off E_DEPRECATED themselves in the short term and in the longer term, they should encrypt their cookies themselves in middleware.

l0gicgate:
Okay, so close #2363 as won't fix?

Issue #334, opened 2012-07-12
Allow encoded URIs as named routes.


deiu:
I am often working with Linked Data (semantic web) with deals with encoded URIs. For this reason I was wondering how I can use them as named routes. Currently, if I try to use a URI for a named route, I get a 404 directly from Apache.

Please consider a short example with the following route: 

http://example.com/person/https%3A%2F%2Fmy-profile.eu%2Fpeople%2Fdeiu%2Fcard%23me

In my application, I would like to define the following route: 

$app->get('/person/:uri', function ($uri) {
    echo 'Requested URI: '.urldecode($uri);
});

which would output: Requested URI: https://my-profile.eu/people/deiu/card#me

StJudeWasHere:
I don't think this is a bug in Slim, the problem here could be caused by the %2F character which apache will automatically translate to /. You can avoid that behavior with the AllowEncodedSlashes directive:

http://httpd.apache.org/docs/2.2/mod/core.html#allowencodedslashes

Using AllowEncodedSlashes NoDecode made your example work for me. Please give it a try and confirm this.

Anyway, if for whatever cause you can't change that directive in your server I'll suggest passing the url as a query parameter.

```
// uri
http://example.com/person?uri=https%3A%2F%2Fmy-profile.eu%2Fpeople%2Fdeiu%2Fcard%23me

// then in the Slim app
$uri = $app->request()->get('uri');
```

Not as pretty but it would work too.

ghost:
Alternatively, you could `urlencode()` it twice.  Not very pretty, but it works too.

deiu:
Thanks! Unfortunately, both suggestions result in URIs that are not 'pretty' and conforming to REST goal of making them easy to read. I'll keep looking.

PR #2586, opened 2019-02-20
Fix for #2568

### LINKED ISSUE #2568: Command for preparing route cache file
Hello guys. I didn't find command for preparing cache file for routes. I have to create custom script which will generate route file before publishing my project to production.
This feature will be very useful, I think.

l0gicgate:
This is a fix for the logic for detecting if user provided router cache file is valid or not.

The following cases are valid:
- If the directory is not writable, but the file already exists and is writable
- If the directory is writable and the file specified does not exist

For the record, if you specify a file that doesn't exist, FastRoute will create it and populate it with the appropriate cache data.

This effectively closes #2568 

The PR for the docs is here [#334](https://github.com/slimphp/Slim-Website/pull/334)


Slim/Router.php
```diff
@@ -130,20 +130,38 @@ public function getBasePath()
...
         return $this;
     }
 
+    /**
+     * This method is in place to facilitate unit tests
+     * @param $cacheFile
+     *
+     * @return bool
+     */
+    protected function isCacheFileWritable($cacheFile)
```
l0gicgate: @akrabat as you can see I had to add this method because I cannot redefine the global function `is_writable()`.

Originally, I tried creating the directory/files required and setting the permissions correctly to hit those test cases but it was not working on the CI or my local windows build, only on local unix builds.

This way at least I can stub this method and make it return what I need to hit those test cases. Let me know if you have a better solution.

Slim/Router.php
```diff
@@ -130,20 +130,51 @@ public function getBasePath()
...
     }
 
+    /**
+     * This method is in place to facilitate unit tests
+     *
+     * @param $cacheFile
+     *
+     * @return bool
+     */
+    protected function isCacheFileReadable($cacheFile)
```
l0gicgate: @akrabat as you can see I had to add these methods `Router::isCacheFileReadable()` and `Router::isCacheDirWritable()` because I cannot redefine the global functions `is_writable()` or `is_readable()`. Which made unit testing almost impossible.

Originally, I tried creating the directory/files required and setting the permissions correctly to hit those test cases but it was not working on the CI or my local windows build, only on local unix builds.

This way at least I can stub this method and make it return what I need to hit those test cases. Let me know if you have a better solution.

Slim/Router.php
```diff
@@ -130,20 +130,51 @@ public function getBasePath()
      */
     public function setCacheFile($cacheFile)
     {
+        $this->cacheFile = $cacheFile;
```
akrabat: Don't set `$cacheFile` to the class property before we've validated it.

Slim/Router.php
akrabat: Technically, as this class is in a namespace, you could mock out `is_readable()`/`is_readable()`. See `tests/Assets/HeaderFunctions.php` - I wrote it up [here](https://akrabat.com/replacing-a-built-in-php-function-when-testing-a-component/).

Slim/Router.php
l0gicgate: why? if it's invalid, an exception is thrown anyway so it doesn't matter that it is set.

akrabat: The user could catch the exception and then continue.

tests/Assets/HeaderFunctions.php
```diff
@@ -57,3 +57,31 @@ function header($string, $replace = true, $statusCode = null)
         ]
     );
 }
+
```
akrabat, line 60: We should rename this file now that it containers functions other than header functions. Maybe `PhpFunctions.php`?

akrabat (approved):

Issue #2585, opened 2019-02-19
Test Suite Proposed Changes


l0gicgate:
We need to tidy up our test suite for **Slim 4**. I would like to propose the following:

**Add pre-defined code coverage directory in phpunit.xml.dist**
We should add coverage-html generation in our phpunit configuration and also add that pre-defined directory to our **.gitignore**
```xml
<phpunit>
    ...
    <logging>
        <log
            type="coverage-html"
            target="./coverage"
            lowUpperBound="20"
            highLowerBound="50"
        />
    </logging>
</phpunit>
```

<br/>

**Get rid of Pimple as a Dev Dependency**
All we need for our test suite is a simple implementation of `ContainerInterface`, we don't to anything but simple key/value setting. Less dependencies to install means faster builds. Plus the Pimple instantiation/synthax is terrible:
```php
use Pimple\Container as Pimple;
use Pimple\Psr11\Container as Psr11Container;

public function testRequiringContainer() {
    $pimple = new Pimple();
    $pimple['key'] = 'value';
    $container = new Psr11Container($pimple);
    ...
}
```

<br/>

**We need to update PHPUnit to 7.5 and remove deprecated methods**
Currently we are using `TestCase::assertAttributeEquals` a lot which is a deprecated method. Will need to refactor all tests using [deprecated methods from this list](https://github.com/sebastianbergmann/phpunit/issues/3338)

<br/>

**Break up AppTest into smaller modules**
This may take some time, but I want to make `App` a little bit more modular in the near future which will hopefully reduce the size of the `AppTest`. A lot of the test cases are from proxying `Router` methods.

<br/>

**Status**
- [x] Remove Pimple
- [x] Upgrade to PHPUnit 7.5
- [x] Add Code Coverage directory to .gitignore

akrabat:
Pimple should be mocked out for 4.x testing.

akrabat:
Agreed re PHPUnit upgrade.

l0gicgate:
Okay and what about the code coverage bit @akrabat? This is more so just for developer user experience, it's annoying when I start a new branch I have to update .gitignore every time, I do generate coverage locally to make sure I've done all the required tests.

At the minimum we should at least add a predefined coverage-html output directory to **.gitignore**

bnf:
I like the idea to remove the pimple dev dependency by mocking `ContainerInterface` in unit tests.
[...]

PR #2587, opened 2019-02-22
4.x Remove Pimple

l0gicgate:
Removing Pimple as per discussion in #2585.

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -8,13 +8,10 @@
  */
 namespace Slim\Tests;
 
-use Pimple\Container as Pimple;
-use Pimple\Psr11\Container as Psr11Container;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\UriInterface;
 use Psr\Http\Server\RequestHandlerInterface;
-use ReflectionClass;
 use Slim\App;
 use Slim\CallableResolver;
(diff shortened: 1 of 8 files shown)


tests/AppTest.php
```diff
@@ -39,8 +37,7 @@ public static function setupBeforeClass()
 
     public function testGetContainer()
     {
-        $pimple = new Pimple();
-        $container = new Psr11Container($pimple);
+        $container = new MockContainer();
```
akrabat, line 40: Why have a `MockContainer` object when we could be using PHPUnit's `createMock()` ?

l0gicgate: We need to have ArrayAccess implemented on the class unless you want to force usage of the `ContainerInterface::set()` method

akrabat: > We need to have ArrayAccess implemented on the class

Why? Slim doesn't set anything into the container anymore does it?

l0gicgate: I misread your question. We can mock a container for that specific test sure. We have a MockContainer though so why not use it

akrabat: Why do we have a `MockContainer` though? We shouldn't need it as we should be able to use PHPUnit's Mocking functionality for every situation where Slim's classes use a container. If we can't then we need to look at that particular Slim code.

l0gicgate:
Will refactor with Mocks instead and re-open.

Issue #2590, opened 2019-02-23
[4.x] CallableResolver is misused as Container provider


bnf:
#2555 introduced a new method `getContainer` for the `CallableResolverInterface`.
This is actually a hack that is used to hide a real dependency to the `ContainerInterface`. Classes that use `CallableResolverInterface::getContainer` need to be refactored to depend on `ContainerInterface` directly and the method `CallableResolverInterface::getContainer` should be removed again.

Here is an example how `Slim\Routable` uses the `CallableResolver` to fetch the container:
`$this->callableResolver->getContainer()`:
https://github.com/slimphp/Slim/blob/5311e952552dcd3c8c49a363c438649031965783/Slim/Routable.php#L63-L64

l0gicgate:
There’s more changes coming to Route/RouteGroups/Routable that is upcoming as I discussed in #2583. It will deal with all of this.

l0gicgate:
Closing #2591 has been merged

PR #2591, opened 2019-02-23
Refactor middleware into stack

bnf:
Improve certain things:

- Do not hide container dependencies in the CallableResolverInterface
  getContainer method (this method is dropped now)
- Use a middleware stack instead of a queue, as slim used to use before #2555
- Use designated request handlers as kernel (as we know them from slim v3)
  instead of middlewares that ignore the second (RequestHandler) parameter
- Do not recreate middlewares in route/route-group middleware combination,
  rather chain middleware dispatchers, thanks to the (now) proper use of
  RequestHandlerInterface of the MiddlewareDispatcher which allows itself
  to be chained into another middleware stack
- Drop some helper middlewares, which is good as they were applicable only
  as internal middleware anyway. This logic is now handled in the middleware
  dispatcher for optimization reasons (because that handler
  create a RequestHandler wrapper for every middleware, it's most performant
  to include lazy/closure dispatch logic in the RequestHandler wrappers).
- DispatchMiddleware (and it's corresponding test) is moved into the
  Router which now acts as a PSR-15 RequestHandler
- Replace middleware related tests that test internal state (e.g. whether
  something is at the bottom of a list.) That is an implementation detail and
  should rather be verified by checking for proper operation
- Rename MiddlewareRunner to MiddlewareDispatcher, as it actually dispatches
  middlewares now
- Add more MiddlewareDispatcher tests

Fixes: #2590


Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+     */
+    public function __construct(
+        RequestHandlerInterface $kernel,
+        ContainerInterface $container = null,
+        array $middlewares = []
+    ) {
+        $this->seedMiddlewareStack($kernel);
+        $this->container = $container;
+
+        foreach ($middlewares as $middleware) {
```
l0gicgate: This should be a separate setter. No logic should be in constructor.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Add a new middleware to the stack
+     *
+     * Middlewares are organized as a stack. That means middlewares
+     * that have been added before will be executed after the newly
+     * added one (last in, first out).
+     *
+     * @param MiddlewareInterface|string|callable $middleware
+     */
+    public function add($middleware): bool
```
l0gicgate: This method should throw, not return a boolean.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+     * Middlewares are organized as a stack. That means middlewares
+     * that have been added before will be executed after the newly
+     * added one (last in, first out).
+     *
+     * @param MiddlewareInterface $middleware
+     */
+    public function addMiddleware(MiddlewareInterface $middleware)
+    {
+        $next = $this->tip;
+        $this->tip = new class($middleware, $next) implements RequestHandlerInterface {
```
l0gicgate, line 109: No class declaration inside functions like this, separate it, helps with testability as well.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+     * Middlewares are organized as a stack. That means middlewares
+     * that have been added before will be executed after the newly
+     * added one (last in, first out).
+     *
+     * @param string $middleware
+     */
+    public function addLazy(string $middleware)
+    {
+        $next = $this->tip;
+        $this->tip = new class($middleware, $next, $this->container) implements RequestHandlerInterface {
```
l0gicgate, line 138: Same comment as before, no class declaration inside functions like this.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+     * Middlewares are organized as a stack. That means middlewares
+     * that have been added before will be executed after the newly
+     * added one (last in, first out).
+     *
+     * @param callable $middleware
+     */
+    public function addCallable(callable $middleware)
+    {
+        $next = $this->tip;
+        $this->tip = new class($middleware, $next) implements RequestHandlerInterface {
```
l0gicgate, line 188: Same comment as before, no class declaration inside functions like this.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Add a new middleware by class name
+     *
+     * Middlewares are organized as a stack. That means middlewares
+     * that have been added before will be executed after the newly
+     * added one (last in, first out).
+     *
+     * @param string $middleware
+     */
+    public function addLazy(string $middleware)
```
l0gicgate: This method name should be something else, `lazy` isn't a good word. Use `deferred` as we already agreed upon that and the terminology is used elsewhere in the repo.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,211 @@
...
+
+            public function __construct(string $middleware, RequestHandlerInterface $next, ContainerInterface $container = null)
+            {
+                $this->middleware = $middleware;
+                $this->next = $next;
+                $this->container = $container;
+            }
+
+
+            public function handle(ServerRequestInterface $request): ResponseInterface
```
l0gicgate, line 153: All the logic contained in this method should be in a resolve method instead. Also, use the logic from `DeferredResolutionMiddleware` as @akrabat and I already agreed upon it.

Slim/Route.php
```diff
@@ -97,6 +105,7 @@ public function __construct(
         $callable,
         ResponseFactoryInterface $responseFactory,
         CallableResolverInterface $callableResolver,
+        ContainerInterface $container = null,
```
l0gicgate, line 114: No null parameters before a required one, put `$invocationStrategy` first.

Slim/RouteGroup.php
```diff
@@ -23,6 +23,13 @@
  */
 class RouteGroup extends Routable implements RouteGroupInterface
 {
+    /**
+     * middlewares
+     *
+     * @var array
+     */
+    protected $middlewares = [];
```
l0gicgate: The plural of `middleware` is still `middleware` and not `middlewares`. Please refactor that.

Slim/Router.php
```diff
@@ -102,12 +111,14 @@ class Router implements RouterInterface
      *
      * @param ResponseFactoryInterface      $responseFactory
      * @param CallableResolverInterface     $callableResolver
+     * @param ContainerInterface            $container
```
l0gicgate: This should be `ContainerInterface|null` as well

Slim/Router.php
```diff
@@ -47,6 +51,11 @@ class Router implements RouterInterface
      */
     protected $callableResolver;
 
+    /**
+     * @var ContainerInterface
```
l0gicgate: This should be `ContainerInterface|null`

Slim/Router.php
```diff
@@ -482,4 +494,29 @@ public function urlFor(string $name, array $data = [], array $queryParams = []):
...
+     * last and it detects whether or not routing has been performed in the user
+     * defined middleware stack. In the event that the user did not perform routing
+     * it is done here
+     *
+     * @param ServerRequestInterface $request
+     * @return ResponseInterface
+     * @throws HttpNotFoundException
+     * @throws HttpMethodNotAllowedException
+     */
+    public function handle(ServerRequestInterface $request): ResponseInterface
```
l0gicgate: I'm not a huge fan of this, the Router's responsibility to handle requests. If @akrabat agrees then sure. I just don't think it belongs here.

tests/MiddlewareDispatcherTest.php
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2018 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+namespace Slim\Tests;
+
+use Pimple\Container as Pimple;
+use Pimple\Psr11\Container as Psr11Container;
+use Prophecy\Argument;
```
l0gicgate, line 13: I don't see the need to add an entire mocking framework for these tests.

tests/MiddlewareDispatcherTest.php
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @link      https://github.com/slimphp/Slim
+ * @copyright Copyright (c) 2011-2018 Josh Lockhart
+ * @license   https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+namespace Slim\Tests;
+
+use Pimple\Container as Pimple;
```
l0gicgate, line 11: Just as an FYI, we are getting rid of Pimple entirely as a dev dependency. See PR #2587. You will need to refactor those tests.

l0gicgate (changes requested):

l0gicgate:
Also please make sure that Travis CI build passes. Thank you.

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,213 @@
...
+    protected $tip;
+
+    /**
+     * @var ContainerInterface
+     */
+    protected $container;
+
+    /**
+     * @param RequestHandlerInterface $kernel
+     * @param ContainerInterface $container
```
l0gicgate: This should be `ContainerInterface|null` also

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,213 @@
...
+{
+    /**
+     * Tip of the middleware call stack
+     *
+     * @var RequestHandlerInterface
+     */
+    protected $tip;
+
+    /**
+     * @var ContainerInterface
```
l0gicgate: This should be `ContainerInterface|null`

l0gicgate (dismissed):

Slim/MiddlewareDispatcher.php
bnf: Well, I disagree. Its fine to put constructing logic into the constructor.

Slim/MiddlewareDispatcher.php
bnf: No, testability is as easy as with seperate classes. Anonymous provide the same level of testability as closure, which we used for the use cases pre #2555.

Slim/Router.php
bnf: yup

tests/MiddlewareDispatcherTest.php
bnf: It's part of phpunit. Really no issue here.

Slim/Router.php
bnf: I'm a huge fan of this. The current approach of misusing middlewares for this is actually wrong. The router dispatches the request to a route. It is the RequestHandler.

tests/MiddlewareDispatcherTest.php
bnf: yes, these tests are copied from the previous DeferredDispatcher tests.. I included them this way to keep it reviewabl.
New tests (as in the second commit) therefore do not the pimple container, but rather mock the container (and request and everything) using prophecies.

Slim/MiddlewareDispatcher.php
l0gicgate: Okay, well I'm requesting that you change it. Thank you.

Slim/MiddlewareDispatcher.php
l0gicgate: I'm requesting that you separate those classes. Thank you.

tests/MiddlewareDispatcherTest.php
l0gicgate: If PHPUnit drops the dependency unexpectedly, then our build fails. So you have to include it in our `composer.json` if you're using it. It's a no from me, but @akrabat can make the final decision.

Slim/Router.php
l0gicgate: Cool, @akrabat can make the call on this one.

Slim/MiddlewareDispatcher.php
froschdesign: > No logic should be in constructor.

A condition is already a logic and is used in many constructors. 😉

tests/MiddlewareDispatcherTest.php
froschdesign: @l0gicgate 
> If PHPUnit drops the dependency unexpectedly…

It's never _unexpectedly_. Everything else means you used the ["death star version constraint" on PHPUnit](https://thephp.cc/news/2017/02/death-star-version-constraint) in your `composer.json`. 😉

Slim/MiddlewareDispatcher.php
l0gicgate: > A condition is already a logic and is used in many constructors. 😉

There's 2 constructors in this entire repository with an if condition. I would like to keep the constructors free of logic as much as possible.

tests/MiddlewareDispatcherTest.php
bnf: phpunit provides prophecy through its `TestCase::prophesize()` method. That means it's mostly an implementation detail, and could only be removed in a major release, which requires changes to slim's composer.json to update. No unexpected build fails..
Anyway, because I'm explicitly using the namespace `Prophecy` here, I'm fine to include prophecy as require-dev dependency – still, this is not a classic transitive dependency issue, as phpunit provides prophecy as part of its API.

Slim/MiddlewareDispatcher.php
bnf: yup, I agree and will adapt.

Slim/MiddlewareDispatcher.php
froschdesign: I found 4 constructors in the 4.x branch here on Github! :stuck_out_tongue_closed_eyes:

froschdesign: Regardless, it would be better to document these requirements in a coding standards. Otherwise, such things will appear again and again – in pull requests and reviews.

l0gicgate: We have a discussion in #2502 about coding standards we’d like to implement. It’s in the works! You’re more than welcome to join the discussion!

tests/MiddlewareDispatcherTest.php
l0gicgate: That was mainly my point, you’re explicitly using that namespace which means we should add it to our `composer.json` file regardless if PHPUnit uses/provides it or not.

On another note, if we’re going to use that style of testing (Prophesizing interfaces) then we should slowly migrate the entire test suite to do this so we maintain cohesiveness. 

Please add input on #2585 @bnf

Slim/MiddlewareDispatcher.php
```diff
@@ -0,0 +1,213 @@
...
+                $this->middleware = $middleware;
+                $this->next = $next;
+                $this->container = $container;
+            }
+
+            public function handle(ServerRequestInterface $request): ResponseInterface
+            {
+                if ($this->container !== null && $this->container->has($this->middleware)) {
+                    $middleware = $this->container->get($this->middleware);
+                } elseif (is_subclass_of($this->middleware, MiddlewareInterface::class)) {
```
akrabat: Currently, if the container factory returns a string that's the name of a class that implements `MiddlewareInterface`, then `4.x` will instantiate that class and use it. In this code, that no longer happens, I think?

akrabat:
A few quick benchmarks using siege indicates that this PR is 15% faster than the current `4.x`.

Slim/MiddlewareDispatcher.php
l0gicgate: That's correct @akrabat. If the container resolves to a class name and not an instantiated object then this will not instantiate the object as was being done in [DeferredResolutionMiddleware](https://github.com/slimphp/Slim/blob/4.x/Slim/Middleware/DeferredResolutionMiddleware.php#L64-L66)
```php
$container = new Container();
$container['myClass'] = MyMiddleware::class; // Not instantiated
```

As I said, the logic from `DeferredResolutionMiddleware` should be implemented.

Slim/Router.php
l0gicgate: @bnf another thing, when you put that logic inside `Router` you're assuming that whoever implements that `RouterInterface` is going to copy that logic and if they don't then the expected behavior doesn't happen. As seen in `App::__construct()` you can provide your own `RouterInterface`. I really believe that logic doesn't belong here.

Slim/MiddlewareDispatcher.php
bnf: To get this discussion resolved for this thread I've dropped this part from the constructor. It was mostly used for tests anyway, which means it's actually unneeded.

Slim/Router.php
bnf: Yup, fixed.

Slim/Router.php
bnf: Hm.. I understand that this seems strange. So let me name my argument: A router actually routes something, this is what it does with `handle` now. Previously the router was only a route-calculator, but not a router. To reflect this change I actually updated the `RouterInterface`.
To sum up: We have two operations: route calculation and route dispatch. That means one middleware and one dispatcher. I changed the `Router` to be a dispatcher now.
Maybe the naming is the problem here. We should probably put the route calculation logic directly into the `RoutingMiddleware` and then have a `RouteDispatcher` and no `Router` at all.

tests/MiddlewareDispatcherTest.php
bnf: I've added the dependency to composer.json now.

Slim/MiddlewareDispatcher.php
bnf: You are right and I have no problem with implementing that (and did so for now), but well, may I ask: what is the usecase to return a `string` or a `Closure` from a DI container that needs to be instantiated manually with `new` then? If one actually adds something to a container, then a proper object implementing `MiddlewareInterface` should and can be easily used, no?
I'm really fine with keeping the current behaviour with this patch, but maybe this behaviour should be revisted within another patch?

Slim/MiddlewareDispatcher.php
bnf: fixed.

Slim/MiddlewareDispatcher.php
l0gicgate: It's the same behavior as `CallableResolverInterface` as seen here https://github.com/slimphp/Slim/blob/4.x/Slim/CallableResolver.php#L73. 

This case only usually happens in the event that no container is present. So the instantiation of the object is deferred to when it gets resolved. It could happen with a container as well but very rarely would a user intentionally do this however some do.

Also see issue #2592 we need to pass container in there to duplicate the callable resolver behavior. I created that issue because someone who's doing testing on `4.x` said that they didn't understand why the same behavior from `CallableResolver` wasn't being replicated in `DeferredResolutionMiddleware` (passing the container to the instantiated object's constructor). [See original message from Slack Discussion](https://slimphp.slack.com/archives/C06SHUR19/p1551046761239000). That makes me thing this feature is being used.

Slim/Router.php
l0gicgate: I think that we should separate the current `Router` into 2 interfaces. `RouteCollectorInterface` and `RouteDispatcher` interface.

The first interface will be `RouteCollectorInterface`:
```php
interface RouteCollectorInterface {
    public function map(array $methods, string $pattern, $handler): RouteInterface;

    public function pushGroup(string $pattern, $callable): RouteGroupInterface;

    public function popGroup();

    public function getNamedRoute(string $name): RouteInterface;

    public function lookupRoute(string $identifier): RouteInterface;

    public function relativePathFor(string $name, array $data = [], array $queryParams = []): string;

    public function pathFor(string $name, array $data = [], array $queryParams = []): string;

    public function setCacheFile(?string $cacheFile): RouterInterface;

    public function setDefaultInvocationStrategy(InvocationStrategyInterface $strategy);
}
```

And a `RouteDispatcherInterface`:
```php
use Psr\Http\Server\RequestHandlerInterface;
use Slim\RoutingResults;

class RouteDispatcherInterface extends RequestHandlerInterface {
    public function dispatch(ServerRequestInterface $request): RoutingResults;

    public function setDispatcher(Dispatcher $dispatcher): RouteDispatcherInterface;
}
```

akrabat: There are three things that need to happen:

1. Registering route paths to handlers
2. Determination of the route handler to execute based on the request's URI
3. Executing the handler

These all need to be separate.

1. Registration happens before `run()`
2. Determination of the route handler to execute needs to occur in the middle of the middleware execution so that the user can register middleware before and after it happens.
3. Execution of the handler happens last

Within Symfony's Routing component, there is a `RouteCollection` object that you register route paths too. . Then `Router::matchRequest()` (which calls `UriMatcher::match()` is used to determine the controller to execute. The actual execution is done completely separately in the HttpKernel component, where `Kernel::handleRaw()` method instantiates a controller class (via an event) and then invokes it using `$controller(...$arguments)`

To circle back to Slim, we need to have these tasks separated too. I don't mind what we call them, but they must be separate. In Slim 3, you register your paths into the Router object and then use it to determine the handler to be  executed. This determination is very clunky in Slim 3's `App` and can only be at the start or end of the middleware chain. The actual dispatching is the application's job and is done in `App::__invoke()`. Due to the way Slim 3's middleware stack is set up, `App::__invoke` is _always_ the innermost middleware of the stack.  

In `4.x` today, `App` is no longer its own middleware and the dispatching is now in `DispatchMiddleware` which is much more explicit as now `DispatchMiddleware` is the innermost middleware. I agree with @bnf that  it should be a request handler, not a middleware though as we're never gonna call `handle`. 

 In `2591`, the Router does the dispatching too as it is a request handler in addition to a router though it doesn't explicitly implement `RequestHandlerInterface`. I'm not so much a fan of this as it makes the router do double duty. However, our Route objects are remarkably confused on this point too mind you.

Slim/MiddlewareDispatcher.php
akrabat: > what is the usecase to return a `string` or a `Closure` from a DI container that needs to be instantiated manually with `new` then? If one actually adds something to a container, then a proper object implementing `MiddlewareInterface` should and can be easily used, no?

This is a good question. I was mostly observing how it currently works rather than asking whether it should do this or not :)

Let's open an issue to consider it separately as this PR is already a bit too big…

Slim/MiddlewareDispatcher.php
akrabat: FWiW, I don't mind the anonymous functions as they are analogous to the closures we used in Slim 3. One advantage is that a user can't instantiate them, so they can never be part of Slim 4's public API and so have to be supported by us. One disadvantage is that they are opaque and not as explicit as a named class.

 I don't hold a strong opinion on this at all and can go either way.

Slim/MiddlewareDispatcher.php
akrabat: I also prefer the _deferred_ over _lazy_.

Slim/Router.php
l0gicgate: @akrabat I agree.

App should be in charge of these concerns:
- Collecting routes using `RouteCollector`
- Then we should have `App` create a `RouteResolver` which takes in the `RouteCollector`, creates the `FastRoute\Dispatcher`.
- Then we need to remove `DispatchMiddleware` (`Router::handle()` in this PR) and simply run middleware signature detection in `App`. We can simply have the `App::addMiddleware()` method look to see if `RoutingMiddleware` has been added to the queue or not, if it hasn't been then we add it to `MiddlewareRunner` right before we call `$this->middlewareRunner->handle($request)` in `App`
- After the `MiddlewareRunner` finished its job `App` can run the appropriate route and return the response.

And voila. You now have gotten rid of all the problems and separated concerns properly, Let me fork this PR and show the changes I would apply.

Slim/MiddlewareDispatcher.php
bnf: Yup, fine by me.

Slim/RouteGroup.php
bnf: No, it's `middlewares`. See https://en.wiktionary.org/wiki/middleware
> middleware (usually uncountable, plural **middlewares**)

Slim/Route.php
bnf: `$invocationStrategy` actually is nullable, no need to refactor this one.

Slim/MiddlewareDispatcher.php
bnf: I think that is out of scope for this patch, as @akrabat already said question regarding resolving should be handled in another issue.

Slim/MiddlewareDispatcher.php
bnf: Ok, fine by me. Changed that to be `addDeferred`.

Slim/MiddlewareDispatcher.php
bnf: That's exactly the reason why I thought anonymous classes would actually be the right choice here. There is really only one reason for using a class at all: We need to fullfill the `RequestHandlerInterface` requirement of PSR-15 middlewares. So the anonymous classes are just an implementation detail of the `MiddlewareDispatcher` here (and don't differ much from the Closure that we used in slim v3).
These classes are basicly closures (of course I do not refer `\Closure`s here, just the semantic meaning) that implement an interface, and the language construct to implement that is `new class`.

Slim/Router.php
bnf: @akrabat ok, fine. I've moved the dispatch logic out of the `Router` again and moved it into a separate `RouteDispatcher` which implements `RequestHandlerInterface`.

No need to fork this pull request..

l0gicgate: @akrabat see PR on @bnf's forked repository https://github.com/bnf/Slim/pull/1

I separated `Router` into the appropriate components. Now you have the following components:

- `RouteCollector` which collects the routes
- `RouteResolver` which creates the dispatcher, computes the `RoutingResults` and resolves the appropriate `Route` when queried
- `RouteRunner` which is added as the kernel of `MiddlewareDispatcher` to see if Routing has been done

Refactored all tests accordingly.

It's much cleaner now.

Slim/MiddlewareDispatcher.php
l0gicgate: Those are good points @akrabat & @bnf. I can live with the internal class declaration, I'll just close my eyes xD

Slim/RouteGroup.php
l0gicgate: @bnf we’re going to go with “middleware”. I don’t care about what wikipedia says. Just like we don’t say “softwares” or “hardwares”, the word “middleware” is uncountable as stated in the wiki you linked. You can leave it, I’ll just change it myself later.

Slim/MiddlewareDispatcher.php
l0gicgate: Ok, I’ll PR it myself later. Don’t worry about it.

bnf: No, sorry, I definitely overlooked that slim always used `middleware` as plural. I've updated my changes to reflect that.

Slim/Router.php
akrabat: > I've moved the dispatch logic out of the `Router` again and moved it into a separate `RouteDispatcher` which implements `RequestHandlerInterface`.

Lovely, thanks! I can't see the commit where you did this though?

tests/MiddlewareDispatcherTest.php
```diff
@@ -122,4 +125,237 @@ public function testResolveThrowsExceptionWithoutContainerAndUnresolvableClass()
...
+        $dispatcher->add($middleware3Prophecy->reveal());
+
+        $response = $dispatcher->handle($requestProphecy->reveal());
+
+        $this->assertSame(['3', '2', '1', '0'], $response->getHeader('X-SEQ-PRE-REQ-HANDLER'));
+        $this->assertSame(['0', '1', '2', '3'], $response->getHeader('X-SEQ-POST-REQ-HANDLER'));
+        $this->assertSame(204, $response->getStatusCode());
+    }
+
+    public function testDoesNotInstantiateLazyMiddlewareInCaseOfAnEarlyReturningOuterMiddleware()
```
l0gicgate: **Should use the keyword "Deferred" instead of "Lazy":**
`testDoesNotInstantiateDeferredMiddlewareInCaseOfAnEarlyReturningOuterMiddleware()`

tests/MiddlewareDispatcherTest.php
```diff
@@ -122,4 +125,237 @@ public function testResolveThrowsExceptionWithoutContainerAndUnresolvableClass()
...
+        $dispatcher->addDeferred(MockSequenceMiddleware::class);
+        $dispatcher->addMiddleware($middlewareProphecy->reveal());
+        $response = $dispatcher->handle($requestProphecy->reveal());
+
+        $this->assertFalse(MockSequenceMiddleware::$hasBeenInstantiated);
+        $this->assertEquals($responseProphecy->reveal(), $response);
+        $kernelProphecy->handle(Argument::type(ServerRequestInterface::class))->shouldNotHaveBeenCalled();
+    }
+
+    public function testThrowsExceptionForLazyNonMiddlewareInterfaceClasses()
```
l0gicgate: **Should use the keyword "Deferred" instead of "Lazy":**
`testThrowsExceptionForDeferredNonMiddlewareInterfaceClasses()`

l0gicgate:
Please add test coverage on:
https://coveralls.io/builds/21930483/source?filename=Slim%2FApp.php#L112
https://coveralls.io/builds/21930483/source?filename=Slim%2FRoute.php#L292
https://coveralls.io/builds/21930483/source?filename=Slim%2FRouteGroup.php#L69

akrabat (approved):

PR #2602, opened 2019-03-03
4.x Replace Pimple by Prophecy based mocks

bnf:
Remove Pimple as discussed in #2585.

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -14,7 +14,6 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\UriInterface;
 use Psr\Http\Server\RequestHandlerInterface;
-use ReflectionClass;
 use Slim\App;
 use Slim\CallableResolver;
 use Slim\Error\Renderers\HtmlErrorRenderer;
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -8,6 +8,7 @@
  */
 namespace Slim\Tests;
 
+use Prophecy\Argument;
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
@@ -31,6 +32,16 @@ class AppTest extends TestCase
     public static function setupBeforeClass()
     {
         ini_set('error_log', tempnam(sys_get_temp_dir(), 'slim'));
+    }
+
+    public function testDoesNotUseContainerAsServiceLocator()
+    {
+        $containerProphecy = $this->prophesize(ContainerInterface::class);
+        $responseFactory = $this->getResponseFactory();
+        $app = new App($responseFactory, $containerProphecy->reveal());
+
+        $containerProphecy->has(Argument::type('string'))->shouldNotHaveBeenCalled();
+        $containerProphecy->get(Argument::type('string'))->shouldNotHaveBeenCalled();
     }
 
     /********************************************************************************
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -8,8 +8,7 @@
  */
 namespace Slim\Tests;
 
-use Pimple\Container as Pimple;
-use Pimple\Psr11\Container as Psr11Container;
+use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\UriInterface;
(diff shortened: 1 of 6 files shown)


tests/AppTest.php
```diff
@@ -39,33 +37,30 @@ public static function setupBeforeClass()
...
-        $pimple = new Pimple();
-        $container = new Psr11Container($pimple);
-        $callableResolver = new CallableResolver($container);
+        $containerProphecy = $this->prophesize(ContainerInterface::class);
+        $callableResolver = new CallableResolver($containerProphecy->reveal());
         $responseFactory = $this->getResponseFactory();
-        $router = new Router($responseFactory, $callableResolver, $container);
-        $app = new App($responseFactory, $container);
+        $router = new Router($responseFactory, $callableResolver, $containerProphecy->reveal());
+        $app = new App($responseFactory, $containerProphecy->reveal());
```
akrabat: Couldn’t we pass in null for the container here? Pretty sure we no longer look in the container for the router.

akrabat:
LGTM 👍

tests/AppTest.php
l0gicgate: Yes @akrabat 

```php
    public function testGetRouter()
    {
        $responseFactory = $this->getResponseFactory();
        $routerProphecy = $this->prophesize(RouterInterface::class);
        $app = new App($responseFactory, null, null, $routerProphecy->reveal());

        $this->assertEquals($routerProphecy->reveal(), $app->getRouter());
    }
```

akrabat: Good. Mocking the router is a separate PR of course :)

l0gicgate (dismissed):
I'm actually a huge fan of Prophecy now thanks to you @bnf. I started refactoring `AppTest` and mocking all the PSR-7 objects using that (upcoming PR for that). Thank you for this work!

tests/AppTest.php
l0gicgate: No need to mock the router in this instance since we're just mocking the getter on `App`.

tests/AppTest.php
```diff
@@ -39,33 +37,30 @@ public static function setupBeforeClass()
...
     public function testGetCallableResolver()
     {
-        $pimple = new Pimple();
-        $container = new Psr11Container($pimple);
-        $callableResolver = new CallableResolver($container);
+        $containerProphecy = $this->prophesize(ContainerInterface::class);
+        $callableResolver = new CallableResolver($containerProphecy->reveal());
         $responseFactory = $this->getResponseFactory();
-        $app = new App($responseFactory, $container);
+        $app = new App($responseFactory, $containerProphecy->reveal());
```
l0gicgate: The `CallableResolver` isn't being passed into `App`. This test should be:
```php
    public function testGetCallableResolver()
    {
        $responseFactory = $this->getResponseFactory();
        $callableResolverProphecy = $this->prophesize(CallableResolverInterface::class);
        $app = new App($responseFactory, null, $callableResolverProphecy->reveal());

        $this->assertEquals($callableResolverProphecy->reveal(), $app->getCallableResolver());
    }
```

tests/AppTest.php
bnf: An alternative to passing `null` would be to inject a `ContainerInterface` mock that asserts that the router isn't retrieved from it, e.g.:
```php
$containerProphecy->has(Router::class)->shouldNotHaveBeenCalled();
$containerProphecy->get(Router::class)->shouldNotHaveBeenCalled();
```

What do you think?

akrabat:
I'm easy either way.

tests/AppTest.php
l0gicgate: The `CallableResolver` isn't being passed into `App`. This test should be:
```php
    public function testGetCallableResolver()
    {
        $responseFactory = $this->getResponseFactory();
        $callableResolverProphecy = $this->prophesize(CallableResolverInterface::class);
        $app = new App($responseFactory, null, $callableResolverProphecy->reveal());

        $this->assertEquals($callableResolverProphecy->reveal(), $app->getCallableResolver());
    }
```

l0gicgate (changes requested):

tests/AppTest.php
l0gicgate: @bnf what is the purpose of doing that I'm curious? Since all those parameters are typed, why would `App` attempt to retrieve `Router` from `ContainerInterface`?

bnf: The intention would be to assert that the class doesn't use the container as service locator (for direct dependencies), but only for lazy (dispatching) work. Anyway, that assertion would be a better fit for a separate test. (will push a commit that does that with a test method called `testDoesNotUseContainerAsServiceLocator`)

Still, I think the container should be passed here to verify that`App` injects the container into the `Router` that it creates.
*But*, I figure that we have two different cases here: Router injection and router creation. Will push a patch to test both cases.
For the second case (router creation, which `testGetRouter` currently tests for), it's important to pass `ContainerInterface` to assert that `App` actually forwards the `ContainerInterface` injection to `Router`.

tests/AppTest.php
bnf: Yup, that test is missing (though I'd use `assertSame` instead of `assertEquals` to verify that the same object is referenced), but I think that the current test is also a valid one (with another name), as it tests the creation of the `CallableResolver`. I've pushed a commit that adds both, see https://github.com/slimphp/Slim/pull/2602/commits/9cb619972896e008e1db361fe6b71c0c7c09e3f4

l0gicgate: Fantastic! Thank you

l0gicgate (approved):

akrabat (approved):

PR #2603, opened 2019-03-03
4.x Refactor AppTest To Use Prophecy

l0gicgate:
This is a continuation of migrating towards the usage of `phpspec/prophecy`.

- Prophesized PSR-7 Object Interfaces
- Prophesized ContainerInterface
- Prophesized MiddlewareInterface
- No more deprecated `$this->assertAttributeEquals()` calls!

@bnf and @akrabat please review!

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -113,130 +113,139 @@ class AppTest extends TestCase
      * Router proxy methods
      *******************************************************************************/
 
-    public function testGetRoute()
+    public function upperCaseRequestMethodsProvider()
     {
-        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
-        $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->get('/', function () {
-        });
-
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['GET'], $methodsProperty->getValue($route));
+        return [
+            ['GET'],
+            ['POST'],
+            ['PUT'],
+            ['PATCH'],
+            ['DELETE'],
+            ['OPTIONS'],
+        ];
     }
 
-    public function testPostRoute()
+    public function lowerCaseRequestMethodsProvider()
     {
-        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
-        $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->post('/', function () {
-        });
-
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['POST'], $methodsProperty->getValue($route));
+        return [
+            ['get'],
+            ['post'],
+            ['put'],
+            ['patch'],
+            ['delete'],
+            ['options'],
+        ];
     }
 
-    public function testPutRoute()
+    /**
+     * @dataProvider upperCaseRequestMethodsProvider()
+     */
+    public function testGetPostPutPatchDeleteOptionsMethods($method)
     {
+        $streamProphecy = $this->prophesize(StreamInterface::class);
+        $streamProphecy->__toString()->willReturn('Hello World');
+
+        $responseProphecy = $this->prophesize(ResponseInterface::class);
+        $responseProphecy->getBody()->willReturn($streamProphecy->reveal());
+
         $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
-        $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->put('/', function () {
+        $responseFactoryProphecy->createResponse()->willReturn($responseProphecy->reveal());
+
+        $uriProphecy = $this->prophesize(UriInterface::class);
+        $uriProphecy->getPath()->willReturn('/');
+
+        $requestProphecy = $this->prophesize(ServerRequestInterface::class);
+        $requestProphecy->getMethod()->willReturn($method);
+        $requestProphecy->getUri()->willReturn($uriProphecy->reveal());
+        $requestProphecy->getAttribute('routingResults')->willReturn(null);
+        $requestProphecy->withAttribute(Argument::type('string'), Argument::any())->will(function ($args) {
+            $clone = clone $this;
+            $clone->getAttribute($args[0])->willReturn($args[1]);
+            return $clone;
         });
 
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['PUT'], $methodsProperty->getValue($route));
-    }
-
-    public function testPatchRoute()
-    {
-        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
+        $methodName = strtolower($method);
         $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->patch('/', function () {
+        $app->$methodName('/', function ($request, ResponseInterface $response) {
+            return $response;
         });
+        $response = $app->handle($requestProphecy->reveal());
 
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['PATCH'], $methodsProperty->getValue($route));
-    }
-
-    public function testDeleteRoute()
-    {
-        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
-        $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->delete('/', function () {
-        });
-
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['DELETE'], $methodsProperty->getValue($route));
-    }
-
-    public function testOptionsRoute()
-    {
-        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
-        $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->options('/', function () {
-        });
-
-        $methodsProperty = new ReflectionProperty(Route::class, 'methods');
-        $methodsProperty->setAccessible(true);
-
-        $this->assertInstanceOf(Route::class, $route);
-        $this->assertEquals(['OPTIONS'], $methodsProperty->getValue($route));
+        $this->assertEquals('Hello World', (string) $response->getBody());
     }
 
     public function testAnyRoute()
     {
+        $streamProphecy = $this->prophesize(StreamInterface::class);
+        $streamProphecy->__toString()->willReturn('Hello World');
+
+        $responseProphecy = $this->prophesize(ResponseInterface::class);
+        $responseProphecy->getBody()->willReturn($streamProphecy->reveal());
+
         $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
+        $responseFactoryProphecy->createResponse()->willReturn($responseProphecy->reveal());
+
         $app = new App($responseFactoryProphecy->reveal());
-        $route = $app->any('/', function () {
+        $app->any('/', function ($request, ResponseInterface $response) {
+            return $response;
         });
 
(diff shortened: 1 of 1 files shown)


tests/AppTest.php
```diff
@@ -917,525 +1115,1133 @@ public function testEmptyGroupWithEmptyNestedGroupAndSegmentRouteWithoutLeadingS
...
+        $responseProphecy = $this->prophesize(ResponseInterface::class);
+        $responseProphecy->getBody()->willReturn($streamProphecy->reveal());
 
-        $containerProphecy = $this->prophesize(ContainerInterface::class);
-        $containerProphecy->has('foo')->willReturn(true);
-        $containerProphecy->get('foo')->willReturn($mock);
+        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
+        $responseFactoryProphecy->createResponse()->willReturn($responseProphecy->reveal());
+
+        $handlerMock = $this->getMockBuilder('MockRouteHandler')
```
bnf: Hm.. I know we had our debate about anonymous classes, but I personally quite dislike `MockBuilder::setMethods` for its different semantics (of course that call to `setMethods` is required here, because we are mocking an non existent class/method, i don't doubt that). Also mocking a non existent class is a bit of a hack (I know it was the same before, when using `stdClass`).

Would you mind to use an anonymous class here?

tests/AppTest.php
```diff
@@ -917,525 +1115,1133 @@ public function testEmptyGroupWithEmptyNestedGroupAndSegmentRouteWithoutLeadingS
...
 
-        $responseFactory = $this->getResponseFactory();
-        $app = new App($responseFactory, $containerProphecy->reveal());
-        $app->get('/foo', 'foo:bar');
+        $handlerMock
+            ->method('foo')
+            ->willReturn($responseProphecy->reveal());
+
+        $containerProphecy = $this->prophesize(ContainerInterface::class);
+        $containerProphecy->has(Argument::exact('handler'))->willReturn(true);
```
bnf: `Argument::exact` is implicit when defining non `\Prophecy\Argument` arguments, so that ` $containerProphecy->has('handler')->willReturn(true);` would be the same, but shorter.
The `phpspec/prophecy` uses the latter in its documentation.
I think it would be good, when we use one of these consistently.
I don't have a strong opinion on this, but I'd go with what is easier to type and what the documentation suggests.

tests/AppTest.php
```diff
@@ -917,525 +1115,1133 @@ public function testEmptyGroupWithEmptyNestedGroupAndSegmentRouteWithoutLeadingS
...
+        $streamBody = '';
+        $streamProphecy = $this->prophesize(StreamInterface::class);
+        $streamProphecy->__toString()->will(function () use (&$streamBody) {
+            return $streamBody;
+        });
+        $streamProphecy->write(Argument::type('string'))->will(function ($args) use (&$streamBody) {
+            $streamBody .= $args[0];
+        });
+        $streamProphecy->read(Argument::exact(11))->will(function () use (&$hasBeenRead, &$streamBody) {
+            $hasBeenRead = true;
```
bnf: As we can mutate state in a promise this should do sth like the following instead of modified a bound boolean:
```php
$this->eof()->willReturn(true);
```

tests/AppTest.php
```diff
@@ -917,525 +1115,1133 @@ public function testEmptyGroupWithEmptyNestedGroupAndSegmentRouteWithoutLeadingS
...
-        $app = new App($responseFactory);
-        $app->get('/', function (ServerRequestInterface $request, ResponseInterface $response) {
+        $hasBeenRead = false;
+        $streamBody = '';
+        $streamProphecy = $this->prophesize(StreamInterface::class);
+        $streamProphecy->__toString()->will(function () use (&$streamBody) {
+            return $streamBody;
+        });
+        $streamProphecy->write(Argument::type('string'))->will(function ($args) use (&$streamBody) {
+            $streamBody .= $args[0];
```
bnf: We should try to mutate state here. Something like:
```php
$body = $this->reveal()->getBody();
$this->getBody()->willReturn($body . $args[0]);
```
Will send a PR to your branch.

bnf (commented):
Nice change, I have some comments, and will send them as pull request to your branch next to the review in here.

tests/AppTest.php
l0gicgate: I have no problems doing that. I've actually come around on the anonymous class issue, it just really triggers my OCD but that's my own problem. I will refactor to use an anonymous class! Thanks

tests/AppTest.php
l0gicgate: Oh okay, gotcha. I'm new to using Prophecy, I will change the usage of `Argument::exact()` to just the string. Let's go with the easier to type version!

tests/AppTest.php
l0gicgate: Haha I didn't think of that! I will change it. That's great!

tests/AppTest.php
l0gicgate: Yeah for sure, that would be much cleaner. I will refactor to do that instead.

l0gicgate:
@bnf please re-review. Just fixed all the requested changes!

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
 
-    public function testGroupSegmentWithSegmentRouteThatDoesNotEndInASlash()
-    {
-        $responseFactory = $this->getResponseFactory();
-        $app = new App($responseFactory);
-        $app->group('/foo', function (App $app) {
-            $app->get('/bar', function (ServerRequestInterface $request, ResponseInterface $response) {
-                // Do something
+        $methods = ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'];
+        foreach ($methods as $method) {
```
bnf: This could (maybe even should) use the `upperCaseRequestMethodsProvider` that you introduced for `testGetPostPutPatchDeleteOptionsMethods`.

Someone once told me that tests shouldn't use loops. It made sense to me, therefore I'd avoid them here as well.

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
-        $responseFactory = $this->getResponseFactory();
-        $app = new App($responseFactory);
-        $app->group('/foo', function (App $app) {
-            $app->group('/baz', function (App $app) {
-                $app->get('', function (ServerRequestInterface $request, ResponseInterface $response) {
-                    // Do something
-                });
-            });
+        $app = new App($responseFactoryProphecy->reveal());
+        $app->map([$method], '/', function ($request, ResponseInterface $response) {
```
bnf: Would be nice to add a type declaration for `$request` as well.

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
+            ['foo'], // Route That Does Not Start With A Slash
+            ['/foo'], // Route That Does Not End In A Slash
+            ['/foo/'], // Route That Ends In A Slash
+        ];
     }
 
-    public function testGroupSingleSlashWithEmptyRoute()
+    /**
+     * @param string $pattern
+     * @dataProvider  routePatternsProvider
```
bnf: double space

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
+    /********************************************************************************
+     * Route Groups
+     *******************************************************************************/
 
-        $router = $app->getRouter();
-        $this->assertAttributeEquals('//baz/bar', 'pattern', $router->lookupRoute('route0'));
+    public function routeGroupsDataProvider()
+    {
+        return [
+            [['', ''], ''], // testEmptyGroupWithEmptyRoute
```
bnf: We could use named data provider entries which allow to use phpunit `--filter` flag in case one needs to debug the tests e.g.:
```
['empty group with segment route that does not end in a slash' => [
    '', '/foo'], '/foo'
],
```
You could automatically convert your current format using some substitutions (vi here):
```
:'<,'>s/\[\(.*\)\], \/\/ test\(.*\)$/'\2' => [^M                \1^M            ],/
:'<,'>s/\([^']\)\([A-Z]\)/\1 \l\2/g
:'<,'>s/'\([A-Z]\)/'\l\1/g
```

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
-                    // Do something
-                });
-            });
-        });
+        $patternProperty = new ReflectionProperty(Route::class, 'pattern');
+        $patternProperty->setAccessible(true);
 
-        $router = $app->getRouter();
-        $this->assertAttributeEquals('//baz', 'pattern', $router->lookupRoute('route0'));
+        $this->assertEquals($pattern, $patternProperty->getValue($route));
```
bnf: `Slim\Routable` has a `getPattern()` method, that means the Reflection is not needed here. :)
This may just be: `$this->assertEquals($pattern, $route->getPattern());`

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
-            $app->get('/', function (ServerRequestInterface $request, ResponseInterface $response) {
-                // Do something
-            });
-        });
+        $patternProperty = new ReflectionProperty(Route::class, 'pattern');
+        $patternProperty->setAccessible(true);
 
-        $router = $app->getRouter();
-        $this->assertAttributeEquals('/', 'pattern', $router->lookupRoute('route0'));
+        $this->assertEquals($expectedPath, $patternProperty->getValue($route));
```
bnf: Same as before, `Slim\Routable` has a `getPattern()` method.
This may just be: `$this->assertEquals($expectedPath, $route->getPattern());`

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
-        $app = new App($responseFactory);
-        $app->group('', function (App $app) {
-            $app->group('/', function (App $app) {
-                $app->get('/bar', function (ServerRequestInterface $request, ResponseInterface $response) {
-                    // Do something
-                });
-            });
+        $middlewareProphecy = $this->prophesize(MiddlewareInterface::class);
+        $middlewareProphecy->process(Argument::cetera())->will(function () use (&$called, $responseProphecy) {
+            $called++;
```
bnf: Would be nice to use the prophecy method `shouldBeCalled` (or even `shouldHaveBeenCalled`) instead.

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
 
-        $router = $app->getRouter();
-        $this->assertAttributeEquals('//bar', 'pattern', $router->lookupRoute('route0'));
-    }
+        $middlewareProphecy2 = $this->prophesize(MiddlewareInterface::class);
+        $middlewareProphecy2->process(
+            Argument::type(ServerRequestInterface::class),
+            Argument::type(RequestHandlerInterface::class)
+        )->will(function ($args) use (&$called, $responseProphecy) {
+            $called++;
```
bnf: Would be nice to use the prophecy method `shouldBeCalled` (or even `shouldHaveBeenCalled`) instead.

tests/AppTest.php
```diff
@@ -100,1342 +113,1498 @@ public function testCreatesRouterWhenNullWithInjectedContainer()
...
-                    $response = $handler->handle($request);
-                    $appendToOutput('Out3');
-                    return $response;
-                });
-            })->add(function ($request, $handler) {
-                $appendToOutput = $request->getAttribute('appendToOutput');
-                $appendToOutput('In2');
-                $response = $handler->handle($request);
-                $appendToOutput('Out2');
-
```
bnf: Missing spaces between `name`, `=>` and `foo`.

bnf, line 1345: Why do you remove the `ServerRequestInterface` type declaration here (also in following test methods)?

bnf, line 1542: Would be nice to use `shouldBeCalled` (or `shouldHaveBeenCalled`) instead.

bnf (changes requested):
Thanks, looks good. I have some minor comments, but generally this is good to go for me.
As already discussed on slack, I like that we actually test public facing code, instead of implementation details now (which will be actually the case once we remove the two remaining property reflections, as per my comments)

tests/AppTest.php
l0gicgate: fixed

tests/AppTest.php
l0gicgate: done. refactored to use `shouldHaveBeenCalled()`

tests/AppTest.php
l0gicgate: fixed

akrabat (approved):

Issue #2583, opened 2019-02-17
Refactor App Router Proxy Methods, RouteGroup, Routable and Route


l0gicgate:
While doing work on #2555 I've noticed that `App` is being passed around in order to provide its proxy methods `get()`, `post()`, `put()`, `patch()`, `delete()`, `options()`, `any()` and `map()`.

As seen here in `App::group()`, `RouteGroup` needs to be invoked so we can pass `App` to it in order for the callable to have access to the Router's proxy method:
```php
    public function group(string $pattern, $callable): RouteGroupInterface
    {
        $router = $this->getRouter();

        /** @var RouteGroup $group */
        $group = $router->pushGroup($pattern, $callable);
        if ($this->callableResolver instanceof CallableResolverInterface) {
            $group->setCallableResolver($this->callableResolver);
        }

        $group($this);
        $router->popGroup();

        return $group;
    }
```

The reference flow of this is terrible, `Router` is in charge of handling groups, `App` shouldn't be. As you see here, `$group($this);` refers to `RouteGroup::__invoke()` which was just a hackish way to pass in `App` to the called function;
```php
    public function __invoke(App $app = null)
    {
        /** @var callable $callable */
        $callable = $this->callable;
        if ($this->callableResolver) {
            $callable = $this->callableResolver->resolve($callable);
        }

        $callable($app);
    }
```

What I propose is that we create a `RouterProxy` that `App` can extend and that `RouteGroup` can extend as well so then we don't have to pass this reference in a hackish way:
```php
abstract class RouterProxy {
    public function get(string $pattern, $callable): RouteInterface
    {
        return $this->map(['GET'], $pattern, $callable);
    }

    public function post(string $pattern, $callable): RouteInterface
    {
        return $this->map(['POST'], $pattern, $callable);
    }

    public function put(string $pattern, $callable): RouteInterface
    {
        return $this->map(['PUT'], $pattern, $callable);
    }

    public function patch(string $pattern, $callable): RouteInterface
    {
        return $this->map(['PATCH'], $pattern, $callable);
    }

    public function delete(string $pattern, $callable): RouteInterface
    {
        return $this->map(['DELETE'], $pattern, $callable);
    }

    public function options(string $pattern, $callable): RouteInterface
    {
        return $this->map(['OPTIONS'], $pattern, $callable);
    }

    public function any(string $pattern, $callable): RouteInterface
    {
[...]

PR #2604, opened 2019-03-04
[DRAFT] 4.x Refactor Routing

l0gicgate:
This is a draft proposal to refactor all routing components as per issue #2583.

**The proposed changes introduce the following interfaces:**
- `RouteCollectorInterface`: Collects all routes and groups, handles cache file
- `RouteCollectorProxyInterface`: This interface is what the `App` components extends and also what is being passed into `RouteGroup`. We will not be exposing all of `App`'s methods when instantiating a `RouteGroup` instead this proxy interface handles communication with the `RouteCollector`
- `RouteInterface` stays the same, minor change is that we have gotten rid of `Routable` all together
- `RouteGroupInterface` this interface now has a `collectRoutes()` method which helps us get rid of the old `RouteGroup::__invoke()`, `Router::pushGroup()` and `Router::popGroup()` methods. `RouteCollectorProxy` has the ability to have a base path which will append it to all the patterns being passed via the proxy methods `get()`, `put()`, `post()`, etc. while collecting the routes from inside the `RouteGroup`'s callable
- `RouteResolverInterface` this component is very simple and its only concerns are to create the `FastRoute` dispatcher, dispatch the `ServerRequestInterface` to it and return the computed `RoutingResults` via the `RouteResolverInterface::computeRoutingResults()` method. This interface also has the responsibility to resolve the appropriate route communicating with `RouteCollector` to achieve that. `RoutingMiddleware` uses this resolver to perform the routing.
<br/>

**Deprecated Components/Methods**
- `Router` and `RouterInterface` are now gone since they've been effectively separated
- `Routable` does not exist anymore as it has been merged into `Route` since it is the only component that was using those methods
- `RouteGroup` does not hold a `$pattern` variable anymore as it is unnecessary due to the state of the pattern residing inside `RouteCollectorProxy` when collecting the inner routes of a group
- `Route::finalize()` does not exist anymore, it has effectively been merged with `Route::prepare()`
- `RouteDispatcher` has been renamed to `RouteRunner` since that's all it does, it calls `Route::run()`
<br/>

 **New Namespace Folder**
New namespace folder called `Routing` which has all of the routing components. 

<br/>

**This PR will be separated into smaller ones**
- [x] Decouple `RouteGroup` & `Route` from `Routable` - #2612 Merged
- [x] Decouple `Router` into `RouteCollector` and `RouteResolver` - #2622 Merged 
- [x] Add `Route::setInvocationStrategy()` to `RouteInterface` - #2634 Merged
- [x] Add `RouteCollector::fullUrlFor()` as per #2493 - #2638 Merged
- [x] Introduce `DispatcherInterface` to decouple from FastRoute Dispatcher - #2639 Merged
- [x] Introduce `RouteParserInterface` to decouple from FastRoute RouteParser - #2640 Merged
- [x] Introduce `RouteCollectorProxy` - #2641 & #2642 Merged


l0gicgate:
@bnf I would like your input on this when you get a minute. Thank you!

akrabat:
This is in danger of being too big. Can we pull out some bits into their own PRs?

PR #2610, opened 2019-03-16
Remove unused 'keys' method from `Collection`

matijabelec:
As far as I can see `CollectionInterface` does not defines `keys` method, therefore, concrete implementation `Collection` does not need `keys` method either.

### DIFF
--- a/Slim/Collection.php
+++ b/Slim/Collection.php
@@ -88,16 +88,6 @@ class Collection implements CollectionInterface
     }
 
     /**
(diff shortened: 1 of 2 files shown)


l0gicgate:
I think that retrieving the keys from a Collection is functionality that is relevant/usable even though it is unused in the project. It’s a no from me, what are your thoughts @akrabat?

akrabat:
We can't remove this method as we'll break BC.

Issue #2600, opened 2019-03-02
Remove MiddlewareDispatcher::getMiddleware()?


akrabat:
In Route::finalize(), rather than getting the Group's middleware and adding it to the MiddlwareDispatcher, maybe we should call $group->addToMiddlewareDipatcher($this->middlewareDispatcher) which does the foreach over the middleware. This removes RouteGroup::getMiddleware()  which feels like an encapsulation leakage.

Issue #2599, opened 2019-03-02
Can we fold Route::finalize() into Route::run()?


akrabat:
I'm not sure what the benefit of `finalize()` is, so maybe we should look at this?

l0gicgate:
This change is addressed in #2604. `finalize()` is being folded into `prepare()`. Please follow discussion there.

PR #2612, opened 2019-03-19
4.x Refactor RouteGroup & Route

### LINKED ISSUE #2598: Why is RouteGroup a Routable when it can't be routed to?
This probably needs tidying up.

l0gicgate:
This is pull 1 out of 7 to complete the goals set in #2604 

- Refactor `RouteGroup` so it doesn't extend `Routable`
- Implement `appendMiddlewareToDispatcher()` method to avoid encapsulation leakage
- Refactor `Route::finalize()` and fold it into `Route::run()`
- Remove `Routable` as it is now obsolete.

This PR effectively closes #2598, #2599 and #2600.

- [x] Discussion
- [x] Write Tests


Slim/Route.php
```diff
@@ -312,61 +313,29 @@ public function prepare(ServerRequestInterface $request, array $arguments): Rout
...
-     */
-    public function finalize(): void
-    {
-        if ($this->finalized) {
-            return;
-        }
-
-        // Chain a new (outer) middleware dispatcher for route-group-middleware
-        // to the existing route-middleware dispatcher
+        // Add Middleware From Groups
```
akrabat: Lower case `f` and `g`

akrabat:
Looks reasonable to me.

Slim/Route.php
l0gicgate: Fixed

akrabat (approved):

akrabat:
Do we need a CHANGELOG entry?

l0gicgate:
@akrabat yeah I think so. The public interface has changed for both `Route` and `RouteGroup` so I'll add an entry.

PR #2616, opened 2019-03-30
Update PhpFunctionOverrides.php

wangbaojin:
modify the namespace


l0gicgate:
That namespace is `Slim` for a purpose. As you can see the build fails, this is used for global function overrides so we can stub them and test them when they're being used within the `Slim` namespace.

PR #2622, opened 2019-04-02
4.x - Split RouterInterface into RouteCollectorInterface and RouteResolverInterface

l0gicgate:
This is pull 2 out of 7 to complete the goals set in #2604 

- Split `RouterInterface` into `RouteCollectorInterface` and `RouteResolverInterface`
- Rename `RouteDispatcher` to `RouteRunner`
- Move all routing components into new `Slim\Routing` aggregate folder

- [x] Discussion
- [x] Write Tests


Slim/Interfaces/RouteCollectorInterface.php
```diff
@@ -10,56 +10,65 @@
...
+     */
+    public function setDefaultInvocationStrategy(InvocationStrategyInterface $strategy): RouteCollectorInterface;
+
+    /**
+     * @return null|string
+     */
+    public function getCacheFile(): ?string;
+
+    /**
+     * Set path to fast route cache file. If this is false then route caching is disabled.
```
akrabat: If what is `false` ?

Slim/Interfaces/RouteCollectorInterface.php
```diff
@@ -10,56 +10,65 @@
...
-     * @return RoutingResults
+     * @param string|null $cacheFile
+     * @return RouteCollectorInterface
      *
-     * @link   https://github.com/nikic/FastRoute/blob/master/src/Dispatcher.php
+     * @throws InvalidArgumentException
+     * @throws RuntimeException
      */
-    public function dispatch(ServerRequestInterface $request): RoutingResults;
+    public function setCacheFile(?string $cacheFile): RouteCollectorInterface;
```
akrabat: Should we have `setCacheFile(string $cacheFile): RouteCollectorInterface` and `clearCacheFile(): void` ?

i.e. I'm not totally sure about the nullable `?string`.

(Just a question. I have no firm opinion, yet.)

Slim/Interfaces/RouteCollectorInterface.php
```diff
@@ -72,13 +81,53 @@ public function popGroup();
...
+    public function removeNamedRoute(string $name): RouteCollectorInterface;
+
     /**
      * @param string $identifier
      *
      * @return RouteInterface
+     *
+     * @throws RuntimeException
      */
     public function lookupRoute(string $identifier): RouteInterface;
```
akrabat, line 100: Is the route's `identifier` the same as its 'name' as used in `removeNamedRoute()`?

Slim/Interfaces/RouteCollectorInterface.php
```diff
@@ -72,13 +81,53 @@ public function popGroup();
...
 
+    /**
+     * Add a route group to the array
+     *
+     * @param string   $pattern The group pattern
+     * @param callable $callable A group callable
+     *
+     * @return RouteGroupInterface
+     */
+    public function pushGroup(string $pattern, $callable): RouteGroupInterface;
```
akrabat, line 110: Should we type hint `$callable`?

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: This is a typo, it should be a string or null. False was for Slim 3 because we didn't have type hinting. Will fix.

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: When would you clear the cache file though? I feel like it's either you set the cache file or you don't.

Slim/Routing/RouteCollector.php
```diff
@@ -7,32 +7,22 @@
...
-/**
- * Router
- *
- * This class organizes Slim application route objects. It is responsible
- * for registering route objects, assigning names to route objects,
- * finding routes that match the current HTTP request, and creating
- * URLs for a named route.
- */
-class Router implements RouterInterface
+class RouteCollector implements RouteCollectorInterface
```
akrabat, line 29: I'd still like a comment explaining what the class does.

akrabat:
Should the namespace be `Slim\Router` ?

Slim/Interfaces/RouteCollectorInterface.php
akrabat: > When would you clear the cache file though? I feel like it's either you set the cache file or you don't.

Why have `$cacheFile` as nullable then?

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: The route identifier is different than the name.

You can set a route's name via `$route->setName()`:
https://github.com/slimphp/Slim/blob/fb928c78eae3879dbc0e60abbdebfe94138fa19d/Slim/Routing/Route.php#L212

And the identifier refers to `RouteCollector`'s `routeCounter` incremental property with the prefix `route`:
https://github.com/slimphp/Slim/blob/fb928c78eae3879dbc0e60abbdebfe94138fa19d/Slim/Routing/Route.php#L148

https://github.com/slimphp/Slim/blob/fb928c78eae3879dbc0e60abbdebfe94138fa19d/Slim/Routing/RouteCollector.php#L296-L305

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: This method will be removed entirely in the next PR along with `popGroup()`. But no, we can't type hint it unless we drop support for Slim 2 callables (which I've been meaning to ask if you'd want to)

Slim/Routing/RouteCollector.php
l0gicgate: Will fix.

Slim/Interfaces/RouteCollectorInterface.php
akrabat: Let's deal with that separately.

l0gicgate:
> Should the namespace be Slim\Router ?

I think `Slim\Routing` is better imo. A router would be a single component, routing is synonymous of all the components being used.

akrabat:
> I think `Slim\Routing` is better imo. A router would be a single component, routing is synonymous of all the components being used.

The combination of the classes is the router. I don't care enough though, so `Routing` it is unless someone else has a compelling opinion.

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: Fixed by removing nullable parameter

Slim/App.php
```diff
@@ -94,12 +94,17 @@ public function __construct(
         RouteCollectorInterface $routeCollector = null,
         RouteResolverInterface $routeResolver = null
     ) {
+        $this->container = $container;
```
akrabat: It's neater if the assignments are in same order as the parameters in the constructor.

l0gicgate: Agreed, fixed!

Slim/Routing/RouteCollector.php
```diff
@@ -314,46 +317,54 @@ public function relativePathFor(string $name, array $data = [], array $queryPara
...
-
         $segments = [];
         $segmentName = '';
-        foreach ($routeDatas as $routeData) {
-            foreach ($routeData as $item) {
-                if (is_string($item)) {
-                    // this segment is a static string
-                    $segments[] = $item;
+
+        /**
```
akrabat: Not a docblock - just a comment, so just one `*` needed.

Slim/Interfaces/RouteCollectorInterface.php
```diff
@@ -10,56 +10,62 @@
...
-     * Dispatch router for HTTP request
+     * Set default route invocation strategy
      *
-     * @param  ServerRequestInterface $request The current HTTP request object
+     * @param InvocationStrategyInterface $strategy
+     * @return RouteCollectorInterface
+     */
+    public function setDefaultInvocationStrategy(InvocationStrategyInterface $strategy): RouteCollectorInterface;
+
+    /**
```
akrabat, line 32: Needs a description

Slim/Routing/RouteCollector.php
l0gicgate: Fixed

Slim/Interfaces/RouteCollectorInterface.php
l0gicgate: Added description

akrabat (approved):

Issue #2548, opened 2018-11-28
`EnvironmentInterface` seems to be a useless type to expect / depend on


Rarst:
`Slim\Container` declares default `environment` service to be an implementation of `EnvironmentInterface`.

```
Slim\App expects a container that implements Psr\Container\ContainerInterface
with these service keys configured and ready for use:

 - environment: an instance of \Slim\Interfaces\Http\EnvironmentInterface

@property-read \Slim\Interfaces\Http\EnvironmentInterface environment
```

Thing is `EnvironmentInterface` doesn’t contain anything actually needed for it to work, only test–related `mock()`.

This means that this ` $app->getContainer()->environment->get('foo');` won’t be understood by any tools, because the interface doesn’t _have_ `get()`.

Also the classes that depend on environment like `Slim\Http\Request::createFromEnvironment(Environment $environment)` all require specific `Slim\Environment` _class_, not the interface. So trying to replace `environment` service with something that implements `EnvironmentInterface` would break the app.

It feels like either:

1. `EnvironmentInterface` should be expanded to comprehensively describe a generic replaceable service, in analogue to `ContainerInterface`. Other components of the app should be changed to depend on the interface, not the implementation.
2. Inline documentation should reflect the de–facto dependency on concrete `Environment` implementation, not the interface.

Rarst:
Hey, I noticed this was tagged as a workable issue... But I could use some feedback on _which_ approach is desirable here?

jdrieghe:
I made a small PR that implements one of your proposals: https://github.com/slimphp/Slim/pull/2625.
If it's decided to go with this, I'm more than willing to do the same for the documentation.

PR #2625, opened 2019-04-05
Remove references to unused EnvironmentInterface

jdrieghe:
One of the proposed solutions in this issue (https://github.com/slimphp/Slim/issues/2548) is to no longer reference the `EnvironmentInterface` in the docs, which I believe extends to type hints in the code.

I don't think removing `EnvironmentInterface` completely is possible without shipping a breaking change as people might have extended the `Environment` specifically implemented the `EnvironmentInterface` again in their extension.

Unless someone is willing to rework the `EnvironmentInterface` and the rest of the framework to accept any implementation of this interface, I think this fix might be serviceable to manage developer expectations.

If this solution is accepted I will create a similar pull request to the documentation to replace these references: https://github.com/slimphp/Slim-Documentation/search?q=EnvironmentInterface&unscoped_q=EnvironmentInterface

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -21,7 +21,7 @@ use Slim\Exception\ContainerException as SlimContainerException;
  * with these service keys configured and ready for use:
  *
  *  - settings: an array or instance of \ArrayAccess
- *  - environment: an instance of \Slim\Interfaces\Http\EnvironmentInterface
+ *  - environment: an instance of \Slim\Http\Environment
  *  - request: an instance of \Psr\Http\Message\ServerRequestInterface
  *  - response: an instance of \Psr\Http\Message\ResponseInterface
  *  - router: an instance of \Slim\Interfaces\RouterInterface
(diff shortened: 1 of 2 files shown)


Slim/Http/Environment.php
```diff
@@ -18,7 +17,7 @@
  * This is particularly useful for unit testing, but it also lets us create
  * custom sub-requests.
  */
-class Environment extends Collection implements EnvironmentInterface
```
Rarst: This bit feels like worth keeping to me, the class does implement the interface and who knows how it's used/hinted downstream.

jdrieghe: Good point. Will put it back later tonight.

akrabat (approved):

akrabat:
Thanks @jdrieghe

jdrieghe:
@akrabat should we change the reference on this page as well? https://github.com/slimphp/Slim-Website/blob/328758f4ecd0400ec14efc280e8d3b9051f39bed/docs/v3/concepts/di.md

akrabat:
> @akrabat should we change the reference on this page as well? https://github.com/slimphp/Slim-Website/blob/328758f4ecd0400ec14efc280e8d3b9051f39bed/docs/v3/concepts/di.md

Yes.

Issue #2627, opened 2019-04-11
Netbeans debug only provides path via a parameter


keithy:
Netbeans insists on using index.php as an access point, and providing the path via the parameter. Slim 3 is not using that parameter, but rather the uri->getPath()

l0gicgate:
This is not a relevant issue to the Slim codebase. This is behavior specific to use of Netbeans.

keithy:
As indicated, I submitted this to document the issue.

It has nothing to do with Netbeans itself. The difficulty with netbeans is that it offers no alternative means to invoke. I placed Netbeans in the issue title to help others who may have had the same problem.

The issue being that when the framework is invoked like so: index.php?path=/some/path.php slim does not pick up the path from the parameter. Whether it should or not is another question. I think it should, you may disagree. If you do, then this would be a suitable place to document it and explain why, because I am not understanding.

PR #2628, opened 2019-04-11
Router - if path is missing get path from parameter

keithy:
https://github.com/slimphp/Slim/issues/2627

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -131,22 +131,16 @@ class Router implements RouterInterface
     public function setCacheFile($cacheFile)
     {
         if (!is_string($cacheFile) && $cacheFile !== false) {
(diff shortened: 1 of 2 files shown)


keithy:
Not expecting this to be merged, just putting it out there.

l0gicgate:
A few comments:

1. I can't understand the actual issue you attempted to describe in #2627 
2. You're not working off of the lastest master branch of `3.x`
3. We're not going to modify the codebase to cater to an editor in particular, in this case Netbeans

Thanks

keithy:
As indicated, I submitted this to document the issue. The issue being that when the framework is invoked like so: index.php?path=/some/path.php slim does not pick up the path from the parameter.

Issue #2632, opened 2019-04-17
getallheaders() might return false


adriansuter:
Hi there

According to the php documentation, the function getallheaders() might return `false`. In fact, on my hosting, the function even returns `null`.

Probably you should check, if the variable `$headers` is an array and skip if not.

https://github.com/slimphp/Slim/blob/eaee12ef8d0750db62b8c548016d82fb33addb6b/Slim/Http/Headers.php#L82

On my hosting I get the following warning:

> array_change_key_case() expects parameter 1 to be array, null given

Thanks
Adrian

PR #2633, opened 2019-04-17
Add type check for $headers

adriansuter:
As $headers may not be an array, we added an if statement to avoid any warnings. 
See #2632

### DIFF
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -76,12 +76,22 @@ class Headers extends Collection implements HeadersInterface
     public static function determineAuthorization(Environment $environment)
     {
         $authorization = $environment->get('HTTP_AUTHORIZATION');
+        if (!empty($authorization)) {
+            return $environment;
+        }
+        
+        if (!is_callable('getallheaders')) {
+            return $environment;
+        }
+        $headers = getallheaders();
+        
+        if (!is_array($headers)) {
+            return $environment;
+        }
 
-        if (empty($authorization) && is_callable('getallheaders') && $headers = getallheaders()) {
-            $headers = array_change_key_case($headers, CASE_LOWER);
-            if (isset($headers['authorization'])) {
-                $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
-            }
+        $headers = array_change_key_case($headers, CASE_LOWER);
+        if (isset($headers['authorization'])) {
+            $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
         }
 
         return $environment;
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -79,9 +79,11 @@ class Headers extends Collection implements HeadersInterface
 
         if (empty($authorization) && is_callable('getallheaders')) {
             $headers = getallheaders();
-            $headers = array_change_key_case($headers, CASE_LOWER);
-            if (isset($headers['authorization'])) {
-                $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
+            if (is_array($headers)) {
+                $headers = array_change_key_case($headers, CASE_LOWER);
+                if (isset($headers['authorization'])) {
+                    $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
+                }
             }
         }
 
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -77,11 +77,7 @@ class Headers extends Collection implements HeadersInterface
     {
         $authorization = $environment->get('HTTP_AUTHORIZATION');
 
-        if (
-            empty($authorization) 
-            && is_callable('getallheaders') 
-            && $headers = getallheaders()
-        ) {
+        if (empty($authorization) && is_callable('getallheaders') && $headers = getallheaders()) {
             $headers = array_change_key_case($headers, CASE_LOWER);
             if (isset($headers['authorization'])) {
                 $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
--- a/Slim/Http/Headers.php
+++ b/Slim/Http/Headers.php
@@ -77,13 +77,14 @@ class Headers extends Collection implements HeadersInterface
     {
         $authorization = $environment->get('HTTP_AUTHORIZATION');
 
-        if (empty($authorization) && is_callable('getallheaders')) {
-            $headers = getallheaders();
-            if (is_array($headers)) {
-                $headers = array_change_key_case($headers, CASE_LOWER);
-                if (isset($headers['authorization'])) {
-                    $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
-                }
+        if (
+            empty($authorization) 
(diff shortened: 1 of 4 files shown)


Slim/Http/Headers.php
```diff
@@ -79,9 +79,11 @@ public static function determineAuthorization(Environment $environment)
 
         if (empty($authorization) && is_callable('getallheaders')) {
```
l0gicgate: It feels like a lot of nested statements. Would this be cleaner?
```php
if (
    empty($authorization) 
    && is_callable('getallheaders') 
    && $headers = getallheaders()
) {
    $headers = array_change_key_case($headers, CASE_LOWER);
    if (isset($headers['authorization'])) {
        $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
    }
}
```

l0gicgate (changes requested):

Slim/Http/Headers.php
adriansuter: Sure.

Slim/Http/Headers.php
```diff
@@ -77,8 +77,7 @@ public static function determineAuthorization(Environment $environment)
     {
         $authorization = $environment->get('HTTP_AUTHORIZATION');
 
-        if (empty($authorization) && is_callable('getallheaders')) {
-            $headers = getallheaders();
+        if (empty($authorization) && is_callable('getallheaders') && $headers = getallheaders()) {
```
akrabat: I know you've just changed this, but it was better before.

Assigning to a variable inside of an `if` statement is difficult for long term maintenance as it looks like it should be an `==` test.

akrabat (changes requested):

Slim/Http/Headers.php
adriansuter: True. How about early returns? I mean
```php
        if (!empty($authorization)) {
                return $environment;
        }
        if (!is_callable('getallheaders')) {
                return $environment;
        }
        $headers = getallheaders();
        if (!is_array($headers)) {
            return $environment;
        }

        $headers = array_change_key_case($headers, CASE_LOWER);
        if (isset($headers['authorization'])) {
            $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
        }

        return $environment;
```

akrabat: Works for me. Personally, I'd combine the first two `if` statements:

```php
if (!empty($authorization) || !is_callable('getallheaders')) {
```

Also, add a blank line between after each `}`

l0gicgate: Sorry for suggestion a variable declaration in a condition. Here's my return early version:
```php
public static function determineAuthorization(Environment $environment)
{
    $authorization = $environment->get('HTTP_AUTHORIZATION');

    if (!empty($authorization) || !is_callable('getallheaders')) {
        return $environment;
    }

    $headers = getallheaders();
    if (is_array($headers)) {
        $headers = array_change_key_case($headers, CASE_LOWER);
        if (isset($headers['authorization'])) {
            $environment->set('HTTP_AUTHORIZATION', $headers['authorization']);
        }
    }

    return $environment;
}
```

l0gicgate (approved):

Slim/Http/Headers.php
```diff
@@ -76,13 +76,18 @@ public static function createFromEnvironment(Environment $environment)
     public static function determineAuthorization(Environment $environment)
     {
         $authorization = $environment->get('HTTP_AUTHORIZATION');
+        if (!empty($authorization) || !is_callable('getallheaders')) {
+            return $environment;
+        }
+        
+        $headers = getallheaders();
+        if (!is_array($headers)) {
```
l0gicgate, line 84: Can I get you to write a test for this if statement please?

adriansuter: Of course, I will give it a try.

So I need to override the built-in function `getallheaders()`. In one test the function would return false, in another test, the function would return an array. So far so good. Do you use any library to mock/override the built-in functions? I haven't found one in your source code.

All I could see is the php-file `PhpFunctionOverrides.php` in `tests/Assets`. So I suppose you use the namespace method to override the built-in php functions. That is fine. Well, the namespace of course for the class is `Slim\Http`. So in order to override the `getallheaders()` method, I need to create a new file for this namespace, needn't I?

Do you know for example [the method to be able to control the behaviour of the overriden function depending on the calling method](http://www.manuel-strehl.de/dev/overwrite_PHP_built-in_functions.en.html)? Would it make sense to implement such a helper `_mock()` function for unit testing? It has a GPL and a MIT license.

l0gicgate: So this is a bit complex I suppose. You do need to override `getallheaders()` but since it is used in the namespace `Slim\Http` you'll need to create a new function override file in `tests\Assets` called `PhpHttpFunctionOverrides.php` and put the contents:

```php
<?php
/**
 * Slim Framework (https://slimframework.com)
 *
 * @link      https://github.com/slimphp/Slim
 * @copyright Copyright (c) 2011-2017 Josh Lockhart
 * @license   https://github.com/slimphp/Slim/blob/3.x/LICENSE.md (MIT License)
 */

namespace Slim\Http;

function getallheaders()
{
    return GETALLHEADERS_RETURN;
}

```

Then in `composer.json` you'll need to add `PhpHttpFunctionOverrides.php` to the `files` property:
```json
"autoload-dev": {
    "files": [
        "tests/Assets/PhpFunctionOverrides.php",
        "tests/Assets/PhpHttpFunctionOverrides.php"
    ]
},
```

Make sure that you do `composer dumpautoload` after you make that change so the override is loaded on your end. Once that's done you can write your tests. In the tests you'll have to define the constant like so:

```php
public function testDetermineAuthorizationReturnsEarlyIfHeadersIsNotArray()
{
    // Define it as false, or ['existing' => 'header']
    define('GETALLHEADERS_RETURN', false);

    ...your assertions...
}
```

Hope this helps!

adriansuter: The test is implemented. I used $GLOBALS in order to be able to set the return value to `false` only for this test. Otherwise it may affect other tests executed afterwards. Is that okay?

adriansuter: To avoid `$GLOBALS` we could also override the function like this
```php
// tests/Assets/PhpHttpFunctionOverrides.php
function getallheaders()
{
    $backtrace = debug_backtrace();
    foreach ($backtrace as $step) {
        if ($step['function'] === 'testDetermineAuthorizationReturnsEarlyIfHeadersIsNotArray') {
            return false;
        }
    }

    return \getallheaders();
}
```
So the function would check the backtrace, whether a caller has the name of the test.

**Problem: When the name of the method gets changed, the test would not work anymore.**

What do you think?

l0gicgate: This is why I'd rather define a constant which explicitly shows in the test what the return value will be. Also `getallheaders()` isn't being used anywhere else in the codebase.

**The override could be:**
```php
function getallheaders()
{
    return defined('GETALLHEADERS_RETURN') ? GETALLHEADERS_RETURN : \getallheaders();
}
```

PR #2634, opened 2019-04-17
4.x - Route InvocationStrategy Setter

### LINKED ISSUE #2597: How do you change the invocation strategy for a specific route?
Should it be possible to change the invocation strategy for a specific route? If so, how would we do so?

l0gicgate:
This is pull 3 out of 7 to complete the goals set in #2604 

This PR adds the capability to set a `Route`'s invocation strategy per route.

Closes #2597

**Example usage:**
```php
use Slim\App;
use Slim\Handlers\Strategies\RequestResponse;

$app = new App(...);

$invocationStrategy = new RequestResponse();
$app
    ->get('/', function () {...})
    ->setInvocationStrategy(...);
```

### DIFF
--- a/tests/Routing/RouteTest.php
+++ b/tests/Routing/RouteTest.php
@@ -134,7 +134,7 @@ class RouteTest extends TestCase
         $route = new Route(['GET'], '/', $callable, $responseFactory, $callableResolver);
         $route->setInvocationStrategy($strategy);
 
-        $this->assertEquals($strategy, $route->getInvocationStrategy());
+        $this->assertSame($strategy, $route->getInvocationStrategy());
     }
 
     public function testGetGroups()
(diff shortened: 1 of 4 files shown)


tests/Routing/RouteTest.php
```diff
@@ -121,6 +121,22 @@ public function testGetInvocationStrategy()
...
+        $callable = function (ServerRequestInterface $request, ResponseInterface $response, $args) {
+            return $response;
+        };
+
+        $responseFactory = $this->getResponseFactory();
+        $callableResolver = new CallableResolver();
+        $strategy = new RequestResponse();
+
+        $route = new Route(['GET'], '/', $callable, $responseFactory, $callableResolver);
+        $route->setInvocationStrategy($strategy);
```
akrabat, line 135: If I comment out this line (`$route->setInvocationStrategy($strategy);`), the test still passes…

l0gicgate: Just fixed the assertion to use `assertSame()` instead I messed up 😂

akrabat (approved):

PR #2635, opened 2019-04-17
Quality improvements

akrabat:
Some general quality improvements as reported by PhpStorm and PHP Inspections (EA Extended)

### DIFF
--- a/Slim/Interfaces/RouteInterface.php
+++ b/Slim/Interfaces/RouteInterface.php
@@ -90,7 +90,7 @@ interface RouteInterface
      *
      * @return string|null
      */
-    public function getArgument(string $name, $default = null): ?string;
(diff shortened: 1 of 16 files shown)


Slim/Interfaces/RouteInterface.php
```diff
@@ -90,7 +90,7 @@ public function getIdentifier(): string;
      *
      * @return string|null
      */
-    public function getArgument(string $name, $default = null);
+    public function getArgument(string $name, $default = null): ?string;
```
l0gicgate: we should type `$default` as well:

```php
public function getArgument(string $name, string $default = null): ?string;
```

akrabat: Done.

l0gicgate (approved):

PR #2637, opened 2019-04-20
3.x - Response function isBadRequest()

KarimChukfeh:
Adds a helper method `$response->isBadRequest()` using `StatusCode::HTTP_BAD_REQUEST`


l0gicgate:
I'm not opposed to this @akrabat. We do have all the other helper methods on the `Response` object and this is just one more. I want your thoughts on this before merging though.

akrabat:
What is the use-case where this method would be needed?

KarimChukfeh:
I'm testing my responses using PHPUnit and [slim-rest-helpers](https://github.com/there4/slim-test-helpers). An assertion look like this `$this->assertTrue($this->client->response->isOk());` 
but there's no helper method for 400 Bad Request. So the 400 assertion has to look like this `$this->assertEquals(400, $this->client->response->getStatusCode());`

Issue #2656, opened 2019-04-24
Docblock param type is string only - array should be added


adriansuter:
The docblock for the following method defines the type of the second param `$value` as `string`.

https://github.com/slimphp/Slim/blob/203dd4e768cb0c1cf634e01e2521a95699017c8d/Slim/Http/Headers.php#L114-L123

But in
https://github.com/slimphp/Slim/blob/203dd4e768cb0c1cf634e01e2521a95699017c8d/Slim/Http/Headers.php#L182
the function `array_merge()` would return an `array`.

So the docblock should allow `array` as well.

PR #2657, opened 2019-04-24
Update docblock

adriansuter:
Add array as type to the second param of set().


adriansuter:
Pull request for issue #2656

Issue #2658, opened 2019-04-24
getBasePath() is not part of PSR-7


adriansuter:
The property `uri` is of type `\Psr\Http\Message\UriInterface`. In the following line the method `getBasePath()` is called. This method is not part of the PSR-7 standard and therefore is not defined in `\Psr\Http\Message\UriInterface`.

https://github.com/slimphp/Slim/blob/203dd4e768cb0c1cf634e01e2521a95699017c8d/Slim/Http/Request.php#L500

However the method is defined in `\Slim\Http\Uri`, of which type the property usually is.

Possible solutions:
- Check that `$this->uri` is of that type.
- Check that `$this->uri` has the method `getBasePath()`.
- The base path has to be injected into the Request from somewhere else.

l0gicgate:
Good catch. Will you create a PR for this? If not, I can.

PR #2659, opened 2019-04-24
Check that the property uri is a Uri instance

adriansuter:


adriansuter:
Pull request for issue #2658.

I decided to check if the property is of type `\Slim\Http\Uri`.

l0gicgate:
Can you write a test case for it please?

Slim/Http/Request.php
```diff
@@ -497,7 +497,11 @@ public function getRequestTarget()
             return '/';
         }
 
-        $basePath = $this->uri->getBasePath();
+        if ($this->uri instanceof Uri) {
+            $basePath = $this->uri->getBasePath();
+        } else {
```
l0gicgate, line 502: Can you add a test case for this else case please?

l0gicgate:
You don't need to create an entire stub to mock that URI. Instead just prophesize the interface:
```php
public function testGetRequestTargetIfUriPsr7Only() {
    $basePath = '/base/path';

    $uriProphecy = $this->prophesize(UriInterface::class);
    $uriProphecy
        ->getBasePath()
        ->willReturn($basePath)
        ->shouldBeCalledOnce();
    
    $request = $this->requestFactory();
    $prop = new ReflectionProperty($request, 'uri');
    $prop->setAccessible(true);
    $prop->setValue($request, $uriProphecy->reveal());
    
    $this->assertEquals($basePath, $request->getRequestTarget());
}
```

Note: In the example above I'm mocking `UriInterface::class` in this test case it should be `Uri::class` since we want to detect that we are being passed a Slim `Uri` object.

adriansuter:
Ah, I didn't know that. But wait, shouldn't we make sure, that the method `getBasePath()` was never called if the uri is only an instance of a class implementing RouterInterface which has no such method.

l0gicgate:
> the uri is only an instance of a class implementing RouterInterface which has no such method

@adriansuter you mean a class implementing `Uri`. You can mock that instead of `UriInterface`. Still the same synthax.

```php
$uriProphecy = $this->prophesize(Uri::class);
```

adriansuter:
No, I mean a class implementing the `UriInterface` directly. That is actually the case for which I have added the `else` statement in `getRequestTarget`.

So in my opinion the test should be, whether the method works in case the uri is **not** of type (or any extending type) of `Uri`. Because that is the case when we can't call `getBasePath()`. Remember that this method is not part of PSR7.

The other case, when the uri **is** a `Uri` (or any extending type) works, as all other tests assure that already. Actually they use `Uri` all the time, as the request factory creates an instance of type `Uri`.

l0gicgate:
**Here's what the two tests should be:**
```php
public function testGetRequestTargetWithSlimPsr7Uri() {
    $basePath = '/base/path';

    $uriProphecy = $this->prophesize(Uri::class);
    $uriProphecy
        ->getBasePath()
        ->willReturn($basePath)
        ->shouldBeCalledOnce();
    
    $request = $this->requestFactory();
    $prop = new ReflectionProperty($request, 'uri');
    $prop->setAccessible(true);
    $prop->setValue($request, $uriProphecy->reveal());
    
    $this->assertEquals($basePath . '/', $request->getRequestTarget());
}
```
```php
public function testGetRequestTargetWithNonSlimPsr7Uri() {
    // We still pass in a UriInterface, which isn't an instance of Slim URI
    $uriProphecy = $this->prophesize(UriInterface::class);
    $uriProphecy
        ->getBasePath()
        ->shouldBeNotBeCalled();
    
    $request = $this->requestFactory();
    $prop = new ReflectionProperty($request, 'uri');
    $prop->setAccessible(true);
    $prop->setValue($request, $uriProphecy->reveal());
    
    $this->assertEquals('/', $request->getRequestTarget());
}
```

Please get rid of that Uri stub.

Thanks

adriansuter:
Thanks.

Okay, you wanted both tests, I see. Of course I will remove the "stub".

(Tomorrow, it's one o'clock in the morning)

adriansuter:
In the second test case `testGetRequestTargetWithNonSlimPsr7Uri()` I could not add a `MethodProphecy` as PHPUnit would immediately return an error because the method `getBasePath()` is not defined in `UriInterface`.

l0gicgate (approved):

Issue #2655, opened 2019-04-24
Risky testMoveToStream()


adriansuter:
PHPUnit marks the following test as risky (no asserts). Could that be removed?

https://github.com/slimphp/Slim/blob/203dd4e768cb0c1cf634e01e2521a95699017c8d/tests/Http/UploadedFilesTest.php#L217

l0gicgate:
Instead of removing it, we should implement it.

adriansuter:
This would be tricky.

Currently the test case uses the stream `php://temp`. But the php documentation states in example 5 of https://www.php.net/manual/en/wrappers.php.php

> **php://memory and php://temp are not reusable**

So we can't use this stream to check if `\Slim\Http\UploadedFile::moveTo()` worked as expected, can we?

A possible solution would be to write a custom stream wrapper or to use the stream `php://output`. So for example:
```php
    public function testMoveToStream()
    {
        $uploadedFile = $this->generateNewTmpFile();
        $contents = file_get_contents($uploadedFile->file);

        ob_start();
        $uploadedFile->moveTo('php://output');
        $movedFileContents = ob_get_clean();

        $this->assertEquals($contents, $movedFileContents);
        $this->assertFileNotExists($uploadedFile->file);
    }
```
Is it okay to use output buffering in a PHPUnit test case?

l0gicgate:
It's totally fine to use output buffering in a test case. That test looks good to me, not sure why it was never implemented in the first place.

adriansuter:
Alright, I will create a pull request.

PR #2661, opened 2019-04-25
Test case testMoveToStream implemented

adriansuter:
We used the stream `php://output` instead of `php://temp` and output buffering.


adriansuter:
Pull request for issue #2655

l0gicgate (approved):

PR #2668, opened 2019-04-26
Add missing @throws to docblocks

### LINKED ISSUE #2666: Slim\App docblocks have missing throws annotations
The class `\Slim\App` (3.x) has many methods that may throw exceptions, but their docblocks do not contain corresponding `throws` annotations.

Should we add those annotations or would that break installations?

RyanNerd:
Related: @return type in NotFound.php had wrong types in docblock (fixed)

Closes #2666

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -112,7 +112,10 @@ class App
      *
      * @param  string $method
      * @param  array $args
+     *
      * @return mixed
+     * @throws \BadMethodCallException when method is not valid
+     * @throws \Interop\Container\Exception\ContainerException
      */
     public function __call($method, $args)
     {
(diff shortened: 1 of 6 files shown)


l0gicgate (changes requested):
Can we import the exceptions at the top instead of using the `\Exception` annotation. I'd like to keep the doc blocs as cohesive as possible. Right now some have them and some don't. Also, put a line return after last `@param` and `@return`

```php
use Exception;

/**
 * @param ...
 *
 * @return ...
 *
 * @throws Exception
 */
```

l0gicgate (approved):

l0gicgate (approved):

PR #2670, opened 2019-04-27
Format Doc Blocks

### LINKED ISSUE #2666: Slim\App docblocks have missing throws annotations

RyanNerd:
- Add missing `@throws`
  - Closes: #2666

- Format DocBlocks to be cohesive
  - Use imports instead of \ClassName
  - Line return after last `@param` and `@return`

  ```
  use Exception;

  /**
   * @param ...
   *
   * @return ...
   *
   * @throws Exception
   */
  ```

  Closes #2668


Slim/Container.php
```diff
@@ -121,7 +122,7 @@ public function get($id)
         }
         try {
             return $this->offsetGet($id);
-        } catch (\InvalidArgumentException $exception) {
+        } catch (InvalidArgumentException $exception) {
             if ($this->exceptionThrownByContainer($exception)) {
                 throw new SlimContainerException(
```
l0gicgate, line 129: Missing the throw tag for this exception.

Slim/App.php
```diff
@@ -8,6 +8,7 @@
  */
 namespace Slim;
 
+use Interop\Container\Exception\ContainerException;
```
l0gicgate: Please order the imports alphabetically and remove any unused imports.

Slim/Container.php
RyanNerd: Fixed in the last commit I just pushed.

Slim/App.php
RyanNerd: This will take me some time to sort out. Right now It's 1:11am my local time. I'll see if tomorrow I can find some time...

RyanNerd: Latest commit and push sorts all imports alphabetically

RyanNerd:
Travis is reporting: Slim\Tests\AppTest::testRunNotAllowedWithoutHandler
ReflectionException: Class MethodNotAllowedException does not exist.
I've tried several times to fix this (remove import and use inline import for example). I'm stuck not sure why this class is not loading or being detected and loaded.

l0gicgate:
See #2671

PR #2672, opened 2019-04-28
Add throws ContainerException annotations

adriansuter:
The `\Psr\Container\ContainerInterface` may throw `\Interop\Container\Exception\ContainerException` exceptions. This PR annotates the methods in `\Slim\App` repspectively.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -100,6 +100,7 @@ class App
      * @return mixed
      *
      * @throws BadMethodCallException
+     * @throws ContainerException
      */
     public function __call($method, $args)
     {
@@ -120,6 +121,8 @@ class App
      * @param  callable|string $callable The route callback routine
      *
      * @return RouteInterface
+     *
+     * @throws ContainerException
      */
     public function get($pattern, $callable)
     {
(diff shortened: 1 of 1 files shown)


adriansuter:
The PR #2668 did not add these annotations.

l0gicgate:
it will never throw those exceptions. The method that throws those is `$container->get()` in the event that the requested offset isn’t present on the container. The thing is, all those offsets will always be present since we define them ourselves in the default service provider.

adriansuter:
Well, that is true. Is it possible, that someone would unset those properties? Maybe by accident?

l0gicgate:
It’s very unlikely that someone would go out of their way to unset those. We’re not going to annotate those exceptions.

adriansuter:
What about a try catch inside the methods?

l0gicgate:
Try/catch and then what? Throw again? It doesn’t make sense. It’s going to stay as is.

adriansuter:
I'm fine with that. Sorry. I just thought to add a try/catch/ignore to make sure, no-one would issue that again. My IDE is actually warning me, because of non-annotated exceptions that could be possibly thrown. Never mind.

l0gicgate:
The thing is if we annotate those exceptions then the end user will have to try/catch all those methods because their IDE is now going to complain. I’d rather us contain it at our level.

Issue #2669, opened 2019-04-27
Unit tests failing building Fedora RPM


trasher:
Hi,

I'm helping on Slim packaging on fedora from a while. Since 3.12, unit tests are failing in the build environment :/
On this try, PHP 7.3.5 RC1 is used, but result is the same with 7.2.17.

Log can be found here:
https://kojipkgs.fedoraproject.org//work/tasks/3317/34483317/build.log

You can find software versions at:
https://kojipkgs.fedoraproject.org//work/tasks/3317/34483317/root.log

Any idea what goes wrong? Thank you :)

l0gicgate:
It's hard to diagnose what's going on since the builds for that release pass just fine. The tests that are failing weren't affected by code from that release. I'm not sure what's going on here.

The failing tests are:
```bash
1) Slim\Tests\AppTest::testResponseReplacesPreviouslySetHeaders
Failed asserting that Array &0 () is identical to Array &0 (
    0 => Array &1 (
        'header' => 'X-Foo: baz1'
        'replace' => true
        'status_code' => null
    )
    1 => Array &2 (
        'header' => 'X-Foo: baz2'
        'replace' => false
        'status_code' => null
    )
    2 => Array &3 (
        'header' => 'HTTP/1.1 200 OK'
        'replace' => true
        'status_code' => 200
    )
).

2) Slim\Tests\AppTest::testResponseDoesNotReplacePreviouslySetSetCookieHeaders
Failed asserting that Array &0 () is identical to Array &0 (
    0 => Array &1 (
        'header' => 'Set-Cookie: foo=bar'
        'replace' => false
        'status_code' => null
    )
    1 => Array &2 (
        'header' => 'Set-Cookie: bar=baz'
        'replace' => false
        'status_code' => null
    )
    2 => Array &3 (
        'header' => 'HTTP/1.1 200 OK'
        'replace' => true
        'status_code' => 200
    )
).

3) Slim\Tests\RouterTest::testCacheFileExistsAndIsNotReadable
Failed asserting that exception of type "\RuntimeException" is thrown.
```

trasher:
Are there chances some of those failing tests requires an internet connexion? The build environment does not have access to internet.

Since project tests are OK, and usage seems ok; I can find a way to ignore them on packaging side at last; this is not a blocker ;)

Thank you.

l0gicgate:
[...]

PR #2673, opened 2019-04-29
non-buffered responses fixes

1ma:
- Fix potential casting from array to string in Response::withHeader()
- Fix non buffered responses when calling Response::withAddedHeader() and Response::withoutHeader()

Reference: https://github.com/slimphp/Slim-Psr7/pull/39#issuecomment-487278458

### DIFF
--- a/Slim/Http/Message.php
+++ b/Slim/Http/Message.php
@@ -230,6 +230,10 @@ abstract class Message implements MessageInterface
         $clone = clone $this;
         $clone->headers->add($name, $value);
 
+        if ($this instanceof Response && $this->body instanceof NonBufferedBody) {
+            header(sprintf('%s: %s', $name, $this->getHeaderLine($name)));
+        }
+
         return $clone;
     }
 
@@ -250,6 +254,10 @@ abstract class Message implements MessageInterface
     {
         $clone = clone $this;
         $clone->headers->remove($name);
+
+        if ($this instanceof Response && $this->body instanceof NonBufferedBody) {
+            header_remove($name);
+        }
 
         return $clone;
     }
--- a/Slim/Http/Message.php
+++ b/Slim/Http/Message.php
@@ -231,7 +231,7 @@ abstract class Message implements MessageInterface
         $clone->headers->add($name, $value);
 
         if ($this instanceof Response && $this->body instanceof NonBufferedBody) {
-            header(sprintf('%s: %s', $name, $this->getHeaderLine($name)));
+            header(sprintf('%s: %s', $name, $clone->getHeaderLine($name)));
         }
 
         return $clone;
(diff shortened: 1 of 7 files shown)


l0gicgate:
Can you PR this on Slim PSR-7 as well please? Also we should add tests cases for the newly introduced code.

1ma:
The merged PR against slim/psr7 that I referenced in the first comment already has these fixes. Also, regarding the difficulty of testing this feature: https://github.com/slimphp/Slim/pull/2504#issuecomment-469035419

Unfortunately neither of the 2 original PRs had any (real) tests for this.

Slim/Http/Message.php
```diff
@@ -230,6 +230,10 @@ public function withAddedHeader($name, $value)
         $clone = clone $this;
         $clone->headers->add($name, $value);
 
+        if ($this instanceof Response && $this->body instanceof NonBufferedBody) {
```
l0gicgate, line 233: We can add tests by overriding the `header()` can't we?

Slim/Http/Message.php
```diff
@@ -251,6 +255,10 @@ public function withoutHeader($name)
         $clone = clone $this;
         $clone->headers->remove($name);
 
+        if ($this instanceof Response && $this->body instanceof NonBufferedBody) {
+            header_remove($name);
```
l0gicgate, line 259: We can also override `header_remove()` here just to make sure it's being called as well.

Slim/Http/Response.php
```diff
@@ -270,13 +270,13 @@ public function getReasonPhrase()
...
     {
-        if ($this->body instanceof NonBufferedBody) {
-            header("$name: $value");
-        }
-
         $clone = clone $this;
         $clone->headers->set($name, $value);
 
+        if ($this->body instanceof NonBufferedBody) {
+            header(sprintf('%s: %s', $name, $this->getHeaderLine($name)));
```
l0gicgate: Same here we can override `header()`.

Slim/Http/Message.php
1ma: I don't know how to do that and Google isn't shedding much light either. Can you think of any example?

1ma: Unless you mean moving the header and header_remove functions to new private methods just for the sake of mocking them (in which case code coverage would still be incomplete)?

1ma:
That'd be it. I've done some cleanup in the testing autoloading, too. These overrides should only be loaded from tests/bootstrap.php, not at each `composer install`. Now to port these new tests back to Slim-Psr7...

l0gicgate (approved):

l0gicgate:
I'm somewhat worried we might break some builds in the wild as seen in #2669 because we're moving the assets into `bootstrap.php` but I guess we'll have to deal with it when the issues come in. Great work @1ma!

Issue #2653, opened 2019-04-23
Slim 4 Alpha Release Feedback


l0gicgate:
**Slim 4 Alpha Release**
See the full [release notes](https://github.com/slimphp/Slim/pull/2665)

**Before doing anything read the docs**
I just finished the first draft of the docs for Slim 4 which are available here. I need feedback please:
http://dev.slimframework.com/docs/v4

**Download an test the 4.0.0-alpha release**
You may also play around with the `4.x` branch and create a simple app with it to test things out.
`composer require slim/slim:4.0.0-alpha`

**Install a PSR-7 Implementation**
You will also need to install a PSR-7 implementation and ServerRequestCreator combo. You can use Slim's PSR-7 Implementation or choose one of the ones described on the `4.x` branch README: https://github.com/slimphp/Slim/blob/4.x/README.md
`composer require slim/psr7:dev-master`

**Slim 4 DDD API Skeleton Prototype**
I also just created a Slim4 Skeleton with a DDD style directory structure with example files and test coverage for everything. I'm not sure if it's the right fit yet as it may be a bit too opinionated but I'd love some feedback:
```bash
git clone https://github.com/l0gicgate/Slim-Skeleton
git checkout 4.x
```

If you have any questions don't hesitate to ping me on Slack or ask questions in this thread directly, I'm available to help!

RyanNerd:
@l0gicgate  Just a little nitpik:
The link to "contributing: Branching Strategy" on this page: http://slim-website.lgse.com/docs/v4/cookbook/uploading-files.html -- Links to **version 3** (http://slim-website.lgse.com/docs/v3/contributors/strategy.html)

l0gicgate:
Good catch @RyanNerd! Fixed

RyanNerd:
@l0gicgate Version 3 has a Tutorial link with a guide to creating a "first application". Slim 4 with all the decoupling and _plug-in_ choices (PSR-7, Containers, Route & ErrorHandling as middleware, etc.) is much more complex than Slim 3. This can be overwhelming to someone new to Slim (and even those like myself that are familiar with version 3).
Perhaps the docs could include a Tutorial for v4 showing how to create a "first application" using bare bones concepts and the  _plug-in_ defaults.

l0gicgate:
@RyanNerd I removed the tutorial from the v4 docs for now. It does need to be written and will be added to the docs later. In the mean time someone can look at the skeleton repository to see how things are implemented in a "real life" scenario.

[...]

PR #2674, opened 2019-04-29
Add suggestions for possible PSR-7 implementations to composer.json

michaelbiberich:
References: #2653, specifically [this comment](https://github.com/slimphp/Slim/issues/2653#issuecomment-487377115).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,5 +76,13 @@
         "phpunit": "php vendor/bin/phpunit",
         "phpcs": "php vendor/bin/phpcs",
         "phpstan": "php -d memory_limit=-1 vendor/bin/phpstan analyse Slim"
+    },
+    "suggest": {
+        "slim/psr7": "This is the Slim Framework projects PSR-7 implementation.",
+        "nyholm/psr-7": "This is the fastest, strictest and most lightweight implementation at the moment.",
+        "nyholm/psr-7-server": "Necessary when using nyholm/psr-7.",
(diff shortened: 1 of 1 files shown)


l0gicgate (changes requested):
I think that we should suggest Slim PSR-7 only and provide a link to the docs instead for more repos maybe?

```
"suggest": {
    "slim/psr7": "Slim PSR-7 implementation. See http://dev.slimframework.com/docs/v4/start/installation.html for more information.",
}
```

michaelbiberich:
The requested changes have been addressed in https://github.com/slimphp/Slim/pull/2674/commits/aa6f672bc8fe5c73f0b3a2220022082c9fefc0b1.

composer.json
```diff
@@ -76,5 +76,8 @@
         "phpunit": "php vendor/bin/phpunit",
         "phpcs": "php vendor/bin/phpcs",
         "phpstan": "php -d memory_limit=-1 vendor/bin/phpstan analyse Slim"
+    },
+    "suggest": {
+        "slim/psr7": "Slim PSR-7 implementation. See http://dev.slimframework.com/docs/v4/start/installation.html for more information."
```
l0gicgate: Let's change that url to `http://www.slimframework.com/docs/v4/start/installation.html`

I don't want to have to PR this again on next release. Sorry 😂

michaelbiberich: No problem, should be addressed in https://github.com/slimphp/Slim/pull/2674/commits/249a5852060edde0c861f68d860b726d112d3ade.

l0gicgate (approved):

l0gicgate:
Thank you!

PR #2677, opened 2019-05-01
Refactoring app factory

asaokamei:
This pull request demonstrates an improvement on factory process (i.e. not intended for merge). 

The improvements are:

1. put all instantiation of objects into one class, `Slim\Factory\ServiceProvider`,
2. allow user's container to provide required objects for Slim, such as `ResponseFactoryInterface`. 

The improvement 2. is what I think is the key improvement. For instase, it allows such code. 

```php
$container = new MyFavoriteContainer();
$container->set(
    ResponseFactoryInterface::class,
    new DecoratedResponseFactory(new ResponseFactory(), new StreamFactory())
);
$app = AppFactory::create($container);
```

The benefits I can think of...

By separating the container setup and app factory, a bootstrap process for slim applications maybe cleaner. 

Maybe easier to increase required objects for Slim in the future. 

Cons are... 

more classes and codes. maybe a bit slower.


l0gicgate:
I don't quite understand how this makes the process of creating an app easier. You can already provide your own `ResponseFactoryInterface` via `AppFactory::setResponseFactory()`.

**Your syntax**:
```php
$provider = new ServiceProvider();
$container = new ServiceContainer($provider);
$container->set(
    ResponseFactoryInterface::class,
    new DecoratedResponseFactory(new ResponseFactory(), new StreamFactory())
);
$builder = new AppFactory($container);
$app = $builder->getApp();
```

**Versus**:
```php
$decoratedResponseFactory = new DecoratedResponseFactory(
    new ResponseFactory(),
    new StreamFactory()
);

AppFactory::setResponseFactory($decoratedResponseFactory);

$app = AppFactory::create();
```

asaokamei:
I do not think the improvement will make things easier to code but cleaner. 
Maybe I should not have use the word cleaner, though. I could not think of another good word... Anyway, let me try to explain... 

In short, the improvement allows a container to configure the slim application. 

This is the only reason I think it is cleaner because I can define all the dependencies of the application in one place. Also I can build the object utilizing the container's functions. 

Let's take a code from `Slim-Skeleton`'s `dependencies.php`. 
I can define the `MyResponseFactoryInterface::class` in the same file where others are defined. 

```php
return function (App $app) {
    /** @var Container $container */
    $container = $app->getContainer();

    $container->set(LoggerInterface::class, function (Container $c) {
        $settings = $c->get('settings');
        ....
        return $logger;
    });
    $container->set(ResponseFactoryInterface::class, function (Container $c) {
        return new MyDecoratedResponseFactory(new ResponseFactory(), new StreamFactory(), $c->get(LoggerInterface::class));
    });
};
```

Hmm, I need to define the container before creating the app. But I hope this shows what I mean cleaner.

l0gicgate:
I don't like the idea of a service provider. This is how Slim 3 was, they used the container for this which is something I wanted to get away from in Slim 4. This also adds an extraneous container for no reason and a layer of complexity that doesn't add any value.

Instead, why not just have a configuration file prior to creating the `App` with `AppFactory`:

**../app/factories.php**
```php
<?php

use Nyholm\Psr7\Factory\Psr17Factory;
use Slim\Factory\AppFactory;
use Slim\Http\Factory\DecoratedResponseFactory;

$decoratedResponseFactory = new DecoratedResponseFactory(
    new Psr17Factory(),
    new Psr17Factory()
);

AppFactory::setResponseFactory($decoratedResponseFactory);
```

**index.php**
```php
<?php

use Slim\Factory\AppFactory;

require __DIR__ . '/../vendor/autoload.php';

require __DIR__ . '/../app/factories.php';

$app = AppFactory::create();
$app->run();
```

asaokamei:
I see your point. I will close this request. 
Thank you for the great framework!

l0gicgate:
Thank you for bringing the discussion forward, I hope this doesn't discourage you from contributing in the future @asaokamei I appreciate the effort.

asaokamei:
Not at all!

PR #2701, opened 2019-05-30
3.x - App::respond() Additional Tests

hopeseekr:
Unfortunately for me, I only saw the stupid 4.x branch when I was preparing the fork.

You **really** ought to do all new development on the `master` branch :-(

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -366,7 +366,7 @@ class App
      *
      * @param ResponseInterface $response
      */
-    public function respond(ResponseInterface $response)
+    public function respond(ResponseInterface $response, $connectionStatus = 'connection_status')
     {
         // Send response
         if (!headers_sent()) {
@@ -415,14 +415,14 @@ class App
 
                     $amountToRead -= strlen($data);
 
-                    if (connection_status() != CONNECTION_NORMAL) {
+                    if ($connectionStatus() != CONNECTION_NORMAL) {
                         break;
                     }
                 }
             } else {
                 while (!$body->eof()) {
                     echo $body->read((int)$chunkSize);
-                    if (connection_status() != CONNECTION_NORMAL) {
+                    if ($connectionStatus() != CONNECTION_NORMAL) {
                         break;
                     }
                 }
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -366,7 +366,7 @@ class App
      *
      * @param ResponseInterface $response
      */
-    public function respond(ResponseInterface $response, $connectionStatus = 'connection_status')
+    public function respond(ResponseInterface $response)
     {
         // Send response
         if (!headers_sent()) {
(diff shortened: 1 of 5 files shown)


carc1n0gen:
That's simply not how this project does things.  There actually is no master branch at all.

Slim/App.php
```diff
@@ -295,8 +295,6 @@ public function run($silent = false)
         try {
```
l0gicgate, line 295: This try/catch block should be removed altogether since it's not catching anything.

Slim/App.php
```diff
@@ -404,7 +366,7 @@ public function process(ServerRequestInterface $request, ResponseInterface $resp
      *
      * @param ResponseInterface $response
      */
-    public function respond(ResponseInterface $response)
+    public function respond(ResponseInterface $response, $connectionStatus = 'connection_status')
```
l0gicgate: Why are you adding this to the function's signature? Instead you should stub the `connection_status` function to achieve this.

l0gicgate (changes requested):
I don't particularly see the purpose of this PR. I'm uneasy about removing things from the `App` core when we're this deep in the release cycle with no actual value added.

Slim/App.php
hopeseekr: Do you know how it is possible to override global functions like `connection_status()` ? I am pretty sure it is not possible except via the Runkit terrible hack.

JoeBengalen: In the 4.x branch it is done like this:
https://github.com/slimphp/Slim/blob/4.x/tests/Assets/PhpFunctionOverrides.php

The trick is that namespace scoped functions takes precedence over global scoped functions.

l0gicgate: The overrides for this namespace are located here https://github.com/slimphp/Slim/blob/3.x/tests/Assets/PhpFunctionOverrides.php

hopeseekr: Thank you so much for teaching me this!

hopeseekr: I have refactored it per your guidance. Thank you again.

Slim/App.php
hopeseekr: https://stackoverflow.com/questions/2614473/does-it-make-sense-to-do-try-finally-without-catch

I think in this context, it's possibly best to leave it.

tests/Assets/PhpFunctionOverrides.php
```diff
@@ -88,3 +88,18 @@ function is_writable($path)
     }
     return true;
 }
+
+function connection_status($toggleFailure = false)
```
l0gicgate: Let's use globals instead of changing the function signature.

See an example of using globals here:
https://github.com/slimphp/Slim-Http/blob/master/tests/Assets/PhpFunctionOverrides.php#L24

tests/Assets/PhpFunctionOverrides.php
```diff
@@ -89,16 +89,10 @@ function is_writable($path)
     return true;
 }
 
-function connection_status($toggleFailure = false)
+function connection_status()
```
l0gicgate, line 97: Please add docblock for this function.

tests/Assets/PhpFunctionOverrides.php
hopeseekr: Done.

tests/Assets/PhpFunctionOverrides.php
hopeseekr: Done.

Slim/App.php
l0gicgate: Please remove this try/finally block. To me it doesn't add any value here, it was originally there to catch an exception.

l0gicgate:
@hopeseekr it's been two weeks that I've requested for the changes are you going to proceed with this? If not, I'm going to close this PR.

Thanks

l0gicgate:
I'm closing this PR as stale since @hopeseekr is not responding.

PR #2702, opened 2019-05-30
4.x - Additional Tests For ResponseEmitter

hopeseekr:
Based off of my work in PR #2701.

### DIFF
--- a/tests/Assets/PhpFunctionOverrides.php
+++ b/tests/Assets/PhpFunctionOverrides.php
@@ -42,6 +42,11 @@ function header($string, $replace = true, $statusCode = null)
     );
 }
 
+/**
+ * Allows the mocking of invalid HTTP states.
+ *
+ * @return int
+ */
 function connection_status()
 {
     if (isset($GLOBALS['connection_status_return'])) {
--- a/tests/Assets/PhpFunctionOverrides.php
+++ b/tests/Assets/PhpFunctionOverrides.php
@@ -41,3 +41,12 @@ function header($string, $replace = true, $statusCode = null)
         ]
     );
 }
+
+function connection_status()
+{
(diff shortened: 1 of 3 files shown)


adriansuter:
And the work of many others 😉

tests/Assets/PhpRoutingFunctionOverrides.php
```diff
@@ -7,32 +7,51 @@
...
-namespace Slim\Routing;
-
-/**
- * Is a file descriptor writable
- *
- * @param $file
- * @return bool
- */
-function is_readable($file)
+namespace Slim
```
l0gicgate: Can you adopt the same pattern that is being adopted with the other overrides instead of including two namespaces in one file. Use globals to override the return value.

See an example of using globals here:
https://github.com/slimphp/Slim-Http/blob/master/tests/Assets/PhpFunctionOverrides.php#L24

l0gicgate (changes requested):

tests/Assets/PhpRoutingFunctionOverrides.php
hopeseekr: Thank you @l0gicgate  so much for teaching me something new and valuable!! I really appreciate it!

tests/Assets/PhpFunctionOverrides.php
```diff
@@ -41,3 +41,12 @@ function header($string, $replace = true, $statusCode = null)
         ]
     );
 }
+
+function connection_status()
```
l0gicgate, line 50: Please add docblock for this function

hopeseekr: Done.

l0gicgate (approved):

PR #2704, opened 2019-06-03
Fix trailing question mark when query values are null

ntzm:
`http_build_query` ignores values that are null, so URLs where all the query data is null leaves a trailing question mark. This PR resolves this.

### DIFF
--- a/Slim/Router.php
+++ b/Slim/Router.php
@@ -382,7 +382,11 @@ class Router implements RouterInterface
         }
         $url = implode('', $segments);
 
-        if ($queryParams) {
+        $hasQueryParams = array_filter($queryParams, function ($value) {
+            return $value !== null;
+        }) !== [];
+
+        if ($hasQueryParams) {
             $url .= '?' . http_build_query($queryParams);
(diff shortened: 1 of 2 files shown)


Antnee:
Need to be careful with this. The trailing `?` has zero effect, but some people may be depending on it. Don't ask me why 😉

l0gicgate:
@Antnee I don't see why someone would rely on that question mark to be there. The original assertion only verifies that that the array is not empty, I'd rather filter the parameters.

@ntzm can you port this to the 4.x branch as well in the `RouteParser` component. Thanks

l0gicgate (approved):

Antnee:
I was referring to the comments about removing the pointless charset from the application/json content type, @l0gicgate 😉

l0gicgate:
@Antnee unfortunately people were relying on that for whatever reason. Oh well, this is what happens with deep cycle releases.. You can't predict everyone's code won't break.

ntzm:
Will port to 4.x :+1:

Issue #2712, opened 2019-06-05
[4.x] Customize the response factory class in the slim http psr17 factory


adriansuter:
Hi there

In order to be able to set the Response Factory Class (which defaults to `Slim\Http\Factory\DecoratedResponseFactory`) in `\Slim\Factory\Psr17\SlimHttpPsr17Factory`, we should add a new static method `setResponseFactory(string $responseFactoryClass)`.

See also https://github.com/slimphp/Slim-Http/issues/95.

adriansuter:
The possibility to customize the response factory class in the slim http psr17 factory (i.e. to customize the http decorator) should be mentioned in the docs. Where would it make sense?

PR #2713, opened 2019-06-05
[4.x] Add method to customize the response factory class in the slim http psr17 factory

adriansuter:
This PR addresses #2712.

### DIFF
--- a/Slim/Factory/Psr17/SlimHttpPsr17Factory.php
+++ b/Slim/Factory/Psr17/SlimHttpPsr17Factory.php
@@ -27,4 +27,15 @@ class SlimHttpPsr17Factory extends Psr17Factory
     ): ResponseFactoryInterface {
         return new static::$responseFactoryClass($responseFactory, $streamFactory);
     }
+
+    /**
+     * Set the response factory class the method `createDecoratedResponseFactory`
+     * should use.
+     *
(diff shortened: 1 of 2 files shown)


adriansuter:
How would we be able to set a custom http decorator?

l0gicgate:
You'd need to create your own `ResponseFactory` then use `AppFactory::setResponseFactory($yourCustomFactory)`

Then create your own `ServerRequestCreator` and do `ServerRequestCreatorFactory::setServerRequestCreator($yourCustomServerRequestCreator)`.

adriansuter:
Wouldn't it be nice, that if you use one of the *built-in* PSR-7 implementations, then you just have to set the custom http decorator response factory to the slim http psr17 factory? Then Slim does automatically decorate as before.

l0gicgate:
How many decorator libraries are there for PSR-7 decoration besides Slim-Http? I don't know of any.

adriansuter:
Agree. So you think that this should not be an option? Either you use Slim-Http or you have to create your own response factory and server request creator.

l0gicgate:
If we were to do it correctly, we'd need to create a decorator provider specifically for that purpose. Opening up the SlimHttp provider with that setter just seems like a hackish way to do it. So for now, unless users end up requesting that feature I don't think it's necessary.

Issue #2710, opened 2019-06-05
[4.x] Decorated Request in case of third-party Server Request Creator


adriansuter:
Hi there

Related to #2709, how is it possible to get a decorated request?

I mean, to get a response factory that would create a decorated response, we could use `\Slim\Factory\Psr17\SlimHttpPsr17Factory`. But I can't use `\Slim\Factory\Psr17\SlimHttpServerRequestCreator` to achieve that for decorated requests, could I?

```php
declare(strict_types=1);

use Nyholm\Psr7\Factory\Psr17Factory;
use Nyholm\Psr7Server\ServerRequestCreator;
use Psr\Http\Message\RequestInterface as Request;
use Psr\Http\Message\ResponseInterface as Response;
use Slim\Factory\AppFactory;
use Slim\Factory\Psr17\SlimHttpPsr17Factory;


require __DIR__.'/../vendor/autoload.php';

$responseFactory = new Psr17Factory();
$app = AppFactory::create(
    SlimHttpPsr17Factory::createDecoratedResponseFactory(
        $responseFactory, $responseFactory
    )
);

$app->get('/', function (Request $request, Response $response, array $args) {
    $response->getBody()->write(get_class($request).', '.get_class($response));

    return $response;
});

// Create the nyholm server request creator.
$creator = new ServerRequestCreator(
    $responseFactory, $responseFactory, $responseFactory, $responseFactory
);

// Create the request.
$request = $creator->fromGlobals();

$app->run($request);
```

Requesting the route would return
```
Nyholm\Psr7\ServerRequest, Slim\Http\Response
```

adriansuter:
I just saw, that instead of using the server request creator, we could use `\Slim\Http\Factory\DecoratedServerRequestFactory`:
```php
$decoratedServerRequestFactory = new DecoratedServerRequestFactory($responseFactory);
$request = $decoratedServerRequestFactory->createServerRequest(
    'GET', $responseFactory->createUri('/'), []
);
```
Unfortunately there is no `createServerRequestFromGlobals()` method, which would therefore be up to the user to implement. Apparently this is not part of PSR-17, and so this would not be possible to be implemented directly in Slim-Http.

adriansuter:
Another workaround would be the following:
```php
declare(strict_types=1);

use Nyholm\Psr7\Factory\Psr17Factory;
use Nyholm\Psr7Server\ServerRequestCreator;
use Psr\Http\Message\RequestInterface as Request;
[...]

PR #2716, opened 2019-06-06
 4.x - Improve AppFactory & ServerRequestCreatorFactory Automatic Decoration

### LINKED ISSUE #2709: [4.x] Decorated Response in case of explicitly set Response Factory
Hi there

## The problem

When providing a response factory explicitly to the `AppFactory::create()` method, then the response would no longer be decorated automatically. Is this the expected behaviour?

Say my `index.php` is as follows (I am using the Nyholm PSR-7 for this example):
```php
declare(strict_types=1);

use Nyholm\Psr7\Factory\Psr17Factory;
use Psr\Http\Message\RequestInterface as Request;
use Slim\Factory\AppFactory;

require __DIR__.'/../vendor/autoload.php';

// Explicitly define the response factory!!
$app = AppFactory::create(
    new Psr17Factory()
);

$app->get('/', function (Request $request, Response $response, array $args) {
    $response->getBody()->write(get_class($request).', '.get_class($response));

    return $response;
});

$app->run();
```

Calling the route returns
```
Slim\Http\ServerRequest, Nyholm\Psr7\Response
```
So the request gets decorated (as expected), but the response does not get decorated.

## Is this the correct solution?

In order to get the response decorated, we need to do as follows, right?
```php
declare(strict_types=1);

use Nyholm\Psr7\Factory\Psr17Factory;
use Psr\Http\Message\RequestInterface as Request;
use Slim\Factory\AppFactory;
use Slim\Factory\Psr17\SlimHttpPsr17Factory;

require __DIR__.'/../vendor/autoload.php';

$responseFactory = new Psr17Factory();
$app = AppFactory::create(
    SlimHttpPsr17Factory::createDecoratedResponseFactory(
        $responseFactory, $responseFactory
    )
);

$app->get('/', function (Request $request, Response $response, array $args) {
    $response->getBody()->write(get_class($request).', '.get_class($response));

    return $response;
});

$app->run();
```

The result was
```
Slim\Http\ServerRequest, Slim\Http\Response
```

If this is the way to go, then this should be mentioned in the documentation I guess. What do you think?

l0gicgate:
**AppFactory's New Behavior**
`AppFactory` will attempt to decorate `ResponseFactory` instances passed via `AppFactory::create()` or set via `AppFactory::setResponseFactory()` given that a `StreamFactoryInterface` has been set via `AppFactory::setStreamFactory()` given that `Slim\Http` is present.

Note that the `AppFactory` setters will override any of the PSR-17 factory provider automatically detected implementations.

```php
<?php
use Slim\Factory\AppFactory;
use Your\Psr7\CustomResponseFactory;
use Your\Psr7\CustomStreamFactory;

// Set your custom StreamFactory first
$myStreamFactory = new CustomStreamFactory();
AppFactory::setStreamFactory($myStreamFactory);

// You can use a setter for your ResponseFactory
$myCustomResponseFactory = new CustomResponseFactory();
AppFactory::setResponseFactory($myCustomResponseFactory);
$app = AppFactory::create();

// Or you can use the AppFactory::create() constructor parameter
$app = AppFactory::create($myCustomResponseFactory);
```

**ServerRequestCreatorFactory's New Behavior**
Similarly to `AppFactory` you can set a default `ServerRequestCreator` which implements `ServerRequestCreatorInterface` that will override the automatically detected server request creators present in the PSR-17 factory provider.

```php
<?php
use Slim\Factory\AppFactory;
use Slim\Factory\ServerRequestCreatorFactory;
use Your\Psr7\CustomServerRequestCreator;

$myCustomServerRequestCreator = new CustomServerRequestCreator();
ServerRequestCreatorFactory::setServerRequestCreator($myCustomServerRequestCreator);

$app = AppFactory::create();
$app->run();
```

This PR effectively closes #2709, #2710, #2712  and #2713

### DIFF
--- a/Slim/Factory/AppFactory.php
+++ b/Slim/Factory/AppFactory.php
@@ -130,8 +130,10 @@ class AppFactory
      * @param StreamFactoryInterface   $streamFactory
      * @return ResponseFactoryInterface
      */
-    protected static function attemptResponseFactoryDecoration(ResponseFactoryInterface $responseFactory, StreamFactoryInterface $streamFactory): ResponseFactoryInterface
-    {
+    protected static function attemptResponseFactoryDecoration(
+        ResponseFactoryInterface $responseFactory,
+        StreamFactoryInterface $streamFactory
+    ): ResponseFactoryInterface {
         if (static::$slimHttpDecoratorsAutomaticDetectionEnabled
             && SlimHttpPsr17Factory::isResponseFactoryAvailable()
         ) {
--- a/Slim/Factory/AppFactory.php
+++ b/Slim/Factory/AppFactory.php
@@ -11,6 +11,7 @@ namespace Slim\Factory;
 
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
 use RuntimeException;
 use Slim\App;
 use Slim\Factory\Psr17\Psr17Factory;
@@ -32,6 +33,11 @@ class AppFactory
      * @var ResponseFactoryInterface|null
      */
     protected static $responseFactory;
+
+    /**
+     * @var StreamFactoryInterface|null
+     */
+    protected static $streamFactory;
 
     /**
      * @var ContainerInterface|null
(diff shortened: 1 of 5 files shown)


adriansuter:
Cool.

PR #2717, opened 2019-06-06
add member variable  for AbstractError classes

teanooki:
For errors I want to use the Slim error handlers. Because I'm using Twig, I have derived my own classes and overwrote the html output methods. The only bad thing was the fixed title 'Slim Application Error'. It was quite deep in the code so that I would have had to do a slow string replace afterwards or double the entire functions. Instead I decided to move this string into a member variable.

### DIFF
--- a/Slim/Handlers/Error.php
+++ b/Slim/Handlers/Error.php
@@ -66,8 +66,6 @@ class Error extends AbstractError
      */
     protected function renderHtmlErrorMessage(Exception $exception)
     {
-        $title = 'Slim Application Error';
-
         if ($this->displayErrorDetails) {
             $html = '<p>The application could not run because of the following error:</p>';
             $html .= '<h2>Details</h2>';
@@ -86,8 +84,8 @@ class Error extends AbstractError
             "<title>%s</title><style>body{margin:0;padding:30px;font:12px/1.5 Helvetica,Arial,Verdana," .
             "sans-serif;}h1{margin:0;font-size:48px;font-weight:normal;line-height:48px;}strong{" .
             "display:inline-block;width:65px;}</style></head><body><h1>%s</h1>%s</body></html>",
-            $title,
-            $title,
+            $this->title,
+            $this->title,
             $html
         );
 
@@ -159,7 +157,7 @@ class Error extends AbstractError
     protected function renderJsonErrorMessage(Exception $exception)
     {
         $error = [
-            'message' => 'Slim Application Error',
+            'message' => $this->title,
         ];
 
         if ($this->displayErrorDetails) {
@@ -189,7 +187,7 @@ class Error extends AbstractError
      */
     protected function renderXmlErrorMessage(Exception $exception)
     {
-        $xml = "<error>\n  <message>Slim Application Error</message>\n";
+        $xml = "<error>\n  <message>" . $this->title . "</message>\n";
         if ($this->displayErrorDetails) {
             do {
                 $xml .= "  <exception>\n";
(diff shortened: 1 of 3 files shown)


l0gicgate:
@adriansuter I don't like this solution, it's kind of hackish. I wonder if we could figure out a better way to do this for the `3.x` branch. Problem is most likely any other solution will introduce some potential BC changes.

adriansuter:
How about adding a new method into `\Slim\Handlers\AbstractError` called `setErrorRenderer()` which gets an instance of an error renderer. To keep it simple, the error renderer will have to define three methods to render all three output types (html, json, xml). For that, we create a new interface `ErrorRendererInterface` which contains the three methods `renderHtmlErrorMessage()`, `renderJsonErrorMessage()` and `renderXmlErrorMessage()`.

The following methods would then check for a custom error renderer and if present (not `null`), then the corresponding methods would be called. If not, then the methods would behave as before (this is to avoid BC changes).
- `\Slim\Handlers\Error::renderHtmlErrorMessage()`
- `\Slim\Handlers\Error::renderJsonErrorMessage()`
- `\Slim\Handlers\Error::renderXmlErrorMessage()`
- `\Slim\Handlers\PhpError::renderHtmlErrorMessage()`
- `\Slim\Handlers\PhpError::renderJsonErrorMessage()`
- `\Slim\Handlers\PhpError::renderXmlErrorMessage()`

Note that if done like this, and if there is a custom error renderer present, some methods would simply be ignored:
- `\Slim\Handlers\Error::renderHtmlException()`
- `\Slim\Handlers\Error::renderHtmlExceptionOrError()`
- `\Slim\Handlers\Error::createCdataSection()`
- `\Slim\Handlers\PhpError::renderHtmlError()`
- `\Slim\Handlers\PhpError::createCdataSection()`
It is the duty of the user to implement those in the custom error renderer, if needed.

## User Manual

All the user has to do is:
1. Write a custom error renderer (for example `MyErrorRenderer`) implementing the `ErrorRendererInterface`.
2. Instantiate the error handler and set the custom error renderer.
3. Put the error handler into the container.
```php
$app = new \Slim\App();
$c = $app->getContainer();
$c['errorHandler'] = function ($c) {
    $error = new Error(
        $c->get('settings')['displayErrorDetails']
    );
    $error->setErrorRenderer(new MyErrorRenderer($c));
    return $error;
};
```

## Questions

Do you think, that would introduce some BC changes?

Is this too complicated?

Slim/Handlers/Error.php
```diff
@@ -189,7 +187,7 @@ protected function renderJsonErrorMessage(Exception $exception)
      */
     protected function renderXmlErrorMessage(Exception $exception)
     {
-        $xml = "<error>\n  <message>Slim Application Error</message>\n";
+        $xml = "<error>\n  <message>" . $this->title . "</message>\n";
```
jacekkarczmarczyk, line 190: I'm pretty sure that it should be htmlspecialchar-ed (same in PhpError.php)

adriansuter: Use the method `\Slim\Handlers\Error::createCdataSection()`.

l0gicgate:
@adriansuter 

> Do you think, that would introduce some BC changes?
Is this too complicated?

At this point, I'm not comfortable making changes like this on the 3.x branch anymore.

I'm not sure what do to with this PR.

adriansuter:
Maybe we should stay away from adding new features to the 3.x branch. Bugfixes only.

l0gicgate:
@adriansuter I agree.

l0gicgate:
I'm closing this as we are no longer accepting pull requests for Slim 3 for anything other than security fixes.

PR #2719, opened 2019-06-09
Change empty to empty and !is_numeric

adriansuter:
If the body is '0' for example, the php method `empty()` would return `true`. But `\Slim\ResponseEmitter::isResponseEmpty` should return `false` in this case. So this PR adds a second condition, `!is_numeric()` to assure this.

### DIFF
--- a/Slim/ResponseEmitter.php
+++ b/Slim/ResponseEmitter.php
@@ -129,7 +129,7 @@ class ResponseEmitter
      */
     public function isResponseEmpty(ResponseInterface $response): bool
     {
-        $contents = (string)$response->getBody();
+        $contents = (string) $response->getBody();
         return (empty($contents) && !is_numeric($contents))
             || in_array($response->getStatusCode(), [204, 205, 304], true);
     }
--- a/Slim/ResponseEmitter.php
+++ b/Slim/ResponseEmitter.php
@@ -129,7 +129,8 @@ class ResponseEmitter
      */
     public function isResponseEmpty(ResponseInterface $response): bool
     {
-        $contents = (string) $response->getBody();
-        return empty($contents) || in_array($response->getStatusCode(), [204, 205, 304], true);
+        $contents = (string)$response->getBody();
+        return (empty($contents) && !is_numeric($contents))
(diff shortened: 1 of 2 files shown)


Slim/ResponseEmitter.php
```diff
@@ -129,7 +129,8 @@ private function emitBody(ResponseInterface $response): void
      */
     public function isResponseEmpty(ResponseInterface $response): bool
     {
-        $contents = (string) $response->getBody();
-        return empty($contents) || in_array($response->getStatusCode(), [204, 205, 304], true);
+        $contents = (string)$response->getBody();
+        return (empty($contents) && !is_numeric($contents))
```
bednic, line 133: Shouldn't this be something like 
```
if(strlen($contents) > 0) 
```
or something similar. Checking if it's number when you cast body as string looks wierd and dangerous. It's same with <code>empty()</code>, which works in php 5.4 and 5.3 different.

l0gicgate (approved):

Slim/ResponseEmitter.php
adriansuter: Not sure if that was a performance decision? But **yes**, `strlen()` would be much clearer.

After all, we are absolutely sure that the variable is set, is a `string`, is not `false` and is not an `array`, which would have been reasons for using `empty()`.

l0gicgate: Yeah definitely an overlook on my part. I suppose we should use `!== null && strlen` instead.

bednic: I think there is no necessity of additional conditions except strlen() > 0 : 

1. On line before we cast body as string, so null will be considered as empty string, ergo <code>''</code>
2. Every content even numeric will end up as string
3. <code>emtpy()</code> really returns true on <code>'0'</code>, which sucks, so I would get rid of it

so the line in return should be:
```
return (strlen($contents) > 0) || in_array($response->getStatusCode(), [204, 205, 304], true);
```
nothing more, nothing less.

adriansuter: Done in PR #2721 .

PR #2728, opened 2019-06-17
[4.x] Make Slim routing as extensible as possible

adriansuter:
This PR would change some properties from `private` to `protected` in order to make the classes extensible.

### DIFF
--- a/Slim/Routing/Dispatcher.php
+++ b/Slim/Routing/Dispatcher.php
@@ -13,12 +13,12 @@ class Dispatcher implements DispatcherInterface
     /**
      * @var RouteCollectorInterface
      */
-    private $routeCollector;
+    protected $routeCollector;
 
     /**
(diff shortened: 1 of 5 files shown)


shadowhand:
I feel like this is a bad direction to go in. [Composition over inheritance][1] states that we should compose functionality instead of creating extension. If anything needs to change in routing, it would be to break up the router into smaller classes so that implementations can be swapped.

[1]: https://en.wikipedia.org/wiki/Composition_over_inheritance

adriansuter:
Interesting, I need to look at that.

l0gicgate:
I don't think we should make these specific components extensible since they're implementations specifically made for use with the FastRoute library. I can't think of a use case where someone would extend these components. Let's leave them as is for now. I actually might change a few things around before Slim 4 major release here routing wise so we can implement other libraries like the Symfony router (TBD).

Issue #2731, opened 2019-06-17
[4.x] Multilingual support for error renderers


adriansuter:
Of course, Slim should stay slim - but I was wondering if it would make sense to add some sort of multilingual support for the following error renderers?

- `\Slim\Error\Renderers\HtmlErrorRenderer`
- `\Slim\Error\Renderers\PlainTextErrorRenderer`

They define some texts (hard-coded in the source code) like the title `Slim Application Error` or `The application could not run because of the following error` or `A website error has occurred. Sorry for the temporary inconvenience` and so on.

Or should that be left to the user? So he/she has to write a custom error handler. I mean, the user might even decide to use Twig or any other templating engine to do the rendering.

_Notice 1:_  
The XML and JSON error renderers do not contain texts that should be translatable.

_Notice 2:_  
The error messages itselves (at least the php error messages) would still be in english, of course. But in production, these error messages should not be shown (`$displayErrorDetails = false`) anyway.

l0gicgate:
I think we need to refactor those two renderers and give the ability to change those hard coded messages. I have not yet thought about how to do it in a way that makes it easy for the end user to customize these. Definitely open to suggestions on how we could implement this.

This is why I haven't merged #2717 yet. I'm trying to think of a better way to do this.

Brammm:
Just thinking out loud here: wouldn't it be easier to put the HTML and strings etc in some kind of template file (just straight up PHP, not using a library) and make those easily replaceable? 

The easiest would be to make it configurable, but I see version 4 has just done away with the configuration, so not sure what the best solution would be. Haven't delved too much in the internals of v4.

l0gicgate:
I think the best way to do this is to include a setter for the renderer on the `ErrorHandler` component that we ship so you don't have to extend it to change the rendering.

https://github.com/slimphp/Slim/blob/4.x/Slim/Handlers/ErrorHandler.php#L85

This solution works for both the `3.x` and `4.x` branch. That way the end user can easily create their own renderer and set it on the error handler.

adriansuter:
The problem is, that we would not know, what renderer we need to set, until the `ErrorHandler` gets invoked. Currently that would be decided in the method
https://github.com/slimphp/Slim/blob/ccefb2ff3eb3ea3e686c111c2ca4d51253764261/Slim/Handlers/ErrorHandler.php#L204

## Idea 1

[...]

PR #2734, opened 2019-06-20
[4.x] Make error renderers in the error handler customizable

adriansuter:
This PR addresses #2731.

The default error handler has now a registry of known content types and their corresponding error renderers. It is possible to register new content types (or update existing content types) using `\Slim\Handlers\ErrorHandler::registerErrorRenderer()`.

The default error renderer can also be defined using `\Slim\Handlers\ErrorHandler::setDefaultErrorRenderer()`.

**I removed the property `$knownContentTypes` as this information could now be read directly from the array keys of the new property `$errorRenderers`. Not sure if that would break things for users.**

## Usage

A user could define a new renderer
```php
class MyErrorRenderer implements \Slim\Interfaces\ErrorRendererInterface
{
    /**
     * {@inheritdoc}
     */
    public function render(Throwable $exception, bool $displayErrorDetails): string
    {
        // TODO Build the custom error string that should be sent as response to the client
        // For example:
        return 'Custom Translated Error';
    }
}
```
and then after instantiating the app, the user could do the following
```php
// Add error handling middleware
$callableResolver = $app->getCallableResolver();
$responseFactory  = $app->getResponseFactory();
$errorMiddleware  = new ErrorMiddleware($callableResolver, $responseFactory, $displayErrorDetails, false, false);

$eh = $errorMiddleware->getDefaultErrorHandler();
$eh->registerErrorRenderer('text/html', MyErrorRenderer::class);
$errorMiddleware->setDefaultErrorHandler($eh);

$app->add($errorMiddleware);
```

## Further work to do

I think it would be nice, if we could get the Container into the renderer (if defined). So that the renderer can get the template engine for example.

Idea: A new interface called `ContainerAwareInterface` that has exactly one method called `setContainer()`. So when the renderer gets instantiated, we could check if the renderer implements this interface and if so, we could inject the container into the renderer.

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -32,16 +32,19 @@ use Throwable;
 class ErrorHandler implements ErrorHandlerInterface
 {
     /**
-     * Known handled content types
-     *
-     * @var array
+     * @var string
      */
-    protected $knownContentTypes = [
-        'application/json',
-        'application/xml',
-        'text/xml',
-        'text/html',
-        'text/plain'
+    protected $defaultErrorRenderer = HtmlErrorRenderer::class;
+
+    /**
+     * @var string[]
+     */
+    protected $errorRenderers = [
+        'application/json' => JsonErrorRenderer::class,
+        'application/xml' => XmlErrorRenderer::class,
+        'text/xml' => XmlErrorRenderer::class,
+        'text/html' => HtmlErrorRenderer::class,
+        'text/plain' => PlainTextErrorRenderer::class,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


l0gicgate:
I wonder if we should just pass `CallableResolver` into `ErrorHandler` and resolve the renderers. Then it gives the ability to someone to inject dependencies via the renderer's constructor. 

Everything looks good though so far for this PR, great work!

adriansuter:
I am currently away. Will have to check that next week.

Thanks for your kind words.

adriansuter:
@l0gicgate 
> I wonder if we should just pass `CallableResolver` into `ErrorHandler` and resolve the renderers. Then it gives the ability to someone to inject dependencies via the renderer's constructor.

I think, that could work. In this case though, the renderer(s) would be instantiated in every request, even if no error gets triggered. Could that lead to a performance issue?

To achieve that, we would have to change the signature of the constructor `\Slim\Handlers\ErrorHandler::__construct()` as we have to be sure that the resolver exists. Could that be a potential breaking change?

l0gicgate:
> To achieve that, we would have to change the signature of the constructor \Slim\Handlers\ErrorHandler::__construct() as we have to be sure that the resolver exists. Could that be a potential breaking change?

If someone extended the base handler and overrode the existing constructor then they'd have to change the constructor signature then I guess we'd break their code but that's a pretty minor fix. We're still in beta so I'd rather break things now if anything.

It's a much cleaner way to instantiate things to just use `CallableResolver`. The logic is all there so.

l0gicgate:
@adriansuter this PR looks good to go, when you remove the WIP from the title I'll merge. Thanks

adriansuter:
What about #2737?

Issue #2749, opened 2019-06-29
[4.x] Error handler does not add content-type header if default renderer is used


adriansuter:
Depending on the error renderer, the error handler should correctly add a content-type header to the response.

**HtmlErrorRenderer**
text/html

**JsonErrorRenderer**
application/json

**PlainTextErrorRenderer**
text/plain

**XmlErrorRenderer**
text/xml

## Proposition A
Add a new method to the renderer interface that would return the content-type.

## Proposition B
Add a method to the renderer interface that receives the response and returns a possibly altered response (for example headers added). Method name to be discussed 😃.

## Caution
Both propositions would introduce a breaking change.

Other ideas or inputs? Could we just use the `$contentType` property from the error handler itself? Or could that be dangerous as this is actually user input? What to do in case of the default error renderer?

l0gicgate:
So I don’t think it should be the renderer’s responsibility to set the Content-Type. Since the error handler generates the response and determines the Content-Type I would say let’s have a setter to force a certain content type, add a parameter to the renderer registering function so you can also set the content type associated with it, and then if the default content type isn’t set we fallback on the detected content type parsed during renderer determination logic. 

What do you think? This doesn’t introduce any breaking changes.

adriansuter:
You are right, the renderer is just responsible for rendering the error.

By the way, I just saw, that actually the error handler is already adding the content-type header.
https://github.com/slimphp/Slim/blob/01f692327ece29bcf930db99bf1c12b84eafa6c9/Slim/Handlers/ErrorHandler.php#L273

Hence the only renderer that the handler does not know the content type of, is the default renderer. Well, currently it appears that the handler assumes the default renderer to be `text/html`. The method `\Slim\Handlers\ErrorHandler::determineContentType()` would return `text/html` if no match could be found between `Accept` header of the request and registered error renderers.

Probably we should return `null` in those cases, i.e. the content type could not be determined (or could not be matched with registered error renderers) successfully. Then we should bring the content type of the default error renderer into the error handler. Probably as new argument to `setDefaultErrorRenderer()`. To make it consistent (with `registerErrorRenderer`) I would suggest to add the content type argument as first argument.
```php
    public function setDefaultErrorRenderer(string $contentType, $errorRenderer): void
    {
        $this->defaultErrorRendererContentType = $contentType;
        $this->defaultErrorRenderer = $errorRenderer;
    }
```

So the `respond()` method would look like
```php
    protected function respond(): ResponseInterface
    {
[...]

PR #2751, opened 2019-07-01
[4.x] Add content-type for the default error renderer

adriansuter:
This PR addresses #2749.

There is one commit https://github.com/slimphp/Slim/commit/419f9bd91fc3f78b2395bb9b1d524ac45e3d3803 that is actually a fix (independent of the issue).

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -69,7 +69,7 @@ class ErrorHandler implements ErrorHandlerInterface
     protected $logErrorDetails;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $contentType;
 
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -33,6 +33,11 @@ use Throwable;
 class ErrorHandler implements ErrorHandlerInterface
 {
     /**
+     * @var string
+     */
+    protected $defaultErrorRendererContentType = 'text/html';
+
+    /**
      * @var ErrorRendererInterface|string|callable
      */
     protected $defaultErrorRenderer = HtmlErrorRenderer::class;
@@ -168,7 +173,7 @@ class ErrorHandler implements ErrorHandlerInterface
      * @param ServerRequestInterface $request
      * @return string
      */
-    protected function determineContentType(ServerRequestInterface $request): string
+    protected function determineContentType(ServerRequestInterface $request): ?string
     {
         $acceptHeader = $request->getHeaderLine('Accept');
         $selectedContentTypes = array_intersect(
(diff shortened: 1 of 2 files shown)


l0gicgate:
So one thing I noticed is that right now we're letting the request `Accept` header dictate which renderer is being used. Let's just take a real world example where I'm building an API and I want to force the renderer to ALWAYS be rendering in `application/json`, don't you agree that we should make the `defaultErrorRenderer` override the determined content type?

adriansuter:
That is absolutely true. In fact, it is the route that could define, what error content type to be expected. Or the user might even set the response content type based on a route param. For example, say we have a route that would usually return html, but if we add `?json` to the url, then the content type should be json. How do we know in the error handler, what content type the user would like to return?

l0gicgate:
>For example, say we have a route that would usually return html, but if we add ?json to the url, then the content type should be json. How do we know in the error handler, what content type the user would like to return?

This is why we used the `Accept` header in the first place.

So what I'm thinking is that we need to introduce two modes into the handler. First mode is `detect` which detects the content type (it would also be the default mode) and then a `force` mode which would force the content type to be what the user defines. 
```php
$errorHandler->forceContentType('application/json')
```



There's a few more things we need to do with this handler. Namely add default headers for the response since it creates an entirely new response and if you need CORS headers then you'd need to extend the `ResponseEmitter` in order to add headers to that response which kind of sucks. I'm not sure if we need to add `setDefaultHeaders` or perhaps include a way to add middleware so we can do something like:
```php
$errorHandler->addResponseMiddleware(function (Response $response) {
    return $response->withHeader(...);
});
```

adriansuter:
Would it be possible to set the forced content type in a route handler? I'm asking because if you mix different content types in the same route handler (or between different route handlers), then it should be possible to do that. Or do you recommend that the client does have to set the accept header correctly in that case?

Concerning the response headers: one callback should be enough. Or do you think that there could be more than one "middleware"? Probably some sort of `ErrorResponseHeaderRenderer` or `ErrorResponseHeaderHandler`.

l0gicgate:
I don't think that's necessary. Also, I'm thinking that maybe we should modify the `ResponseEmitter` instead to give the opportunity to append default headers via there since all content that gets emitted gets funneled through there. Lets' just add the `detect` and the `force` mode to the error handler and I think we'll be good to go. We can dive into appending default headers in a separate PR.

l0gicgate:
Is this PR good to go @adriansuter? or are you going to add the detect/force mode?

adriansuter:
I will add the force mode.

adriansuter:
@l0gicgate The `forceContentType()` is added to the error handler. I think that this should be added to the docs somewhere.

l0gicgate:
Yes we should probably put it in the advanced error handling section

l0gicgate (approved):

PR #2759, opened 2019-07-19
4.x - Add route context helper

### LINKED ISSUE #2758: Accessing RouteParser From Route Callables
We need to figure out a sane way for users to access the `RouteParser` component from a route callable. Right now you have to jump through hoops to get access to it from within a route and it feels unnatural to have to map it via DI.

Perhaps an internal middleware, or perhaps creating a route callable wrapper which would provide the component.

Open to suggestions.

### LINKED ISSUE #2761: Routing tests have impossible case
`RoutingMiddlewareTest::testRouteIsNotStoredOnMethodNotAllowed()` expects an exception, but it also includes inspection of the request. The [inspection part of the test][1] is never reached, because `RoutingMiddleware` throws an exception for a "method not allowed" error.

[1]: https://github.com/slimphp/Slim/blob/c4bdfa0bf05b6391f8bfe44852e289dba7cc40f4/tests/Middleware/RoutingMiddlewareTest.php#L72-L81

shadowhand:
Creates a new `RouteContext` class that can be used to gain access to the current request route, route parser, and routing results.

Fixes #2758 
Fixes #2761

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -68,7 +68,7 @@ class App extends RouteCollectorProxy implements RequestHandlerInterface
         );
 
         $this->routeResolver = $routeResolver ?? new RouteResolver($this->routeCollector);
-        $routeRunner = new RouteRunner($this->routeResolver);
+        $routeRunner = new RouteRunner($this->routeResolver, $this->routeCollector->getRouteParser());
 
         $this->middlewareDispatcher = new MiddlewareDispatcher($routeRunner, $container);
     }
--- a/Slim/Middleware/RoutingMiddleware.php
+++ b/Slim/Middleware/RoutingMiddleware.php
@@ -16,6 +16,7 @@ use Psr\Http\Server\RequestHandlerInterface;
 use RuntimeException;
 use Slim\Exception\HttpMethodNotAllowedException;
 use Slim\Exception\HttpNotFoundException;
+use Slim\Interfaces\RouteParserInterface;
 use Slim\Interfaces\RouteResolverInterface;
 use Slim\Routing\RoutingResults;
 
@@ -27,11 +28,18 @@ class RoutingMiddleware implements MiddlewareInterface
     protected $routeResolver;
 
     /**
-     * @param RouteResolverInterface $routeResolver
+     * @var RouteParserInterface
      */
-    public function __construct(RouteResolverInterface $routeResolver)
+    protected $routeParser;
+
+    /**
+     * @param RouteResolverInterface $routeResolver
+     * @param RouteParserInterface   $routeParser
+     */
+    public function __construct(RouteResolverInterface $routeResolver, RouteParserInterface $routeParser)
     {
         $this->routeResolver = $routeResolver;
+        $this->routeParser = $routeParser;
     }
 
     /**
@@ -45,6 +53,7 @@ class RoutingMiddleware implements MiddlewareInterface
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
+        $request = $request->withAttribute('routeParser', $this->routeParser);
         $request = $this->performRouting($request);
         return $handler->handle($request);
     }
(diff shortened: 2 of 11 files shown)


adriansuter:
Looks good, althoug I have to check in more depth.

Could you please add the slim file header comment to the two new files?

shadowhand:
@adriansuter ✅

adriansuter:
In `RouteContex::fromRequest()`, should we check if the three calls to 
`$serverRequest->getAttribute()` did not return `null`?

Slim/App.php
```diff
@@ -68,11 +77,20 @@ public function __construct(
         );
 
         $this->routeResolver = $routeResolver ?? new RouteResolver($this->routeCollector);
-        $routeRunner = new RouteRunner($this->routeResolver);
+        $this->routeParser = $routeParser ?? new RouteParser($this->routeCollector);
```
l0gicgate: You can access the `RouteParser` via `$this->getRouteCollector()->getRouteParser()` no need to add it to the constructor or instantiate it here.

Slim/App.php
```diff
@@ -68,11 +77,20 @@ public function __construct(
...
+        $routeRunner = new RouteRunner($this->routeResolver, $this->routeParser);
 
         $this->middlewareDispatcher = new MiddlewareDispatcher($routeRunner, $container);
     }
 
+    /**
+     * @return RouteParserInterface
+     */
+    public function getRouteParser(): RouteParserInterface
```
l0gicgate: Let's remove this getter as it is accessible via `$app->getRouteCollector()->getRouteParser()`

l0gicgate (changes requested):

shadowhand:
@adriansuter and do what? Throw an exception? PHP will already throw a `TypeError` if the values are `null` when trying to construct the `RouteContext`.

l0gicgate:
@shadowhand we need to throw an exception to notify the user that routing hasn't been performed if those attributes are null. Probably throw a `RuntimeException`

adriansuter:
Or an `InvalidArgumentException` as actually the given argument does not have sufficient information.

But then, a call to this static method has to be wrapped into a try-catch block. Probably it is better to throw a `RuntimeException`.

shadowhand:
@l0gicgate @adriansuter updated with your requested changes.

Slim/App.php
```diff
@@ -18,8 +18,10 @@
 use Slim\Factory\ServerRequestCreatorFactory;
 use Slim\Interfaces\CallableResolverInterface;
 use Slim\Interfaces\RouteCollectorInterface;
+use Slim\Interfaces\RouteParserInterface;
```
adriansuter: This can be removed.

Slim/App.php
```diff
@@ -18,8 +18,10 @@
...
 use Slim\Interfaces\RouteResolverInterface;
 use Slim\Routing\RouteCollectorProxy;
+use Slim\Routing\RouteParser;
```
adriansuter: This can be removed.

adriansuter (changes requested):
Some imports can be removed.

Slim/Middleware/RoutingMiddleware.php
```diff
@@ -76,6 +84,7 @@ public function performRouting(ServerRequestInterface $request): ServerRequestIn
                     ->prepare($routeArguments);
                 return $request
                     ->withAttribute('route', $route)
+                    ->withAttribute('routeParser', $this->routeParser)
```
l0gicgate: I just noticed that we are only adding the `routingResults` and `routeParser` attributes when a route is found. They should be included in both the `MethodNotAllowed` and `Found` cases. In the `MethodNotAllowed` case there won't be a `route` attribute so we need to make that attribute nullable on the `RouteContext`

Slim/App.php
shadowhand: Would be nice if phpcbf would pick that up. 😕

Slim/Middleware/RoutingMiddleware.php
shadowhand: #2761 mentioned this too. I updated this PR to fix it.

l0gicgate (approved):
Looks good to me @adriansuter can you review one last time before I merge please?

adriansuter (approved):

adriansuter:
Thanks - everything is fine.

adriansuter:
We should not forget to update the docs about this new feature.

l0gicgate:
@adriansuter absolutely

PR #2763, opened 2019-07-20
Remove code duplication

mapogolions:


l0gicgate (approved):

Slim/CallableResolver.php
```diff
@@ -51,24 +51,33 @@ public function resolve($toResolve)
...
 
-        if (!is_string($toResolve)) {
-            $this->assertCallable($toResolve);
-        }
-
-        // check for slim callable as "class:method"
-        if (preg_match(self::CALLABLE_PATTERN, $toResolve, $matches)) {
-            $resolved = $this->resolveCallable($matches[1], $matches[2]);
-            $this->assertCallable($resolved);
+        $resolved = $toResolve;
```
adriansuter, line 54: Is there a reason to assign this here instead of inside an `else`-clause in the next conditional block? Because if done like this, this assignment might be unnecessary if `$toResolve` is a `string`.

mapogolions: It's just default value. You're right, IF statement can override that. Of course, you can do something like this.
```php
$resolved = null;
if (...) {
  $resolved = ...;
} else {
  $resolved = ...;
}
...
```
but i prefer a more concise version

adriansuter: You don't even need to initialize the variable, because `$resolved` is guranteed to be assigned during the whole conditional block. But I am fine with that. Thanks for your work.

adriansuter (approved):

Slim/CallableResolver.php
```diff
@@ -80,7 +89,7 @@ public function resolve($toResolve)
      *
      * @throws RuntimeException if the callable does not exist
      */
-    protected function resolveCallable($class, $method = '__invoke')
+    protected function resolveCallable($class, $method)
```
adriansuter, line 92: @l0gicgate Could that be a possible BC?

l0gicgate: Well considering it's a protected function and it's a final class, I don't believe so.

adriansuter:
In my opinion, this PR can be merged.

Issue #2762, opened 2019-07-19
Simplifying queuing internal middleware


l0gicgate:
I think that we should add helpers on the `App` component to instantiate and add middleware to the queue.

Example to add the `ErrorMiddleware`:
```php
$callableResolver = $app->getCallableResolver();
$responseFactory = $app->getResponseFactory();
$errorMiddleware = new ErrorMiddleware($callableResolver, $responseFactory, true, true, true);
$app->add($errorMiddleware);
```

The only real input we need from the user is for the last 3 parameters:
```php
/*
 * @param bool $displayErrorDetails - Should be set to false in production
 * @param bool $logErrors - Parameter is passed to the default ErrorHandler
 * @param bool $logErrorDetails - Display error details in error log
 */
```

Should we do something like:
```php
$app->addErrorMiddleware(true, true, true);
```

Same goes for the `RoutingMiddleware`:
```php
// Currently
$routingMiddleware = new RoutingMiddleware($routeResolver);
$app->add($routingMiddleware);

// Could be
$app->addRoutingMiddleware();
```

It would remove a lot of the cluttering statements

shadowhand:
Instead of adding more things to `App`, what about making a factory?

```php
$app->add(SlimMiddleware::createRouting($app));
$app->add(SlimMiddleware::createErrorHandler($app));
```

adriansuter:
Good idea.

Probably we should also make sure, that the error middleware and the routing middleware would not be added more than once.

adriansuter:
Or some sort of a setup (helper) class that helps setting up the slim framework with the default core built-in middlewares?

```php
SlimSetup::addDefaultErrorMiddleware($app, true, true, true) : MiddlewareInterface;
```

shadowhand:
@adriansuter I actually like that more, duplicating `$app` twice in a line is awkward. With defaults, it gets real clean:

```php
SlimMiddleware::addRouting($app);
SlimMiddleware::addErrorHandling($app);
```

adriansuter:
@shadowhand Beware! A class name ending by `Middleware` which is not really a middleware might be confusing for users.

shadowhand:
@adriansuter true, `SlimSetup` is probably reasonable. I feel like `AppFactory` could also be a reasonable place for these helpers:

```php
$app = AppFactory::create();
AppFactory::addRoutingMiddleware($app);
AppFactory::addErrorHandlingMiddleware($app);
```

adriansuter:
[...]

PR #2765, opened 2019-07-29
[4.x]  Simplifying queuing internal middleware

adriansuter:
This PR addresses #2762.

Not sure, if the test cases test everything that needs to be tested.

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -103,6 +103,11 @@ class App extends RouteCollectorProxy implements RequestHandlerInterface
         return $this;
     }
 
+    /**
+     * Add the slim built-in routing middleware to the app middleware stack
+     *
+     * @return RoutingMiddleware
+     */
     public function addRoutingMiddleware(): RoutingMiddleware
     {
         $routingMiddleware = new RoutingMiddleware(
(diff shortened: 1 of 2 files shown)


l0gicgate:
Looks good to me @adriansuter, great work! As for the tests, the middleware test suite was already covering most of this anyway but it's good you went above and beyond with it! We just need to update the docs to reflect those changes. I'm going to merge.

l0gicgate (approved):

Slim/App.php
```diff
@@ -101,6 +103,46 @@ public function addMiddleware(MiddlewareInterface $middleware): self
...
+     * @param bool $displayErrorDetails
+     * @param bool $logErrors
+     * @param bool $logErrorDetails
+     *
+     * @return ErrorMiddleware
+     */
+    public function addErrorMiddleware(
+        bool $displayErrorDetails,
+        bool $logErrors,
+        bool $logErrorDetails
```
shadowhand, line 133: In the interest of simplification, should these default to `true`?

l0gicgate: I think we need to keep the signature the same as the middleware implementation. Wasn't there someone claiming that true defaults were the root of all evil here not long ago? @akrabat 😂

adriansuter: Also the defaults would depend on the environment. In development a user might want to have all set to `true`, but in production one should not display error details (security issue). So I think, we should keep it as is. So the user has to think about these flags.

akrabat: Defaults are difficult  :)

akrabat: Do we actually need `logErrorDetails`? i.e. what's the use-case for logging errors, but not the details?

adriansuter: @akrabat Probably disk memory limitation concerns?

PR #2776, opened 2019-08-05
4.x - Fix group middleware on multiple nested groups

### LINKED ISSUE #2775: Group middleware doesn't apply to all nested groups
I came across a strange bug in Slim 4.0.0. It seems like group middleware is only running for the first nested group. I.e. `/api/works/test` outputs "Middleware ran successfully", but `/api/does_not_work/test` returns an empty response. Here's my code:

```php
use Psr\Http\Server\RequestHandlerInterface as RequestHandler;
use Slim\Factory\AppFactory;
use Slim\Routing\RouteCollectorProxy;

require '../../vendor/autoload.php';

$app = AppFactory::create();

$app->group('/api', function (RouteCollectorProxy $app) {
    $app->group('/works', function (RouteCollectorProxy $app) {
        $app->get('/test', function (Request $request, Response $response) {
            return $response;
        });
    });

    $app->group('/does_not_work', function (RouteCollectorProxy $app) {
        $app->get('/test', function (Request $request, Response $response) {
            return $response;
        });
    });
})->add(function (Request $request, RequestHandler $handler) {
    $response = $handler->handle($request);
    $response->getBody()->write('Middleware ran successfully');
    return $response;
});

$app->run();
```

theodorejb:
Resolves #2775

### DIFF
--- a/Slim/Routing/RouteCollector.php
+++ b/Slim/Routing/RouteCollector.php
@@ -252,7 +252,7 @@ class RouteCollector implements RouteCollectorInterface
         $this->routeGroups[] = $routeGroup;
 
         $routeGroup->collectRoutes();
-        array_shift($this->routeGroups);
+        array_pop($this->routeGroups);
 
         return $routeGroup;
     }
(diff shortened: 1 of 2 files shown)


l0gicgate:
Good catch! Thanks for the PR.

l0gicgate (approved):

Issue #2780, opened 2019-08-06
4.2 - Middleware Callable Resolver Support


adriansuter:
See #2779.

We should modify the `CallableResolver` (and its interface) in order to be able to use it in the `MiddlewareDispatcher`. There should be 3 methods:
* Resolve a regular callable (For any invocables)
* Resolve a route callable
* Resolve a middleware callable

 Currently, in 4.0.x and 4.1.x, the logic of resolving the middleware callable has been hard coded into the dispatcher. This has to be replaced.

**Warning:** This is a BC! New users of Slim 4 should, if possible, not mess around with the callable resolver.

At the same time, the arguments of the `App` constructor should be extended by a `MiddlewareDispatcherInterface` such that users could provide their own dispatcher implementation.

l0gicgate:
I know we are going against typical versioning practices by doing this but it's too early in the release cycle to wait until the next major version to fix this.

This BC break will be seamless for most users unless you:
- Implemented your own `CallableResolverInterface`
- You instantiated the app manually instead of using `AppFactory`

l0gicgate:
We might be able to implement this functionality without having to change the interface after some thoughts. Since the method `CallableResolverInterface::resolve()` is not typed, we can pass in a wrapper which will indicate what kind of callable needs to be resolved.

Example of a typed callable (this would be for internal usage only anyway):
```php
<?php
declare(strict_types=1);

namespace Slim;

class TypedCallable {
    public const GenericCallable = 0;
    public const RouteCallable = 1;
    public const MiddlewareCallable = 2;

    /**
     * @var callable|string
     */
    private $callable;

    public function __construct(int $type, $callable) {
        $this->type = $type;
        $this->callable = $callable;
    }

    /**
     * @return int
     */
    public function getType(): int
    {
        return $this->type;
    }

    /**
     * @return callable|string
     */
    public function getCallable()
    {
        return $this->callable;
    }

    /**
     * @param $callable
     * @return TypedCallable
     */
    public static function genericCallable($callable): TypedCallable
    {
        return new static(self::GenericCallable, $callable);
    }

    /**
     * @param $callable
     * @return TypedCallable
     */
    public static function routeCallable($callable): TypedCallable
    {
        return new static(self::RouteCallable, $callable);
    }

    /**
     * @param $callable
     * @return TypedCallable
     */
    public static function middlewareCallable($callable): TypedCallable
    {
        return new static(self::MiddlewareCallable, $callable);
    }
}
```

[...]

PR #2787, opened 2019-08-09
4.x - Add an advanced callable resolver

adriansuter:
This PR addresses #2780.

It defines a new `AdvancedCallableResolver` and a corresponding interface which extends the existing `CallableResolverInterface`. The advanced callable resolver would be created by default in the `App` constructor (if none had been passed).

The `MiddlewareDispatcher` would get the callable resolver as third optional param (non-BC). If it has one, and if the resolver is advanced, then the dispatcher would call the `resolveMiddleware` method. If the resolver is not advanced or if there is not even a resolver available, the dispatcher would execute its own hard coded resolving.

The `\Slim\Routing\Route` and the `\Slim\Routing\RouteGroup` would check if the resolver is advanced. If so, they call `resolveRoute` to perform the resolving. Otherwise they do as before.

Note: All the test cases for the callable resolver had been copied and implemented for the advanced callable resolver. Some of them were "triplicated", because two had to throw exceptions and one not.


Slim/MiddlewareDispatcher.php
```diff
@@ -32,15 +34,23 @@ class MiddlewareDispatcher implements RequestHandlerInterface
...
+    /**
+     * @param RequestHandlerInterface        $kernel
+     * @param ContainerInterface|null        $container
+     * @param CallableResolverInterface|null $callableResolver
      */
     public function __construct(
         RequestHandlerInterface $kernel,
-        ?ContainerInterface $container = null
+        ?ContainerInterface $container = null,
+        ?CallableResolverInterface $callableResolver = null
```
l0gicgate: I think that we should make this parameter non-optional.

Slim/MiddlewareDispatcher.php
```diff
@@ -141,53 +151,66 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                    }
 
-                $callable = $instance ?? $resolved;
-                if ($instance && $method) {
-                    $callable = [$instance, $method];
+                    $callable = $instance ?? $resolved;
+                    if ($instance && $method) {
+                        $callable = [$instance, $method];
+                    }
+                } elseif ($this->callableResolver instanceof AdvancedCallableResolverInterface) {
```
l0gicgate: This should be pushed to the top of this function, we should always prioritize the `AdvancedCallableResolver` to resolve first and fallback on the other logic if it's not present.

Slim/AdvancedCallableResolver.php
```diff
@@ -0,0 +1,157 @@
...
+        return [$instance, $method];
+    }
+
+    /**
+     * @param mixed                        $resolved
+     * @param string|object|array|callable $toResolve
+     *
+     * @return callable
+     */
+    private function checkResolvedAndBindContainerIfClosure($resolved, $toResolve): callable
```
l0gicgate: let's change this method name to
```php
private function assertCallableAndBindClosureToContainer($resolved, $toResolve): callable
```

l0gicgate (changes requested):
@adriansuter that's a really good solution so we entirely avoid any BC breaks. It's also opt-in for developers to implement which is awesome. Great work

Slim/AdvancedCallableResolver.php
adriansuter: Sure.

Slim/MiddlewareDispatcher.php
adriansuter: Wouldn't that be a BC?

Slim/MiddlewareDispatcher.php
adriansuter: Agree. Currently, if the callable resolver is given but not an advanced one, then this whole `handle()` method would always throw an exception. Do you think we should, in case the callable resolver is non-advanced, process the hard-coded resolving instead of just throwing the exception?

Slim/MiddlewareDispatcher.php
l0gicgate: Nope since currently the MiddlewareDispatcher was instantiated internally and not accessible to the end user.

Slim/MiddlewareDispatcher.php
l0gicgate: In the case that a non-advanced one is passed that would mean it’s a custom one which means it’s up to them to resolve these objects how they want to. The whole point of this change is to force the dispatcher to use `CallableResolver`. 

What we could do is in case of a non advanced resolver we wrap the `resolve` method in a try catch and fall back on that resolution logic in case that it can’t resolve it. What do you think?

Slim/MiddlewareDispatcher.php
adriansuter: So we need to change the whole signature of the constructor? Into something like
```php
    public function __construct(
        RequestHandlerInterface $kernel,
        CallableResolverInterface $callableResolver,
        ?ContainerInterface $container = null
    ) {
```
Are you sure about that? We will have to change a lot of test cases (as they obviously don't pass a callable resolver so far).

l0gicgate: Oh it will break a lot of internals for sure. No external BCs though.

adriansuter:
Okay, there are a lot of changes in the unit tests. I tried to make sure that they test the non-advanced callable resolver as well as the advanced callable resolver.

Two tests did not work as expected. But only if the callable resolver was non-advanced:
* `\Slim\Tests\MiddlewareDispatcherTest::testDeferredResolvedClosureIsBoundToContainer`
* `\Slim\Tests\MiddlewareDispatcherTest::testFetchesMiddlewareFromContainer`

Any idea why? I think there is something wrong in the `\Slim\MiddlewareDispatcher::addDeferred() -> class::handle()` method. Just couldn't find the bug.

adriansuter:
@l0gicgate Merged your PR. Slim is now using the `\Slim\Interfaces\AdvancedCallableResolverInterface` for the `\Slim\CallableResolver` by default.

l0gicgate (approved):

PR #2789, opened 2019-08-10
Cookie-header detection must be case insensitive

mapogolions:
Follow the [official guide](http://www.slimframework.com/docs/v4/start/installation.html) to create a simple application and reproduce the bug.
```php
// imports
$app->get('/', function (Request $request, Response $response, $args) {
    header('Set-Cookie: name=foo');
    $response = $response->withHeader('Set-Cookie', 'name=bar');
    $response->getBody()->write("Hello world!");
    return $response;
});
$app->run();
```
While everything is fine, the browser receives two headers with the same fields but different values.
```sh
Set-Cookie: name=foo
Set-Cookie: name=bar
```
However, when we start using case insensitive names, the behavior changes.
```php
// imports
$app->get('/', function (Request $request, Response $response, $args) {
    header('Set-Cookie: name=foo');
    $response = $response->withHeader('set-cOOkie', 'name=bar');
    $response->getBody()->write("Hello world!");
    return $response;
});
$app->run();
```
```sh
set-cOOkie: name=bar
```

### DIFF
--- a/Slim/ResponseEmitter.php
+++ b/Slim/ResponseEmitter.php
@@ -58,7 +58,7 @@ class ResponseEmitter
     private function emitHeaders(ResponseInterface $response): void
     {
         foreach ($response->getHeaders() as $name => $values) {
-            $first = $name !== 'Set-Cookie';
+            $first = strtolower($name) !== 'set-cookie';
             foreach ($values as $value) {
                 $header = sprintf('%s: %s', $name, $value);
                 header($header, $first);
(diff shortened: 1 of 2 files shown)


l0gicgate:
That’s not a Slim issue. That’s an issue with the PSR-7 implementation being used.

l0gicgate (approved):

mapogolions:
Why are you using this line of code `$first = $name !== 'Set-Cookie'`. A usual user does't keep in mind the rules for formatting http-headers. Why there should be different results in almost identical cases
- `$response->withHeader('set-cookie', ['name=foo', 'bar=0'])`
- `$response->withHeader('Set-Cookie', ['name=foo', 'bar=0'])`
The built-in function `header` makes no such difference.

adriansuter (approved):

Issue #2778, opened 2019-08-05
4.x - Configure the application via the container?


mnapoli:
Following #2770 I am working on the PHP-DI / Slim integration and I have a question: was it considered to use the container to provide all the services needed to initialize the application? (I mean the route collector, callable resolver, router, etc)

Because at the moment everything is configured via a global static locator (the `AppFactory` class). Given everything is static in there, it is not compatible with the dependency injection (I cannot inject things coming from the container, or using the container). It also means that if I create 2 applications (for example in tests) the configuration from one app will bleed into the next one.

If we made everything come from the container, then building an application would be simpler:

```php
$app = new App($container);
```

Then if you want to overload anything, you can do it by defining the service in the container.

Such behavior would also be more in line with what other frameworks are doing, which would follow the principle of least astonishment.

Anyway, maybe you've already considered it? Would it be too late to change? Maybe not?

adriansuter:
To me this seems plausible. But we have to pay attention, as Slim 4 might be used without container.

mnapoli:
Absolutely, and maybe a "default" container could be a simple PHP array? This is, after all, a container, even though a poor one. (and I'm not talking about reproducing Pimple at all, just a plain simple array with objects, not closures)

But again most users that don't use a container and want to use the defaults will keep doing `new App()` and that's good for them.

l0gicgate:
You’ll just have to create your own factory or extend ours if you want to retain the PSR-17 detection. 

As for everything else it can be worked around by extending the `App` as you were doing before. 

Like @adriansuter said, we want to leave the possibility of not having to use a container at all.

mnapoli:
> You’ll just have to create your own factory or extend ours if you want to retain the PSR-17 detection.

I'm not sure what you mean or how it relates to this issue to be honest. Maybe I missed something?

> As for everything else it can be worked around by extending the App as you were doing before.

That doesn't address the problems I've raised around the global static locator that is now included in Slim 4 AFAICT.

> Like @adriansuter said, we want to leave the possibility of not having to use a container at all.

Agreed, what I suggest should be compatible with that.

l0gicgate:
You need a way to provide `Request` and `Response` for your DI bridge to work. Hence why I’m saying you might need to implement your own set of PSR17 factories in order to be able to inject them in the container (or use middleware) for your Invoker to work. 

[...]

PR #2793, opened 2019-08-10
Configure the application via a container

mnapoli:
This is an implementation for what we discussed in #2778.

The idea is to simplify how to configure the application when using a PSR-11 container.

Here is an example with PHP-DI, using the default services from Slim:

```php
$container = new DI\Container;

$app = AppFactory::createFromContainer($container);
```

Here is an example when we want to overload one service:

```php
$container = new DI\Container;
$container->set(CallableResolverInterface::class, $customInstance);

$app = AppFactory::createFromContainer($container);
```

(in reality I would recommend using a config file, but the example here is simpler)

instead of the current way of doing it:

```php
$container = new DI\Container;

AppFactory::setContainer($container);
AppFactory::setCallableResolver($customResolver);
$app = AppFactory::create($container);
```

The advantages:

- `AppFactory` is no longer storing global state, meaning subsequent calls to `AppFactory::create()` will not be affected by any previous call to `AppFactory`
- containers can come pre-configured for Slim

The 2nd point is interesting for PHP-DI: the bridge could provide a pre-configured container for Slim. Users would just have to call `AppFactory::createFromContainer($container)`, which is pretty cool.


tests/Factory/AppFactoryTest.php
```diff
@@ -228,4 +229,40 @@ public function testResponseAndStreamFactoryIsBeingInjectedInDecoratedResponseFa
...
+    public function testCreateAppWithEmptyContainer()
+    {
+        $container = new MockContainer([]);
+
+        $app = AppFactory::createFromContainer($container);
+
+        $this->assertSame($container, $app->getContainer());
+    }
+
+    public function testCreateAppWithContainerThatProvidesAllImplementations()
```
l0gicgate: Just to keep the test suite cohesive can you use prophecies instead of mocks please? This also removes the need for a `MockContainer`

tests/Mocks/MockContainer.php
```diff
@@ -0,0 +1,26 @@
+<?php declare(strict_types=1);
+
+namespace Slim\Tests\Mocks;
+
+use Psr\Container\ContainerInterface;
+
+class MockContainer implements ContainerInterface
```
l0gicgate: As mentioned in my comment above. There is no need for this mock class. Please use prophecies.

l0gicgate (changes requested):
This is great, thank you for this contribution. I would like to mention that I’d like to work closely with third-party library devs like you to make Slim flexible and sensical in terms of approaching different problems. I highly value your input @mnapoli

tests/Factory/AppFactoryTest.php
mnapoli: 👍 	roger on the prophecy mocks. However the MockContainer class makes everything simpler, the test will be much more complex and brittle here. Here we have an interface let's take advantage from it? :)

mnapoli: Capture d’écran 2019-08-10 à 22 44 08

I'm giving it a try, I got no autocompletion because of prophecy's magic. I don't know these mocks. Do you have any pointer on how to make it return the correct object corresponding to the service name that is passed to `get()`?

Edit: btw I've pushed the fix for the other mocks.

l0gicgate (approved):

l0gicgate:
Looks good to me. Please have a quick look before I merge @adriansuter!

adriansuter (approved):

adriansuter:
@l0gicgate Can be merged.

We need to make sure that this possibility gets its way into the documentation.

l0gicgate:
@adriansuter 

> We need to make sure that this possibility gets its way into the documentation.

See https://github.com/slimphp/Slim-Website/issues/411

PR #2798, opened 2019-08-13
Add BodyParsingMiddleware

akrabat:
This middleware will parse the body of a PSR-7 ServerRequest and if the content-type is known and a parser is registered, it will add the parsed data to the request object before passing to the next middleware. 

Usage:

```php
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;
use Slim\Factory\AppFactory;
use Slim\Middleware\BodyParsingMiddleware;
use Slim\Psr7\Response;

$app = AppFactory::create();
$app->addBodyParsingMiddleware();

$app->post('/foo', function (ServerRequestInterface $request): ResponseInterface {
    $data = $request->getParsedBody();

    $response = new Response();
    $response->getBody()->write(
        print_r($data, true)
    );
    return new Response();
});

$app->run();
```

### DIFF
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -23,8 +23,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     protected $bodyParsers;
 
     /**
-     * Constructor.
-     *
      * @param callable[] $bodyParsers list of body parsers as an associative array of mediaType => callable
      */
     public function __construct(array $bodyParsers = [])
@@ -53,8 +51,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
     /**
-     * Register media type parser.
-     *
      * @param string   $mediaType A HTTP media type (excluding content-type params).
      * @param callable $callable  A callable that returns parsed contents for media type.
      * @return self
@@ -66,8 +62,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
     /**
-     * Is a body parser registered for this media type?
-     *
      * @param string   $mediaType A HTTP media type (excluding content-type params).
      * @return boolean
      */
@@ -77,10 +71,9 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
     /**
-     * Retrieve media type parser.
-     *
      * @param string    $mediaType A HTTP media type (excluding content-type params).
      * @return callable
+     * @throws RuntimeException
      */
     public function getBodyParser(string $mediaType): callable
     {
@@ -91,9 +84,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
 
-    /**
-     * Register the default body parsers
-     */
     protected function registerDefaultBodyParsers(): void
     {
         $this->registerBodyParser('application/json', function ($input) {
@@ -168,8 +158,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
     /**
-     * Get the request's media type, if known.
-     *
      * @param ServerRequestInterface $request
      * @return string|null The serverRequest media type, minus content-type params
      */
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -170,8 +170,6 @@ class BodyParsingMiddleware implements MiddlewareInterface
     /**
      * Get the request's media type, if known.
      *
-     * Note: This method is not part of the PSR-7 standard.
-     *
      * @param ServerRequestInterface $request
      * @return string|null The serverRequest media type, minus content-type params
      */
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -18,14 +18,14 @@ use RuntimeException;
 class BodyParsingMiddleware implements MiddlewareInterface
 {
     /**
-     * @var array
+     * @var callable[]
      */
     protected $bodyParsers;
 
     /**
      * Constructor.
      *
-     * @param array $bodyParsers list of body parsers as an associative array of mediaType => callable
+     * @param callable[] $bodyParsers list of body parsers as an associative array of mediaType => callable
      */
     public function __construct(array $bodyParsers = [])
     {
@@ -94,7 +94,7 @@ class BodyParsingMiddleware implements MiddlewareInterface
     /**
      * Register the default body parsers
      */
-    protected function registerDefaultBodyParsers()
+    protected function registerDefaultBodyParsers(): void
     {
         $this->registerBodyParser('application/json', function ($input) {
             $result = json_decode($input, true);
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -106,42 +106,29 @@ class BodyParsingMiddleware implements MiddlewareInterface
             return $result;
         });
 
-        $this->registerBodyParser('application/xml', function ($input) {
-            $backup = libxml_disable_entity_loader(true);
-            $backup_errors = libxml_use_internal_errors(true);
-            $result = simplexml_load_string($input);
-
-            libxml_disable_entity_loader($backup);
-            libxml_clear_errors();
-            libxml_use_internal_errors($backup_errors);
-
-            if ($result === false) {
-                return null;
-            }
-
-            return $result;
-        });
-
-        $this->registerBodyParser('text/xml', function ($input) {
-            $backup = libxml_disable_entity_loader(true);
-            $backup_errors = libxml_use_internal_errors(true);
-            $result = simplexml_load_string($input);
-
-            libxml_disable_entity_loader($backup);
-            libxml_clear_errors();
-            libxml_use_internal_errors($backup_errors);
-
-            if ($result === false) {
-                return null;
-            }
-
-            return $result;
-        });
-
         $this->registerBodyParser('application/x-www-form-urlencoded', function ($input) {
             parse_str($input, $data);
             return $data;
         });
+
+        $xmlCallable = function ($input) {
+            $backup = libxml_disable_entity_loader(true);
+            $backup_errors = libxml_use_internal_errors(true);
+            $result = simplexml_load_string($input);
+
+            libxml_disable_entity_loader($backup);
+            libxml_clear_errors();
+            libxml_use_internal_errors($backup_errors);
+
+            if ($result === false) {
+                return null;
+            }
+
+            return $result;
+        };
+
+        $this->registerBodyParser('application/xml', $xmlCallable);
+        $this->registerBodyParser('text/xml', $xmlCallable);
     }
 
     /**
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -77,7 +77,7 @@ class BodyParsingMiddleware implements MiddlewareInterface
     }
 
     /**
-     * Retreive media type parser.
+     * Retrieve media type parser.
      *
      * @param string    $mediaType A HTTP media type (excluding content-type params).
      * @return callable
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -9,11 +9,11 @@ declare(strict_types=1);
 
 namespace Slim\Middleware;
 
-use RuntimeException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use RuntimeException;
 
 class BodyParsingMiddleware implements MiddlewareInterface
 {
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -1,0 +1,202 @@
+<?php
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
+namespace Slim\Middleware;
+
+use RuntimeException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class BodyParsingMiddleware implements MiddlewareInterface
+{
+    /**
+     * @var array
+     */
+    protected $bodyParsers;
+
+    /**
+     * Constructor.
+     *
+     * @param array $bodyParsers list of body parsers as an associative array of mediaType => callable
+     */
+    public function __construct(array $bodyParsers = [])
+    {
+        $this->registerDefaultBodyParsers();
+
+        foreach ($bodyParsers as $mediaType => $parser) {
+            $this->registerBodyParser($mediaType, $parser);
+        }
+    }
+
+    /**
+     * @param ServerRequestInterface  $request
+     * @param RequestHandlerInterface $handler
+     * @return ResponseInterface
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $parsedBody = $request->getParsedBody();
+        if ($parsedBody === null || empty($parsedBody)) {
+            $parsedBody = $this->parseBody($request);
+            $request = $request->withParsedBody($parsedBody);
+        }
+
+        return $handler->handle($request);
+    }
+
+    /**
+     * Register media type parser.
+     *
+     * @param string   $mediaType A HTTP media type (excluding content-type params).
+     * @param callable $callable  A callable that returns parsed contents for media type.
+     * @return self
+     */
+    public function registerBodyParser(string $mediaType, callable $callable): self
+    {
+        $this->bodyParsers[$mediaType] = $callable;
+        return $this;
+    }
+
+    /**
+     * Is a body parser registered for this media type?
+     *
+     * @param string   $mediaType A HTTP media type (excluding content-type params).
+     * @return boolean
+     */
+    public function hasBodyParser(string $mediaType): bool
+    {
+        return isset($this->bodyParsers[$mediaType]);
+    }
+
+    /**
+     * Retreive media type parser.
+     *
+     * @param string    $mediaType A HTTP media type (excluding content-type params).
+     * @return callable
+     */
+    public function getBodyParser(string $mediaType): callable
+    {
+        if (!isset($this->bodyParsers[$mediaType])) {
+            throw new RuntimeException('No parser for type ' . $mediaType);
+        }
+        return $this->bodyParsers[$mediaType];
+    }
+
+
+    /**
+     * Register the default body parsers
+     */
+    protected function registerDefaultBodyParsers()
+    {
+        $this->registerBodyParser('application/json', function ($input) {
+            $result = json_decode($input, true);
+
+            if (!is_array($result)) {
+                return null;
+            }
+
+            return $result;
+        });
+
+        $this->registerBodyParser('application/xml', function ($input) {
+            $backup = libxml_disable_entity_loader(true);
+            $backup_errors = libxml_use_internal_errors(true);
+            $result = simplexml_load_string($input);
+
+            libxml_disable_entity_loader($backup);
+            libxml_clear_errors();
+            libxml_use_internal_errors($backup_errors);
+
+            if ($result === false) {
+                return null;
+            }
+
+            return $result;
+        });
+
+        $this->registerBodyParser('text/xml', function ($input) {
+            $backup = libxml_disable_entity_loader(true);
+            $backup_errors = libxml_use_internal_errors(true);
+            $result = simplexml_load_string($input);
+
+            libxml_disable_entity_loader($backup);
+            libxml_clear_errors();
+            libxml_use_internal_errors($backup_errors);
+
+            if ($result === false) {
+                return null;
+            }
+
+            return $result;
+        });
+
+        $this->registerBodyParser('application/x-www-form-urlencoded', function ($input) {
+            parse_str($input, $data);
+            return $data;
+        });
+    }
+
+    /**
+     * @param ServerRequestInterface $request
+     * @return null|array|object
+     */
+    protected function parseBody(ServerRequestInterface $request)
+    {
+        $mediaType = $this->getMediaType($request);
+        if ($mediaType === null) {
+            return null;
+        }
+
+        // Check if this specific media type has a parser registered first
+        if (!isset($this->bodyParsers[$mediaType])) {
+            // If not, look for a media type with a structured syntax suffix (RFC 6839)
+            $parts = explode('+', $mediaType);
+            if (count($parts) >= 2) {
+                $mediaType = 'application/' . $parts[count($parts) - 1];
+            }
+        }
+
+        if (isset($this->bodyParsers[$mediaType])) {
+            $body = (string)$request->getBody();
+            $parsed = $this->bodyParsers[$mediaType]($body);
+
+            if (!is_null($parsed) && !is_object($parsed) && !is_array($parsed)) {
+                throw new RuntimeException(
+                    'Request body media type parser return value must be an array, an object, or null'
+                );
+            }
+
+            return $parsed;
+        }
+
+        return null;
+    }
+
+    /**
+     * Get the request's media type, if known.
+     *
+     * Note: This method is not part of the PSR-7 standard.
+     *
+     * @param ServerRequestInterface $request
+     * @return string|null The serverRequest media type, minus content-type params
+     */
+    protected function getMediaType(ServerRequestInterface $request): ?string
+    {
+        $contentType = $request->getHeader('Content-Type')[0] ?? null;
+
+        if (is_string($contentType) && trim($contentType) != '') {
+            $contentTypeParts = explode(';', $contentType);
+            return strtolower(trim($contentTypeParts[0]));
+        }
+
+        return null;
+    }
+}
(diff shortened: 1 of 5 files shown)


akrabat:
Rebased to tip of `4.x` branch

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
+namespace Slim\Middleware;
+
+use RuntimeException;
```
adriansuter: If the use statements should be in alphabetical order, then `RuntimeException` should be the last one.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+{
+    /**
+     * @var array
+     */
+    protected $bodyParsers;
+
+    /**
+     * Constructor.
+     *
+     * @param array $bodyParsers list of body parsers as an associative array of mediaType => callable
```
adriansuter: As the array consists of `callable` elements only, we could actually type-hint the param as `callable[]`. The same holds for the `$bodyParsers` property.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+     * @param string   $mediaType A HTTP media type (excluding content-type params).
+     * @return boolean
+     */
+    public function hasBodyParser(string $mediaType): bool
+    {
+        return isset($this->bodyParsers[$mediaType]);
+    }
+
+    /**
+     * Retreive media type parser.
```
adriansuter: Typo "Retrieve"?

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+            throw new RuntimeException('No parser for type ' . $mediaType);
+        }
+        return $this->bodyParsers[$mediaType];
+    }
+
+
+    /**
+     * Register the default body parsers
+     */
+    protected function registerDefaultBodyParsers()
```
adriansuter: We should type-hint the return.
```php
protected function registerDefaultBodyParsers(): void
```

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+                return null;
+            }
+
+            return $result;
+        });
+
+        $this->registerBodyParser('application/xml', function ($input) {
+            $backup = libxml_disable_entity_loader(true);
+            $backup_errors = libxml_use_internal_errors(true);
+            $result = simplexml_load_string($input);
```
adriansuter: If we are using these functions we should add the following lines to `composer.json` in the `require` section:
```json
        "ext-libxml": "*",
        "ext-simplexml": "*"
```

Then `ext-simplexml` can be removed from the `require-dev` section.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+            libxml_use_internal_errors($backup_errors);
+
+            if ($result === false) {
+                return null;
+            }
+
+            return $result;
+        });
+
+        $this->registerBodyParser('text/xml', function ($input) {
```
adriansuter: As the callables for `application/xml` and `text/xml` are doing the same thing, we could simplify
```php
        $xmlCallable = function ($input) {
            $backup = libxml_disable_entity_loader(true);
            $backup_errors = libxml_use_internal_errors(true);
            $result = simplexml_load_string($input);

            libxml_disable_entity_loader($backup);
            libxml_clear_errors();
            libxml_use_internal_errors($backup_errors);

            if ($result === false) {
                return null;
            }

            return $result;
        };

        $this->registerBodyParser('application/xml', $xmlCallable);

        $this->registerBodyParser('text/xml', $xmlCallable);
```

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+            return $parsed;
+        }
+
+        return null;
+    }
+
+    /**
+     * Get the request's media type, if known.
+     *
+     * Note: This method is not part of the PSR-7 standard.
```
adriansuter: Does this "Note" make any sense here?

tests/Middleware/BodyParsingMiddlewareTest.php
```diff
@@ -0,0 +1,210 @@
...
+                return 10; // invalid - should return null, array or object
+            },
+        ];
+        $middleware = new BodyParsingMiddleware($parsers);
+
+        $this->expectException(RuntimeException::class);
+        $middleware->process($request, $this->createRequestHandler());
+    }
+
+    public function testSettingAndGetttingAParser()
```
adriansuter: Typo? Getting has tripple `t`.

adriansuter (changes requested):
Good work. Some minor change requests.

Slim/Middleware/BodyParsingMiddleware.php
akrabat: I'm not sure about this. `BodyParsingMiddleware` is optional and if not used, then you don't need the xml extensions to run Slim.

Slim/Middleware/BodyParsingMiddleware.php
akrabat: No - copy/paste error from Slim-Http :)

Slim/Middleware/BodyParsingMiddleware.php
froschdesign: If a package or extension is not strictly required then use the `suggest` section: https://getcomposer.org/doc/04-schema.md#suggest

akrabat: Added to the `suggest` section.

akrabat:
@adriansuter Review items addressed.

adriansuter (approved):
Wonderful.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Retrieve media type parser.
+     *
+     * @param string    $mediaType A HTTP media type (excluding content-type params).
+     * @return callable
+     */
+    public function getBodyParser(string $mediaType): callable
+    {
+        if (!isset($this->bodyParsers[$mediaType])) {
+            throw new RuntimeException('No parser for type ' . $mediaType);
```
l0gicgate, line 81: Can you add `@throws RuntimeException` to the doc bloc

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+        if ($parsedBody === null || empty($parsedBody)) {
+            $parsedBody = $this->parseBody($request);
+            $request = $request->withParsedBody($parsedBody);
+        }
+
+        return $handler->handle($request);
+    }
+
+    /**
+     * Register media type parser.
```
l0gicgate: Let's remove this comment. The method name speaks for itself.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+     * @return self
+     */
+    public function registerBodyParser(string $mediaType, callable $callable): self
+    {
+        $this->bodyParsers[$mediaType] = $callable;
+        return $this;
+    }
+
+    /**
+     * Is a body parser registered for this media type?
```
l0gicgate: Same here. The method name speaks for itself.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+     * @param string   $mediaType A HTTP media type (excluding content-type params).
+     * @return boolean
+     */
+    public function hasBodyParser(string $mediaType): bool
+    {
+        return isset($this->bodyParsers[$mediaType]);
+    }
+
+    /**
+     * Retrieve media type parser.
```
l0gicgate: And same here. The method name speaks for itself.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+    {
+        if (!isset($this->bodyParsers[$mediaType])) {
+            throw new RuntimeException('No parser for type ' . $mediaType);
+        }
+        return $this->bodyParsers[$mediaType];
+    }
+
+
+    /**
+     * Register the default body parsers
```
l0gicgate: Let's remove this comment as well.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            return $parsed;
+        }
+
+        return null;
+    }
+
+    /**
+     * Get the request's media type, if known.
```
l0gicgate: Remove this comment.

tests/Middleware/BodyParsingMiddlewareTest.php
```diff
@@ -0,0 +1,210 @@
...
+            public function handle(ServerRequestInterface $request): ResponseInterface
+            {
+                $this->response->request = $request;
+                return $this->response;
+            }
+        };
+    }
+
+    /**
+     * Helper function to create a ServerRequest with a body and a content-type header
```
l0gicgate: Let's remove this comment

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,187 @@
...
+
+class BodyParsingMiddleware implements MiddlewareInterface
+{
+    /**
+     * @var callable[]
+     */
+    protected $bodyParsers;
+
+    /**
+     * Constructor.
```
l0gicgate: Can we remove this from the doc bloc, it's implied that it's the constructor. I want to keep comments cohesive throughout the code base. Only comments that are absolutely necessary should be present, which in most cases they are not if your code is explicit.

l0gicgate (changes requested):
Minor changes. Mainly just doc bloc stuff. This is a great addition to our middleware suite.

akrabat:
@l0gicgate Review items addressed.

shadowhand:
@akrabat what about adding `App::addBodyParsers(array $parsers = [])` like we have for error handling, etc.

akrabat:
> @akrabat what about adding `App::addBodyParsers(array $parsers = [])` like we have for error handling, etc.

A key reason that `addErrorMiddleware()` exists is that there's a number of dependencies to the constructor that the app instance can fill in for you.

That's not the case for this one, so we'd be adding it for the dev experience of:

```php
$app->addErrorMiddleware(true, true, true);
$app->addBodyParsing();
```

looking nicer than:

```php
$app->addErrorMiddleware(true, true, true);
$app->add(new Slim\Middleware\BodyParsingMiddleware());
```

I'm easy either way and defer to @l0gicgate on it.

shadowhand:
Before this is merged, I would like to give a more complete review of it. I think it can be improved and be made more friendly to IoC.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,175 @@
...
+        if (!isset($this->bodyParsers[$mediaType])) {
+            // If not, look for a media type with a structured syntax suffix (RFC 6839)
+            $parts = explode('+', $mediaType);
+            if (count($parts) >= 2) {
+                $mediaType = 'application/' . $parts[count($parts) - 1];
+            }
+        }
+
+        if (isset($this->bodyParsers[$mediaType])) {
+            $body = (string)$request->getBody();
```
anfly0, line 145: What if the stream is not seekable (rewindable)?
This would be a problem if any subsequent middleware, for whatever reason, needed to access the raw body.
An example would be authentication via HMAC.

One solution would be to simply copy the non-seekable stream to a new seekable stream.

akrabat: > What if the stream is not seekable (rewindable)?
> This would be a problem if any subsequent middleware, for whatever reason, needed to access the raw body.
> An example would be authentication via HMAC.

`__toString()` will rewind (if it can) and leaves the pointer back where it was before the call, so the stream won't look any different for subsequent middleware. (If it can't rewind, then nothing's changed of course.)


> One solution would be to simply copy the non-seekable stream to a new seekable stream.

If the stream is not rewindable, then there's no guarantee that the data at the beginning is anywhere in memory now, so I don't see any advantage to copying to seekable stream?

At the end of the day,  I think that if you're using a non-rewindable string in your ServerRequest, then you're incompatible with getParsedBody() and won't need or want this middleware.

anfly0: > 
> `__toString()` will rewind (if it can) and leaves the pointer back where it was before the call, so the stream won't look any different for subsequent middleware. (If it can't rewind, then nothing's changed of course.)

According to PSR-7 it is not a requirement to attempt to rewind the stream __after__ the data has been read by ` __to_string()`.  PSR-7 only requires the stream to be rewound __before__ the data is read. 
 
> If the stream is not rewindable, then there's no guarantee that the data at the beginning is anywhere in memory now, so I don't see any advantage to copying to seekable stream?
> 
> At the end of the day, I think that if you're using a non-rewindable string in your ServerRequest, then you're incompatible with getParsedBody() and won't need or want this middleware.

Well most often you will see something like `$streamFactory->createStreamFromFile('php://input')` when creating the body stream. 
The problem is that there is no guarantee `php://input` is seekable.
So the only real solution if the stream is not seekable and the raw data needs to access more than once is to copy the stream.

akrabat: > Well most often you will see something like `$streamFactory->createStreamFromFile('php://input')` when creating the body stream.
> The problem is that there is no guarantee `php://input` is seekable.
> So the only real solution if the stream is not seekable and the raw data needs to access more than once is to copy the stream.

Right. I see what you're saying now. The problem is that we can't copy the stream as there's no method in `StreamInterface` that provide access to the underlying stream object.

Thoughts?

akrabat: A further note. It looks like this has to be solved at the PSR-7 implementation level. e.g. Diactoros's [PhpInputStream](https://github.com/zendframework/zend-diactoros/blob/master/src/PhpInputStream.php) which is then used by [ServerRequest](https://github.com/zendframework/zend-diactoros/blob/master/src/ServerRequest.php#L95-L97).

l0gicgate:
@akrabat 

> $app->addBodyParsing();

I'd maybe rename that to `addBodyParsingMiddleware()`, I'd like to keep that suffix for all helper methods.

I mean theoretically we should probably do a helper for all internal middleware:
- ErrorMiddleware - Already Done
- RoutingMiddleware - Already Done
- BodyParsingMiddleware - We should add it
- ContentLengthMiddleware - We should add it
- MethodOverrideMiddleware - We should add it

What do you think @adriansuter

Slim/Middleware/BodyParsingMiddleware.php
l0gicgate: @akrabat this seems like such an oversight from the PSR-7 spec. It makes for unpredictable behavior. As discussed, we should leverage caching on `Slim-Psr7` to mitigate some of this unpredictability.

anfly0: @akrabat My thinking is that the simplest way to copy the stream would be something like:
```
$oldStream = $request->getBody();
while (!$oldStream->eof()) {
     $newStream->write($oldStream->read(4096));
}
$request = $request->withBody($newStream);
```
Maybe not the cleanest, but reasonably efficient.

adriansuter:
As long as there aren't dozens of internal middleware, why not. This would make things definitely more developer friendly.

But we need to pay attention not to blow up our `App` class.

Unfortunately this works only for internal middleware. So we would mix two "styles" of adding middleware. I'm not too happy with that, as I would prefer to make things consistent. But that is some sort of compromise, I guess.

Slim/Middleware/BodyParsingMiddleware.php
akrabat: > @akrabat My thinking is that the simplest way to copy the stream would be something like:
> 
> ```
> $oldStream = $request->getBody();
> while (!$oldStream->eof()) {
>      $newStream->write($oldStream->read(4096));
> }
> $request = $request->withBody($newStream);
> ```
> 
> Maybe not the cleanest, but reasonably efficient.

(Update: ignore that comment - you’re not cloning)

How do we create $newStream and how do we do know what type of stream resource $oldStream has inside it?

If it’s php://input, we can’t read it twice, so we’d have to convert to a string and then create a $newStream that was no longer a php://input, but now a php://temp or something.

Diactoros solves this with a special Stream class for php://input, which seems quite locgical. I’m going to see if I can do something for Slim’s Psr7 implementation .

juliangut:
> * ErrorMiddleware - Already Done
> 
> * RoutingMiddleware - Already Done
> 
> * BodyParsingMiddleware - We should add it
> 
> * ContentLengthMiddleware - We should add it
> 
> * MethodOverrideMiddleware - We should add it

Why not basic auth, cors, trailing slash, http cache or gzip/deflate encoding middlewares? I mean routing and error must of course be bundled with Slim but apart from those, there are already great PSR15 middleware out there for all those things

I was surprised yesterday when I saw this PR, I guess it is because you normally receive questions about how to support X payload, but that can be documented

Just my thoughts if they are of any use

akrabat:
> Why not basic auth, cors, trailing slash, http cache or gzip/deflate encoding middlewares? I mean routing and error must of course be bundled with Slim but apart from those, there are already great PSR15 middleware out there for all those things

The ones that @l0gicgate listed all functionality that was built into Slim 3. With Slim 4 we have moved them into separate components so that the user can now choose to replace with more appropriate solutions for their use-case. 

> I was surprised yesterday when I saw this PR, I guess it is because you normally receive questions about how to support X payload, but that can be documented

This one has come about because the solution we thought we had with Slim-Http does't work as well as we'd like with PSR-15, so this PR addresses that gap.

I don't think we're against supplying other middleware as optional components if we feel that it's something that Slim should come with out of the box and there's not a wide-used, simple-to-use option out there already.

Slim/Middleware/BodyParsingMiddleware.php
anfly0: The only interesting thing about the underlying stream, in this case, is if it is seekable or not.
This can of course easily be checked with `$stream->isSeekable()`.
If it is not seekable, copy all the content to a new seekable stream (based on php//temp).
This approach would allow the PSR-7 implementation to use any resource when creating the request object.

My final thoughts is that:
As long as PSR-7 doesn't mandate that the stream is seekable, any middleware touching the stream will have to check if the stream is seekable and make sure that the stream is rewound before passing it to the handler.
In this way, the middleware is not coupled to any specific PSR-7 implementation.

l0gicgate:
@adriansuter 

> I would prefer to make things consistent

I'm hoping that this will be it for our internal middleware suite. I like consistency as well. So with that line of thinking I guess we should add helper methods for all of the internal middleware.

@akrabat 

We can do that in a different PR though.

Slim/Middleware/BodyParsingMiddleware.php
l0gicgate: @anfly0 but what if the buffer is bigger than 4096 bytes? We need a better solution or just leave it as is.

akrabat: > We need a better solution or just leave it as is.

My personal preference is to put caching into Slim-Psr7's Stream and turn it on for php://input. I have a proof of concept for this working locally and it seems to do what I expect.

akrabat:
> We can do that in a different PR though.

I've added App::addBodyParsingMiddleware() to this PR. I agree that separate PRs for the other middleware makes sense.

shadowhand:
I'm not sure if I am 👍 on this, because as @juliangut says, [middlewares/payload][1] already exists and works for any PSR-15 system.

[1]: https://packagist.org/packages/middlewares/payload

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,175 @@
...
+
+        return $handler->handle($request);
+    }
+
+    /**
+     * @param string   $mediaType A HTTP media type (excluding content-type params).
+     * @param callable $callable  A callable that returns parsed contents for media type.
+     * @return self
+     */
+    public function registerBodyParser(string $mediaType, callable $callable): self
```
shadowhand, line 58: Rather than a generic `callable`, why not create an interface?

```php
interface BodyParser
{
    public function canParse(string $contentType): bool;

    /**
     * @return mixed
     */
    public function parse(StreamInterface $stream);
}
```

Then you can do:

```php
$middleware = new BodyParsingMiddleware();
$middleware->register(new JsonParser());
$middleware->register(new XmlParser());
// etc
```

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,175 @@
...
+        if (!isset($this->bodyParsers[$mediaType])) {
+            throw new RuntimeException('No parser for type ' . $mediaType);
+        }
+        return $this->bodyParsers[$mediaType];
+    }
+
+
+    protected function registerDefaultBodyParsers(): void
+    {
+        $this->registerBodyParser('application/json', function ($input) {
```
shadowhand, line 89: This is fine for basic applications, but what about users of JSON-API that use `application/vnd.api+json` as the content type? The parsing is exactly the same but this hard lock on `application/json` won't work.

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -0,0 +1,175 @@
...
+        if ($mediaType === null) {
+            return null;
+        }
+
+        // Check if this specific media type has a parser registered first
+        if (!isset($this->bodyParsers[$mediaType])) {
+            // If not, look for a media type with a structured syntax suffix (RFC 6839)
+            $parts = explode('+', $mediaType);
+            if (count($parts) >= 2) {
+                $mediaType = 'application/' . $parts[count($parts) - 1];
```
shadowhand, line 140: Okay, I get that this is supposed to handle `application/vnd.api+json` using `application/json` parser... that behavior is not particularly obvious though, and would be better (IMO) handled on a case by case basis in a specific parser.

shadowhand (commented):
As @juliangut says, [middlewares/payload][1] already exists and works for any PSR-15 system.

At minimum, I feel like this kind of solution would be better published as a separate package (or additional functionality contributed to `middlewares/payload`) that anyone can use with PSR-15 dispatchers. To me this largely feels like "not invented here" and, while pretty easy to use for simple applications, is not particularly flexible.

[1]: https://packagist.org/packages/middlewares/payload

Slim/Middleware/BodyParsingMiddleware.php
l0gicgate: I don't disagree with you on this. This to me feels like it should be extracted in its own package if we're going to go that route so we can contain the files contextually. The middleware we ship right now are self-contained into one file.

Slim/Middleware/BodyParsingMiddleware.php
anfly0: @l0gicgate That's what the while loop is for. It will copy 4096 byte chunks until $oldStream->eof() is true. This is not a very uncommon way to copy stream-like objects.
Example:
https://gist.github.com/anfly0/f23bb8418d5309492d47ec4b463ad1df

anfly0: @akrabat Regardless of when and where the data gets copied, I think it's a good idea to only copy the data if the stream is not seekable. My _feeling_ is that copying the stream can be rather expensive and should be avoided if possible.

l0gicgate: @anfly0 my bad I didn't see the while, oversight on my part.

akrabat: > @akrabat Regardless of when and where the data gets copied, I think it's a good idea to only copy the data if the stream is not seekable. My _feeling_ is that copying the stream can be rather expensive and should be avoided if possible.

I’ll have a play & see what I can do.

akrabat: @anfly0 I'm confused trying to do this.

If the stream is not seeable, then we can't rewind it and hence can't parse it in this middleware. Have I missed something?

akrabat: So maybe I copy the stream if it's not seekable _and_ `$stream->tell()` is `0`?

anfly0: @akrabat I'm no 100 on this, but I think that `$stream->tell()` can return 0 on a not seekable stream even if some of the data has been read.
Offsets have no real meaning if the stream is not seekable.

One idea is to copy the stream if it's not seekable and then compare the size of the new stream with what's in the Content-Length header.
If the match it's reasonably safe to assume that all data was copied and if they don't some error handling can be done.  
And then parse the contents.

akrabat: How do we get a non-seeable stream into a `ServerRequest`?

l0gicgate (approved):

akrabat:
@l0gicgate I'm happy with this as the first cut. The only potential outstanding issue is related to non-seekable streams and I'd like to see a real use-case for a non-seeable stream in a `ServerRequest` before try to solve it. The most common case is `php://input` which reports as seekable, but isn't really for PUT, so that needs to be handled at the PSR-7 implementation level..

Slim/Middleware/BodyParsingMiddleware.php
anfly0: I'm not sure I understand the question.
The PSR-7 implementation used is free to use whatever resource or method to create the stream.
It is an edge case, but it is entirely possible for the stream not to be seekable.

esetnik:
Are we supposed to use `$app->addBodyParsingMiddleware();` if we are also using `slim-http`? It seems this functionality is already included in `slim-http` so maybe it can be removed now that it's available in slim itself? See https://github.com/slimphp/Slim-Http/issues/99

l0gicgate:
@esetnik no need to use it no. And we are not removing it as you may use Slim-Http without using Slim

akrabat:
It is beneficial to have it in Slim-Http to help migration from Slim 3.

PR #2801, opened 2019-08-13
4.x - Add setLogErrorRenderer() to ErrorHandler

### LINKED ISSUE #2797: 4.x - Add pseudo content type for log renderer in ErrorHandler
The `ErrorHandler` is currently instantiating a `PlainTextErrorRenderer` in the `writeToErrorLog`-method. So a user would not be able to provide a custom log renderer.

I suggest to add a new pseudo content type "slim-log" to the array of error renderers or to add a new method `setLogErrorRenderer()` which takes one param `ErrorRendererInterface|string|callable`.

I prefer the second solution, as the first one might be a little bit hackish.

What are your thoughts?

RyanNerd:
Please review. If the logic looks good I'll add unit tests for this feature.

Closes #2797

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -281,9 +281,8 @@ class ErrorHandler implements ErrorHandlerInterface
      */
     protected function writeToErrorLog(): void
     {
-        /** @var ErrorRendererInterface $renderer */
-        $renderer = $this->callableResolver->resolve($this->logErrorRenderer);
-        $error = $renderer->__invoke($this->exception, $this->logErrorDetails);
+        $renderer = $this->logErrorRenderer;
+        $error = $renderer($this->exception, $this->logErrorDetails);
         $error .= "\nView in rendered output by enabling the \"displayErrorDetails\" setting.\n";
         $this->logError($error);
     }
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -54,6 +54,11 @@ class ErrorHandler implements ErrorHandlerInterface
     ];
 
     /**
+     * @var ErrorRendererInterface|string|callable
+     */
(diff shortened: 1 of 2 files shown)


Slim/Handlers/ErrorHandler.php
```diff
@@ -259,14 +264,25 @@ public function setDefaultErrorRenderer(string $contentType, $errorRenderer): vo
...
      * Write to the error log if $logErrors has been set to true
      *
      * @return void
      */
     protected function writeToErrorLog(): void
     {
-        $renderer = new PlainTextErrorRenderer();
+        /** @var ErrorRendererInterface $renderer */
+        $renderer = $this->callableResolver->resolve($this->logErrorRenderer);
         $error = $renderer->__invoke($this->exception, $this->logErrorDetails);
```
l0gicgate: `$this->callableResolver->resolve($this->logErrorRenderer)` will return a `callable`. You'll need to use `$renderer(...)` instead of calling `__invoke()` directly.

If you rant the test suite locally, you could have seen the failure. The travis-ci build fails

RyanNerd: @l0gicgate Sorry, a misunderstanding on my part. I've fixed this and pushed -- travis-ci build no longer fails.

RyanNerd: @l0gicgate  I'm having trouble getting the test suite working locally, but I should have paid attention to the failed travis-ci. In any case this has been fixed and no longer fails travis-ci.

adriansuter:
The logic looks good to me. We should add unit tests.

l0gicgate:
@adriansuter definitely. @RyanNerd do you need help with writing the tests?

RyanNerd:
@l0gicgate I need to find some time to write the unit tests. Things got busy for me. I'll see if I can get something out there in the next 7 days.

l0gicgate:
@RyanNerd okay no problem. I might just get antsy and write the tests for you if I have 5 mins 😂

PR #2803, opened 2019-08-15
Emit non seekable streams

mapogolions:
Every time when we emit any response we read all data from a stream twice.
- https://github.com/slimphp/Slim/blob/4.x/Slim/ResponseEmitter.php#L37
- https://github.com/slimphp/Slim/blob/4.x/Slim/ResponseEmitter.php#L49

There may be problems with slow streams or responses with too big body. In addition, with a such `ResponseEmitter` work, it is impossible to send a response with non seekable body.

### DIFF
--- a/Slim/ResponseEmitter.php
+++ b/Slim/ResponseEmitter.php
@@ -133,7 +133,11 @@ class ResponseEmitter
         if (in_array($response->getStatusCode(), [204, 205, 304], true)) {
             return true;
         }
-        $contents = (string) $response->getBody();
-        return !strlen($contents);
+        $stream = $response->getBody();
+        $seekable = $stream->isSeekable();
+        if ($seekable) {
+            $stream->rewind();
+        }
+        return $seekable ? $stream->read(1) === '' : $stream->eof();
     }
 }
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -1564,6 +1564,7 @@ class AppTest extends TestCase
             $body .= $args[0];
             $this->__toString()->willReturn($body);
         });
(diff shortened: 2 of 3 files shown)


Slim/ResponseEmitter.php
```diff
@@ -133,7 +133,11 @@ public function isResponseEmpty(ResponseInterface $response): bool
...
             return true;
         }
-        $contents = (string) $response->getBody();
-        return !strlen($contents);
+        $stream = $response->getBody();
+        $seekable = $stream->isSeekable();
+        if ($seekable) {
+            $stream->rewind();
+        }
+        return $seekable ? $stream->read(1) === '' : $stream->eof();
```
shadowhand, line 141: Could this be made easier with ... ?

```php
return $stream->getSize() > 0;
```

mapogolions: Your logic condition need to invert. `return !$stream->getSize() > 0`. Please, run unit tests with your code. You will see some errors (AppTest errors can be ignored). The most interesting is the error `Slim\Tests\ResponseEmitterTest::testRespondIndeterminateLength`. It shows that reading from a stream of at least one byte is the most reliable way. I myself don’t like such excessive complexity.

l0gicgate (approved):

Slim/ResponseEmitter.php
mapogolions: I found yet another example. Resource `$resource = popen('echo 12', 'r')` will be defined as empty, but in reality has content.

adriansuter (approved):

Issue #2512, opened 2018-10-02
4.x - Base path handling


akrabat:
We need to have base path handling working consistently without confusion in Slim 4.

See also: https://github.com/slimphp/Slim/issues/1964

l0gicgate:
This should be moved to Slim-Psr7 repo along with #1964

akrabat:
This essentially means we should support running Slim in a subdirectory.

i.e. Slim should work for a URL like _www.example.com/myapp/hello_  where the route in Slim is `$app->get('/hello', HelloAction::class');`

I would like this to be PSR-7 implementation independent.

tuupola:
Slim already works when running from a subdirectory. I do it all the time. Just need to set the `RewriteBase` in `.htaccess` file. For example for the above example it would be:

```
RewriteBase /myapp/
```

l0gicgate:
@tuupola wouldn't it be better to make it server environment independent somehow? We could maybe add a setting to `App` which appends a path prefix to all the routes when they're mapped maybe?

tuupola:
@l0gicgate I don't have any strong feelings about that. 

However one cannot get rid of `RewriteBase`. It is not about prefixing Slim urls. It is about telling Apache where to find the `index.php` file.

Lets assume the document root is `/var/www/public` and Slim app is in subfolder `foo`. 

```
/var/www/public
/var/www/public/foo
/var/www/public/foo/.htaccess
/var/www/public/foo/index.php
```

```php
$app = new \Slim\App;
$app->get("/hello", function ($request, $response, $args) {
    print "Hello!";
});
$app->run();
```

```
RewriteEngine On
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^ index.php [QSA,L]
```

Without `RewriteBase` when request is made to https://example.com/foo/hello Apache tries to access `/var/www/public/index.php` which is wrong. You will get a 404 response from Apache with something like:

```
The requested URL /var/www/public/index.php was not found on this server.
```

With `RewriteBase` when request is made to https://example.com/foo/hello Apache accesses `/var/www/public/foo/index.php` which is correct and you will get a response from Slim.

```
RewriteEngine On
RewriteBase /foo/
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^ index.php [QSA,L]
```

[...]

PR #2805, opened 2019-08-16
Base path handling

bigtunacan:
Attempt at fixing base path handling for Slim projects that live in
sub-directories.  Based on work by @esetnik.

Maintains backwards compatibility for users that don't need this.

Relates to issues #2512 and #2770

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -105,11 +105,12 @@ class App extends RouteCollectorProxy implements RequestHandlerInterface
      *
      * @return RoutingMiddleware
      */
-    public function addRoutingMiddleware(): RoutingMiddleware
+    public function addRoutingMiddleware(?string $basePath = null): RoutingMiddleware
     {
         $routingMiddleware = new RoutingMiddleware(
             $this->getRouteResolver(),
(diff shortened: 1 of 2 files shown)


adriansuter:
The route collector does actually have the base path. Did you try `$app->setBasePath(..)`?

l0gicgate:
@bigtunacan thank you for attempting a solution but this won't be the route we would be taking to fix this. Also I think there's two different issues here:

You can achieve what you want by:
- Rewriting the base path via Apache config
- Use `$app->getRouteCollector()->setBasePath()` to append the base path to every route that's being mapped.

Let's defer back to the issue thread as again, I think this is a config issue and not a Slim issue.

piotr-cz:
@l0gicgate 
That solution doesn't work: `$app->getRouteCollector()` returns `\Slim\Routing\RouteCollector` while for routing is used `\FastRoute\RouteCollector`

piotr-cz:
Update:
using `$app->getRouteCollector()->setBasePath($basePath)` works in Slim 4.1 but not in 4.2 anymore

Issue #2804, opened 2019-08-16
4.x - Deferred callable as string without resolver


adriansuter:
Dear reader

In `DeferredCallable` we can pass a `string` as callable. But the callable resolver is allowed to be `null` and therefore the `__invoke` will fail. Shouldn't we throw a `RuntimeException` if that happens? If yes, already in the constructor or later in the invocation?

mapogolions:
I briefly looked - https://github.com/slimphp/Slim/blob/4.x/Slim/DeferredCallable.php#L40. It seems correct. Please, give a counterexample.

```php
function add($a, $b) {
    return $a + $b;
}
$callable = new DeferredCallable('add');
echo $callable(10, 20) . PHP_EOL;
```

adriansuter:
For example if the *callable string* is in Slim notation `ClassName:method`.
```php
class Calculator
{
    function add($a, $b)
    {
        return $a + $b;
    }
}

$callable = new DeferredCallable('Calculator:add', new \Slim\CallableResolver());
echo $callable(10, 20).PHP_EOL;

$callable = new DeferredCallable('Calculator:add');
try {
    echo $callable(10, 20).PHP_EOL;
} catch (Error $e) {
    echo $e->getMessage().PHP_EOL;
}
```
This would output
```
30
Call to undefined function Calculator:add()
```
The first one is working as it has a callable resolver.

Although. I am not sure if this would ever happen.

mapogolions:
In case of `Slim:notation` it's really is. But `DeferredCallable` entities are mainly used in a middleware stack. Therefore, they can be added through the following method calls:
- `MiddlewareDispatcher::add`
- `MiddlewareDispatcher::addDeferred`
I think, created `DeferredCallable's` must recieve `$callableResolver` from - https://github.com/slimphp/Slim/blob/4.x/Slim/MiddlewareDispatcher.php#L48. This field is not nullable.

mapogolions:
I found case. 
```php
$middleware = new DeferredCallable('Slim:notation');
$middlewareDispatcher->add($middleware) or addCallable($middleware);
```
I think it doesn't look natural.

l0gicgate:
> Shouldn't we throw a RuntimeException if that happens? If yes, already in the constructor or later in the invocation?

@adriansuter yes, let's throw a `RuntimeException` as soon as possible.

adriansuter:
@l0gicgate Sure. Is this class actually used somewhere? I haven't found any references.

[...]

PR #2807, opened 2019-08-16
4.x - Add check for Slim callable notation if no resolver given

adriansuter:
PR for #2804 

The `\Slim\DeferredCallable` can be instantiated without callable resolver. If the callable is a `string`, then in many cases, this should work as expected. For example for function names, or `ClassName::MethodName` notation. But the Slim callable notation `ClassName:MethodName` would not work. This PR makes sure that we throw an exception if that is the case.

### DIFF
--- a/Slim/DeferredCallable.php
+++ b/Slim/DeferredCallable.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Slim;
 
+use RuntimeException;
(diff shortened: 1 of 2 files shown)


l0gicgate:
As mentioned in #2804 we should remove `DeferredCallable` altogether. It's just unnecessary.

adriansuter:
Closing as we should rather remove the `DeferredCallable`.

PR #2809, opened 2019-08-17
4.x - Eliminate a duplicate code via HOF

mapogolions:
- reduce responsibility of some methods
- move duplicate code into a separate private function
- add more strict type hinting


l0gicgate (approved):
It looks tidier for sure. @adriansuter please review before I merge.

Slim/CallableResolver.php
```diff
@@ -4,7 +4,6 @@
  *
  * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
  */
-
 declare(strict_types=1);
```
adriansuter: Could you please add an empty line between the header comment and the `declare(..,)`.

adriansuter (changes requested):
**Wonderful work.**

One minor change request and one question: Do you think it would make sense to take the Closures out of the `resolveRoute` and `resolveMiddleware` methods? Otherwise php has to create the Closure over and over again, whenever the middleware or route get resolved.

For example:
```php
    protected function routePredicate($it): bool
    {
        return $it instanceof RequestHandlerInterface;
    }

    protected function middlewarePredicate($it): bool
    {
        return $it instanceof MiddlewareInterface;
    }

    /**
     * {@inheritdoc}
     */
    public function resolveRoute($toResolve): callable
    {
        return $this->resolveByPredicate($toResolve, [$this, 'routePredicate'], 'handle');
    }

    /**
     * {@inheritdoc}
     */
    public function resolveMiddleware($toResolve): callable
    {
        return $this->resolveByPredicate($toResolve, [$this, 'middlewarePredicate'], 'process');
    }
```
Of course the signature of `resolveByPredicate` would have to be adapted.

Slim/CallableResolver.php
mapogolions: Oh, PHP with its pretty ugly lambda expression :). I like FP style, something like `$this->method(x => x instanceof SomeInterface)`. For sure, i can make changes.

mapogolions:
@adriansuter  we were able to remove duplicate code, thanks to ideas from FP. In particular - higher order function. As you know, a function is treated like a __Fist Class Citizen__ (as any variable). Creating a local function with every function call is a common practice. Other languages have more concise syntax, like `it.map(x => x + 1)`

adriansuter:
Of course. I am open to any solution. Other languages do not only have more concise syntax, but they often have compilers and optimizers too. I am not sure, whether PHP does recognize the Closure to be the same during a second call. And I don't know if there would be any negative performance impact. But I can imagine, there is. Not sure, how big though.

Nevertheless, writing the Closures in the two methods would have the advantage, that everything "needed" is written inside the methods. So that might be better for maintainability.

As I said, I don't actually have a strong opinion for one or the other solution. Both have their advantages and disadvantages.

mapogolions:
Since I already made a commit that removes predicates from methods. Let's it be as it is. I think this is not so fundamental.

adriansuter (approved):

PR #2812, opened 2019-08-18
4.x - Travis-CI: Require coveralls library only if in analysis mode

adriansuter:
Travis-CI job for PHP nightly always fails because coveralls library does not work with PHP8.

This PR would make sure that the coveralls library would only be required in the ANALYSIS job.


adriansuter:
It appears, that this did not work as expected as many other libraries are not PHP8 ready - including Slim 😄 .

Other libraries that would not work with PHP8:
* `nyholm/psr7`
* `nyholm/psr7-server`
* `slim/http`
* `slim/psr7`
* `phpunit/phpunit`

Should we remove the php-nightly job for now?

Link to the job: https://travis-ci.org/slimphp/Slim/jobs/573533899

l0gicgate:
I’m indifferent to leaving it / removing it since it doesn’t affect anything.

l0gicgate (approved):

l0gicgate:
We should do that to all the repos that have coveralls enabled

PR #2813, opened 2019-08-19
4.x - Avoid that Travis-CI runs phpunit two times in analysis mode

adriansuter:
The Travis-CI job with analysis-mode (PHP 7.1) currently runs phpunit two times. One for the coverage report and then again without coverage.

This PR simply wraps the second phpunit into an if-statement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_script:
 
 script:
 - if [[ "$ANALYSIS" == 'true' ]]; then vendor/bin/phpunit --coverage-clover clover.xml ; fi
-- vendor/bin/phpunit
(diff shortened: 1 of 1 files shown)


l0gicgate (approved):

l0gicgate:
We need to fix this on all other repos as well. It'll make for faster builds. We also switched to PHP 7.3 for analysis on Slim-Psr7 I wonder if we should do the same for all the repos

adriansuter:
Sure. Given that PHP 7.1 would reach end of life soon, and that 7.3 is faster and more robust than 7.1, we should set the analysis to 7.3.

Issue #2815, opened 2019-08-20
$app->redirect() crashes when using DI\Container


slokhorst:
The following example causes a crash when visiting `/test`:

```php
<?php
use Psr\Http\Message\ResponseInterface as Response;
use Psr\Http\Message\ServerRequestInterface as Request;
use Slim\Factory\AppFactory;

require __DIR__ . '/../vendor/autoload.php';

/* CAUSES CRASH */
$container = new DI\Container();
AppFactory::setContainer($container);
/* CAUSES CRASH */

$app = AppFactory::create();
$app->get('/', function (Request $request, Response $response, array $args) {
    $response->getBody()->write('Hello World');
    return $response;
});

$app->redirect('/test', '/');
$app->run();
```

Crash message:
```
Details
Type: Error
Code: 0
Message: Call to a member function createResponse() on null
vendor/slim/slim/Slim/Routing/RouteCollectorProxy.php
Line: 199
Trace
#0 vendor/slim/slim/Slim/Handlers/Strategies/RequestResponse.php(42): Slim\Routing\RouteCollectorProxy->Slim\Routing\{closure}(Object(Slim\Http\ServerRequest), Object(Slim\Http\Response), Array)
#1 vendor/slim/slim/Slim/Routing/Route.php(374): Slim\Handlers\Strategies\RequestResponse->__invoke(Object(Closure), Object(Slim\Http\ServerRequest), Object(Slim\Http\Response), Array)
#2 vendor/slim/slim/Slim/MiddlewareDispatcher.php(65): Slim\Routing\Route->handle(Object(Slim\Http\ServerRequest))
#3 vendor/slim/slim/Slim/MiddlewareDispatcher.php(65): Slim\MiddlewareDispatcher->handle(Object(Slim\Http\ServerRequest))
#4 vendor/slim/slim/Slim/Routing/Route.php(339): Slim\MiddlewareDispatcher->handle(Object(Slim\Http\ServerRequest))
#5 vendor/slim/slim/Slim/Routing/RouteRunner.php(65): Slim\Routing\Route->run(Object(Slim\Http\ServerRequest))
#6 vendor/slim/slim/Slim/Middleware/ErrorMiddleware.php(89): Slim\Routing\RouteRunner->handle(Object(Slim\Http\ServerRequest))
[...]

PR #2816, opened 2019-08-20
4.x - Fix RouteCollectorProxy::redirect() Bug

l0gicgate:
Fix for #2815

When a container is present the callable gets bound to the container when it's resolved. So the `$this` context becomes the container and not `RouteCollectorProxy` anymore.

```php
$handler = function () use ($to, $status) {
    $response = $this->responseFactory->createResponse($status);
    return $response->withHeader('Location', (string) $to);
};
```

We're using `$this->responseFactory()` when we should just pass it in via `use()`

### DIFF
--- a/Slim/Routing/RouteCollectorProxy.php
+++ b/Slim/Routing/RouteCollectorProxy.php
@@ -195,8 +195,10 @@ class RouteCollectorProxy implements RouteCollectorProxyInterface
      */
     public function redirect(string $from, $to, int $status = 302): RouteInterface
     {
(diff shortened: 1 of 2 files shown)


adriansuter:
Is there a reason for the container prophecy in the test case? `CallableResolver` can be used without.

Other than that, everything looks fine. Can't officially approve as my phone app (OctoDroid) has a bug.

adriansuter:
Ah, should have read the issue. Everything is good!

PR #2817, opened 2019-08-20
4.x - MiddlewareDispatcherInterface

### LINKED ISSUE #2785: 4.x - MiddlewareDispatcherInterface
We will need to introduce a new interface in order to enable users to drop in their own `MiddlewareDispatcher`

Since we won't be able to rely on the constructor, we need methods to set the following:
- `add()` - Untyped method to add a `MiddlewareInterface`, a `string` for deferred callables or a `callable`
- `addMiddleware()` - Typed method to add only `MiddlewareInterface` for people who enjoy stricter typed approaches
- `seedMiddlewareStack()` Set the kernel or "tip" of the middleware stack
- ~~`setCallableResolver()` Set a `CallableResolver`, for resolving all middleware class names / callables. Currently we are injecting `ContainerInterface` inside of `MiddlewareDispatcher` and duplicating some of the logic from within `CallableResolver`. That needs to be changed once we modify `CallableResolver` to handle all that logic.~~

Here is the proposed interface:
```php
<?php
declare(strict_types=1);

namespace Slim\Interfaces;

use Psr\Http\Server\MiddlewareInterface;
use Psr\Http\Server\RequestHandlerInterface;

interface MiddlewareDispatcherInterface extends RequestHandlerInterface
{
    /**
     * Add a new middleware to the stack
     *
     * Middleware are organized as a stack. That means middleware
     * that have been added before will be executed after the newly
     * added one (last in, first out).
     *
     * @param MiddlewareInterface|string|callable $middleware
     * @return self
     */
    public function add($middleware): self;

    /**
     * Add a new middleware to the stack
     *
     * Middleware are organized as a stack. That means middleware
     * that have been added before will be executed after the newly
     * added one (last in, first out).
     *
     * @param MiddlewareInterface $middleware
     * @return self
     */
    public function addMiddleware(MiddlewareInterface $middleware): self;

    /**
     * Seed the middleware stack with the inner request handler
     *
     * @param RequestHandlerInterface $kernel
     * @return void
     */
    public function seedMiddlewareStack(RequestHandlerInterface $kernel): void;
}
```

l0gicgate:
Closes #2785 

This PR enables users to inject their own `MiddlewareDispatcher` into the `App` component.

The new proposed interface:
```php
<?php
declare(strict_types=1);

namespace Slim\Interfaces;

use Psr\Http\Server\MiddlewareInterface;
use Psr\Http\Server\RequestHandlerInterface;

interface MiddlewareDispatcherInterface extends RequestHandlerInterface
{
    /**
     * Add a new middleware to the stack
     *
     * Middleware are organized as a stack. That means middleware
     * that have been added before will be executed after the newly
     * added one (last in, first out).
     *
     * @param MiddlewareInterface|string|callable $middleware
     * @return self
     */
    public function add($middleware): self;

    /**
     * Add a new middleware to the stack
     *
     * Middleware are organized as a stack. That means middleware
     * that have been added before will be executed after the newly
     * added one (last in, first out).
     *
     * @param MiddlewareInterface $middleware
     * @return self
     */
    public function addMiddleware(MiddlewareInterface $middleware): self;

    /**
     * Seed the middleware stack with the inner request handler
     *
     * @param RequestHandlerInterface $kernel
     * @return void
     */
    public function seedMiddlewareStack(RequestHandlerInterface $kernel): void;
}
```

### DIFF
--- a/Slim/MiddlewareDispatcher.php
+++ b/Slim/MiddlewareDispatcher.php
@@ -18,8 +18,9 @@ use Psr\Http\Server\RequestHandlerInterface;
 use RuntimeException;
 use Slim\Interfaces\AdvancedCallableResolverInterface;
 use Slim\Interfaces\CallableResolverInterface;
+use Slim\Interfaces\MiddlewareDispatcherInterface;
 
-class MiddlewareDispatcher implements RequestHandlerInterface
+class MiddlewareDispatcher implements MiddlewareDispatcherInterface
 {
     /**
      * Tip of the middleware call stack
@@ -29,7 +30,7 @@ class MiddlewareDispatcher implements RequestHandlerInterface
     protected $tip;
 
     /**
-     * @var CallableResolverInterface
+     * @var CallableResolverInterface|null
      */
     protected $callableResolver;
 
@@ -54,12 +55,9 @@ class MiddlewareDispatcher implements RequestHandlerInterface
     }
 
     /**
-     * Seed the middleware stack with the inner request handler
-     *
-     * @param RequestHandlerInterface $kernel
-     * @return void
+     * {@inheritdoc}
      */
-    protected function seedMiddlewareStack(RequestHandlerInterface $kernel): void
+    public function seedMiddlewareStack(RequestHandlerInterface $kernel): void
     {
         $this->tip = $kernel;
     }
@@ -83,9 +81,9 @@ class MiddlewareDispatcher implements RequestHandlerInterface
      * added one (last in, first out).
      *
      * @param MiddlewareInterface|string|callable $middleware
-     * @return self
+     * @return MiddlewareDispatcherInterface
      */
-    public function add($middleware): self
+    public function add($middleware): MiddlewareDispatcherInterface
     {
         if ($middleware instanceof MiddlewareInterface) {
             return $this->addMiddleware($middleware);
@@ -113,9 +111,9 @@ class MiddlewareDispatcher implements RequestHandlerInterface
      * added one (last in, first out).
      *
      * @param MiddlewareInterface $middleware
-     * @return self
+     * @return MiddlewareDispatcherInterface
      */
-    public function addMiddleware(MiddlewareInterface $middleware): self
+    public function addMiddleware(MiddlewareInterface $middleware): MiddlewareDispatcherInterface
     {
         $next = $this->tip;
         $this->tip = new class($middleware, $next) implements RequestHandlerInterface
--- a/Slim/MiddlewareDispatcher.php
+++ b/Slim/MiddlewareDispatcher.php
@@ -40,13 +40,13 @@ class MiddlewareDispatcher implements MiddlewareDispatcherInterface
     protected $container;
 
     /**
-     * @param RequestHandlerInterface   $kernel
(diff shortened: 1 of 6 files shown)


Slim/MiddlewareDispatcher.php
```diff
@@ -29,7 +30,7 @@ class MiddlewareDispatcher implements RequestHandlerInterface
     protected $tip;
 
     /**
-     * @var CallableResolverInterface
+     * @var CallableResolverInterface|null
```
adriansuter, line 33: How can that be `null`?

Slim/MiddlewareDispatcher.php
```diff
@@ -54,12 +55,9 @@ public function __construct(
...
 
     /**
-     * Seed the middleware stack with the inner request handler
-     *
-     * @param RequestHandlerInterface $kernel
-     * @return void
+     * {@inheritdoc}
      */
-    protected function seedMiddlewareStack(RequestHandlerInterface $kernel): void
+    public function seedMiddlewareStack(RequestHandlerInterface $kernel): void
```
adriansuter, line 60: If someone has extended the `MiddlewareDispatcher`, then this might be a BC. But I don't think that this should be a problem, what do you think?

tests/AppTest.php
```diff
@@ -118,6 +119,44 @@ public function testCreatesRouteCollectorWhenNullWithInjectedContainer()
...
+        $app = new App(
+            $responseFactoryProphecy->reveal(),
+            null,
+            null,
+            null,
+            null,
+            $middlewareDispatcherProphecy->reveal()
+        );
+
+        $this->assertSame($middlewareDispatcherProphecy->reveal(), $app->getMiddlewareDispatcher());
```
adriansuter, line 140:
```php
$this->assertSame($middlewareDispatcherProphecy->reveal(), $app->getMiddlewareDispatcher());
```
Don't we test that already in the `testGetMiddlewareDispatcherReturnsInjectedInstance` test case? Maybe this line can be removed.

adriansuter (changes requested):
Excellent.

Slim/MiddlewareDispatcher.php
l0gicgate: It cannot right now, but as I mentioned before, with the new logic inside of `addDeferred()` this can now be nullable and not depend on `CallableResolver`. I forgot to null out the parameter in the constructor.

Slim/MiddlewareDispatcher.php
l0gicgate: Yeah that would be an internal BC. Considering there was no way to replace that component before I'm doubtful anyone went out of their way to extend this.

Slim/MiddlewareDispatcher.php
adriansuter: Okay.

tests/AppTest.php
l0gicgate: I've combined the two tests instead. Still asserting that the stack gets seeded.

adriansuter: Even better.

adriansuter (approved):

Issue #2795, opened 2019-08-12
4.x - Fix message in ErrorHandler


l0gicgate:
https://github.com/slimphp/Slim/blob/4.x/Slim/Handlers/ErrorHandler.php#L271

We need to come up with a more appropriate message.

> View in rendered output by enabling the "displayErrorDetails" setting.

This is confusing to me

RyanNerd:
What is the purpose of appending this? (or any additional text to the error log): 

> View in rendered output by enabling the "displayErrorDetails" setting.

to the `error_log()` in the first place?

@l0gicgate This confuses me as well. What are the arguments for appending **anything** additional here? https://github.com/slimphp/Slim/blob/4.x/Slim/Handlers/ErrorHandler.php#L271

If there aren't any compelling arguments then I propose the line should be removed.

l0gicgate:
@RyanNerd so if someone was to look at the logs and see that there are no error details (stack trace). It would indicate to them that by turning on `displayErrorDetails` on `ErrorHandler` they could see the stack trace via a browser/http client by toggling that on. It's really just a "tip"

RyanNerd:
This is marked as an easy pick but changing the message to something else is very subjective. Perhaps something like:

> To see error details set displayErrorDetails to true in the ErrorHandler.

vaga:
Perhaps, something like:
`Tips: To display error details in HTTP response set "displayErrorDetails" to true in the ErrorHandler constructor.`

And you can add this line only if "$displayErrorDetails" is equal to `false`.

PR #2821, opened 2019-08-25
Fix error message, issue #2795

QinisoM:
This is an update to the error message as reported

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -283,7 +283,7 @@ class ErrorHandler implements ErrorHandlerInterface
     {
         $renderer = $this->callableResolver->resolve($this->logErrorRenderer);
         $error = $renderer($this->exception, $this->logErrorDetails);
-        $error .= "\nView in rendered output by enabling the \"displayErrorDetails\" setting.\n";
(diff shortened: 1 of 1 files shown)


l0gicgate (commented):
I think we need to add wording along the lines of “in the browser/client”. That’s what we’re trying to convey. I don’t know how to word it properly but:

To see the error details from your browser/client set “displayErrorDetails” to “true” in the ErrorHandler’s constructor. 


🆘 not an English major. Send help.

RyanNerd:
In Slim 3.x this is what the [verbiage looks like](https://github.com/slimphp/Slim/blob/200c6143f15baa477601879b64ab2326847aac0b/Slim/Handlers/AbstractError.php#L48):
 `$message .= PHP_EOL . 'View in rendered output by enabling the "displayErrorDetails" setting.' . PHP_EOL;`

@l0gicgate Once it is decided how this should be worded in 4.x should 3.x also be changed to the same or would that be a b/c?

l0gicgate:
We can leave it for Slim 3 since it is long-standing and it’s an actual app setting.

odan:
The response must always be the same. It should be irrelevant on which operating system the error message is created. We should keep `\n` and never use `PHP_EOL` in this context. Thanks.

l0gicgate:
We need to discuss further in #2795 before PRs are raised. Follow there

PR #2823, opened 2019-08-27
 Fix error message, issue #2795

QinisoM:
Updated message as suggestd


odan:
Changing `\n` to `PHP_EOL` would cause different response under windows. This would make it harder to test. The response must always be the same, independent where the server is running.

l0gicgate:
We need to discuss further in #2795 before PRs are raised. Follow there

PR #2832, opened 2019-09-09
4.x - Refactor `FastRouteDispatcher`

adriansuter:
This PR does refactor `\Slim\Routing\FastRouteDispatcher` as there were basically some code duplicates.

So I created a new method `routingResults()` that tries to find routing results given an http method and a uri. The method would check the static route map as well as the variable route data.

I have removed the protected method `routingResultsFromVariableRouteResults()` - probably a BC that affects users?

### DIFF
--- a/Slim/Routing/FastRouteDispatcher.php
+++ b/Slim/Routing/FastRouteDispatcher.php
@@ -21,64 +21,56 @@ class FastRouteDispatcher extends GroupCountBased
     /**
      * @param string $httpMethod
      * @param string $uri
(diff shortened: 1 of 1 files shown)


l0gicgate:
> I have removed the protected method routingResultsFromVariableRouteResults() - probably a BC that affects users?

This won't be a BC unless someone extended and overrode.. I'm willing to risk it haha.

Looks much tidier in there now good work!

l0gicgate (approved):

PR #2841, opened 2019-09-19
Update README.md

mokevnin:
It looks like the _index.php_ is inside the _public_ directory, but there is no information about it.


froschdesign:
> It looks like the index.php is inside the public directory, but there is no information about it.

Good catch, but the `vendor` directory does not belong in the public area. My suggestion: add a filename before the code example or as commend in the code example.

mokevnin:
One more time)

l0gicgate (approved):

adriansuter (approved):

Issue #2367, opened 2017-12-13
The App overrides static basePath on the router


morozov:
When a Slim application works behind a reverse proxy and generates absolute URLs, it generates them according to the internal web server hostname and port, not the external ones. Even if the base path is statically defined in container configuration, it gets overridden during request processing:

https://github.com/slimphp/Slim/blob/3ab1a154e45546fd8654f87708577096391b50f7/Slim/App.php#L340-L342

Currently, it's only possible to override this behavior by introducing a middleware which kicks in after `App::process()` and restores the needed base path.

It would be nice if the `App` only set the base path if it's not set yet.

geggleto:
I don't think we will be fixing this in v3... but we will probably address this in v4.

morozov:
Should this be reopened for Slim 4? From the DX perspective, it looks like a bug while the described approach is more of a workaround, not a recommended solution.

PR #2844, opened 2019-09-24
Handle base path by routeCollector instead of RouteCollectorProxy

### LINKED ISSUE #2842: 4.x - Inconsistent `setBasePath` Behavior
- When using `$app->setBasePath('/public')` it breaks `RouteParser::relativeUrlFor` output:
  ```
  echo $routerParser->urlFor('about');
  /public/about
  ```

  ```
  echo $routerParser->relativeUrlFor('about');
  about <- expected
  /public/about <- result
  ```

- Using only `$app->getRouteCollector()->setBasePath('/public')`, RouteParser produces corrent urls for both urlFor and relativeUrlFor methods but breaks current path detection
- Using both `$app->setBasePath` and `$routeCollector->setBasePath` produces invalid url paths for all methods

_Originally posted by @piotr-cz in https://github.com/slimphp/Slim/issues/2512#issuecomment-534056851_

piotr-cz:
At this moment, basePath is handled by both **RouteCollectorProxy** and **RouteCollector**.

This PR:
- [respects](https://github.com/slimphp/Slim/pull/2844/commits/017cb003316b732f250b567e908426264722fc6e#diff-37b87065bdf2692920fc31e79882d28eR41-R44) `RouteCollector::$basePath` when creating routes via Dispatcher
- [changes](https://github.com/slimphp/Slim/pull/2844/commits/017cb003316b732f250b567e908426264722fc6e#diff-f43987d6169165b72946d742f3062eb3R104-R117) `RouteCollectorProxy::setBasePath` and `RouteCollectorProxy::getBasePath` methods which now call RouteCollector methods
- renames `RouteCollectorProxy::$basePath` to `$groupPattern` as [that's what it is](https://github.com/slimphp/Slim/blob/017cb003316b732f250b567e908426264722fc6e/Slim/Routing/RouteCollector.php#L248)

There are no B/C changes.

Fixes https://github.com/slimphp/Slim/issues/2842 and possibly https://github.com/slimphp/Slim/issues/2512

For test case, see [Slim-4-basepath-issue](https://github.com/piotr-cz/Slim-4-basepath-issue)

### DIFF
--- a/Slim/Routing/Dispatcher.php
+++ b/Slim/Routing/Dispatcher.php
@@ -38,8 +38,10 @@ class Dispatcher implements DispatcherInterface
         }
 
         $routeDefinitionCallback = function (RouteCollector $r) {
+            $basePath = $this->routeCollector->getBasePath();
+
             foreach ($this->routeCollector->getRoutes() as $route) {
-                $r->addRoute($route->getMethods(), $route->getPattern(), $route->getIdentifier());
+                $r->addRoute($route->getMethods(), $basePath . $route->getPattern(), $route->getIdentifier());
             }
         };
 
--- a/Slim/Routing/RouteCollectorProxy.php
+++ b/Slim/Routing/RouteCollectorProxy.php
@@ -43,27 +43,27 @@ class RouteCollectorProxy implements RouteCollectorProxyInterface
     /**
      * @var string
      */
-    protected $basePath;
+    protected $groupPattern;
 
     /**
      * @param ResponseFactoryInterface     $responseFactory
      * @param CallableResolverInterface    $callableResolver
      * @param RouteCollectorInterface|null $routeCollector
      * @param ContainerInterface|null      $container
-     * @param string                       $basePath
+     * @param string                       $groupPattern
      */
     public function __construct(
         ResponseFactoryInterface $responseFactory,
         CallableResolverInterface $callableResolver,
         ?ContainerInterface $container = null,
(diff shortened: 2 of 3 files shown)


adriansuter (approved):

morozov:
> There are no B/C changes.

FWIW, there is a BC break. Prior to this PR, the base path, if set using a middleware, was only effective when generating URLs but not when routing requests. Now it's effective with routing as well.

While it may have been a bug, it could be used for in case when the Slim API sits behind a reverse proxy (see https://github.com/slimphp/Slim/issues/2367).

piotr-cz:
As I understand it, this is a bug as documentation mentions only the `App::setBasePath` method [here](http://www.slimframework.com/docs/v4/start/web-servers.html#run-from-a-sub-directory).

I guess if you would like to use different base path for creating routes, easiest way would be to mutate the request URI in the middleware.

This may not be semantically correct but the trick is that once the routing phase is done mutated URI object will be used to render routes.

I'm attaching a code I've been using in custom middleware before this PR
```php
public function process(Request $request, RequestHandler $handler): ResponseInterface
{
    $basePath = static::resolveBasePath();

    $uri = $request->getUri();
    $path = $uri->getPath();

    // Strip base (don't use ltrim)
    $path = substr($path, strlen($basePath));

    // Update request
    $request = $request->withUri($uri->withPath($path));

    return $handler->handle($request);
}
```

PR #2848, opened 2019-09-29
add status code 422

xleliu:
Use for parameters validate


l0gicgate:
The exceptions within the core library are mainly for internal purposes. We are not going to add this unfortunately as it is out of the scope of the core library.

Issue #2837, opened 2019-09-16
4.x - Get base path within a route callback


adriansuter:
Say you have a static file for which you would like to build the url. So the url is not a route. As there might be a base path present, we can't just write `/static-file.html`. How do we get the base path or even the base url from within a route callback?

Of course we could inject the base path to the class containing the route callback. But there should be a better solution.

Something like
```php
$absoluteUrl = RouteContext::fromRequest($request)->getRouteParser()->basePath() . 'static-file.html';
$fullQualifiedUrl = RouteContext::fromRequest($request)->getRouteParser()->fullBasePath($uri) . 'static-file.html';
```

I am not sure if that belongs to the route parser though. If it does, we could maybe instead of the above create two methods that are similar to `urlFor` and `fullUrlFor` but instead of using a route name, we could pass a path (relative to the application end point).
```php
public function url(string $path, array $queryParams = []): string;
public function fullUrl(UriInterface $uri, string $path, array $queryParams = []): string;
```

So
```php
RouteContext::fromRequest($request)->getRouteParser()->url('static-file.html');
RouteContext::fromRequest($request)->getRouteParser()->fullUrl($uri, 'static-file.html');
```

What do you think?

adriansuter:
Any other ideas or opinions on that?

odan:
@adriansuter What would be the difference compared to the existing Uri::getBaseUrl?

https://github.com/slimphp/Slim-Http#decorated-uri-object-methods

> The decorated UriInterface provides the following additional methods:
> Uri::getBaseUrl()
> Returns the fully qualified base URL of the underlying uri object.

I think first we should get a better overview of what we have now (and where) and point out the differences and uses cases for each of that.

adriansuter:
@odan if your app lives in a subdirectory, say `app`, that is you would access your root route `/` with `http://example.com/app`, then the base path is `/app`. Say I have a file `script.js`, how can we build a path inside a route callback that works even if we change the base path later on? Would that work with the http decorator `Uri`?

l0gicgate:
> Any other ideas or opinions on that?

@adriansuter I think we could maybe provide the basepath via middleware when routing is done? Then maybe add onto the `RouteContext` object?

odan:
@adriansuter The currently possible way would be to define a "name" for the `/` route, e.g. "root" and use this name to determine the url with the basePath.

```php
$app->setBasePath('/app');

[...]

PR #2852, opened 2019-10-01
Adding RouteParser::getBasePath method

piotr-cz:
See #2837

At this moment it's not trivial to retrieve `basePath` from code that doesn't have access to App or RouteCollector instances (and in my opinion code responsible for rendering responses shouldn't have).

RouteParser class seems like good place for it as

- it contains other related methods (`relativeUrlFor`, `urlFor`, `fullUrlFor`)
- it can be available as a route attribute (and indirectly in middleware) when using `App::addRoutingMiddleware`

I haven't implemented in this PR other methods that could be helpful like baseUrl by purpose as I'm not sure if it should be implemented in Slim or view classes.

### DIFF
--- a/Slim/Interfaces/RouteParserInterface.php
+++ b/Slim/Interfaces/RouteParserInterface.php
@@ -61,5 +61,5 @@ interface RouteParserInterface
      *
      * @return string
      */
-    public function basePath(): string;
+    public function getBasePath(): string;
 }
--- a/Slim/Interfaces/RouteParserInterface.php
+++ b/Slim/Interfaces/RouteParserInterface.php
@@ -55,4 +55,11 @@ interface RouteParserInterface
      * @return string
      */
     public function fullUrlFor(UriInterface $uri, string $routeName, array $data = [], array $queryParams = []): string;
+
+    /**
+     * Get base path
+     *
+     * @return string
+     */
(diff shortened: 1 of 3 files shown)


adriansuter (approved):

odan:
IMHO: For the sake of "symmetry", the method name `getBasePath` would be more appropriate and meaningful.

piotr-cz:
@odan This is I was not sure about, as there also are methods like [`relativeUrlFor`](https://github.com/slimphp/Slim/blob/4.2.0/Slim/Routing/RouteParser.php#L42) and not `getRelativeUrlFor`.

odan:
@piotr-cz Ah yes, that's right.

piotr-cz:
@odan 
I've renamed method to `getBasePath` like you suggested.
It is used in other places in application and different one could confuse people.

l0gicgate:
This is a BC change since we’re adding onto the interface. @akrabat correct?

akrabat:
> This is a BC change since we’re adding onto the interface. @akrabat correct?

Yes.  Ouch.

piotr-cz:
What about keeping method in implementation without adding it to interface?
We could add it to interface next major version

piotr-cz:
Superseeded by #2859

Issue #2856, opened 2019-10-03
Truncated stream


mirkobriga:
When I write:

`
return $response->withStatus($guzzle_response->getStatusCode())->withBody($guzzle_response->getBody());
`

ResponseEmitter > emitBody > while - exit before sends all chunks. speed solution here 
https://github.com/mirkobriga/Slim/commit/de2497ef22424076d64123620365b81443b7de52

l0gicgate:
Why don’t you just emit the Guzzle response? Not sure why you’re wrapping it

mirkobriga:
Because it's an api call to another service. I'm not wrapping I set that body stream like slim response body stream. I think withBody should work with every stream.

mirkobriga:
Please investigate... I found that in stream eof function that wrapping PHP feof function It return true, but there are other characters before EOF.

mirkobriga:
All these failed returning truncated response

`
<?php

use Psr\Http\Message\ResponseInterface as Response;
use Psr\Http\Message\ServerRequestInterface as Request;
use Slim\Psr7\Stream;
use Slim\Factory\AppFactory;
use GuzzleHttp\Client;

require __DIR__ . '/../vendor/autoload.php';

$app = AppFactory::create();

$app->get('/api/test1', function (Request $request, Response $response, $args) {
    $client = new Client();
    $called_url_response = $client->request('GET', 'http://api.plos.org/search?q=title:%22Drosophila%22%20and%20body:%22RNA%22&fl=id&start=1&rows=1000');   
    return $response->withStatus($called_url_response->getStatusCode())->withBody($called_url_response->getBody());
});

$app->get('/api/test2', function (Request $request, Response $response, $args) {
    $client = new Client();
    $called_url_response = $client->request('GET', 'http://api.plos.org/search?q=title:%22Drosophila%22%20and%20body:%22RNA%22&fl=id&start=1&rows=1000');   
    $stream=new Stream($called_url_response->getBody()->detach());
    return $response->withStatus($called_url_response->getStatusCode())->withBody($stream);
});

$app->get('/api/test3', function (Request $request, Response $response, $args) {
    $client = new Client();
    $resource= fopen('http://api.plos.org/search?q=title:%22Drosophila%22%20and%20body:%22RNA%22&fl=id&start=1&rows=1000', 'rw+');
    $stream=new Stream($resource);
    return $response->withStatus($called_url_response->getStatusCode())->withBody($stream);
});

[...]

PR #2857, opened 2019-10-03
bugfix on response withBody StreamInterface

mirkobriga:


l0gicgate:
This is supposed to be a bug fix for #2856 but OP should instead emit the guzzle response directly instead.

PR #2858, opened 2019-10-03
Strategy handler response factory refactoring

iSize1ce:
The response factory is used only in two of the three strategies, and it is passed as an argument in the interface. Perhaps it would be more logical to pass the response factory to the constructor as a dependency

### DIFF
--- a/Slim/Interfaces/InvocationStrategyInterface.php
+++ b/Slim/Interfaces/InvocationStrategyInterface.php
@@ -22,7 +22,6 @@ interface InvocationStrategyInterface
      *
      * @param callable               $callable       The callable to invoke using the strategy.
      * @param ServerRequestInterface $request        The request object.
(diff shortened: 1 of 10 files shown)


Slim/Interfaces/InvocationStrategyInterface.php
```diff
@@ -22,15 +22,13 @@ interface InvocationStrategyInterface
...
      * @param ServerRequestInterface $request        The request object.
-     * @param ResponseInterface      $response       The response object.
      * @param array                  $routeArguments The route's placeholder arguments
      *
      * @return ResponseInterface The response from the callable.
      */
     public function __invoke(
         callable $callable,
         ServerRequestInterface $request,
-        ResponseInterface $response,
```
l0gicgate, line 33: We cannot change the interface as this is a bc break.

PR #2859, opened 2019-10-04
Adding RouteParser::getBasePath method (no changes to RouteParserInterface)

### LINKED ISSUE #2837: 4.x - Get base path within a route callback

piotr-cz:
This superseeds #2852 and partially resolves #2837

Only difference is that there are no changes to `RouteParserInterface` to not break backward compatibility.

### DIFF
--- a/Slim/Routing/RouteParser.php
+++ b/Slim/Routing/RouteParser.php
@@ -125,4 +125,14 @@ class RouteParser implements RouteParserInterface
         $protocol = ($scheme ? $scheme . ':' : '') . ($authority ? '//' . $authority : '');
         return $protocol . $path;
     }
+
+    /**
+     * Get base path
+     *
+     * @return string
+     */
+    public function getBasePath(): string
+    {
+        return $this->routeCollector->getBasePath();
+    }
 }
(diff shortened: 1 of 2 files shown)


adriansuter:
That feels a bit hackish.

`\Slim\Routing\RouteContext::getRouteParser()` returns a `\Slim\Interfaces\RouteParserInterface` and therefore you would have to check if that is actually a `\Slim\Routing\RouteParser` and only then you could call that method.

l0gicgate:
Fixed in #2860

PR #2860, opened 2019-10-04
[4.x] Add base path to `$request` and use `RouteContext` to read

adriansuter:
Another attempt for #2837. This is an alternative solution to #2859.

It uses the route runner to bring the base path to the `$request` object and then the route context helper class to read that attribute from the `$request`.


piotr-cz:
Thank you!

piotr-cz:
As the `basePath` attribute is set in RouteRunner, request passed to middlewares and indirectly to error handlers doesn't have have it set.

Only place is the Route callable.

adriansuter:
@piotr-cz Should we add the basePath attribute directly in the `App->run()` method? If someone is overriding this method, then it is important to make sure that this attribute will be written to the request. Otherwise we can't access it.

@l0gicgate Do you think we should/could undo the RouteRunner modification?

adriansuter:
A middleware run before the routing middleware would not have access to any route context instance (route parser, etc), would it? So this solution would only work in a route callback.

piotr-cz:
After this PR attributes  `route`, `routingResults` and `basePath` are set on the request by the [RouteRunner](https://github.com/slimphp/Slim/blob/4.3.0/Slim/Routing/RouteRunner.php#L66-L81) which is available in route callable.

As I understand it, if you want to get route attributes in the middleware, you have to add [RoutingMiddleware](https://github.com/slimphp/Slim/blob/4.3.0/Slim/Middleware/RoutingMiddleware.php) (see [docs](http://www.slimframework.com/docs/v4/middleware/routing.html)). But this sets only `route` and `routingResults` attributes.

Another option could be to set the `basePath` request attribute in [App::handle](https://github.com/slimphp/Slim/blob/4.3.0/Slim/App.php) method.

To be honest, I'm not sure what is the best way out of this.
I've decided to set base path on the view renderer during [dependecy initialization](https://github.com/slimphp/Slim-Skeleton/blob/4.0.3/app/dependencies.php).

adriansuter:
@piotr-cz Injecting the base path into the middleware, error handlers and route callables is the preferred way to go. At least if you use dependency injection. The solution to use request attributes should IMO only be used in small projects in case of absence of dependency injection.

piotr-cz:
okay, thanks for explanation

PR #2862, opened 2019-10-08
Optionally handle subclasses of exceptions in custom error handler

nickdnk:
Hello guys

I thought this could be a nice addition, so a particular app can decide to handle all subclasses of an exception with one handler. This is nice way to keep things separated.

I added this as an option that defaults to false so this *should* not be a breaking change, but I'm not 100% sure since `$handlers` is `protected` and not `private`.

What do you think?

### DIFF
--- a/tests/Middleware/ErrorMiddlewareTest.php
+++ b/tests/Middleware/ErrorMiddlewareTest.php
@@ -99,24 +99,127 @@ class ErrorMiddlewareTest extends TestCase
         $this->assertInstanceOf(ErrorHandler::class, $handler);
     }
 
+    public function testSuperclassExceptionHandlerHandlesExceptionWithSubclassExactMatch()
+    {
+        $responseFactory = $this->getResponseFactory();
+        $app = new App($responseFactory);
+        $callableResolver = $app->getCallableResolver();
+        $mw = new ErrorMiddleware($callableResolver, $this->getResponseFactory(), false, false, false);
+        $app->add(function ($request, $handler) {
+            throw new \LogicException('This is a LogicException...');
+        });
+        $mw->setErrorHandler(\LogicException::class, (function (ServerRequestInterface $request, $exception) {
+            $response = $this->createResponse();
+            $response->getBody()->write($exception->getMessage());
+            return $response;
+        })->bindTo($this), true); // - true; handle subclass but also LogicException explicitly
+        $mw->setDefaultErrorHandler((function () {
+            $response = $this->createResponse();
+            $response->getBody()->write('Oops..');
+            return $response;
+        })->bindTo($this));
+        $app->add($mw);
+        $app->get('/foo', function (ServerRequestInterface $request, ResponseInterface $response) {
+            $response->getBody()->write('...');
+            return $response;
+        });
+        $request = $this->createServerRequest('/foo');
+        $app->run($request);
+        $this->expectOutputString('This is a LogicException...');
+    }
+
+    public function testSuperclassExceptionHandlerHandlesSubclassException()
+    {
+        $responseFactory = $this->getResponseFactory();
+        $app = new App($responseFactory);
+        $callableResolver = $app->getCallableResolver();
+
+        $mw = new ErrorMiddleware($callableResolver, $this->getResponseFactory(), false, false, false);
+
+        $app->add(function ($request, $handler) {
+            throw new \InvalidArgumentException('This is a subclass of LogicException...');
+        });
+
+        $mw->setErrorHandler(\LogicException::class, (function (ServerRequestInterface $request, $exception) {
+            $response = $this->createResponse();
+            $response->getBody()->write($exception->getMessage());
+            return $response;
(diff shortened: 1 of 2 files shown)


nickdnk:
Force pushed for typo.

nickdnk:
We could mitigate the BC by doing this instead:

```
if (isset($this->handlers[$type])) {
    return $this->callableResolver->resolve(
        is_array($this->handlers[$type]) ? $this->handlers[$type]['handler'] : $this->handlers[$type]
    );
} else {
    foreach ($this->handlers as $class => $handler) {
        if (is_array($handler) && $handler['handle_subclass'] === true) {
            if (is_subclass_of($type, $class)) {
                return $this->callableResolver->resolve($handler['handler']);
            }
        }
    }
}
```

Then even if a someone has subclassed ErrorMiddleware *and* overriden $handlers to contain callables while not overriding the `setErrorHandler` method, it still wouldn't break. Why someone would do that I don't know though.

l0gicgate:
@adriansuter I think this would be a good addition and there's no BC breaks. What do you think about it?

nickdnk:
Is there a reason that `$handlers` is `protected` and not `private` ? Given how close this variable is tied to the `setErrorHandler` method, should it be visible to a subclass?

l0gicgate:
> Is there a reason that $handlers is protected and not private ? Given how close this variable is tied to the setErrorHandler method, should it be visible to a subclass?

In case one wants to extend `ErrorMiddleware` we left it as protected.

adriansuter (commented):
That is a good idea.

Another solution would be to add a new protected property `subClassHandlers` and then just write 
```php
    public function setErrorHandler(string $type, $handler, bool $handleSubclasses = false): self
    {
        if ($handleSubclasses) {
            $this->subClassHandlers[$type] = $handler;
        } else {
            $this->handlers[$type] = $handler;
        }

        return $this;
    }
```

That way, we do not change the data logic of the `handlers` property which is expected to be a plain array of handlers.

Of course we would have to change `getErrorHandler()` too.
```php
    public function getErrorHandler(string $type)
    {
        if (isset($this->handlers[$type])) {
            return $this->callableResolver->resolve($this->handlers[$type]);
        } elseif (isset($this->subClassHandlers[$type])) {
            return $this->callableResolver->resolve($this->subClassHandlers[$type]);
        } else {
            foreach ($this->subClassHandlers as $class => $handler) {
                if (is_subclass_of($type, $class)) {
                    return $this->callableResolver->resolve($handler);
                }
            }
        }

        return $this->getDefaultErrorHandler();
    }
```

That way, we would also get rid of the double `if`-statements in the `foreach` loop.

What do you mean, @l0gicgate and @nickdnk ? If you want to leave it as requested in this PR, then I am fine too.

nickdnk:
I think yours is a good workaround, @adriansuter.

l0gicgate:
@adriansuter I like your solution. We should proceed with that.

nickdnk:
Updated the PR.

Also, I added phpdocs for the handler arrays based on the value of `$defaultErrorHandler` - but I'm not entirely sure that you can annotate them as `ErrorHandlerInterface` since they're functions? Edit: Aha, so if you pass that interface in then it will behave as a function when __invoke is called. Got it.

l0gicgate:
Need to write tests as well for the subclass case. Are you able to do that?

nickdnk:
I did. I don't know why it claims it wasn't covered. There is a test for both with and without subclass use.

adriansuter:
Did you add a test case for the case you set an error handler as subclass handler, but the exception thrown is not a subclass but exactly that class?

`} elseif (isset($this->subClassHandlers[$type])) {`

nickdnk:
I think that did it.

tests/Middleware/ErrorMiddlewareTest.php
```diff
@@ -99,24 +99,129 @@ public function testGetErrorHandlerWillReturnDefaultErrorHandlerForUnhandledExce
...
+            return $response;
+        });
+        $request = $this->createServerRequest('/foo');
+        $app->run($request);
+        $this->expectOutputString('This is a LogicException...');
+    }
+
+    public function testSuperclassExceptionHandlerHandlesSubclassException()
+    {
+
```
l0gicgate: Can you remove this blank line please.

tests/Middleware/ErrorMiddlewareTest.php
```diff
@@ -99,24 +99,129 @@ public function testGetErrorHandlerWillReturnDefaultErrorHandlerForUnhandledExce
...
+        $request = $this->createServerRequest('/foo');
+        $app->run($request);
+
+        $this->expectOutputString('This is a subclass of LogicException...');
+    }
+
+    public function testSuperclassExceptionHandlerDoesNotHandleSubclassException()
+    {
+
```
l0gicgate: Can you remove this blank line please.

l0gicgate (changes requested):

l0gicgate (approved):

PR #2865, opened 2019-10-17
php-di/slim-bridge added in composer -> suggest

shahariaazam:
We need to give proper hints where possible for the developer specially who are newbie. So they can get started with appropriate tools and library.
For the documentation `http://www.slimframework.com/docs/v4/concepts/di.html`. Most of the people will start using implementing Container without installing any container library. So I guess, it would be great to show suggestions for the developer to install DI bridge  to work with container.


l0gicgate:
We should probably suggest a container library like PHP-DI. The library you’re suggesting is complementary.

PR #2884, opened 2019-11-17
Added github action yml

dgrng:
Added github action php.yml


l0gicgate:
We're currently using Travis for this. Maybe in the future we can move to github actions.

PR #2885, opened 2019-11-26
Replace abandoned `container-interop/container-interop` package with `psr/container`

Ayesh:
PSR-11 `container-interop/container-interop` package is replaced by `psr/container`, but Slim v3 branch still uses it.

There were just a few uses of interfaces internally used by Slim, and this PR replaces them with the `psr/container` counterparts. 

Quoting from `container-interop/container-interop` project page:

> ## Deprecation warning!
> 
> Starting Feb. 13th 2017, container-interop is officially deprecated in favor of [PSR-11](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md).
> Container-interop has been the test-bed of PSR-11. From v1.2, container-interop directly extends PSR-11 interfaces.
> Therefore, all containers implementing container-interop are now *de-facto* compatible with PSR-11.
> 
> - Projects implementing container-interop interfaces are encouraged to directly implement PSR-11 interfaces instead.
> - Projects consuming container-interop interfaces are very strongly encouraged to directly type-hint on PSR-11 interfaces, in order to be compatible with PSR-11 containers that are not compatible with container-interop.
> 
> Regarding the delegate lookup feature, that is present in container-interop and not in PSR-11, the feature is actually a design pattern. It is therefore not deprecated. Documentation regarding this design pattern will be migrated from this repository into a separate website in the future.

I do not think this will be a BC break (in semantic versioning sense) because the `container-interop` package implements `psr/container` internally.

Travis CI tests: https://travis-ci.org/Ayesh/Slim/builds/617400679

Thank you.

### DIFF
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -9,7 +9,7 @@ namespace Slim;
 
 use ArrayAccess;
 use Psr\Container\ContainerInterface;
-use Psr\Container\ContainerExceptionInterface as ContainerException;
+use Psr\Container\ContainerExceptionInterface;
 use InvalidArgumentException;
 use Pimple\Container as PimpleContainer;
 use Slim\Exception\ContainerException as SlimContainerException;
@@ -95,9 +95,10 @@ class Container extends PimpleContainer implements ContainerInterface
      * @return mixed
      *
      * @throws InvalidArgumentException         Thrown when an offset cannot be found in the Pimple container
-     * @throws SlimContainerException           Thrown when an exception is not an instance of ContainerException
+     * @throws SlimContainerException           Thrown when an exception is
+     *         not an instance of ContainerExceptionInterface
      * @throws ContainerValueNotFoundException  No entry was found for this identifier.
-     * @throws ContainerException               Error while retrieving the entry.
+     * @throws ContainerExceptionInterface      Error while retrieving the entry.
      */
     public function get($id)
     {
@@ -153,9 +154,10 @@ class Container extends PimpleContainer implements ContainerInterface
      * @return mixed
      *
      * @throws InvalidArgumentException         Thrown when an offset cannot be found in the Pimple container
-     * @throws SlimContainerException           Thrown when an exception is not an instance of ContainerException
+     * @throws SlimContainerException           Thrown when an exception is not
+     *         an instance of ContainerExceptionInterface
      * @throws ContainerValueNotFoundException  No entry was found for this identifier.
-     * @throws ContainerException               Error while retrieving the entry.
+     * @throws ContainerExceptionInterface      Error while retrieving the entry.
      */
     public function __get($name)
     {
--- a/Slim/Container.php
+++ b/Slim/Container.php
@@ -8,8 +8,8 @@
 namespace Slim;
 
 use ArrayAccess;
-use Interop\Container\ContainerInterface;
-use Interop\Container\Exception\ContainerException;
+use Psr\Container\ContainerInterface;
+use Psr\Container\ContainerExceptionInterface as ContainerException;
 use InvalidArgumentException;
 use Pimple\Container as PimpleContainer;
 use Slim\Exception\ContainerException as SlimContainerException;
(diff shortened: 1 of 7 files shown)


Slim/Container.php
```diff
@@ -8,8 +8,8 @@
 namespace Slim;
 
 use ArrayAccess;
-use Interop\Container\ContainerInterface;
-use Interop\Container\Exception\ContainerException;
+use Psr\Container\ContainerInterface;
+use Psr\Container\ContainerExceptionInterface as ContainerException;
```
jacekkarczmarczyk: What is the reason for aliasing? I think the word `Interface` should still be there

Ayesh: Thanks for your review. 

In `psr/container`, the exception interface is named `ContainerExceptionInterface`. There are some code in the container class that refers to the `container-interop` pacakge as `ContainerException`. With the aliasing, we were skipping changes in a few other places in the file. 

I have pushed another commit with changes in the aliases. We not use the `psr/container`-provided Exception interface directly.

l0gicgate:
I think this is fine. @akrabat @adriansuter  I don't foresee a BC break from this am I blind? 😂

akrabat:
LGTM. I'm in favour anyway! #YOLO 😂

l0gicgate (approved):

Ayesh:
Awesome thanks a lot @jacekkarczmarczyk , @akrabat and @l0gicgate!

lcharette:
Note this introduced a breaking change in our package (UserFrosting). Since we depend on Slim V3, we referenced `container-interop/container-interop` in our code. 

We've already planned to switch to PSR for out next minor version, but in the meantime, we have to lock Slim version so it doesn't impact our users.

Ayesh:
@lcharette would it work if you directly require `container-interop/container-interop` package in your `composer.json`? If you were running slim `v3.12`, you are already using `psr/container` because `container-interop` requires it.

lcharette:
@Ayesh Yes and no. Works with Composer. But no, because `Slim\Container` still doesn't implement the right instance anymore (Our class expected Slim\Container to be an instance of `container-interop/container-interop`). 

In any case, updating our composer.json can be considered a Breaking Change IMO.

akrabat:
Probably should have considered that it would be type hinted against. I saw it as an implementation detail.

Harumaro:
We had an issue following the new minor version update [from 3.12.2 to 3.12.3](https://github.com/slimphp/Slim/compare/3.12.2...3.12.3) and had to change our namespace imports for `ContainerInterface` from `Interop` to `Psr`. This definitely broke our BC and I wouldn't have expected this kind of change on a minor tag 😟... just a heads up.

l0gicgate:
I had a feeling this was going to happen, unfortunately this sounds like a downstream issue that's not under the scope of this package. Also, having abandoned packages are a security issue. You can always lock the Slim version to 3.12.2, this branch is in maintenance mode only anyway

Ayesh:
@Harumaro I do apologize for the BC breaks (I opened the PR 😊). 

Similar to what Rob and Pierre said, I was under the assumption that the use of `container-interop` package was contained within Slim, and downstream packages should not be using it in favor of the `psr/container` package.

lcharette:
It should be noted that while Interop was deprecated since Feb. 13th 2017, the "abandoned" status on Packagist/Composer is fairly recent, which didn't help.

SvenRtbg:
Happy to see this getting released. Yes, it is a BC break technically, but one that can get resolved very quickly because the container-interop interface is exactly the same as the psr/container interface, except for the name. 

And what could be done about it? Version 4 has already been released, so that's not an option. Minor version bump? Same BC issue. Limiting the maximum version if you have issues, or adapt and rename your interface consumption in the code. 

Thanks again!

Dillon-Brown:
> It should be noted that while Interop was deprecated since Feb. 13th 2017, the "abandoned" status on Packagist/Composer is fairly recent, which didn't help.

Agree with this, I only just managed to track this issue down today which caused a BC change for us. While it may be resolved quickly, it's not immediately alarming for some older legacy applications with LTS support to still be supporting things like PHP 5.5 which is likely going to be throwing so many unavoidable composer issues at you that an additional "abandoned" status on Packagist/Composer isn't going to make developers drop everything and look to see if SemVer is being followed or not.....

Issue #2883, opened 2019-11-16
Some reserved keywords are not allowed as route placeholders


adriansuter:
The following route handler fails and throws an exception:
``` php
$app->get('/{route}', function (Request $request, Response $response, array $args): Response {
    $context = RouteContext::fromRequest($request);

    return $response;
});
```

That is because the default invocation strategy `\Slim\Handlers\Strategies\RequestResponse` puts the route arguments to the request attributes. So the argument `route` would overwrite the attribute with the same key set in the routing middleware.

The exception thrown is even misleading: "Uncaught RuntimeException: Cannot create RouteContext before routing has been completed"

Do you have an idea, how we could avoid that or should we add a hint to the docs?

Currently I guess the following keywords are reserved:
- `route`
- `routeParser`
- `routingResults`
- `basePath`

adriansuter:
Maybe we could check in `\Slim\Handlers\Strategies\RequestResponse::__invoke` whether a request attribute already exists and if so, we could throw a runtime exception "Route argument x is now allowed".

shahariaazam:
No, I don't think so. Here what I just tested after seeing your issue.

```php
$app->get('/test/{route}', function (Slim\Psr7\Request $request, Slim\Psr7\Response $response, array $args): Response {
    $response->getBody()->write($request->getAttribute("route"));
    return $response;
});
```

```
GET /test/2
```
Returns `2`

It's working perfectly. I think there shouldn't be any reserved word in route placeholder. Also I couldn't find anything about it in docs.

So can you please post your full trace to take a look to get any hints?

adriansuter:
@shahariaazam Now put the following line into your route handler:
`$context = RouteContext::fromRequest($request); `.

You will get an error. But if you change the placeholder name, everything is working as expected.

My explanations above are not based on the docs, but on the source code.

odan:
I tried the example from @adriansuter and I get the same error: 

```
Type: TypeError
Code: 0
Message: Argument 1 passed to Slim\Routing\RouteContext::__construct() must implement 
interface Slim\Interfaces\RouteInterface or be null, string given, called in 
RouteContext.php on line 34
File: RouteContext.php
Line: 63
```

The same error occurs for route placeholders such as: `routeParser`, `routingResults` and `basePath`.

Unfortunately, it is a general problem when route argument names overlap with request attribute names.

l0gicgate:
I feel like we should have underscored those:
- `__route__`
- `__routeParser__`
[...]

PR #2889, opened 2019-11-28
Change route context attribute names

adriansuter:
PR for #2883 

The attribute names are even configurable - although I am not sure if that should be an option. If not, let me know - I will remove all the unnecessary calls to static methods and instead hard code the attribute names.

If yes, maybe we could find another solution than using static methods and properties in the `RouteContext` class - I am not totally happy with this *static* solution.

### DIFF
--- a/Slim/Routing/RouteContext.php
+++ b/Slim/Routing/RouteContext.php
@@ -16,67 +16,13 @@ use Slim\Interfaces\RouteParserInterface;
 
 final class RouteContext
 {
-    private static $routeAttributeName = '__route__';
+    public const ROUTE_ATTRIBUTE_NAME = '__route__';
 
-    private static $routeParserAttributeName = '__routeParser__';
+    public const ROUTE_PARSER_ATTRIBUTE_NAME = '__routeParser__';
 
-    private static $routingResultsAttributeName = '__routingResults__';
+    public const ROUTING_RESULTS_ATTRIBUTE_NAME = '__routingResults__';
 
-    private static $basePathAttributeName = '__basePath__';
-
-    public static function setRouteAttributeName(string $routeAttributeName): void
-    {
-        self::$routeAttributeName = $routeAttributeName;
-    }
-
-    public static function setRouteParserAttributeName(string $routeParserAttributeName): void
-    {
-        self::$routeParserAttributeName = $routeParserAttributeName;
-    }
-
-    public static function setRoutingResultsAttributeName(string $routingResultsAttributeName): void
-    {
-        self::$routingResultsAttributeName = $routingResultsAttributeName;
-    }
-
-    public static function setBasePathAttributeName(string $basePathAttributeName): void
-    {
-        self::$basePathAttributeName = $basePathAttributeName;
-    }
-
-    public static function attachRoute(ServerRequestInterface $request, RouteInterface $route): ServerRequestInterface
-    {
-        return $request->withAttribute(self::$routeAttributeName, $route);
-    }
-
-    public static function grabRoute(ServerRequestInterface $request): ?RouteInterface
-    {
-        return $request->getAttribute(self::$routeAttributeName);
-    }
-
-    public static function attachRouteParser(
-        ServerRequestInterface $request,
-        RouteParserInterface $routeParser
-    ): ServerRequestInterface {
-        return $request->withAttribute(self::$routeParserAttributeName, $routeParser);
-    }
-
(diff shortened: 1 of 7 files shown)


adriansuter:
# Proposition

Pass an optional `RouteContextFactoryInterface` to the App. In `App::handle()` use that factory to create a route context and attach it to the request.

So we need
```php
RouteContextFactoryInterface::createRouteContext():RouteContextInterface;
RouteContext::attach(ServerRequestInterface, RouteContextInterface):ServerRequestInterface;
```

Then, whenever we need that route context, we can grab it from the `$request`.

```php
RouteContext::fromRequest(ServerRequestInterface):RouteContextInterface
```

The `RouteContextInterface` has methods to set and get route parser, routing results, base path and route.

Warning: A lot of phpunit test cases would need rewrites.

Slim/Routing/RouteContext.php
```diff
@@ -16,16 +16,78 @@
...
 {
+    private static $routeAttributeName = '__route__';
+
+    private static $routeParserAttributeName = '__routeParser__';
+
+    private static $routingResultsAttributeName = '__routingResults__';
+
+    private static $basePathAttributeName = '__basePath__';
+
+    public static function setRouteAttributeName(string $routeAttributeName): void
```
l0gicgate: I don't think these setters are necessary since it's for internal purposes.

Slim/Routing/RouteContext.php
```diff
@@ -16,16 +16,78 @@
 
 final class RouteContext
 {
+    private static $routeAttributeName = '__route__';
```
l0gicgate: We should make these public constants so we can use them in our test cases, makes it easier to refactor if needed.

tests/Middleware/RoutingMiddlewareTest.php
```diff
@@ -41,17 +41,17 @@ public function testRouteIsStoredOnSuccessfulMatch()
         $responseFactory = $this->getResponseFactory();
         $mw = (function (ServerRequestInterface $request) use ($responseFactory) {
             // route is available
-            $route = $request->getAttribute('route');
+            $route = $request->getAttribute('__route__');
```
l0gicgate: as mentioned above, we should use public constants on the `RouteContext` object.

Slim/Routing/RouteContext.php
```diff
@@ -16,16 +16,78 @@
...
+    {
+        self::$routingResultsAttributeName = $routingResultsAttributeName;
+    }
+
+    public static function setBasePathAttributeName(string $basePathAttributeName): void
+    {
+        self::$basePathAttributeName = $basePathAttributeName;
+    }
+
+    public static function attachRoute(ServerRequestInterface $request, RouteInterface $route): ServerRequestInterface
```
l0gicgate: If we use constants we can get rid of all the `attachX` methods also

l0gicgate (changes requested):
Let's remove the `_ATTRIBUTE_NAME` suffix from the constants, then it's good to go!

adriansuter:
Good to go.

l0gicgate (approved):

jacekkarczmarczyk:
I'm not familar with v4 yet so I wasn't sure if this attribute names where a public this, but apparently they are (https://github.com/slimphp/Slim/issues/2892), so changing them is definitely a breaking change. 

I'd rather see these names as configurable thing (with default names unchanged, `'route'` etc) in `RouteContextConfig` or something like this, which could be injected to classes that use the attributes. Probably  there could be also `RouteContextFactory` that would have `RouteContextConfig` injected in the constructor, otherwise you should pass the config in `RouteContext::fromRequest` arguments

N1ghteyes:
can  confirm that this is a breaking change as per my issue - referencing the docs is one thing (my original code came from v4 docs - one of the basic examples i think) but if possible i would like to see Slim avoid breaking changes in minor releases.

l0gicgate:
@N1ghteyes

> can confirm that this is a breaking change as per my issue

Your issue is unrelated to this. As I mentioned, pattern matching is done via FastRoute, which isn't code that was touched by this PR. 

> i would like to see Slim avoid breaking changes in minor releases.

We do too however using the request attributes directly isn't the correct way to access the route/routing results/base path/etc. While it was originally stated that way in the earlier version of the v4 docs, this was a mistake and has since been changed. Using the `RouteContext` object which was in the original 4.0.0 release would have avoided any potential breaks.

Issue #2746, opened 2019-06-28
4.x - Error renderers do not handle the displayErrorDetails uniformly


adriansuter:
The output of the error renderers can be controlled using `$displayErrorDetails`. But the slim error renderers would handle that differently.

**HtmlErrorRenderer**
- `false`: Standard error message without any information about the exception.
- `true`: Details like type, code, message, file, line and stack trace of the exception thrown. No information about previous exception(s).

**JsonErrorRenderer**
- `false`: The error message of the exception thrown.
- `true`: Details like type, code, message, file and line of the exception thrown as well as of the previous exception(s). But no stack trace for any exception(s).

**PlainTextErrorRenderer** (See also #2744)
- `false`: Details like type, code, message, file, line and stack trace of the exception thrown.
- `true`: Details like type, code, message, file, line and stack trace of the exception thrown as well as of all previous exception(s).

**XmlErrorRenderer**
- `false`: The message of the exception thrown.
- `true`: Additionally type, code, message, file and line of the exception thrown as well as of all previous exception(s). But no stack trace for any exception(s).

What do you think the expected behaviour should be?

l0gicgate:
We should try to keep it as consistent as possible if we can.

The HTML and Plain Text renderers should behave the same and the JSON and XML should behave the same.

I'm not sure what we should consider "error details". To me that would mean file, line and previous exception. Basic information should be Code (if other than 0) and Message.

What do you think?

adriansuter:
In production it might be better not to reveal too much of details. Maybe the code is enough. I wouldn't make any difference between the content types.

I suggest: If `false`, then the code and a standard error message. If `true`, then the details (type, code, message, file, line) of every exception and the full stack trace of the first exception.

l0gicgate:
Yes I agree. Let’s just keep it uniform across all content types.

shahariaazam:
> Yes I agree. Let’s just keep it uniform across all content types.

I agree too. Error should be rendered uniformly.

rotexdegba:
Below is a function I use for displaying exceptions and errors in my applications, It may be helpful (the **PHP_EOL** can be substituted with **\<br\>** where applicable):
 
```php
function getThrowableAsStr(\Throwable $e, string $eol=PHP_EOL): string {

        $previous_throwable = $e;
        $message = '';

        do {
            $message .= "Exception / Error Code: {$previous_throwable->getCode()}"
                . $eol . "Exception / Error Class: " . get_class($previous_throwable)
[...]

PR #2896, opened 2019-12-12
Render errors uniformly

adbrvn:
In this PR solved several problems 

* Render error uniformly in different formats (#2746)
* Render http exceptions as custom error pages
* Hide backtraces in plain format too

### DIFF
--- a/Slim/Error/AbstractErrorRenderer.php
+++ b/Slim/Error/AbstractErrorRenderer.php
@@ -9,7 +9,9 @@ declare(strict_types=1);
 
 namespace Slim\Error;
 
+use Slim\Exception\HttpException;
 use Slim\Interfaces\ErrorRendererInterface;
+use Throwable;
 
 /**
  * Abstract Slim application error renderer
@@ -19,4 +21,31 @@ use Slim\Interfaces\ErrorRendererInterface;
  */
 abstract class AbstractErrorRenderer implements ErrorRendererInterface
 {
+    /**
+     * @var string
+     */
+    protected $defaultErrorTitle = 'Slim Application Error';
+
+    /**
+     * @var string
+     */
+    protected $defaultErrorDescription = 'A website error has occurred. Sorry for the temporary inconvenience.';
+
+    protected function getErrorTitle(Throwable $exception): string
+    {
+        if ($exception instanceof HttpException) {
+            return $exception->getTitle();
+        }
+
+        return $this->defaultErrorTitle;
+    }
+
+    protected function getErrorDescription(Throwable $exception): string
+    {
+        if ($exception instanceof HttpException) {
+            return $exception->getDescription();
+        }
+
+        return $this->defaultErrorDescription;
+    }
 }
--- a/tests/Error/AbstractErrorRendererTest.php
+++ b/tests/Error/AbstractErrorRendererTest.php
@@ -62,23 +62,23 @@ class AbstractErrorRendererTest extends TestCase
 
     public function testHTMLErrorRendererRenderHttpException()
     {
-        $exceptionTitle = 'exception-title';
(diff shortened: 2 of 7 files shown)


l0gicgate:
This is great, thank you for this contribution.

tests/Error/AbstractErrorRendererTest.php
```diff
@@ -10,12 +10,16 @@
 namespace Slim\Tests\Error;
 
 use Exception;
+use function json_decode;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Http\Message\ServerRequestInterface;
```
adriansuter: The `ServerRequestInterface` could be removed.

tests/Error/AbstractErrorRendererTest.php
```diff
@@ -10,12 +10,16 @@
 namespace Slim\Tests\Error;
 
 use Exception;
+use function json_decode;
```
adriansuter: This `use function json_decode;` can be removed, as Slim is not using root namespaced php function calls.

adriansuter (changes requested):
Great, thanks. Only two minor issues.

Slim/Error/AbstractErrorRenderer.php
```diff
@@ -19,4 +21,21 @@
  */
 abstract class AbstractErrorRenderer implements ErrorRendererInterface
 {
+    protected function getErrorTitle(Throwable $exception): string
+    {
+        if ($exception instanceof HttpException) {
+            return $exception->getTitle();
+        }
+
+        return 'Slim Application Error';
```
l0gicgate: We should actually make this a protected variable so it can be overridden without having to override the entire method

tests/Error/AbstractErrorRendererTest.php
adbrvn: fixed

tests/Error/AbstractErrorRendererTest.php
adbrvn: fixed

Slim/Error/AbstractErrorRenderer.php
adbrvn: fixed

adriansuter:
Travis fails because of `prophecy` - see https://github.com/slimphp/Slim/issues/2901

Slim/Error/AbstractErrorRenderer.php
```diff
@@ -19,4 +21,32 @@
...
+        return $this->defaultErrorTitle;
+    }
+
+    protected function getErrorDescription(Throwable $exception): string
+    {
+        if ($exception instanceof HttpException) {
+            return $exception->getDescription();
+        }
+
+
```
l0gicgate: remove this extra white line please

l0gicgate:
This PR is good to go. We just need to wait for #2901 to merge

adriansuter (approved):

l0gicgate:
@adriansuter we could prophesize the tests also instead of using mockbuilder for uniformity. We have time since we have to wait anyway. What do you think? I or you can do it since he probably doesn't know how to

adriansuter:
@l0gicgate Why not, uniformity is always a good thing. I think we will have plenty of time indeed - and then there is also Christmas and New Year coming. Feel free to prophesize the mocks.

Slim/Error/AbstractErrorRenderer.php
adbrvn: fixed

l0gicgate:
I'm locking Prophecy version to 1.9 and merging this. We will address issue in #2901. Won't release 4.4.0 until that's resolved.

PR #2900, opened 2019-12-17
Added ability to add handled exceptions as an array

nickdnk:
Hey

As an addition to my latest PR I found this to be helpful as well. Some people may want to handle multiple exceptions the same way even though they are not subclasses of the same parent. This solves that problem without breaking anything at all.

Personally I handle `HttpNotFoundException` and `HttpMethodNotAllowedException` the same way in my API. Just an example.

This may also be overdoing it, so let me know.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,11 @@
     },
     "autoload": {
         "psr-4": {
-            "Slim\\": "Slim",
+            "Slim\\": "Slim"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
             "Slim\\Tests\\": "tests"
         }
     },


l0gicgate:
I think this is a good addition, but I do think we could make a separate method instead of mixing types.

- `setErrorHandler()` - Single `string $type`
- `setMultipleErrorHandlers` - Multiple `array $types` 

I'm not married to the method name, if anyone has a better name I'm up for it. I would like to preserve the typing though.

nickdnk:
> I think this is a good addition, but I do think we could make a separate method instead of mixing types.
> 
> * `setErrorHandler()` - Single `string $type`
> * `setMultipleErrorHandlers` - Multiple `array $types`
> 
> I'm not married to the method name, if anyone has a better name I'm up for it. I would like to preserve the typing though.

What's the advantage of a new method? Wouldn't it be more confusing? The phpdocs clearly indicate how it can be used like this. I can go either way, but it doesn't seem unusual to me that a method can take an array of parameters of the same type as the original parameter.

l0gicgate:
> it doesn't seem unusual to me that a method can take an array of parameters of the same type as the original parameter.

~~I'd rather have it as separate methods so we can enforce the type.~~
~
~~Let's do `mapErrorHandlers`~~

I misunderstood the usage of the array.. I thought we were talking about doing an associative array of handlers e.g.:
```php
$handlers = [
    Exception::class => function () { ... },
    RuntimeException::class => function () { ... },
];
```

I'm merging this.

PR #2902, opened 2019-12-19
Fix prophecies

### LINKED ISSUE #2901: Not an issue, just a note - tests fail
PHPUnit seems to fail when using the currently newest `phpsec/prophecy` version 1.10.0.

See https://github.com/phpspec/prophecy/issues/457 and https://github.com/phpspec/prophecy/issues/458

adriansuter:
Fixes #2901.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "nyholm/psr7": "^1.1",
         "nyholm/psr7-server": "^0.3.0",
         "phpunit/phpunit": "^7.5",
-        "phpspec/prophecy": "^1.8",
+        "phpspec/prophecy": "^1.10",
         "phpstan/phpstan": "^0.11.5",
         "slim/http": "^0.7",
         "slim/psr7": "^0.3",
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -1429,6 +1429,13 @@ class AppTest extends TestCase
         $requestProphecy = $this->prophesize(ServerRequestInterface::class);
         $requestProphecy->getMethod()->willReturn('GET');
         $requestProphecy->getUri()->willReturn($uriProphecy->reveal());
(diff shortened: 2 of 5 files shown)


l0gicgate:
Should we update the composer file as well to `1.10`?

https://github.com/slimphp/Slim/blob/4.x/composer.json#L52

adriansuter:
Sure, I don't think anyone will get hurt.

froschdesign:
@adriansuter @l0gicgate 
First: great to see your quick reaction on this topic.
But so far it is not clear if it is a problem in Prophecy and / or a BC break. Maybe it would help if you joined the discussion to the [related issue report](https://github.com/phpspec/prophecy/issues/457). This could be helpful, also for others.

adriansuter:
I just wrote to the corresponding issue https://github.com/phpspec/prophecy/issues/458. Let's see and wait.

In the meantime, how about updating `composer.json` into
```
        "phpspec/prophecy": ">1.8 <1.10",
```

l0gicgate:
> "phpspec/prophecy": ">1.8 <1.10",

@adriansuter this is a good temporary solution. 

Using `Argument::is()` is also a good solution but it makes me worried about all the changes we have to do to all the other repos.

Thank you for the quick PR by the way. I think we should probably wait to merge this until we hear back on https://github.com/phpspec/prophecy/issues/457

t0mmy742:
[phpspec/prophecy#460](https://github.com/phpspec/prophecy/pull/460) is going to resolve this issue.

But, as mentionned in the PR, it is a best practice to use Argument::is() for strict comparison.

It is maybe better for Slim to use Argument::is() (we don't have that much changes to do in other repos).

l0gicgate:
Okay @t0mmy742 I'm going to merge this then.

PR #2904, opened 2019-12-22
Feature Abstract Controller

ABGEO:
**:tada: I implemented an Abstract controller feature.**

I created the abstract class `Slim\Controller\AbstractController`, which contains the basic functions for the controller.

**Now each controller must extend the `Slim\Controller\AbstractController`**.

## Methods

- `has()` - Test if the container can provide something for the given name.
- `get()` - Returns an entry of the container by its name.
- `getLogger()` - Get Logger.
- `getRequest()` - Get current request object.
- `getResponse()` - Get current response object.
- `resolveArg()` - Resolve current request arguments.
- `json()` - Return JSON response.
- `render()` - Render a view.

## Basic usage

```php
// app/routes.php

// ...
return function (App $app) {
    $app->get('/hello/{name}-{surname}', 'App\Controller\DefaultController:hello');
};
// ...
```

```php
// src/Controller/DefaultController.php

<?php
declare(strict_types=1);

namespace App\Controller;

use Slim\Controller\AbstractController;

class DefaultController extends AbstractController
{

    public function helloAction()
    {
        $name = $this->resolveArg('name');
        $surname = $this->resolveArg('surname');
        $age = $this->resolveArg('age', 18);

        return $this->json([$name, $surname, $age]);
    }
}
```

Render the view
```php
// src/Controller/DefaultController.php

// ...
return $this->render('hello.html.twig', [
     'name' => $name,
     'surname' => $surname,
     'age' => $age,
]);
// ...
```

### DIFF
--- a/Slim/Controller/AbstractController.php
+++ b/Slim/Controller/AbstractController.php
@@ -70,9 +70,9 @@ abstract class AbstractController
     {
         $name .= 'Action';
 
-        $this->request = $arguments[0];
-        $this->response = $arguments[1];
-        $this->args = $arguments[2];
+        $this->request = $arguments[0] ?? null;
+        $this->response = $arguments[1] ?? null;
+        $this->args = $arguments[2] ?? [];
 
         if (!method_exists($this, $name)) {
             throw new BadMethodCallException(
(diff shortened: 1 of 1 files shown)


l0gicgate:
This isn't something that we are interested in incorporating in this library. We also have something similar already in Slim-Skeleton repository.

https://github.com/slimphp/Slim-Skeleton/blob/master/src/Application/Actions/Action.php

Thank you for your contribution though.

PR #2905, opened 2019-12-23
PHP 7.2 as minimal dependency

t0mmy742:
As mentionned [here](https://www.php.net/supported-versions.php), PHP 7.1 reached its end of life 3 weeks ago.

With this PR, I proposed to update our composer.json to PHP ^7.2.
I updated PHPUnit to 8.5 (which support PHP 7.4) and corrected some deprecated functions.

I can send PR for other repos which need an update.

Maybe we could release a new version of Slim before merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .idea
+.phpunit.result.cache
 composer.lock
 phpunit.xml
 clover.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 language: php
 
-dist: trusty
-
 matrix:
   include:
-  - php: 7.1
   - php: 7.2
   - php: 7.3
+  - php: 7.4
     env: ANALYSIS='true'
   - php: nightly
 
@@ -14,7 +12,7 @@ matrix:
   - php: nightly
 
 before_script:
-- if [[ "$ANALYSIS" == 'true' ]]; then composer require php-coveralls/php-coveralls:^2.1.0 ; fi
+- if [[ "$ANALYSIS" == 'true' ]]; then composer require php-coveralls/php-coveralls:^2.2.0 ; fi
 - composer install -n
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ matrix:
   include:
   - php: 7.2
   - php: 7.3
-  - php: 7.4
     env: ANALYSIS='true'
+  - php: 7.4
   - php: nightly
 
   allow_failures:
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     "require": {
         "ext-json": "*",
         "nikic/fast-route": "^1.3",
-        "php": "^7.1",
+        "php": "^7.2",
         "psr/container": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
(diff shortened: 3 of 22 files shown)


t0mmy742:
One line is no longer covered. This problem only occurred with PHP 7.4 (not before).
In reality, it is covered 10 times. The line is not covered during generation by PHPUnit. I can't find if this is a problem with PHPUnit or something changes with PHP 7.4.

**EDIT :** Finally, it's a problem with Xdebug (see [sebastianbergmann/php-code-coverage#711](https://github.com/sebastianbergmann/php-code-coverage/issues/711)). I tested with another driver for code coverage (PCOV) and it gave me 100% code covered. Xdebug already has a fix, merged and ready for next version (2.9.1).
For now, we can choose to test on PHP 7.3, and wait Xdebug 2.9.1 to be released on Travis to test on PHP 7.4.

l0gicgate:
I have a feeling that this is a BC change. @akrabat can we bump up minimum PHP version via a minor release or does this require a new major?

t0mmy742:
@l0gicgate I wanted to see how other projects were doing. I came across several links including [this one](https://blog.whiteoctober.co.uk/index.html%3Fp=3415.html) (which gives two links to the Doctrine website and the Zend Github). And I agree, here we are not touching our code, but just modifying dependencies. So there is no BC change. We should only increase minor version.

Just a quote from [here](https://github.com/zendframework/zend-diactoros/issues/271#issuecomment-330558976) :
> Your ability to upgrade to the latest version is not an indication of whether or not the version is still backwards compatible.

mathmarques:
IMHO this is a BC. This change will break any `composer install on PHP` <= 7.1.
As a comparsion Slim 3.X [kept the PHP 5.5](https://github.com/slimphp/Slim/blob/3.x/composer.json#L31) min-req even after EOL.

t0mmy742:
This will not break anything with composer if you have PHP 7.1.x. It will just not install any update of Slim while you have an outdated version of PHP, but get only the last version of Slim compatible with PHP 7.1 (even if you are doing your first composer install).

Slim 3.x is not recommended to use. Everybody should migrate to Slim 4 if they can.

akrabat:
My position on this has moved from an absolute "no" to be more "on the fence". 

It's true that a composer update/install will ignore the PHP 7.2 version if it is run on PHP 7.1 (or `platform` is set to PHP 7.1. in composer.json), so people on PHP 7.1. can still use an older, presumably more buggy, version of Slim 4.

However, many people use the version of PHP that comes with their hosting supplier or that which comes with the LTS version of their distribution. This means that a proportion of our user base will be unable to get any new versions including security fixes without moving hosts or installing an unsupported 3rd party PHP binary onto their system.

I'm not sure how comfortable I am leaving those users on a buggy version of Slim and unable to upgrade to a bug-fixed version, especially if they are on a distro that (supposedly) ensures that the shipped version of PHP is kept secure.

Moving to PHP 7.2 as a minimum would cut out Debian 9, but not Ubuntu 18.04 or RHEL 8 as both of those shipped with 7.2, so moving to PHP 7.2. as a minimum dependency isn't too traumatic by the nature of timing. _However_ I am much more uncomfortable with raising to a PHP 7.3 minimum and cutting out Ubuntu 18.04 and RHEL, even in a year's time as people don't upgrade LTS versions very fast.

If we do update the minimum PHP version, then it should be in a minor version so that we at least have the option of issue bug-fixes and security updates for the previous min PHP version with patch versions on the previous minor version.

l0gicgate:
So @akrabat if we release this via `4.4` we could still release security patches via `4.3.x`? I think that's a pretty good compromise right?

akrabat:
> So @akrabat if we release this via `4.4` we could still release security patches via `4.3.x`? I think that's a pretty good compromise right?

Yeah.

l0gicgate (approved):

PR #2906, opened 2019-12-24
Fix #2890

### LINKED ISSUE #2890: PHPDoc of Slim\Http\Request::getParams it's wrong
Hello, In my project we use [PHPStan](https://github.com/phpstan/phpstan) for the static analysis, and throw me an error from the method `getParams` from `\Slim\Http\Request` because say the method can return null or array.

But i reviewd the two returns of the method and in both only return an array, that are or `$onlyParams` or `$params`, is the PHPDoc `return`wrong?

t0mmy742:
Closes #2890

### DIFF
--- a/Slim/Http/Request.php
+++ b/Slim/Http/Request.php
@@ -1186,7 +1186,7 @@ class Request extends Message implements ServerRequestInterface
      *
      * @param array|null $only list the keys to retrieve.
      *
-     * @return array|null
+     * @return mixed[]
      */
     public function getParams(array $only = null)
     {


Slim/Http/Request.php
```diff
@@ -1186,7 +1186,7 @@ public function getQueryParam($key, $default = null)
      *
      * @param array|null $only list the keys to retrieve.
      *
-     * @return array|null
+     * @return mixed[]
```
l0gicgate, line 1189: wouldn't it be an array of strings only? It can't be anything else since it's parsed from a string.

```php
string[]|null
```

jacekkarczmarczyk: What `?foo[]=1&foo[]=2` will be parsed to?

l0gicgate:
```php
string|string[]|null
```

I'm not sure what all the combinations are to be honest, I'd rather not use `mixed` if we can define all of them instead.

t0mmy742: It can't return null. No param will return an empty array.
This array can contains string and/or arrays.

?foo[]=1&foo[]=2 will be parsed to [foo[1, 2]]

l0gicgate: > It can’t return null

What if the key doesn’t exist?

t0mmy742: `public function getParams(array $only = null)`

If `only` is not null, and values in `only` does not exist, `getParams()` return an empty array.

l0gicgate: My bad @t0mmy742 the diff's header had me confused for a second. This is correct.

l0gicgate (approved):

PR #2907, opened 2019-12-31
PSR-12

t0mmy742:
PSR-12 expanded and replaced PSR-2. Can Slim switch to PSR-12 ?
[PSR-12: Extended Coding Style](https://www.php-fig.org/psr/psr-12/)

I also add fully-qualified function calls (`use function xxxx`). I didn't add 5 functions (`headers_sent`, `header`, `connection_status`, `is_readable` and `is_writable`) since we override them during tests (used on `ResponseEmitter` and `Routing/RouteCollector` classes).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ pull request for each branch. This allows me to review and pull in new features 
 
 ## Style Guide
 
-All pull requests must adhere to the [PSR-2 standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md).
+All pull requests must adhere to the [PSR-12 standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-12-extended-coding-style-guide.md).
 
 ## Unit Testing
 
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Slim Framework (https://slimframework.com)
  *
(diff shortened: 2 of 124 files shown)


t0mmy742:
> I didn't add 5 functions (`headers_sent`, `header`, `connection_status`, `is_readable`and `is_writable`) since we override them during tests (used on `ResponseEmitter` and `Routing/RouteCollector` classes).

To complete this PR, I created a "sub-branch". You can have a diff just [here](https://github.com/t0mmy742/Slim/compare/cc5a3d4...ad88a41). I'm using uopz extension to change the return value of our 5 functions during tests.
Let me know what you prefer : using uopz extension (it does NOT slow down our tests) or never call our functions with fully-qualified names. Then, I can merge or not this "sub-branch".

l0gicgate:
This also effectively resolves #2881 right? I'm in favor of this. Will need to do on Slim-Http, Slim-Psr7 and Slim-Skeleton

I don't see any added value from using the UOPZ extension, I suppose it makes the code more uniform. What do you think @adriansuter? This PR works both ways.

adriansuter:
Yes, let's move to PSR-12.

Concerning the override, I once started a project that could be used to override global (root) namespace functions. But it is not completed yet.

https://github.com/adriansuter/php-autoload-override

The idea is, that we can define classes for which we can define functions that should be overridden. The tool then intercepts the loading of these classes (using stream wrappers) and overwrites the function calls. It actually works.

t0mmy742:
Yes it closes #2881 

@adriansuter your library is amazing, and really easy to use. It is what we need here !

Do you want to keep all that code in the bootstrap.php file ? Or move code in Assets folder ?

l0gicgate:
I think what we can do for now is merge this and we can decide/finish the rest in a different PR. The scope of this PR is already pretty big.

Issue #2911, opened 2020-01-02
Laminas PSR17 Factory


adriansuter:
Should we add `\Slim\Factory\Psr17\LaminasDiactorosPsr17Factory`?

See https://github.com/adriansuter/Slim/tree/patch-laminas and https://github.com/adriansuter/Slim/blob/patch-laminas/Slim/Factory/Psr17/LaminasDiactorosPsr17Factory.php

Do you think we should remove `\Slim\Factory\Psr17\ZendDiactorosPsr17Factory`? I think that would be a BC, wouldn't it?

l0gicgate:
We should add Laminas and leave Zend until the next major.

t0mmy742:
Maybe we can deprecate Zend and inform users with phpdoc to move to Laminas.

l0gicgate:
@t0mmy742 we should do that yes

PR #2912, opened 2020-01-02
Add Laminas PSR17 Factory

adriansuter:
PR for #2911

### DIFF
--- a/tests/Factory/AppFactoryTest.php
+++ b/tests/Factory/AppFactoryTest.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Slim\Tests\Factory;
 
 use Http\Factory\Guzzle\ResponseFactory as GuzzleResponseFactory;
+use Laminas\Diactoros\ResponseFactory as LaminasResponseFactory;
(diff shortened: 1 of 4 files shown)


tests/Factory/AppFactoryTest.php
```diff
@@ -34,7 +36,7 @@
 use Slim\Routing\RouteCollector;
 use Slim\Tests\Mocks\MockPsr17FactoryWithoutStreamFactory;
 use Slim\Tests\TestCase;
-use Zend\Diactoros\ResponseFactory as ZendDiactorosResponseFactory;
+use Zend\Diactoros\ResponseFactory as ZendResponseFactory;
```
l0gicgate: We should keep the Diactoros keyword in both Zend/Laminas.

- `ZendDiactorosResponseFactory`
- `LaminasDiactorosResponseFactory`

adriansuter: Sure. So I renamed both in `\Slim\Tests\Factory\ServerRequestCreatorFactoryTest` too.

l0gicgate (approved):

PR #2916, opened 2020-01-05
Rename phpcs.xml to phpcs.xml.dist

ghost:


l0gicgate (approved):

l0gicgate (changes requested):
~~We need to add this filename to `.gitattributes` as well so it doesn't get exported during releases~~

ghost:
> We need to add this filename to `.gitattributes` as well so it doesn't get exported during releases

Already added?

l0gicgate (approved):

PR #2917, opened 2020-01-05
Update .editorconfig

ghost:


l0gicgate (approved):

l0gicgate:
This should also be done on all other repos. Slim-Psr7, Slim-Http, etc.

PR #2922, opened 2020-01-16
Create ci.yml for GIthub Actions

Smolevich:
- Add ci.yml for supporting GIthub Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,70 @@
+name: Run ci
+
+on:
+  push:
+  pull_request:
+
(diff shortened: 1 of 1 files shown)


Smolevich:
@l0gicgate can you review please?

l0gicgate:
@Smolevich I don’t see the benefits in this lateral move to GA to be honest. This just means that we have to do it on all the other repos and it has no benefits.

Smolevich:
I think that benefits are exist, but if you try in another projects and don't find benefits for self, i don't want to discuss

PR #2937, opened 2020-04-03
Add support for PSR-3

### LINKED ISSUE #2930: PSR-3 support
## Introduction

The `logError()` method of the `Slim\Handler\ErrorHandler` class currently [calls](https://github.com/slimphp/Slim/blob/2058eea946d8569323104d188814c0ba2b1d1d2d/Slim/Handlers/ErrorHandler.php#L314) the native `error_log()` function directly.

Overriding this behavior is rather tedious, as it requires:

* Subclassing `Slim\Handler\ErrorHandler` to override `logError()`
* Duplicating a lot of the [boilerplate](https://github.com/slimphp/Slim/blob/2058eea946d8569323104d188814c0ba2b1d1d2d/Slim/App.php#L154) contained in the `addErrorMiddleware()` method of `Slim\App` to manually add an instance of the `ErrorHandler` subclass

## Proposal

Modify `Slim\Handler\ErrorHandler` to receive a dependency that implements the `Psr\Log\LoggerInterface` interface from [PSR-3](https://www.php-fig.org/psr/psr-3/).

To maintain backward compatibility, make this dependency optional. If no value is specified for it, use a default implementation of the interface that implements the current behavior of calling the `error_log()` function.

Add the logger instance as a parameter of the `addErrorMiddleware()` method of `Slim\App`. Again, make this parameter optional to maintain backward compatibility.

At this point, customizing the logging behavior can be done by calling `addErrorMiddleware()` with an instance of the desired logger implementation.

## Rationale

Slim 4.x already heavily utilizes PSR standards to allow for customization including [PSR-7](https://www.php-fig.org/psr/psr-7/), [PSR-11](https://www.php-fig.org/psr/psr-11/), and [PSR-15](https://www.php-fig.org/psr/psr-15/). It also has a need for error logging, as shown in the code referenced in the previous section, and PSR-3 addresses this need.

Additionally, an attempt was made to implement PSR-3 support in Slim 3.x (see #525), so the same need has existed in at least one previous major version of Slim.

## Backward Incompatible Changes

No changes that break backward compatibility are strictly necessary to implement this change.

In a future release where backward compatibility breaks are allowed, the `$logErrors` parameter of the `__invoke()` method of `Slim\Handler\ErrorHandler` could be removed, since the logger itself could be configured to log messages or not (e.g. by using [`NullLogger`](https://github.com/php-fig/log/blob/5628725d0e4d687e29575eb41f9d5ee7de33a84c/Psr/Log/NullLogger.php)).

## Proposed Version(s)

It should be possible to include this change as suggested in Slim 4.x.

The suggestion of removing the `$logErrors` parameter could be done in Slim 5.x if desired.

## Implementation

No implementation has been created yet. However, depending on the reception of this proposal, I'd be happy to create a pull request myself.

elazar:
Fixes slimphp/Slim#2930

### DIFF
--- a/Slim/Logger.php
+++ b/Slim/Logger.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Slim;
+
+use Psr\Log\AbstractLogger;
+
+class Logger extends AbstractLogger
+{
+    /**
+     * @param mixed  $level
+     * @param string $message
+     * @param array  $context
+     *
+     * @return void
+     *
+     * @throws \Psr\Log\InvalidArgumentException
+     */
+    public function log($level, $message, array $context = [])
+    {
+        error_log($message);
+    }
+}
--- a/Slim/Logger.php
+++ b/Slim/Logger.php
@@ -1,5 +1,13 @@
 <?php
 
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
 namespace Slim;
 
 use Psr\Log\AbstractLogger;
--- a/Slim/Logger.php
+++ b/Slim/Logger.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 namespace Slim;
 
 use Psr\Log\AbstractLogger;
+use Psr\Log\InvalidArgumentException;
 
 class Logger extends AbstractLogger
 {
@@ -21,7 +22,7 @@ class Logger extends AbstractLogger
      *
      * @return void
      *
-     * @throws \Psr\Log\InvalidArgumentException
+     * @throws InvalidArgumentException
      */
     public function log($level, $message, array $context = [])
     {
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -19,6 +19,7 @@ use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UriInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Log\LoggerInterface;
 use ReflectionClass;
 use ReflectionProperty;
 use RuntimeException;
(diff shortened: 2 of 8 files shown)


Slim/Logger.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
```
l0gicgate, line 2: The header for this file should be as follows:
```php
<?php

/**
 * Slim Framework (https://slimframework.com)
 *
 * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
 */

declare(strict_types=1);
```

tests/AppTest.php
```diff
@@ -724,11 +725,14 @@ public function testAddErrorMiddleware()
         /** @var ResponseFactoryInterface $responseFactory */
         $responseFactory = $this->prophesize(ResponseFactoryInterface::class)->reveal();
 
+        /** @var LoggerInterface $logger */
+        $logger = $this->createMock(LoggerInterface::class);
```
l0gicgate: Let's stick with `$this->prophesize()` for the sake of cohesion

tests/Middleware/ErrorMiddlewareTest.php
```diff
@@ -274,4 +284,9 @@ public function testErrorHandlerHandlesThrowables()
 
         $this->expectOutputString('Oops..');
     }
+
+    private function getMockLogger(): LoggerInterface
```
l0gicgate: This method should be put at the top of the file before all the tests.

tests/Handlers/ErrorHandlerTest.php
```diff
@@ -391,4 +392,9 @@ public function testLogErrorRenderer()
         $writeToErrorLogMethod->setAccessible(true);
         $writeToErrorLogMethod->invoke($handler);
     }
+
+    private function getMockLogger(): LoggerInterface
```
l0gicgate: This method should be put at the top of the file before all the tests.

l0gicgate (changes requested):
I like this PR a lot, thank you for this contribution!

l0gicgate (approved):

Slim/Logger.php
```diff
@@ -0,0 +1,30 @@
...
+class Logger extends AbstractLogger
+{
+    /**
+     * @param mixed  $level
+     * @param string $message
+     * @param array  $context
+     *
+     * @return void
+     *
+     * @throws \Psr\Log\InvalidArgumentException
```
adriansuter: @l0gicgate Is there a rule for Slim concerning fully qualified class names in docblocks? Should they be imported?

adriansuter (approved):
Great piece of work. Just one comment/question. Otherwise, this is good to merge.

Slim/Logger.php
elazar: Strictly speaking, it may be possible to remove this particular annotation given that it was copied over from another logger implementation -- `NullLogger`, if memory serves -- but doesn't actually apply here.

In any event, if all that's required is to move the FQCN into a `use` directive and update this reference accordingly, I'm happy to do so.

l0gicgate: I missed that, we should import the exception along with all the other imports and change this to the shorthand `@throws InvalidArgumentException`. Let's keep the throw statement there.

elazar: Resolved in 40eebcd.

odan:
Great 👍 Thank you!

PR #2942, opened 2020-04-15
Fix PHPdoc for error handlers in ErrorMiddleware 

TiMESPLiNTER:
This PR adds type hints for `string` beside `callable` and `ErrorHandlerInterface` as the `$this->callableResolver->resolve` can resolve strings if they are a registered service id in the di container.

### DIFF
--- a/Slim/Middleware/ErrorMiddleware.php
+++ b/Slim/Middleware/ErrorMiddleware.php
@@ -132,7 +132,7 @@ class ErrorMiddleware implements MiddlewareInterface
      * occurs when processing the current request.
      *
      * @param string $type Exception/Throwable name. ie: RuntimeException::class
-     * @return callable|ErrorHandler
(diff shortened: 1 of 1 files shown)


l0gicgate:
I'm curious, is this necessary since a callable can in fact be a string?

TiMESPLiNTER:
> I'm curious, is this necessary since a callable can in fact be a string?

If the string is a callable function, you're right yes. But a container id is not a callable but just a string which later on gets resolved to a callable.

**Valid callable**
```php
function myHandler() { }
$errorMiddleware->setErrorHandler(HttpException::class, 'myHandler', true);
```

**Invalid callable**
```php
$errorMiddleware->setErrorHandler(HttpException::class, 'my_handlers_container_id', true);
```

l0gicgate:
Fair enough. Merging.

PR #2951, opened 2020-05-05
Minor optimizations in if() blocks

Ayesh:
A PR with a few micro-optimizations for performance and readability.

1. `if()` expression order

Expressions in `if()` are run left-to-right, and evaluating computationally smaller expressions can improve the performance a little bit. This PR contains two of such improvements where it puts `=== null` call and a truthy evaluation ahead of function calls. 

2. `if()`-`elseif()`-`else` branching optimizations. 

If such a block `return`s something, the execution of that function is complete, so we often do not need `elseif` or `else` blocks. While they don't provide any performance benefit, we take nested code outside, which reduces the cognitive load on developers.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -99,7 +99,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
         }
         if (is_string($toResolve)) {
             [$instance, $method] = $this->resolveSlimNotation($toResolve);
-            if ($predicate($instance) && $method === null) {
+            if ($method === null && $predicate($instance)) {
                 $method = $defaultMethod;
             }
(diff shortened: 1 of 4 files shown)


l0gicgate (approved):
Hey thanks for this PR @Ayesh! I suppose there is probably quite of a few instances of this throughout all the codebases. I’m a bit surprised that static analysis doesn’t notify us of those.

Ayesh:
Thanks a lot @l0gicgate :)
I will see for more similar optimizations and send a PR. Most of the `elseif` blocks are easy to refactor and remove the besting without affecting anything else.  

I also sent https://github.com/slimphp/Slim-Csrf/pull/116 and an update to it if you could kindly take a look. Thank you.

PR #2960, opened 2020-05-06
Fix phpstan 0.12 errors

adriansuter:
This PR fixes #2952 .


l0gicgate (approved):
You're a legend @adriansuter. We need to do this on all the other repos as well. I will try to do them soon.

PR #2978, opened 2020-06-30
Split the routing results to its own function.

dlundgren:
This allows the core logic of the RoutingMiddleware to be extended, without having to copy & paste the core logic.

### DIFF
--- a/Slim/Middleware/RoutingMiddleware.php
+++ b/Slim/Middleware/RoutingMiddleware.php
@@ -72,10 +72,7 @@ class RoutingMiddleware implements MiddlewareInterface
      */
     public function performRouting(ServerRequestInterface $request): ServerRequestInterface
     {
-        $routingResults = $this->routeResolver->computeRoutingResults(
-            $request->getUri()->getPath(),
-            $request->getMethod()
-        );
+        $routingResults = $this->resolveRouteFromRequest($request);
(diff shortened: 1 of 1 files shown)


Slim/Middleware/RoutingMiddleware.php
```diff
@@ -101,4 +98,18 @@ public function performRouting(ServerRequestInterface $request): ServerRequestIn
...
         }
     }
+
+    /**
+     * Resolves the route from the given request
+     *
+     * @param  ServerRequestInterface $request
+     * @return RoutingResults
+     */
+    protected function resolveRouteFromRequest(ServerRequestInterface $request)
```
l0gicgate: Let's add the return type `RoutingResults` and rename the method to `resolveRoutingResultsFromRequest()`

l0gicgate (changes requested):
@adriansuter I don't foresee any breaking changes with this. Would you mind having a look when you have a minute?

Slim/Middleware/RoutingMiddleware.php
dlundgren: The method has been renamed, and return type added.

adriansuter (approved):

l0gicgate (approved):

l0gicgate:
Thank you for your contribution @dlundgren!

PR #3008, opened 2020-10-15
if else for get_magic_quotes_gpc usage

twitnic:
`get_magic_quotes_gpc` is deprecated since php7.4
https://www.php.net/manual/en/function.get-magic-quotes-gpc.php

With an if else condition, could this method used by 7.4 and greater ...


l0gicgate:
The Slim 2 release cycle has been closed for a few years now. We will unfortunately not be merging this in. 

Please upgrade to Slim 4 or fork and patch yourself. 

Thank you

Issue #2977, opened 2020-06-29
Preparations for PHP 8


Ayesh:
Now with PHP 8 alpha released, I'd like to suggest that we start to test Slim on PHP 8. 

As of now, Slim 3.x only has Travis CI tests for PHP 7.3. I think we should add 7.4 first, followed by a allow-fail nightly PHP builds. 
For the PHP nightly builds, we will need `--ignore-platform-reqs` or [`--ignore-platform-req=php` if we can use Composer 2](https://php.watch/articles/composer-ignore-platform-req). 

PHPUnit 9.3 (next release) should allow it being installed in PHP 8. It is not ready yet, hence the use of `--ignore-platform-reqs`.

l0gicgate:
Slim `3.x` is in maintenance mode only now just as an FYI.

This would be applicable for the Slim 4 branch though, we will need to bring it up to speed for PHP 8.

edudobay:
Hi! I would be interested in helping make **Slim 4** compatible with PHP 8. (I'm not sure if I should open a new issue to discuss this, as this one was originally about Slim 3.) I already have a proposal for some small code changes that are needed, but I see that dependencies are a bigger issue to tackle and would like to understand the maintainers' view and see if I can help with that too.

The biggest issue seems to be PHPUnit, though I haven't looked so deeply at other packages that Slim depends on for testing. As for production dependencies, all of them use `>=` constraints so they seem fine for now.

To be compatible with PHP `^7.2 || ^8.0`, we must support PHPUnit 8 and 9 at the same time (PHPUnit 8 will not support PHP 8, and PHPUnit 9 requires PHP `>= 7.3`). However, as already discussed in https://github.com/slimphp/Slim/issues/2936, PHPUnit 9 deprecates `prophesize()` and to remove the deprecation warnings we must use `phpspec/prophecy-phpunit` (requires PHP `>= 7.3`).

> `PHPUnit\Framework\TestCase::prophesize()` is deprecated and will be removed in PHPUnit 10. Please use the trait provided by `phpspec/prophecy-phpunit`.

(For the record, I just got 255 of these deprecation warnings (out of 417 tests) running the test suite with PHPUnit 9.4.0.) That's why PHPUnit 9 was not adopted, as I understood from that discussion. I'm not sure how we could solve that while remaining compatible with PHP 7.2.

[...]

PR #3015, opened 2020-11-01
PHP 8 support

edudobay:
This PR allows Slim to be installed in PHP 8 and passes all the tests and checks.

History: I've raised some points in https://github.com/slimphp/Slim/issues/2977#issuecomment-703194522 and since that time I've worked around other repositories to make this possible. This is also my first open-source contribution to a major project :)

**Test fixes:**

* Added a test for an uncovered branch of `MiddlewareDispatcher->addDeferred`. PHPUnit 8 was miscalculating coverage – see https://coveralls.io/builds/33553304/source?filename=Slim/MiddlewareDispatcher.php.

**My questions:**

* I implemented `disableXmlEntityLoader` as a `private static method`. Is this okay? I thought about making it `protected` or extracting it to a class, but I'm unsure about creating new "contracts".
* Coverage – conditional call to `libxml_disable_entity_loader`. Is it okay to just `@codeCoverageIgnore` some part that will only run in PHP 7 or PHP 8? How should situations like this be handled?

**Dependencies:**  I have gone through all dependencies that are blocking the PHP 8 upgrade and temporarily linked to development branches where the PHP 8 issues are solved. Surely this must be addressed before this can be merged (this is isolated in a single commit that can be removed). Those are the open PRs, some by me and some by others:

* [x] https://github.com/slimphp/Slim-Psr7/pull/169
* [x] https://github.com/slimphp/Slim-Http/pull/142
* [x] https://github.com/php-fig/http-message-util/pull/19
* [x] https://github.com/Nyholm/psr7-server/pull/40
* [x] https://github.com/laminas/laminas-diactoros/pull/46

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -182,6 +182,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
             $callable = $callable[0];
         }
         if ($this->container && $callable instanceof Closure) {
+            /** @var Closure $callable */
             $callable = $callable->bindTo($this->container);
         }
         return $callable;
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -31,6 +31,8 @@ use function simplexml_load_string;
 use function strtolower;
 use function trim;
 
+use const LIBXML_VERSION;
+
 class BodyParsingMiddleware implements MiddlewareInterface
 {
     /**
@@ -118,11 +120,11 @@ class BodyParsingMiddleware implements MiddlewareInterface
         });
 
         $xmlCallable = static function ($input) {
-            $backup = libxml_disable_entity_loader(true);
+            $backup = self::disableXmlEntityLoader(true);
             $backup_errors = libxml_use_internal_errors(true);
             $result = simplexml_load_string($input);
 
-            libxml_disable_entity_loader($backup);
+            self::disableXmlEntityLoader($backup);
             libxml_clear_errors();
             libxml_use_internal_errors($backup_errors);
 
@@ -187,5 +189,18 @@ class BodyParsingMiddleware implements MiddlewareInterface
         }
 
         return null;
+    }
+
+    protected static function disableXmlEntityLoader(bool $disable): bool
+    {
+        if (LIBXML_VERSION >= 20900) {
+            // libxml >= 2.9.0 disables entity loading by default, so it is
+            // safe to skip the real call (deprecated in PHP 8).
+            return true;
+        }
+
+        // @codeCoverageIgnoreStart
+        return libxml_disable_entity_loader($disable);
+        // @codeCoverageIgnoreEnd
     }
 }
--- a/Slim/MiddlewareDispatcher.php
+++ b/Slim/MiddlewareDispatcher.php
@@ -287,6 +287,7 @@ class MiddlewareDispatcher implements MiddlewareDispatcherInterface
         $next = $this->tip;
 
         if ($this->container && $middleware instanceof Closure) {
+            /** @var Closure $middleware */
             $middleware = $middleware->bindTo($this->container);
         }
 
--- a/Slim/Routing/Route.php
+++ b/Slim/Routing/Route.php
@@ -369,10 +369,13 @@ class Route implements RouteInterface, RequestHandlerInterface
         }
         $strategy = $this->invocationStrategy;
 
+        /** @var string[] $strategyImplements */
+        $strategyImplements = class_implements($strategy);
+
         if (
             is_array($callable)
             && $callable[0] instanceof RequestHandlerInterface
-            && !in_array(RequestHandlerInvocationStrategyInterface::class, class_implements($strategy))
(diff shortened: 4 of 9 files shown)


l0gicgate:
@edudobay thank you for this PR

> I implemented disableXmlEntityLoader as a private static method. Is this okay? I thought about making it protected or extracting it to a class, but I'm unsure about creating new "contracts".

I think that we could just make that method `protected` but does not need to be `static` unless I'm missing something

> Coverage – conditional call to libxml_disable_entity_loader. Is it okay to just @codeCoverageIgnore some part that will only run in PHP 7 or PHP 8? How should situations like this be handled?

I don't think there's any way around this one in this particular case. We would have to overload the `LIBXML_VERSION` constant somehow to be able to stub the behavior.

> Dependencies: I have gone through all dependencies that are blocking the PHP 8 upgrade and temporarily linked to development branches where the PHP 8 issues are solved. Surely this must be addressed before this can be merged (this is isolated in a single commit that can be removed). Those are the open PRs, some by me and some by others:

This particular issue bothers me a bit because we're going to have to do this work all over again when all the other libraries have official releases to support PHP 8. I would rather just do all the work at once than do it twice. We can just leave this PR open for now until the other libraries release official support for PHP 8.

edudobay:
@l0gicgate Sure, I'll make it `protected` then. But an instance method would not work inside the `static` anonymous function. I suppose that function is static for performance reasons. When I was implementing these changes, I benchmarked several kinds of function calls and I got differences up to 10x in memory usage — even though I don't think this is performance-critical code, I would be careful not to introduce a 10x factor in resource usage. I'll redo these benchmarks (I think I threw it away at the time), but I'll be happy to discuss other solutions too.

Regarding dependencies, I'll just watch them for releases and ping a comment when they're all ready.

edudobay:
Hi! I've updated the source branch and fixed some PHPStan issues that surfaced recently. I'm not sure if my solution was the way to go — the type hints that could go wrong are the ones for `Closure::bindTo()`, which can fail if called on a closure with static scope.

edudobay:
Hi @l0gicgate. All set for external projects! When https://github.com/slimphp/Slim-Psr7/pull/169 is merged and release tags are ready for it and for `slim/http`, we can remove the "TEMP" commit to go back to tagged release constraints.

I guess we will need to wait a few days for an official PHP 8.0 tag to be available in Travis CI, so we can fix that too, right? I'm not sure if we should otherwise remove "allowed to fail" from the nightly build.

l0gicgate:
@edudobay

> I guess we will need to wait a few days for an official PHP 8.0 tag to be available in Travis CI, so we can fix that too, right? I'm not sure if we should otherwise remove "allowed to fail" from the nightly build.

Once they release the official PHP 8.0 tag we can add it to the CI builds and leave the nightly build as "allowed to fail" as we will need it to examine the behavior of future versions on CI e.g: PHP 9 (long time from now)

I will try and release the other repos this weekend!

edudobay:
@l0gicgate Great, let me know if I can help with anything!

l0gicgate:
@edudobay see the releases below:

[Slim-Psr7 1.3.0 Release](https://github.com/slimphp/Slim-Psr7/releases/tag/1.3.0)
[Slim-Http 1.2.0 Release](https://github.com/slimphp/Slim-Http/releases/tag/1.2.0)

We should be good to go for this PR after!

edudobay:
@l0gicgate I've updated composer.json to require these latest `slim/http` and `slim/psr7` if that's ok.

Slim/CallableResolver.php
```diff
@@ -183,6 +183,7 @@ private function bindToContainer(callable $callable): callable
         }
         if ($this->container && $callable instanceof Closure) {
             $callable = $callable->bindTo($this->container);
+            /** @var Closure $callable */
```
l0gicgate: can we put the comment above the variable?

Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -188,4 +188,17 @@ protected function getMediaType(ServerRequestInterface $request): ?string
 
         return null;
     }
+
+    protected static function disableXmlEntityLoader(bool $disable): bool
+    {
+        if (\LIBXML_VERSION >= 20900) {
```
l0gicgate: Let's add this to the imports. We don't do `\` in the codebase

```php
use LIBXML_VERSION;
```

Slim/MiddlewareDispatcher.php
```diff
@@ -288,6 +288,7 @@ public function addCallable(callable $middleware): self
 
         if ($this->container && $middleware instanceof Closure) {
             $middleware = $middleware->bindTo($this->container);
+            /** @var Closure $middleware */
```
l0gicgate: Let's put the comment above the variable

l0gicgate (commented):
Just a few little things then I'll merge

Slim/CallableResolver.php
edudobay: sure! Sorry I missed these little things :)

edudobay:
@l0gicgate I've fixed the issues you mentioned!

l0gicgate (approved):
Thank you for this @edudobay!

PR #3025, opened 2020-11-25
Declare compatibility with PHP 8.x

odan:
Slim 4 should not be declared as compatible with PHP 8.


odan:
I have just seen #3015. So this PR is not necessary.

PR #3027, opened 2020-11-30
Fix: FastRoute dispatcher and data generator should match

edudobay:
While testing with the `master` branch of [FastRoute](https://github.com/nikic/FastRoute), I found the following issue:

FastRoute has recently changed the default dispatcher to MarkBased, and Slim is based on GroupCountBased. The data generator should always match the dispatcher, so the data generator must be explicitly given to avoid breaking with future FastRoute releases.


l0gicgate:
@edudobay great catch! Let's get this one into the 4.7 release. Thank you!

l0gicgate (approved):

PR #3029, opened 2020-12-01
weirdan/prophecy-shim must be a dev requirement

j0k3r:
As it is to support both PHPUnit 8 & 9, that dependency must be in dev requirement.
It breaks code from project using Slim with PHPUnit-Bridge for example. Anyway a PHPUnit related deps must not be in `require` but in `require-dev`.

Following https://github.com/slimphp/Slim/pull/3015
Related to that commit https://github.com/slimphp/Slim/pull/3015/commits/d971acb358660c171a9b61d2be28ec94f0414fd8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,8 +51,7 @@
         "psr/http-message": "^1.0",
         "psr/http-server-handler": "^1.0",
         "psr/http-server-middleware": "^1.0",
-        "psr/log": "^1.1",
-        "weirdan/prophecy-shim": "^1.0 || ^2.0.2"
+        "psr/log": "^1.1"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


edudobay:
@l0gicgate sorry, I missed this!

j0k3r:
@l0gicgate I think it should go into 4.7.1 rather than 4.8.0. In my case, I can't upgrade to 4.7.0 because the inclusion of that deps breaks the build of my project.

I might not be the only one. Also it's a huge addition in the composer.lock:
image

What do you think?

l0gicgate:
@j0k3r I just changed the milestone to 4.7.1. Will be releasing tonight.

PR #3037, opened 2020-12-03
Replace Travis by GitHub Actions

### LINKED ISSUE #3035: Move CI to GitHub Actions

t0mmy742:
I just don't know if you want to use differents jobs for our differents tools (phpcs, phpstan, phpunit,...)

Fix #3035


.github/workflows/tests.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: ${{ runner.os }}-composer-
+
+      - name: Install dependencies
+        run: composer install -n
```
j0k3r: It's simpler

```suggestion
      - name: Install dependencies with Composer
        uses: ramsey/composer-install@v1
        with:
          composer-options: "--optimize-autoloader --prefer-dist"
```

t0mmy742: `--prefer-dist`: it is already the default for stable version.
`--optimize-autoloader`: I'm not really sure that we spend less time for testing with this option.

l0gicgate:
It seems that this PR is still in draft. Looks good to me though, I'd like to see the actions run before merging.

t0mmy742:
@l0gicgate you can see GitHub Actions running just [here](https://github.com/t0mmy742/Slim/actions?query=branch%3Agithub-ci)
It looks good to me, I rebased everything.

j0k3r:
@l0gicgate I think you might need to enable GitHub Actions somewhere in the settings of the slimphp's organization

l0gicgate:
@j0k3r they are already enabled. Maybe after this is merged they will run. I'll try it now.

l0gicgate:
Thanks for your contribution @t0mmy742

l0gicgate:
First run went fantastic. We should do this on all the other repos!

https://github.com/slimphp/Slim/actions/runs/409168671

PR #3050, opened 2021-01-18
removing obsolete get_magic_quotes_gpc usage

gabema:
I know that Slim V2 is no longer supported and we actively use Slim V4 mostly but I still support some isolated Slim V2 dependent components that need to be updated to PHP 7.4 but would require dramatically more effort to update to Slim V4. Any chance we can provide a small update to the Slim V2 project so it's functional with PHP 7.4?

### DIFF
--- a/Slim/Http/Util.php
+++ b/Slim/Http/Util.php
@@ -55,14 +55,9 @@ class Util
      * @param  bool            $overrideStripSlashes
      * @return array|string
      */
-    public static function stripSlashesIfMagicQuotes($rawData, $overrideStripSlashes = null)
+    public static function stripSlashesIfMagicQuotes($rawData, $overrideStripSlashes = false)
     {
-        $strip = is_null($overrideStripSlashes) ? get_magic_quotes_gpc() : $overrideStripSlashes;
-        if ($strip) {
-            return self::stripSlashes($rawData);
-        }
-
-        return $rawData;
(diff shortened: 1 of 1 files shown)


l0gicgate:
At this point @gabema I don't feel confident doing any further releases on the Slim 2 branch especially when we're introducing breaking changes like this. @akrabat what are your thoughts?

gabema:
Fair enough. I definitely didn't expect much or any support for the 2.x branch. I know it's two major versions behind and its on our roadmap to replace the existing implementation. From testing the legacy component with the slim v2 dependency on PHP 7.4 this has been the only code change I've needed to make so far. Just wanted to post the code changes in case others find themselves in the same unfortunate predicament. I'll just use a private patched version of slim 2.x.

akrabat:
I have a private patch too, so there's a reasonable argument for working up a non-BC breaking solution.

I added this before the "`$strip = '" line:

```
        if ($overrideStripSlashes === null) {
            // get_magic_quotes_gpc() always returns false as of PHP 5.4 when the magic quotes feature was removed,
            // so we can just return $rawData immediately
            return $rawData;
        }
```

However, I think we'd want to add in a check for `PHP_VERSION_ID >= 50400`. 

The biggest issue would be testing and CI of course, which may prove the biggest problem.

l0gicgate:
I'm closing this as won't fix unfortunately. Sorry

PR #3069, opened 2021-04-17
Classname/Method Callable Arrays

ddrv:
These changes will allow the route handler to be specified as an array. This allows you to open the handler method in PhPStorm in ctrl + click, and also allows you to avoid the "unused method" type of bindings. 

For example:

```php
<?php

$app->get('/', [App\Controllers\PageController::class, 'home'])->setName('home');
```

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -51,6 +51,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
      */
     public function resolve($toResolve): callable
     {
+        $toResolve = $this->prepareToResolve($toResolve);
         if (is_callable($toResolve)) {
             return $this->bindToContainer($toResolve);
         }
@@ -90,6 +91,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
      */
     private function resolveByPredicate($toResolve, callable $predicate, string $defaultMethod): callable
     {
+        $toResolve = $this->prepareToResolve($toResolve);
         if (is_callable($toResolve)) {
             return $this->bindToContainer($toResolve);
         }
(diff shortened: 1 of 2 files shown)


Slim/CallableResolver.php
```diff
@@ -187,4 +189,22 @@ private function bindToContainer(callable $callable): callable
...
+    /**
+     * @param string|callable $toResolve
+     * @return string|callable
+     */
+    private function prepareToResolve($toResolve)
+    {
+        if (!is_array($toResolve)) {
+            return $toResolve;
+        }
+        $array = $toResolve;
```
l0gicgate: can we rename this variable to `$candidate` please and add a blank line after the `if` statement

tests/CallableResolverTest.php
```diff
@@ -474,7 +508,7 @@ public function testMiddlewareClassNotFoundThrowException()
     public function testCallableClassNotFoundThrowException()
     {
         $this->expectException(RuntimeException::class);
-        $this->expectExceptionMessage('is not resolvable');
+        $this->expectExceptionMessage('Callable Unknown does not exist');
```
l0gicgate: I think that we could make these error messages more relevant. We should include the method as well in the message if it exists. Right now we only show the class name.

l0gicgate (changes requested):
Great work, I like this. We're going to have to add documentation for this as well.

ddrv:
> We're going to have to add documentation for this as well.

- [x] [Pull Request](https://github.com/slimphp/Slim-Website/pull/560)

l0gicgate (approved):

l0gicgate:
@ddrv great work, thank you for your contribution

PR #3120, opened 2021-10-15
Add a new PSR-17 factory to Psr17FactoryProvider

solventt:
There is curious performance benchmark of the PSR-7 and PSR-17 implementations (Guzzle, HttpSoft, Laminas, Nyholm, Slim) - https://github.com/devanych/psr-http-benchmark.

The fastest is HttpSoft, the slowest - Slim.

So I've added `HttpSoftPsr17Factory` to the Slim's factory set to give users the possibility to use the `HttpSoft` implementation of the PSR-7.

What was generally done:

**1\)** Adding `HttpSoftPsr17Factory::class` to the `$factories` property of the `Slim\Factory\Psr17\Psr17FactoryProvider` class

```php
protected static $factories = [
    SlimPsr17Factory::class,
    HttpSoftPsr17Factory::class,
    NyholmPsr17Factory::class,
    LaminasDiactorosPsr17Factory::class,
    GuzzlePsr17Factory::class,
];
```
I placed the class for the second position because it's the fastest implementation of the PSR-7 according to the benchmark. But the Slim's implementation remains in first place because I think it's normal to give priority to packages from the Slim's ecosystem.

**2\)** Adding the new `HttpSoftPsr17Factory` class to the Slim/Factory/Psr17 folder

**3\)** Adding the [httpsoft/http-message](https://github.com/httpsoft/http-message) and [httpsoft/http-server-request](https://github.com/httpsoft/http-server-request) packages to the `composer.json` file (--dev section)

**4\)** Adding `[HttpSoftPsr17Factory::class, HttpSoftServerRequest::class]` to the `provideImplementations()` method in the `ServerRequestCreatorFactoryTest` class

**5\)** Adding `[HttpSoftPsr17Factory::class, HttpSoftResponseFactory::class]` to the `provideImplementations()` method in the `AppFactoryTest` class

**6\)** Editing `README.md` (added a mention of the HttpSoft PSR-7 implementation along with other implementations)

### About tests

- There were: 420 tests, 779 assertions
- Became: 422 tests, 781 assertions

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ This will install Slim and all required dependencies. Slim requires PHP 7.3 or n
 
 Before you can get up and running with Slim you will need to choose a PSR-7 implementation that best fits your application. A few notable ones:
 - [Slim-Psr7](https://github.com/slimphp/Slim-Psr7) - This is the Slim Framework PSR-7 implementation
-- [Nyholm/psr7](https://github.com/Nyholm/psr7) & [Nyholm/psr7-server](https://github.com/Nyholm/psr7-server) - This is the fastest, strictest and most lightweight implementation available
+- [httpsoft/http-message](https://github.com/httpsoft/http-message) & [httpsoft/http-server-request](https://github.com/httpsoft/http-server-request) - This is the fastest, strictest and most lightweight implementation available
+- [Nyholm/psr7](https://github.com/Nyholm/psr7) & [Nyholm/psr7-server](https://github.com/Nyholm/psr7-server) - Performance is almost the same as the HttpSoft implementation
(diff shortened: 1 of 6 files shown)


l0gicgate:
This is a great contribution, I was not aware that our implementation was this slow. We should probably do some digging since the benchmarking code just seems to be instantiating the objects.

I also wasn't aware of the HttpSoft implementation!

We won't be able to merge this until we bump PHP lowest version support to 7.4 in about a month unfortunately

l0gicgate:
We can deprecate PHP 7.3 as of Today @solventt if you would like to add that to this PR

solventt:
@l0gicgate The conflict in composer.json has been resolved

l0gicgate (approved):
Thank you for this contribution @solventt

PR #3122, opened 2021-10-19
Add a fallback route method

solventt:
In [Slim's docs](https://www.slimframework.com/docs/v4/cookbook/enable-cors.html#the-simple-solution) I noticed the following example:

```php
/**
 * Catch-all route to serve a 404 Not Found page if none of the routes match
 * NOTE: make sure this route is defined last
 */
$app->map(['GET', 'POST', 'PUT', 'DELETE', 'PATCH'], '/{routes:.+}', function ($request, $response) {
    throw new HttpNotFoundException($request);
});
```

But there is no need to use the example above, there is a better solution. Let's look at the following examples:

**1\)** `Slim\Routing\FastRouteDispatcher` - the `dispatch()` method:

```php
...
// If nothing else matches, try fallback routes
$routingResults = $this->routingResults('*', $uri);
if ($routingResults[0] === self::FOUND) {
  return $routingResults;
}
...
```

**2\)** `FastRoute\Dispatcher\RegexBasedAbstract` - the original `dispatch()` method:

```php
...
// If nothing else matches, try fallback routes
if (isset($this->staticRouteMap['*'][$uri])) {
  $handler = $this->staticRouteMap['*'][$uri];
  return [self::FOUND, $handler, []];
}
if (isset($varRouteData['*'])) {
  $result = $this->dispatchVariableRoute($varRouteData['*'], $uri);
  if ($result[0] === self::FOUND) {
    return $result;
  }
}
...
```

### What follows from the given examples? It follows that:

- We can define the fallback route anywhere in the route definitions, even in the first place before all other routes - this will really work.

- We can define the fallback route in a more elegant way than the one currently is shown in Slim's docs.

So I've added:

**1\)** In `Slim\Routing\RouteCollectorProxy`:
```php
/**
 * {@inheritdoc}
 */
public function fallback($callable): RouteInterface
{
  return $this->map(['*'], '/{routes:.+}', $callable);
}
```

**2\)** In `Slim\Interfaces\RouteCollectorProxyInterface`:
```php
/**
 * Add a route that will be executed when no other route matches the incoming request
 *
 * @param  callable|string $callable The route callback routine
 *
 * @return RouteInterface
 */
public function fallback($callable): RouteInterface;
```

**3\)** The `testFallback()` unit test.

And now during route definition we can use:
```php
$app->fallback(function ($request, $response) {
  // your logic, e.g. throwing an exception
  throw new HttpNotFoundException($request);
});

// other routes
$app->get('/signup', [SignupController::class, 'request']);

$app->post('/signup', [SignupController::class, 'signup']);
...
```

### DIFF
--- a/Slim/Interfaces/RouteCollectorProxyInterface.php
+++ b/Slim/Interfaces/RouteCollectorProxyInterface.php
@@ -157,4 +157,13 @@ interface RouteCollectorProxyInterface
      * @return RouteInterface
      */
     public function redirect(string $from, $to, int $status = 302): RouteInterface;
+
+    /**
+     * Add a route that will be executed when no other route matches the incoming request
+     *
+     * @param  callable|string $callable The route callback routine
+     *
+     * @return RouteInterface
+     */
+    public function fallback($callable): RouteInterface;
 }
--- a/Slim/Routing/RouteCollectorProxy.php
+++ b/Slim/Routing/RouteCollectorProxy.php
@@ -206,4 +206,12 @@ class RouteCollectorProxy implements RouteCollectorProxyInterface
 
         return $this->get($from, $handler);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fallback($callable): RouteInterface
+    {
+        return $this->map(['*'], '/{routes:.+}', $callable);
+    }
 }
--- a/tests/Routing/RouteCollectorProxyTest.php
+++ b/tests/Routing/RouteCollectorProxyTest.php
@@ -455,4 +455,37 @@ class RouteCollectorProxyTest extends TestCase
 
         $routeCollectorProxy->group($pattern, $callable);
     }
+
+    public function testFallback()
+    {
+        $responseFactoryProphecy = $this->prophesize(ResponseFactoryInterface::class);
+        $callableResolverProphecy = $this->prophesize(CallableResolverInterface::class);
+
+        $pattern = '/{routes:.+}';
+        $callable = function () {
+        };
+
+        $routeProphecy = $this->prophesize(RouteInterface::class);
+        $routeProphecy
+            ->getPattern()
+            ->willReturn($pattern)
+            ->shouldBeCalledOnce();
+
+        $routeCollectorProphecy = $this->prophesize(RouteCollectorInterface::class);
+        $routeCollectorProphecy
+            ->map(['*'], $pattern, Argument::is($callable))
+            ->willReturn($routeProphecy->reveal())
+            ->shouldBeCalledOnce();
+
+        $routeCollectorProxy = new RouteCollectorProxy(
+            $responseFactoryProphecy->reveal(),
+            $callableResolverProphecy->reveal(),
+            null,
+            $routeCollectorProphecy->reveal()
+        );
+
+        $route = $routeCollectorProxy->fallback($callable);
+
+        $this->assertEquals($pattern, $route->getPattern());
+    }
 }


t0mmy742:
Your fallback method is for **every** HTTP methods. Your exemple is about **some** (`['GET', 'POST', 'PUT', 'DELETE', 'PATCH']`). Maybe that's not what we want to do.

`\Slim\Routing\FastRouteDispatcher::dispatch` method can also return `METHOD_NOT_ALLOWED` HTTP code. We should at least move this part before fallback call.

solventt:
@t0mmy742 Quote: "Your exemple is about some (['GET', 'POST', 'PUT', 'DELETE', 'PATCH']). Maybe that's not what we want to do."

OK, what is the shortest way?
```php
$app->fallback(fn ($request, $response) => throw new HttpNotFoundException($request));
$app->map(['GET', 'POST', 'PUT', 'DELETE', 'PATCH'], '/{routes:.+}', fn ($request, $response) => throw new HttpNotFoundException($request));
```
Quote: "\Slim\Routing\FastRouteDispatcher::dispatch method can also return METHOD_NOT_ALLOWED HTTP code."

If Slim and FastRoute placed the fallback code before processing the allowed methods, then it is correct.

l0gicgate:
I've never understood why this is being used in the first place. The docs are wrong. `HttpNotFoundException` will always be thrown if a route is not found:

https://github.com/slimphp/Slim/blob/4.x/Slim/Middleware/RoutingMiddleware.php#L90-L91
```php
case RoutingResults::NOT_FOUND:
    throw new HttpNotFoundException($request);
```

The only relevant case for this is if you want to throw something other than `HttpNotFoundException` when a route isn't found.

That exception is handled by `ErrorMiddleware` if present, the logic for handling any rendering of this exception should be done in an error handler.

Can someone enlighten me on what is the pertinent case for using the catchall/fallback method?

solventt:
@l0gicgate Quote: "Can someone enlighten me on what is the pertinent case for using the catchall/fallback method?"

I'd like to respond in your own words: "The only relevant case for this is if you want to throw something other than HttpNotFoundException when a route isn't found".

A person may not only want to throw another exception, but also display some page:
```php
$app->fallback(function ($request, $response) {
  $view = $request->getAttribute('twig');
  return $view->render('template.twig');
});
```
Yes, you can say that this logic can be written in a custom error handler or error renderer. But there is one thing to consider - most people dont want to write their custom ErrorHandler - It requires a more advanced level of mastery of the SLim framework. They want a simpler solution - the fallback method could be just that.

For example, Laravel has Route::fallback. When this function was introduced, it was described as: "Basically, if no route is matched, then fallback function is a way to override default 404 page and introduce additional logic".

https://laraveldaily.com/route-fallback-if-no-other-route-is-matched/

https://laravel.com/docs/8.x/routing#fallback-routes

l0gicgate:
> Yes, you can say that this logic can be written in a custom error handler or error renderer. But there is one thing to consider - most people dont want to write their custom ErrorHandler - It requires a more advanced level of mastery of the SLim framework. They want a simpler solution - the fallback method could be just that.

You don't need any kind of advanced mastery. The syntax goes as follows:

```php
$errorMiddleware = $app->addErrorMiddleware(true, true, true);

$errorMiddleware->addErrorHandler(HttpNotFoundException::class, function (Request $request, Throwable $t): Response {
  $view = $request->getAttribute('twig');
  return $view->render('template.twig');
});
```

Not sure if this just isn't documented properly or people are stuck on Slim 3.x patterns but this is unnecessary

**Edit:**
In the example I used `addErrorHandler()` but it should be `setErrorHandler()` as per the docs.

t0mmy742:
> OK, what is the shortest way?

I was talking about HTTP methods contained by your `fallback` method. Quoted documentation lists some methods (mainly those used by REST API), but the `fallback` method also use OPTIONS, HEAD, TRACE... But that does not matter; as @l0gicgate said, documentation is wrong :smile:

> Not sure if this just isn't documented properly or people are stuck on Slim 3.x patterns but this is unnecessary

~~We should add some documentation about 404 Not Found (something like your example).~~

EDIT:
[It is already mentioned but on Upgrade section](https://www.slimframework.com/docs/v4/start/upgrade.html#new-not-found--and-not-allowed-handler)

solventt:
> You don't need any kind of advanced mastery. The syntax goes as follows:
> 
> ```
> $errorMiddleware = $app->addErrorMiddleware(true, true, true);
> 
> $errorMiddleware->addErrorHandler(HttpNotFoundException::class, function (Request $request, Throwable $t): Response {
>   $view = $request->getAttribute('twig');
>   return $view->render('template.twig');
> });
 
But in order to use the method you wrote, a person must first learn internal code of the Slim framework, because `addErrorHandler()` is marked as an **internal method**, quote from the Slim code: "Used internally to avoid code repetition when passing multiple exceptions to setErrorHandler()". It is not a public API. But there is a public:

```php
$errorMiddleware->setErrorHandler(HttpNotFoundException::class, ...)
```
**But this method is also not mentioned in the documentation**. Docs contain only `setDefaultErrorHandler()`. So it turns out that to use the tools of the framework properly, one have to learn the internal Slim's code - that's what I meant by advanced mastery.

Ok, Slim has  `setErrorHandler`, `setDefaultErrorHandler`, `ErrorMiddleware`, erorr renderers. How does a user not get confused by all this?

l0gicgate:
This is just a documentation issue once again. We should document this better to avoid confusion. 

I think our documentation has been lacking since the Slim 4 release and I’m very much to blame for this. Writing docs is tedious and I admittedly suck at it.

t0mmy742:
I edited my previous comment:
[It is already mentioned but on Upgrade section](https://www.slimframework.com/docs/v4/start/upgrade.html#new-not-found--and-not-allowed-handler)

We just need to organize documentation 😄

l0gicgate:
@t0mmy742 we should add that exact section from the upgrade docs in the `ErrorMiddleware` docs:

https://www.slimframework.com/docs/v4/middleware/error-handling.html

solventt:
@t0mmy742 That's a good thing. Except that information about error handling is scattered all over the documentation:

- https://www.slimframework.com/docs/v4/start/upgrade.html#new-error-handling-middleware
- https://www.slimframework.com/docs/v4/objects/application.html#advanced-notices-and-warnings-handling
- https://www.slimframework.com/docs/v4/middleware/error-handling.html

solventt:
@l0gicgate thanks

PR #3124, opened 2021-10-27
Test against php 8.1 in not experimental mode

sergiy-petrov:


t0mmy742:
CI already passed successfully with PHP 8.1. It will not break anything.
PHP 8.1 is not official yet. It should be officially released [Nov 25 2021](https://wiki.php.net/todo/php81) (but RC5 already released, so nothing should change but bugfix).
@l0gicgate do you think we should do analyses (PHPCS & PHPStan & coverage) on PHP 8.1 too ? Or maybe we should wait for official release.

sergiy-petrov:
@t0mmy742 I've added option to run phpstan on php 8.1 in separate branch (it's on my fork). Seems it has a lot of errors however job is passed. Not sure why does it behave like that. You could see it on my fork:

https://github.com/sergiy-petrov/Slim/runs/4046667668?check_suite_focus=true

t0mmy742:
It seems like phpstan is [not ready for PHP 8.1 yet](https://github.com/phpstan/phpstan/issues/5558).
We should not do analyses on this version of PHP (I'm answering myself, I know x) )

l0gicgate:
It looks like PHPStan 1.0.0 was just released. Is that what we're waiting for? 🤔

PR #3129, opened 2021-12-02
Deprecate PHP 7.3

l0gicgate:


.github/workflows/tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.3, 7.4, 8.0, 8.1]
+        php: [7.4, 7.4, 8.0, 8.1]
```
vlakoff, line 13: There are two occurrences of "7.4"

vlakoff: Already fixed, sorry

PR #3131, opened 2021-12-03
fix: compatibility of php8.1 for slim3

everyx:
function `explod` and `preg_replace_callback`

`PHP message: PHP Deprecated:  xxxx(): Passing null to parameter #n ($xxx) of type xxx is deprecated /.../vendor/slim/slim/Slim/.../xxx.php on line nnn`

See: https://www.php.net/manual/en/migration81.deprecated.php

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,10 +40,12 @@ jobs:
               run: vendor/bin/phpstan
 
             - name: Tests
+              id: tests
+              if: matrix.php < 8 || !('3.x' == github.event.pull_request.base.ref || 'refs/heads/3.x' == github.ref)
               run: vendor/bin/phpunit --coverage-clover clover.xml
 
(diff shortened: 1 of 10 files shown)


everyx:
@l0gicgate Can we disable `phpunit` with php8 for slim3? The `phpunit` that `slim3` depends on is not compatible with php8, I just want to make my slim3 application can continue to run under php8.1

everyx:
@l0gicgate I just disable phpunit test in `matrix.php >= 8` and pull request target branch or pushed branch is `3.x` in `tests.yml`

l0gicgate:
We are not doing any more releases besides security fixes to Slim 3. It's a rabbit hole to try and make this happen.

everyx:
> We are not doing any more releases besides security fixes to Slim 3. It's a rabbit hole to try and make this happen.

Ok, I will maintain in my own fork, thank you, close this PR

PR #3133, opened 2021-12-31
Improve typehinting for `RouteParserInterface`

jerowork:
# Description
Previous typehinting (by docblock) in `RouteParserInterface` was not sufficient, it only defined the value of the array. However, both `$data` as `$queryParams` are arrays where the key as well as value is a string.

```php
/**
 * @param string[] ... // < only defines the type of the value
 * @param array<string, string> ... // < defines the type of both key and value
 */
```

By improving the typehinting for these values, e.g. static analysis tools will recognize this instead of failing due to type mismatch.

## Changes
- [x] Improved typehinting for `$data` and `$queryParams` in `RouteParserInterface`

## Links
- https://psalm.dev/docs/annotating_code/type_syntax/array_types/
- https://phpstan.org/writing-php-code/phpdoc-types#general-arrays

### DIFF
--- a/Slim/Interfaces/RouteParserInterface.php
+++ b/Slim/Interfaces/RouteParserInterface.php
@@ -19,9 +19,9 @@ interface RouteParserInterface
     /**
      * Build the path for a named route excluding the base path
      *
-     * @param string   $routeName   Route name
-     * @param string[] $data        Named argument replacement data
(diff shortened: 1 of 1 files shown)


l0gicgate:
Thank you for this contribution!

PR #3139, opened 2022-01-10
Fixed #1730 (reintroduced in 4.x)

### LINKED ISSUE #1730: Can't create 200 OK response with Location header

adoy:


Midel3:
This should also resolve this issue: https://github.com/bshaffer/oauth2-server-php/issues/254 

PHP tends to overwrite the status code to 401 whenever a WWW-Authenticate header is present, while 400, 403 and 405 are allowed according to RFC6750

odan:
It looks good to me. This is also how it is done in Laminas, see [SapiEmitter](https://github.com/laminas/laminas-httphandlerrunner/blob/2.2.x/src/Emitter/SapiEmitter.php#L23-L24).

l0gicgate (approved):

PR #3178, opened 2022-03-15
fix: php81 deprecation message

incisakura:
For the change in `Slim/Http/Uri.php#L129`, that is because that  `preg_replace_callback` wouldn't accept the third parameter as `null`. As `parse_url` would return `null` on not found, converting the return value of `parse_url` to `string` should be the best solution.

### DIFF
--- a/Slim/Collection.php
+++ b/Slim/Collection.php
@@ -109,6 +109,7 @@ class Collection implements CollectionInterface
      *
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


l0gicgate:
This branch is in maintenance mode, we are not releasing any more patches other than vulnerability patches. For PHP 8.1 compatibility, please use Slim 4.

PR #3186, opened 2022-04-13
Fix PHP 8.1 incompatibilities

barasimumatik:
While migrating to PHP 8.1 I got fatal errors in the two files changed in this commit. Both changes were suggested by the PHP interpreters diagnostic message and when I fixed them my applicaton worked as before (barring an unrelated dependency or two). It has been running under PHP 7.3 and PHP 7.4 before this.

The first change suggested simply adds the #[\ReturnTypeWillChange] attribute to the methods derived from the ArrayAccess class.

The second change relates to type safety since the preg_replace_callback no longer accepts a $subject that is NULL due to changes in PHP 8.1 (as I understood it).

I made sure the existing tests passed when using PHP 5.6 and PHP 7.4 (I didn't manage to install PHP 5.5 though, but I hope that's OK).
The tests won't run in PHP 8.1 due to incompatibilities with the PHPUnit version, and I tried to upgrade to PHPUnit 8.5 and even PHPUnit 9 with no luck.

### DIFF
--- a/Slim/Collection.php
+++ b/Slim/Collection.php
@@ -109,6 +109,7 @@ class Collection implements CollectionInterface
      *
      * @return bool
      */
+    #[\ReturnTypeWillChange]
     public function offsetExists($key)
     {
         return $this->has($key);
@@ -121,6 +122,7 @@ class Collection implements CollectionInterface
      *
      * @return mixed The key's value, or the default value
      */
+    #[\ReturnTypeWillChange]
     public function offsetGet($key)
     {
         return $this->get($key);
@@ -132,6 +134,7 @@ class Collection implements CollectionInterface
      * @param string $key   The data key
      * @param mixed  $value The data value
      */
+    #[\ReturnTypeWillChange]
     public function offsetSet($key, $value)
     {
         $this->set($key, $value);
@@ -142,6 +145,7 @@ class Collection implements CollectionInterface
      *
      * @param string $key The data key
      */
+    #[\ReturnTypeWillChange]
     public function offsetUnset($key)
     {
         $this->remove($key);
@@ -152,6 +156,7 @@ class Collection implements CollectionInterface
      *
      * @return int
      */
+    #[\ReturnTypeWillChange]
     public function count()
     {
         return count($this->data);
@@ -162,6 +167,7 @@ class Collection implements CollectionInterface
      *
      * @return ArrayIterator
      */
+    #[\ReturnTypeWillChange]
     public function getIterator()
     {
         return new ArrayIterator($this->data);
--- a/Slim/Http/Uri.php
+++ b/Slim/Http/Uri.php
@@ -718,7 +718,7 @@ class Uri implements UriInterface
             function ($match) {
                 return rawurlencode($match[0]);
             },
-            $query
+            is_null($query) ? "" : $query
         );
     }
 


l0gicgate:
@barasimumatik the `3.x` branch is in maintenance mode only unfortunately. If you want PHP 8.x compatibility use Slim 4.

lcharette:
@l0gicgate  Unless another issue makes Slim not compatible with PHP 8.1, this does fall in the "maintenance" category IMO. 

Upgrading to Slim 4 might not be a solution for everyone depending on such a major framework as slim is. It's even more true considering the whole dependency injection change in Slim 4 that makes it not an easy upgrade for some. Compared to the simplicity of this fix...

Otherwise, `composer.json` should be updated to reflect PHP 8.1 incompatibility ?

barasimumatik:
> @l0gicgate Unless another issue makes Slim not compatible with PHP 8.1, this does fall in the "maintenance" category IMO.
> 
> Upgrading to Slim 4 might not be a solution for everyone depending on such a major framework as slim is. It's even more true considering the whole dependency injection change in Slim 4 that makes it not an easy upgrade for some. Compared to the simplicity of this fix...
> 
> Otherwise, `composer.json` should be updated to reflect PHP 8.1 incompatibility ?

Yeah, I agree. This should fall under maintenance, but this is not my project so it's not up to me 🤷‍♂️.

In our case we can't just switch - it will involve a lot of work and it's hard enough to get some time for simply upgrading PHP and making sure everything works as before.
I hoped I could just upgrade PHP, but since Slim and a couple of other dependencies broke, I tried to fix any issues that stopped my application from working.

l0gicgate:
@lcharette @barasimumatik when I say maintenance, I mean security fixes only. I should have been clearer.

The risk of breaking things downstream when releasing anything from this branch now is too great and there's too many unknowns.

barasimumatik:
@l0gicgate I understand. 

I have one question though (if you don't mind): what do you mean by breaking things downstream? Do you mean breaking things for users of Slim (like me), or do you mean something else? The only thing I had an issue with was PHPUnit not being compatible with PHP 8.1 (or perhaps even 8.0). 

I guess what I'm wondering is if there is essentially no way PHP 8+ compatibility can be achieved if someone is willing to put up the work. What is blocking this, if we disregard the "maintenance mode" status, so to speak? (I'm asking mostly out of curiosity)

dregad:
@l0gicgate I'm a bit confused...

According to the [3.12.4 release notes](https://github.com/slimphp/Slim/releases/tag/3.12.4), this issue has been fixed, yet I don't see the updated *Slim/Collection.php* with added `#[\ReturnTypeWillChange]` attribute.

Am I missing something ?

With regards to https://github.com/slimphp/Slim/pull/3186#issuecomment-1108757685
> The risk of breaking things downstream [...] is too great and there's too many unknowns.

I must admit that I fail to see what could possibly break by adding the attribute, which is basically just a comment and was designed by PHP team to prevent breaking compatibility.

l0gicgate:
@dregad it's a mistake in the release notes. This PR was never merged but it was in the milestone. For PHP 8 support, please upgrade to Slim 4.

dregad:
Sadly I can't do that at the moment due to other constraints. But thanks for taking the time to reply!

patrickradius:
So, we are still running into these issues as well, and to be honest the mention of this PR in the 3.12.4 release notes was totally throwing us off and cost us a lot of head scratching.
We even opened a new PR with kind of the same changes because we thought this Collection class was just not done yet.

Could you _please_ reconsider merging this PR into the 3.x branch, because it will help out a lot of people.

dregad:
Considering the strong stance that was taken by @l0gicgate in this PR against patching 3.x (*For PHP 8 support, please upgrade to Slim 4.*), and the minor changes that were proposed here, I find it kind of funny to find out that nearly 3 years later Slim 3.13.0 was released for PHP 8.1-8.5 support with much more involved changes...

PR #3188, opened 2022-04-18
Update Collection.php

esase:


odan:
See here: https://github.com/slimphp/Slim/pull/3186#issuecomment-1098254226

PR #3193, opened 2022-05-04
Handlers types

ashleycoles:
Some more property type hints! This time for `Handlers`

A few notes on the changes made to `ErrorHandler`:

- The `method` property was being accessed before init, so I've switched it to be a nullable string with a default of null
- Both the `logErrorDetails` and `displayErrorDetails` properties were being accessed before init, so I've given them both default values of false

If there is a more sensible solution for the above, please let me know!

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -61,7 +61,7 @@ class ErrorHandler implements ErrorHandlerInterface
     /**
      * @var array<string|callable>
      */
-    protected $errorRenderers = [
+    protected array $errorRenderers = [
         'application/json' => JsonErrorRenderer::class,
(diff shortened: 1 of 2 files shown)


nbayramberdiyev (approved):

l0gicgate (approved):

dakujem:
Adding type to a base class causes fatal when child classes are not updated accordingly.

```
Compile Error: Type of MyErrorHandler::$errorRenderers must be array (as in class Slim\Handlers\ErrorHandler)
```

Should this not have been a major version bump? @l0gicgate 
Maybe not - a protected property is not public interface. I just wanted to point out however, it might have been communicated in the release notes or similar, since I see many other places the types have been added in `4.11` release. But don't get me wrong - I'm glad you are adding the types.

PR #3196, opened 2022-05-04
Adding property type declarations for Factory classes

ashleycoles:
Added type declarations for all properties in `Slim\Factory` classes excluding a union in `ServerRequestCreator` to preserve php7.4 compatibility


nbayramberdiyev (approved):

l0gicgate (approved):
Thank you for this @ashleycoles!

PR #3197, opened 2022-05-08
Remove redundant docblock types

theodorejb:
This will make it easier to read and refactor the code going forward.

Also added a couple missing property types and removed a redundant null check.


l0gicgate (approved):
Thank you for this contribution

PR #3209, opened 2022-07-04
Final public method for abstract class

MathiasReker:
All public methods of abstract classes should be final. Enforce API encapsulation in an inheritance architecture. If you want to override a method, use the Template method pattern.


l0gicgate:
We can't do this as we extend these factories to override the methods.

PR #3210, opened 2022-07-04
Static lambda

MathiasReker:
Lambdas not (indirect) referencing $this must be declared static.


l0gicgate:
We can't do this as we can't bind $this to static closures. Unfortunately there's no way to detect if a closure was declared static before we attempt to bind $this to it.

PR #3212, opened 2022-07-04
Get class to class keyword

MathiasReker:
Replace get_class calls on object variables with class keyword syntax.


nbayramberdiyev:
We need to delay this since `::class` is not supported on objects in PHP 7.

PR #3214, opened 2022-07-04
Is null

MathiasReker:
Replaces is_null($var) expression with null === $var.

### DIFF
--- a/Slim/Middleware/BodyParsingMiddleware.php
+++ b/Slim/Middleware/BodyParsingMiddleware.php
@@ -154,7 +154,7 @@ class BodyParsingMiddleware implements MiddlewareInterface
             $body = (string)$request->getBody();
             $parsed = $this->bodyParsers[$mediaType]($body);
 
-            if (null !== $parsed && !is_object($parsed) && !is_array($parsed)) {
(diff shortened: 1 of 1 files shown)


Slim/Middleware/BodyParsingMiddleware.php
```diff
@@ -154,7 +154,7 @@ protected function parseBody(ServerRequestInterface $request)
             $body = (string)$request->getBody();
             $parsed = $this->bodyParsers[$mediaType]($body);
 
-            if (!is_null($parsed) && !is_object($parsed) && !is_array($parsed)) {
+            if (null !== $parsed && !is_object($parsed) && !is_array($parsed)) {
```
l0gicgate: No yoda conditions. Let's do `$parsed !== null` instead

l0gicgate (changes requested):

Slim/Middleware/BodyParsingMiddleware.php
MathiasReker:
```suggestion
            if ($parsed !== null && !is_object($parsed) && !is_array($parsed)) {
```

nbayramberdiyev (approved):

l0gicgate (approved):

PR #3218, opened 2022-07-06
Refactor some code

mehdihasanpour:


tests/bootstrap.php
```diff
@@ -39,16 +39,16 @@
     ],
     RouteCollector::class => [
         'is_readable' => function (string $file): bool {
-            if (stripos($file, 'non-readable.cache') !== false) {
-                return false;
-            }
-            return true;
+            return ! (stripos($file, 'non-readable.cache') !== false)
```
nbayramberdiyev: You can simply do this:
```php
return stripos($file, 'non-readable.cache') === false;
```

tests/bootstrap.php
```diff
@@ -39,16 +39,16 @@
...
+                 
+            
+             ;
         },
         'is_writable' => function (string $path): bool {
-            if (stripos($path, 'non-writable-directory') !== false) {
-                return false;
-            }
-            return true;
+            return ! (stripos($path, 'non-writable-directory') !== false)
```
nbayramberdiyev: Same as above.

nbayramberdiyev (changes requested):

tests/bootstrap.php
mehdihasanpour: Thanks, this is a better way.

nbayramberdiyev: Can you also unwrap the code parentheses?

mehdihasanpour: I unwrapped the code parentheses.

nbayramberdiyev (approved):

l0gicgate (approved):

PR #3226, opened 2022-08-02
Fix for PHP 8.1

M3sca:


l0gicgate:
Sorry @M3sca but the 2.x branch is in maintenance mode only. We are not doing any more releases.

If you want PHP 8 support you need to migrate to Slim 4.

PR #3228, opened 2022-08-31
Using assertSame to let assert equal be restricted

peter279k:
# Changed log

- As title, it should use the `assertSame` to let assert equals be restricted.


flangofas (approved):

akrabat:
Thanks!

PR #3235, opened 2022-11-05
Persist routes indexed by name in RouteCollector for improved performance.

BusterNeece:
In our application, we have what I think is a reasonable number of routes defined (a few dozen), and we're often calling `RouteParserInterface::relativeUrlFor` and its related functions. Under the hood, these all call `RouteCollectorInterface::getNamedRoute`, and right now, that's iterating through _every_ single route, checking the name, and returning the result if it finds it.

Individually, these calls aren't a big deal, but they add up fast. If, say, you're returning an API response with several records with several links in each row, this tends to result in a _huge_ number of calls to `RouteInterface::getName()`.

As a result of their exposed methods, the nature of the `RouteCollectorInterface` and `RouteParserInterface` means that it isn't possible to implement a lookup table like this in one's own code without reimplementing the entire `RouteParserInterface::relativeUrlFor` functionality.

However, fortunately that also means that this change can be made at the Slim level without any change to the interface. The logic here is identical to how it functions currently (the first route with a given name is returned by `getNamedRoute()`).

### DIFF
--- a/Slim/Routing/RouteCollector.php
+++ b/Slim/Routing/RouteCollector.php
@@ -60,6 +60,13 @@ class RouteCollector implements RouteCollectorInterface
     protected array $routes = [];
 
     /**
+     * Routes indexed by name
+     *
+     * @var RouteInterface[]
+     */
+    protected array $routesByName = [];
+
+    /**
      * Route groups
      *
      * @var RouteGroup[]
@@ -172,7 +179,8 @@ class RouteCollector implements RouteCollectorInterface
     public function removeNamedRoute(string $name): RouteCollectorInterface
     {
         $route = $this->getNamedRoute($name);
-        unset($this->routes[$route->getIdentifier()]);
+
+        unset($this->routesByName[$route->getName()], $this->routes[$route->getIdentifier()]);
         return $this;
     }
 
@@ -181,11 +189,10 @@ class RouteCollector implements RouteCollectorInterface
      */
     public function getNamedRoute(string $name): RouteInterface
     {
-        foreach ($this->routes as $route) {
-            if ($name === $route->getName()) {
-                return $route;
-            }
+        if (isset($this->routesByName[$name])) {
+            return $this->routesByName[$name];
         }
+
         throw new RuntimeException('Named route does not exist for name: ' . $name);
     }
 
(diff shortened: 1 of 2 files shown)


Slim/Routing/RouteCollector.php
```diff
@@ -181,11 +189,10 @@ public function removeNamedRoute(string $name): RouteCollectorInterface
...
     public function getNamedRoute(string $name): RouteInterface
     {
-        foreach ($this->routes as $route) {
-            if ($name === $route->getName()) {
-                return $route;
-            }
+        if (isset($this->routesByName[$name])) {
+            return $this->routesByName[$name];
         }
+
```
akrabat, line 207: I'm pretty sure that a name can be assigned to a route after adding to the route collector, so can we keep the iteration if the lookup fails please? If the iteration finds a route that is not in `routesByName`, then add it for the next lookup.

Slim/Routing/RouteCollector.php
```diff
@@ -181,11 +189,10 @@ public function removeNamedRoute(string $name): RouteCollectorInterface
      */
     public function getNamedRoute(string $name): RouteInterface
     {
-        foreach ($this->routes as $route) {
-            if ($name === $route->getName()) {
-                return $route;
-            }
+        if (isset($this->routesByName[$name])) {
+            return $this->routesByName[$name];
```
akrabat: I think that it's possible that route has been renamed since it was mapped. Can we check the current name of the route in `routesByName[$name]` and if it's not the same as `$name`, remove it from `routesByName` and then fall through to the iteration search?

Slim/Routing/RouteCollector.php
```diff
@@ -229,6 +236,12 @@ public function map(array $methods, string $pattern, $handler): RouteInterface
     {
         $route = $this->createRoute($methods, $pattern, $handler);
         $this->routes[$route->getIdentifier()] = $route;
+
+        $routeName = $route->getName();
+        if (null !== $routeName && !isset($this->routesByName[$routeName])) {
```
akrabat: `s/null !== $routeName/$routeName !== null/` as we don't do Yoda conditions here.

akrabat (changes requested):
Thanks for this! Just a few changes that would be useful.

BusterNeece:
@akrabat Changes should be incorporated now.

akrabat (approved):

akrabat:
Thanks

PR #3251, opened 2023-02-21
HTTP Response status line refactoring

twent:
Made emitStatusLine method refactoring in ResponseEmitter. Used core PHP function for returning http response status line - http_response_code()
Some strict typing was added in tests/Assets/HeaderStack.php


odan:
The `ReasonPhrase` and `ProtocolVersion` from the response object are no longer considered after this change. This change would break existing code because the behavior is no longer the same.

l0gicgate:
As @odan pointed out, this would be a breaking change. We cannot proceed with this PR.

Issue #3272, opened 2023-07-12
version of `App::VERSION`


hashimoto-AJ:
I installed [3.12.4](https://github.com/slimphp/Slim/tree/3.x) by composer, but display of `App::VERSION` is **3.12.3**

~~~
 $ composer show -i
 :
 slim/slim                          3.12.4  Slim is a PHP micro framework that helps you quickly write simple yet po...
 :
~~~

~~~
use Slim\App;
 :
echo App::VERSION
 :
~~~

> 3.14.3

![slim3 12 4](https://github.com/slimphp/Slim/assets/74279799/11fb09d5-31bc-4ebf-b717-d675d307b8ac)

l0gicgate:
Release https://github.com/slimphp/Slim/releases/tag/3.12.5

PR #3274, opened 2023-07-23
Fix 3.x App::VERSION

l0gicgate:
Closes https://github.com/slimphp/Slim/issues/3272


l0gicgate:
Closes https://github.com/slimphp/Slim/issues/3272

PR #3278, opened 2023-08-03
Remove HttpGoneException executable flag

flavioheleno:
This PR performs a simple `chmod -x` on `Slim/Exception/HttpGoneException.php` as it was the only file in Slim's source to have the executable flag and I could not see a reason why it should have.

In other words, `Slim/Exception/HttpGoneException.php` permission flags went from `0755` to `0644`.


l0gicgate:
Thank you for this contribution @flavioheleno

PR #3293, opened 2023-10-23
Improve compatibility with PHP8.1

Vicgzj:
We are facing some compatibility issues with PHP 8.1 regarding the Collection Class and the returns types. Right now we cannot upgrade to version 4.x, but this will fix the current situation with 3.x version. Thanks!

### DIFF
--- a/Slim/Collection.php
+++ b/Slim/Collection.php
@@ -109,7 +109,7 @@ class Collection implements CollectionInterface
      *
      * @return bool
      */
-    public function offsetExists($key)
+    public function offsetExists($key): bool
     {
         return $this->has($key);
     }
@@ -121,6 +121,7 @@ class Collection implements CollectionInterface
      *
      * @return mixed The key's value, or the default value
      */
+    #[\ReturnTypeWillChange]
     public function offsetGet($key)
     {
         return $this->get($key);
@@ -132,6 +133,7 @@ class Collection implements CollectionInterface
      * @param string $key   The data key
      * @param mixed  $value The data value
      */
+    #[\ReturnTypeWillChange]
     public function offsetSet($key, $value)
     {
         $this->set($key, $value);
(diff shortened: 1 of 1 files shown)


odan:
I guess this change is not compatible with PHP 5.5 (and PHP 5.6) as defined in composer.json within the `3.x` branch.

* https://github.com/slimphp/Slim/blob/3.x/composer.json#L31
* https://3v4l.org/ZY1Vc#v5.5.38

```
// Parse error: syntax error, unexpected ':', expecting ';' or '{' in /in/coZnH on line 170
public function offsetExists($key): bool
```

The minimum version would then require PHP >=7.0 for the v3 branch to work.

Vicgzj:
> I guess this change is not compatible with PHP 5.5 (and PHP 5.6) as defined in composer.json within the `3.x` branch.
> 
> * https://github.com/slimphp/Slim/blob/3.x/composer.json#L31
> * https://3v4l.org/ZY1Vc#v5.5.38
> 
> ```
> // Parse error: syntax error, unexpected ':', expecting ';' or '{' in /in/coZnH on line 170
> public function offsetExists($key): bool
> ```
> 
> The minimum version would then require PHP >=7.0 for the v3 branch to work.

We are trying to solve the deprecated errors from our code when we use Slim 3.x on PHP>=7.0. We were checking and we saw another PR that was rejected with the solution that we also propose in this PR. https://github.com/slimphp/Slim/pull/3186
Can you please consider the last comment in that PR from my co-worker Patrick? Thanks

odan:
I've read this post. But as far as I know the Slim 3.x branch is only for essential security updates, but not for new features such as PHP 8 support. To get PHP 8 support, you need to upgrade to Slim 4.

Vicgzj:
> I've read this post. But as far as I know the Slim 3.x branch is only for essential security updates, but not for new features such as PHP 8 support. To get PHP 8 support, you need to upgrade to Slim 4.

It's not possible for us right now to upgrade to PHP 8.1. We already have PHP 8.0.

l0gicgate:
As @odan mentioned, Slim 3.x will not get PHP 8.x support.

Closing this as resolved

PR #3299, opened 2023-11-23
Update tests and add PHP 8.3 to the CI matrix

akrabat:
A few things that we were doing in the tests are deprecated and so have been updated to be correct.

- We now use `setStaticPropertyValue()` to set static properties on a reflected class rather than `setValue()` on a reflected property.
- We no longer dynamically add a `request` property to the response class in `BodyParsingMiddlewareTest`, instead we add it to our handler class.
- We no longer create an unused reflected property in `ErrorHandlerTest`.

I have also added PHP 8.3 to the CI matrix and moved our static analysis tests to PHP 8.2.

Note that we'll need to wait for `phpspec/prophecy` to support PHP 8.3 before we can merge.


l0gicgate:
@akrabat it appears installing the dependencies fails on PHP 8.3
CleanShot 2023-11-23 at 22 45 36@2x

akrabat:
> @akrabat it appears installing the dependencies fails on PHP 8.3 

Yes. See "note" in PR description :)

l0gicgate:
@akrabat my apologies 😅 I guess we'll just have to be patient!

stemd:
Seems that phpspec/prophecy now supports PHP 8.3 (commit 3 days ago):

https://github.com/phpspec/prophecy/commit/03cfe36857643bd13ba00442fcb857b2ba2d64a3?diff=unified&w=0

akrabat:
Thanks @stemd. I have re-run the PHP 8.3 CI test and it passed. Over to @l0gicgate for review.

l0gicgate (changes requested):
For cohesion, can we please import `ReflectionClass` in the import statements instead of using `\ReflectionClass`. Looks great otherwise, thank you for this contribution.

akrabat:
> For cohesion, can we please import `ReflectionClass` in the import statements instead of using `\ReflectionClass`.

Done.

l0gicgate (approved):

PR #3302, opened 2023-12-12
Add support for psr/http-message ^2.0

### LINKED ISSUE #3296: PSR 7 http-message version requirement
I noticed that version **1.1** of **psr/http-message** is what has been required in the **composer.json** ( https://github.com/slimphp/Slim/blob/fdb6d614700a4e9390c3b08dde96e1f9147ec3fc/composer.json#L51 ) for Slim 4. 

Is it possible to have **"^1.1 || ^2.0"** as the requirement instead?

Version 2.0  ( https://github.com/php-fig/http-message/compare/1.1...2.0 ) of **psr/http-message** has explicit return type typing which is better. Both versions **1.1** & **2.0** of **psr/http-message** require PHP **7.2** or **8.0**.

rotexdegba:
Changed **psr/http-message** requirement to **^1.1 || ^2.0** in composer.json.

Note that in dev environment (ie. cloning this repo with this pull request merged and running the tests in PHP 7.4) **psr/http-message ^1.1** always gets pulled in by composer for PHP 7.4 users because PSR 7 implementations like **laminas/laminas-diactoros ^2.17** in the require-dev section of https://github.com/slimphp/Slim/blob/4.x/composer.json requires **psr/http-message ^1.0**. 

For users requiring **slim/slim** in their applications and using a PSR 7 implementation like **nyholm/psr7 ^1.8** which requires **psr/http-message ^1.1 || ^2.0** ( https://github.com/Nyholm/psr7/blob/1.8.1/composer.json ), **psr/http-message ^2.0** will be pulled in for those applications.

Closes https://github.com/slimphp/Slim/issues/3296

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "nikic/fast-route": "^1.3",
         "psr/container": "^1.0 || ^2.0",
         "psr/http-factory": "^1.0",
-        "psr/http-message": "^1.1",
+        "psr/http-message": "^2.0 || ^1.1",
         "psr/http-server-handler": "^1.0",
         "psr/http-server-middleware": "^1.0",
         "psr/log": "^1.1 || ^2.0 || ^3.0"
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "nikic/fast-route": "^1.3",
         "psr/container": "^1.0 || ^2.0",
         "psr/http-factory": "^1.0",
-        "psr/http-message": "^2.0 || ^1.1",
+        "psr/http-message": "^1.1 || ^2.0",
         "psr/http-server-handler": "^1.0",
         "psr/http-server-middleware": "^1.0",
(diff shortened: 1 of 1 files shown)


kosciuk:
composer require slim/psr7

Problem 1
    - slim/psr7[0.1.0, ..., 0.6] require php ^7.1 -> your php version (8.3.0) does not satisfy that requirement.
    - slim/psr7[1.0.0, ..., 1.2.0] require php ^7.2 -> your php version (8.3.0) does not satisfy that requirement.
    - slim/psr7[1.3.0, ..., 1.6.1] require psr/http-message ^1.0 -> found psr/http-message[1.0, 1.0.1, 1.1] but the package is fixed to 2.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - Root composer.json requires slim/psr7 * -> satisfiable by slim/psr7[0.1.0, ..., 0.6, 1.0.0, ..., 1.6.1].

PR #3320, opened 2024-05-02
Add enable-log-details-tip to response body Fixes #3317

### LINKED ISSUE #3317: Using wrong flag to write to error log in ErrorHandler
In `ErrorHandler` at line [262](https://github.com/slimphp/Slim/blob/038fd5713d5a41636fdff0e8dcceedecdd17fc17/Slim/Handlers/ErrorHandler.php#L262), the condition is checked against the wrong flag.

the condition must check against `$this->logErrorDetails` rather than `$this->displayErrorDetails`. so the line would be:

`if (!$this->logErrorDetails) {`

If the bug is valid, I can create a pull request to fix it.

More details: It seems `writeToErrorLog()` is used to save the logs properly somewhere and `$this->displayErrorDetails` is used by the renderer to show the error details in realtime on the client

mpouyandeh:
We must add the tip when sending a response to the client (i.e. browser) only.

### DIFF
--- a/Slim/Handlers/ErrorHandler.php
+++ b/Slim/Handlers/ErrorHandler.php
@@ -259,10 +259,6 @@ class ErrorHandler implements ErrorHandlerInterface
     {
         $renderer = $this->callableResolver->resolve($this->logErrorRenderer);
         $error = $renderer($this->exception, $this->logErrorDetails);
-        if (!$this->displayErrorDetails) {
-            $error .= "\nTips: To display error details in HTTP response ";
-            $error .= 'set "displayErrorDetails" to true in the ErrorHandler constructor.';
-        }
         $this->logError($error);
     }
 
@@ -299,6 +295,11 @@ class ErrorHandler implements ErrorHandlerInterface
         $renderer = $this->determineRenderer();
         $body = call_user_func($renderer, $this->exception, $this->displayErrorDetails);
         if ($body !== false) {
+            if (!$this->displayErrorDetails) {
+                $error .= "\nTips: To display error details in HTTP response ";
+                $error .= 'set "displayErrorDetails" to true in the ErrorHandler constructor.';
+            }
+            
             /** @var string $body */
             $response->getBody()->write($body);
         }


akrabat:
As per https://github.com/slimphp/Slim/issues/3317#issuecomment-2089786428, the tip about the `displayErrorDetails` functionality is intended to be in the log entry, not the response.  

This PR moves it to the response which is incorrect.

mpouyandeh:
Dear Rob. Don't just close issues and pull requests. We gathered here to share and fix issues.

As per [Your comment](https://github.com/slimphp/Slim/issues/3317#issuecomment-2088962397).
> The idea here is that we we add a tip to the log to remind the developer that if they want to see the error in their browser, they need to set `displayErrorDetails` to `true`.

But now:

> As per [#3317 (comment)](https://github.com/slimphp/Slim/issues/3317#issuecomment-2089786428), the tip about the `displayErrorDetails` functionality is intended to be in the log entry, not the response.
> 
> This PR moves it to the response which is incorrect.

**Your comments are not compatible and your understanding is far from truth**

On the other hand, as per [your comment here](https://github.com/slimphp/Slim/issues/3317#issuecomment-2089786428) you said that:

> The important thing is that
> 
> ```
> if (!$this->displayErrorDetails) {
>     $body .= "\nTips: To display error details in HTTP response ";
>     $body .= 'set "displayErrorDetails" to true in the ErrorHandler constructor.';
> }
> ```
> 
> is output to the log, not the the response.


But to me, it seems you didn't even bother to read the tip. it clearly states to display error details in HTTP response.

I already commented on the details [here](https://github.com/slimphp/Slim/issues/3317#issuecomment-2089073180).

I officially ask you to open the PR and I officially ask for another reviewer. 
Dear @akrabat , You make me put time on this a lot more than needed. please open the PR and ask for someone else to review. thank you

akrabat:
I'm very sorry that we seem to have a communication failure as I'm not sure how I could have been clearer.

The message text is 

> To display error details in HTTP response set "displayErrorDetails" to true in the ErrorHandler constructor.

This is an _instructional_ message to the developer that they can see in the log output that provides instructions on how to enable errors in their browser. It is displayed in the logs as the default browser output for an error message in Slim is designed to be read by the users of a Slim website, not developers of that website. Therefore, this instructional message cannot be part of the response as it is irrelevant to a user. This message dates all the way back to the original PR #2398.

 This is my last comment on this specific matter.

PR #3322, opened 2024-05-11
Add template generics for PSR-11 implementations in PHPStan and Psalm

limarkxx:
Slim v4 abstracts out DI to PSR-11, which means given a Slim/App it is unclear what its DI container is capable of.
Psalm and PHPStan both provide mechanisms to express the types generically, allowing them to ensure that you are calling the correct [set method](https://php-di.org/doc/container.html#set) on a DI container received from [getContainer](https://github.com/slimphp/Slim/blob/4.x/Slim/Interfaces/RouteCollectorProxyInterface.php#L23C21-L23C33)

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -120,8 +120,10 @@ final class CallableResolver implements AdvancedCallableResolverInterface
      */
     private function resolveSlimNotation(string $toResolve): array
     {
-        if(!CallableResolver::$callablePattern) {
-            throw new RuntimeException("CallableResolver::\$callablePattern was modified to an invalid value");
+        if (!CallableResolver::$callablePattern) {
+            throw new RuntimeException(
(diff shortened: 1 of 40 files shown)


Slim/CallableResolver.php
```diff
@@ -120,11 +127,10 @@ private function isMiddleware($toResolve): bool
      */
     private function resolveSlimNotation(string $toResolve): array
     {
+        /** @psalm-suppress ArgumentTypeCoercion */
         preg_match(CallableResolver::$callablePattern, $toResolve, $matches);
         [$class, $method] = $matches ? [$matches[1], $matches[2]] : [$toResolve, null];
 
-        /** @var string $class */
```
limarkxx, line 126: note: psalm considers this annotation redundant. They can be included if findUnusedCode is set to false

akrabat (approved):

akrabat:
Thanks!

PR #3351, opened 2024-11-10
Adding more debug information

mamazu:
## What I did

I added some more debug info to the "Bad method" exception.

I also updated the tests because the tests would return the error message `Method "GET" not allowed. Allowed methods: GET"`.

## Why
When triggering an endpoint with a wrong method it correctly throws this error and shows me what methods are actually available but not what method was used to cause the exception.

### DIFF
--- a/Slim/Exception/HttpMethodNotAllowedException.php
+++ b/Slim/Exception/HttpMethodNotAllowedException.php
@@ -10,6 +10,9 @@ declare(strict_types=1);
 
 namespace Slim\Exception;
 
+use Psr\Http\Message\ServerRequestInterface;
+use Throwable;
+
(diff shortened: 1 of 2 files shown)


odan:
I think it could be potentially dangerous to print such detailed information in error messages, especially in a production environment. It’s generally recommended to limit detailed error information to development and debugging environments only. In production, a generic message (like "Method not allowed") would be the proper way.

mamazu:
Totally agree, but if security is the point here we should probably disable the info on what methods are available on that route. But I think returning the method that was used shouldn't expose any additional info.

odan:
Closing this, because this change cannot be added. Thanks.

PR #3363, opened 2024-12-30
Update PlainTextErrorRenderer.php

Alpine418:
Make method protected to support inheritance for custom renderer.


odan:
Thanks for the PR, but this  change cannot be merged, because in the future this class (and other) might be "final" in the future.

Alpine418:
Why final? I'm a big fan of using inheritance to extend existing classes with custom functionality. Especially the core classes of Slim. 

Setting core classes to final restricts extensibility and flexibility massively. At the moment, I don't see any advantage in making core classes final.

PR #3379, opened 2025-03-30
Improve types on routes

nbayramberdiyev:
My routes are defined as follows: `->get('/', [HomeController::class, 'index']);`, which is a [valid](https://www.slimframework.com/docs/v4/objects/routing.html#container-resolution) route definition. However, the latest version of [Psalm](https://psalm.dev) emits the following error:

```
Argument 2 of Slim\App::get expects a public static callable, but a non-static callable provided (see https://psalm.dev/004)
```

This PR addresses this issue by adding the `array{class-string, string}` type to the route handler definition.

### DIFF
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -79,7 +79,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
     }
 
     /**
-     * @param string|callable $toResolve
+     * @param callable|array<class-string, string>|string $toResolve
      *
      * @throws RuntimeException
      */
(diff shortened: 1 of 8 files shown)


odan:
Hi! For Slim, we only use phpstan and not psalm. We also do not intend to support the psalm-specific "syntax" for the docblocks because it would be too much effort to maintain both. Thank you for your understanding.

nbayramberdiyev:
> Hi! For Slim, we only use phpstan and not psalm. We also do not intend to support the psalm-specific "syntax" for the docblocks because it would be too much effort to maintain both. Thank you for your understanding.

I'm having some trouble understanding this. The syntax isn't specific to Psalm, it's also used by PHPStan. That said, the current definition of `string|callable` for the route handler is incomplete for either static analysis tool. Just to clarify, I'm not trying to introduce any Psalm-specific syntax here.

PR #3388, opened 2025-05-15
Added exception for when the headers are send

Matthiasklaasse:
Once in a blue moon yo will put something before the php tag and then your headers break and you can spend a hour trying to figure out why the headers wont do what they are supposed to.

Especially if you don't know about php's useful behavior with the headers it can cause a headache.

### DIFF
--- a/Slim/ResponseEmitter.php
+++ b/Slim/ResponseEmitter.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 namespace Slim;
 
 use Psr\Http\Message\ResponseInterface;
+use RuntimeException;
 
 use function connection_status;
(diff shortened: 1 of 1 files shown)


akrabat:
Thanks.  I have a concern that this might be a BC break.

Please can you provide an example of the failure case preferably in as few files as possible so that I can see what we currently do as I couldn't cause a "headers sent" error with a few quick tests on [akrabat/slim4-starter](https://github.com/akrabat/slim4-starter)

odan:
Closing this as "won't implement" because this change would be a BC break and this feature could also be implemented with a custom ResponseEmitter class.

PR #3391, opened 2025-05-27
Fix text and eol attributes for * selector

bebehr:
In `.gitattributes`: `* text=lf` is not valid syntax.
Use `* text=auto eol=lf` instead to enforce Unix newlines.
Docs: https://git-scm.com/docs/gitattributes


akrabat (approved):
LGTM. Thanks.

PR #3399, opened 2025-07-04
Update http urls in composer.json

bebehr:
Update unsecure urls in composer.json from HTTP to HTTPS.


odan:
It looks like that https://gabrielmanricks.com gives a warning. We should keep the http url for this entry.

bebehr:
Left http://gabrielmanricks.com/ untouched.
Update the other http links to https.

datamweb (approved):

PR #3404, opened 2025-08-18
Drop setAccessible() usage for PHP 8.5 support

arshidkv12:
Remove `setAccessible()` calls to support PHP 8.5 and avoid deprecation

### DIFF
--- a/tests/AppTest.php
+++ b/tests/AppTest.php
@@ -698,18 +698,24 @@ class AppTest extends TestCase
 
         // Check that the routing middleware really has been added to the tip of the app middleware stack.
         $middlewareDispatcherProperty = new ReflectionProperty(App::class, 'middlewareDispatcher');
-        $middlewareDispatcherProperty->setAccessible(true);
+        if (PHP_VERSION_ID < 80100) {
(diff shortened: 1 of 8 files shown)


flangofas (approved):

odan:
There is a lot of redundant code. Could you move this into a method?

arshidkv12:
ok.

akrabat:
If we drop support for anything less than PHP 8.1, we could just delete the calls to setAccessible(). Unfortunately, 8.1 was only release 4 years ago, which might be a little too recent. What do you think @odan ?

odan:
@akrabat I think keeping PHP 8.1 adds extra maintenance overhead (CI jobs, tests, etc.). 

I would suggest aligning with PHP’s official security timeline and keeping 8.1 support only until 31 Dec 2025 ([supported versions](https://www.php.net/supported-versions.php)). 
That’s just one month away, so we could already prepare to drop it in the next release.

odan:
Thanks for the PR. We are working on the changes for PHP 8.5 in this PR: #3415 
 Closing this as duplicate.

PR #3406, opened 2025-08-23
Fix MissingOverrideAttribute errors by adding #[\Override]

sno-windy:
This PR addresses [`MissingOverrideAttribute`](https://psalm.dev/docs/running_psalm/issues/MissingOverrideAttribute/) errors reported by Psalm by adding `#[\Override]` where appropriate.

- No functional changes were made; this should have no impact on existing behavior.
- This is my first contribution to the project, so please let me know if there are any issues with the approach or conventions.

Thank you for your time and consideration!

### DIFF
--- a/Slim/App.php
+++ b/Slim/App.php
@@ -204,6 +204,7 @@ class App extends RouteCollectorProxy implements RequestHandlerInterface
      * @param ServerRequestInterface $request
      * @return ResponseInterface
      */
+    #[\Override]
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
         $response = $this->middlewareDispatcher->handle($request);
--- a/Slim/CallableResolver.php
+++ b/Slim/CallableResolver.php
@@ -47,6 +47,7 @@ final class CallableResolver implements AdvancedCallableResolverInterface
     /**
      * {@inheritdoc}
      */
+    #[\Override]
     public function resolve($toResolve): callable
     {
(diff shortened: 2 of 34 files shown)


s-edy (approved):

akrabat:
This seems useful as a way to spot when the method of a parent class has been changed and the children haven't been all updated. 

I  have fixed-up the PR to be current and enabled the relevant PHPStan configuration so that CI would pick it up.

The only thing I'm unsure of is whether this is a problem we have that wouldn't be caught in another way.

odan:
I think in most cases we define interfaces to avoid such issues. If a method or the signature does not fit, PHP (and PHPStan) will report this anyway. Tests should also cover such mistakes. So I don´t see a huge profit for now.

akrabat:
> I think in most cases we define interfaces to avoid such issues. If a method or the signature does not fit, PHP (and PHPStan) will report this anyway. Tests should also cover such mistakes. So I don't see a huge profit for now.

I agree. While reflecting on it while playing with it it seems to me that  `#[\Override]` is most useful for methods in a class where the parent is in another package  and isn't covered by an interface.

Internally within Slim we should not be in a situation where we are override a non-interface method in a child class.

akrabat:
I'm closing this based on the discussion. This is no reflection on your work @sno-windy and entirely related to what we want in the project. 

Thank you for working with Slim and I hope that this doesn't put you off contributing in the future.

PR #3410, opened 2025-09-02
Slim 2.x compatible with PHP8.3

junior-paytech:
Added compatibility with PHP 8.3

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,3 +1,17 @@
+#Attention
+This fork was intended to update version 2.x of Slim Framework just to make it compatible with PHP 8.x
+
+## Changes made in this fork was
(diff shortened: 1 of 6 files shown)


junior-paytech:
## Changes made in this PR was
* Add explicit return types into methods that implements interface \ArrayIterator, \IteratorAggregate and Countable
* Removed use of magic_quotes_gpc

**Files Changed**
/Slim/Environment.php
/Slim/Helper/Set.php
/Slim/Middleware/Flash.php
/Slim/Http/Response.php
/Slim/Http/util.php

odan:
Hi @junior-paytech, the 2.x branch is in maintenance mode only.
PHP 8.x support is available only in Slim 4, so a migration is required.

PR #3424, opened 2026-02-11
Add Conditional Middleware Support

rakibdevs:
Added Conditional Middleware, a new PSR-15 compliant middleware that enables conditional execution of middleware based on request properties. 

**Use Cases**
1. Path-based Middleware Application
```php
// Authentication only for admin routes
$app->add(
    ConditionalMiddleware::forPath($authMiddleware, '/admin')
);
```
2. Method-based Middleware Application
```php
// CSRF protection only for state-changing methods
$app->add(
    ConditionalMiddleware::forMethods($csrfMiddleware, ['POST', 'PUT', 'PATCH', 'DELETE'])
);
```
3. Combined Conditions
```php
// Rate limiting only for API POST/PUT requests
$app->add(
    ConditionalMiddleware::forPathAndMethods(
        $rateLimitMiddleware,
        '/api',
        ['POST', 'PUT']
    )
);
```
4. Attribute-based Conditions
```php
// Audit logging only for authenticated users
$app->add(
    new ConditionalMiddleware(
        $auditMiddleware,
        fn($request) => $request->getAttribute('user') !== null
    )
);
```
5. Complex Business Logic
```php
// Apply middleware based on multiple criteria
$app->add(
    new ConditionalMiddleware(
        $specialMiddleware,
        fn($request) => 
            $request->getAttribute('subscription') === 'premium' &&
            !str_starts_with($request->getUri()->getPath(), '/public')
    )
);
```

### DIFF
--- a/Slim/Middleware/ConditionalMiddleware.php
+++ b/Slim/Middleware/ConditionalMiddleware.php
@@ -1,0 +1,130 @@
+<?php
+
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/5.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
+namespace Slim\Middleware;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Conditional Middleware
+ *
+ * This middleware allows you to conditionally execute another middleware based on
+ * a given condition (callable). If the condition returns true, the wrapped middleware
+ * is executed. If it returns false, the request is passed directly to the next handler.
+ *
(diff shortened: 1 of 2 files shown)


odan:
Hi @rakibdevs Thanks for the PR.

Adding middleware for paths is already fully supported by defining a route or route group and then calling the “add” or “addMiddleware” method on it. I therefore see no additional benefit in adding multiple ways to achieve the same thing. 
I will therefore close this PR.

mikespub:
Hi @odan 

as you mentioned the specific forPath() and forMethod() methods may overlap with what you can define via routes, but I believe there's still a valid use case for the generic callable() approach.

In particular, the concept of allowing you to re-use existing middleware as is, and only invoke them when certain conditions apply in a standard way seems interesting to me.

Those conditions could go beyond the route-based criteria (path, method) or what each middleware itself allows you to configure (custom rules), and this lets you define those conditions in a consistent way across any middlewares...

Now whether that should become part of Slim itself or be published as a separate middleware is a choice I'll leave to the Slim team ;-)

PR #3426, opened 2026-02-25
Ensure that Nyholm PSR17 factory is fully available

andrewnicols:
We currently use Guzzle for PSR17/PSR7.

We're also now adding in open-telemetry (`open-telemetry/exporter-otlp`), which makes use of `nyholm/psr7-server`, but it does not depend on `nyholm/psr7`.

As a result we now have the Nyholm server request creator, but not the response Factory class.

`\Slim\Factory\Psr17\Psr17Factory::isServerRequestCreatorAvailable()` returns a truthy value for Nyholm, but when `getServerRequestCreator()` is called, it errors out.

The Nyholm PSR7 packages are split into a server package, and a client package. Both are required for the NyholmPsr17Factory to be able to return an instance of the `ServerRequestCreator` class.

This change amends the `NyholmPsr17Factory:: isServerRequestCreatorAvailable()` method to also check if the response factory class provided by the client package is available.

I can't think of a way of testing this with automated testing. For manual testing you can install `nyholm/psr7-server` without `nyholm/psr7` -- before this change we were hitting an exception.

### DIFF
--- a/Slim/Factory/Psr17/NyholmPsr17Factory.php
+++ b/Slim/Factory/Psr17/NyholmPsr17Factory.php
@@ -33,4 +33,15 @@ class NyholmPsr17Factory extends Psr17Factory
 
         return new ServerRequestCreator($serverRequestCreator, static::$serverRequestCreatorMethod);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function isServerRequestCreatorAvailable(): bool
+    {
+        return (
+            parent::isServerRequestCreatorAvailable()
+            && class_exists(static::$responseFactoryClass)
+        );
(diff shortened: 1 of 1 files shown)


odan:
Thanks for the PR. It looks like coveralls.io is not available at the moment: https://status.coveralls.io/ 
I'll review the PR again once it's back online.

andrewnicols:
Hi @odan,

Coveralls is back up - any chance of a review?

Thanks

odan:
Hi @andrewnicols Please check this issue: https://github.com/slimphp/Slim/actions/runs/22377160579/job/65296930027?pr=3426

PR #3430, opened 2026-03-20
Extend RouteCollectorProxy::redirect() to serve non-GET requests too

uhle:
While `RouteCollectorProxy::redirect()` is quite handy and useful, it only works for GET requests.  Therefore I propose to extend it by another argument to set the accepted HTTP methods.  Its default value should be `['GET']` so that legacy code can run unchanged.  Then it would be possible to also use it like that:
```php
$app->redirect('/books', '/library', 307, ['POST']);
```

### DIFF
--- a/Slim/Interfaces/RouteCollectorProxyInterface.php
+++ b/Slim/Interfaces/RouteCollectorProxyInterface.php
@@ -120,7 +120,10 @@ interface RouteCollectorProxyInterface
     /**
      * Add a route that sends an HTTP redirect
      *
-     * @param string|UriInterface $to
+     * @param string $from The route URI pattern
+     * @param string|UriInterface $to The redirect URI
+     * @param int $status The HTTP status code
+     * @param string[] $methods Numeric array of HTTP methods
      */
-    public function redirect(string $from, $to, int $status = 302): RouteInterface;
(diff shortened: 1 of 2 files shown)


EmadAlmahdi:
This feels correct 👍

odan:
Thanks for the PR. I understand the use case, but I don’t think we should extend (and break RouteCollectorProxyInterface)  `redirect()` for this.

`redirect()` is currently a small convenience helper for simple GET/browser redirects. For non-GET requests, especially when the original method and body should be preserved, the behavior should be explicit because the correct status code matters. A `302 Found` redirection may cause clients to change a redirected `POST` request to `GET` if method preservation is required, `307 Temporary Redirect` or `308 Permanent Redirect` should be used instead.

Slim already provides the necessary "tools" for this through regular routes, `map()`, and middleware. For example:

```php
$app->map(['POST'], '/books', function ($request, $response) {
    return $response
        ->withHeader('Location', '/library')
        ->withStatus(307);
});
```

or

```php
$app->post('/books', function ($request, $response) {
    return $response
        ->withHeader('Location', '/library')
        ->withStatus(307);
});
```

If an application is accidentally redirecting non-GET requests because of misrouting or configuration, that should be fixed at the application level.

For these reasons, I would prefer to keep `redirect()` GET-only and use explicit routes or middleware for non-GET redirects.

PR #3432, opened 2026-03-24
Add BasePathMiddleware for subdirectory deployments

KalimeroMK:
This PR adds a new `BasePathMiddleware` that automatically strips the base path from request URIs, allowing Slim applications to run in subdirectories without modifying route definitions.

---

### Typical Scenario:
```apache
# Apache configuration with Alias
Alias /myapp /var/www/myapp/public
```

```php
// Route is defined as:
$app->get('/api/users', function ($request, $response) {
    return $response->withJson(['users' => []]);
});

// But request comes to:
// GET /myapp/api/users

// Result: ❌ 404 Not Found
// Slim looks for exact match with /myapp/api/users
// But route is defined as /api/users
```

### Current workarounds are inelegant:
- Manual path stripping before adding to Slim
- `.htaccess` hacks that don't work on all servers
- Defining routes with variable prefix: `$app->get($basePath . '/api/users', ...)`

---

## 💡 Solution

`BasePathMiddleware` strips the base path before the request reaches the router:

```
Request: GET /myapp/api/users
    ↓
BasePathMiddleware (/myapp)
    ↓
Slim Router sees: /api/users ✅
    ↓
Route match: /api/users → 200 OK
```

---

### Option 1: Explicit Base Path

When you know the exact base path:

```php
<?php
// index.php

use Slim\Factory\AppFactory;
use Slim\Middleware\BasePathMiddleware;

$app = AppFactory::create();

// Add the middleware first in the stack!
$app->add(new BasePathMiddleware('/myapp'));

// Now routes work without prefix
$app->get('/api/users', function ($request, $response) {
    return $response->withJson(['users' => []]);
});

$app->run();
```

**Apache configuration:**
```apache
Alias /myapp /var/www/html/myapp/public
<Directory /var/www/html/myapp/public>
    AllowOverride All
    Require all granted
    
    <IfModule mod_rewrite.c>
        RewriteEngine On
        RewriteCond %{REQUEST_FILENAME} !-d
        RewriteRule ^ index.php [QSA,L]
    </IfModule>
</Directory>
```

---

### Option 2: Auto-Detection

When base path varies (dev vs production):

```php
<?php
// index.php

use Slim\Middleware\BasePathMiddleware;

$app = AppFactory::create();

// Auto-detect from $_SERVER['SCRIPT_NAME']
// If it's /myapp/public/index.php → /myapp
$app->add(BasePathMiddleware::fromRequest($request));

$app->get('/api/users', ...);
$app->run();
```

---

### Option 3: Environment Variable

```php
$app->add(new BasePathMiddleware($_ENV['APP_BASE_PATH'] ?? ''));
```

---

### Access Base Path in Routes

The middleware stores the original base path in request attributes:

```php
$app->get('/api/users', function ($request, $response) {
    $basePath = $request->getAttribute('basePath'); // '/myapp'
    
    // Useful for URL generation
    $url = $basePath . '/api/users/123';
    
    return $response->withJson([
        'basePath' => $basePath,
        'url' => $url
    ]);
});
```

---

### Shared Hosting (cPanel, etc.)

```php
// If application is in /public_html/myapp/
$app->add(new BasePathMiddleware('/myapp'));
```

### Docker with Subdirectory

```yaml
# docker-compose.yml
nginx:
  image: nginx:alpine
  volumes:
    - ./nginx.conf:/etc/nginx/conf.d/default.conf
```

```nginx
# nginx.conf
location /api/ {
    alias /var/www/public/;
    try_files $uri $uri/ /index.php?$query_string;
    
    location ~ \.php$ {
        fastcgi_pass php:9000;
        fastcgi_param SCRIPT_FILENAME $request_filename;
        include fastcgi_params;
    }
}
```

```php
// index.php
$app->add(new BasePathMiddleware('/api'));
```

### Multi-tenant Application

```php
// For /client1/ prefix
$app->add(new BasePathMiddleware('/client1'));

// All routes are accessible at /client1/api/...
```

---

## 🔄 Backwards Compatibility

- ✅ **Fully backwards compatible**
- ✅ No breaking changes
- ✅ Optional middleware
- ✅ Doesn't affect existing routes

---

### Unit Tests
```bash
vendor/bin/phpunit tests/Middleware/BasePathMiddlewareTest.php
```

### Integration Testing
```bash
# Apache configuration for testing
Alias /testapp /path/to/test/public

# Start application
php -S localhost:8080 -t public/

# Test
curl http://localhost:8080/testapp/api/users
```

---
## 🎯 Goal of This PR

The goal of this PR is to:

1. **Simplify deployment** in subdirectory environments
2. **Lower barrier** for using Slim on shared hosting
3. **Standardize** a solution for the most common 404 error problem
4. **Enable** clean route definitions without hardcoded prefixes
---

## 🙏 Thank You

Hope this middleware helps the Slim community simplify their deployments! 🚀

### DIFF
--- a/Slim/Middleware/BasePathMiddleware.php
+++ b/Slim/Middleware/BasePathMiddleware.php
@@ -1,0 +1,126 @@
+<?php
+
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/4.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
+namespace Slim\Middleware;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+use function array_pop;
+use function dirname;
+use function end;
+use function explode;
+use function implode;
+use function is_string;
+use function ltrim;
+use function rtrim;
+use function str_starts_with;
+use function strlen;
+use function substr;
+
+/**
+ * Middleware to handle applications running in a subdirectory.
+ *
+ * Strips the base path from the request URI so routes can be defined
+ * without the base path prefix.
+ *
+ * @package Slim\Middleware
+ * @api
+ */
+class BasePathMiddleware implements MiddlewareInterface
+{
+    /**
+     * @var string The base path to strip (e.g., '/myapp')
+     */
+    private string $basePath;
+
+    /**
+     * @param string $basePath The base path to strip (e.g., '/myapp')
+     */
+    public function __construct(string $basePath = '')
+    {
+        $this->basePath = rtrim($basePath, '/');
+    }
+
+    /**
+     * Create middleware with auto-detected base path from request server params.
+     *
+     * Detects the base path from SCRIPT_NAME server parameter.
+     * If SCRIPT_NAME points to /public/index.php, it strips the /public part.
+     *
+     * @param ServerRequestInterface $request The request to detect base path from
+     * @return self
+     */
+    public static function fromRequest(ServerRequestInterface $request): self
+    {
+        $serverParams = $request->getServerParams();
+        $scriptName = $serverParams['SCRIPT_NAME'] ?? '';
+
+        if (!is_string($scriptName) || $scriptName === '') {
+            return new self('');
+        }
+
+        // Get the directory of the script
+        $basePath = rtrim(dirname($scriptName), '/');
+
+        // If the script is in /public directory, strip it
+        if (str_starts_with($basePath, '/')) {
+            $parts = explode('/', $basePath);
+            if (end($parts) === 'public') {
+                array_pop($parts);
+                $basePath = implode('/', $parts);
+            }
+        }
+
+        return new self($basePath);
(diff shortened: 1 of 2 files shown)


odan:
Hi!

Thanks for the PR. I reviewed it and noticed a few conceptual issues.

First, Slim already provides a way to set the basePath via `setBasePath()`, which also works with Slim's internal routing:

```php
$app = AppFactory::create();
$app->setBasePath('/myapp');
```

So this:

`$app->add(new BasePathMiddleware($_ENV['APP_BASE_PATH'] ?? ''));`

could already be written as:

`$app->setBasePath($_ENV['APP_BASE_PATH'] ?? '');`

And this:

`$app->add(new BasePathMiddleware('/myapp'));`

could already be written as:

`$app->setBasePath('/myapp');`

The auto-detection example will also not work in practice, because the `$request` object is only available after the application lifecycle has started. Therefore, it is not available during middleware registration.

So this will not work:

```php
$app = AppFactory::create();
$app->add(BasePathMiddleware::fromRequest($request)); // $request is not available here
```

Last but not least, this middleware sets a request attribute, but it does not set the Slim base path for the router.

Therefore, I would prefer to close this PR and recommend using [selective/basepath](https://github.com/selective-php/basepath), which works with Slim 4 when auto-detection for the base path is needed.

PR #3449, opened 2026-06-04
docs: remove $ prefix from bash snippets

ErfanMomeniii:
The prefix is a copy-paste hazard — pasting it into a shell produces `command not found: $` — and it was already applied inconsistently, since the Slim-Http install snippet omits it. Documentation-only change with no code impact.


ErfanMomeniii:
Hi @akrabat, please review it.

PR #3451, opened 2026-06-06
feat(routing): make request path decoding configurable

### LINKED ISSUE #3374: Routing does not match (slash %2F in path parameter)
When you have the following route:

```php
$router->any('/something/{magic}/{foo}', function (ServerRequestInterface $request, Response $response, array $args) use ($app) {
   // do something 
});
```

And Call this URL:

```
https://some.host/something/magic/foo%2Fbar
```

It will not match. I tried to check what went wrong. In `$_SERVER` I get:
```
["REQUEST_URI"]=>
  string(32) "/something/magic/foo%2Fbar"
```

It however matches when commenting out: https://github.com/slimphp/Slim/blob/4.x/Slim/Routing/RouteResolver.php#L42

Not sure why this is there tbh? If you want the route parameter to be decoded, shouldn't this be done after routing is done?

ErfanMomeniii:
The routing middleware decodes the request path with `rawurldecode()` before dispatch, which collapses `%2F` into `/` and prevents routes like `/something/{magic}/{foo}` from matching `/something/magic/foo%2Fbar`. This PR adds a decodePath parameter on `App::addRoutingMiddleware();` passing false skips that decoding, so encoded reserved characters survive routing and reach handlers untouched. The default remains true, so existing applications are unaffected.

Fixes #3374


ErfanMomeniii:
Hi @odan, please review it.

ErfanMomeniii:
Hi @odan , if you agree, I think it would be better to fix this in this tag.

odan:
The decoding currently happens in `FastRouteDispatcher::dispatch()` via `rawurldecode($uri)`. 

See here: https://github.com/slimphp/Slim/blob/5.x/Slim/Routing/FastRouteDispatcher.php#L30

Since the dispatcher implements the `DispatcherInterface` applications that need different path handling can provide their own dispatcher implementation and remove the `rawurldecode()` there and then register it in the DI container.

ErfanMomeniii:
Thanks for the review. One detail worth flagging is that there are currently two `rawurldecode()` calls in the routing pipeline, not one.  `RoutingMiddleware::process()` (line 49) decodes the path before passing it to the dispatcher, and `FastRouteDispatcher::dispatch()` (line 30) decodes it again. Because the middleware decodes first, a user-provided `DispatcherInterface` implementation alone can't fix #3374 — the encoded slashes are already collapsed by the time the custom dispatcher receives the URI.
@odan

ErfanMomeniii:
I'd suggest two possible directions here: either remove the redundant `rawurldecode()` from one of the two layers so a custom `DispatcherInterface` becomes the sole extension point (smaller diff, no new API, but users need to know to swap the dispatcher in DI), or keep it configurable on `RoutingMiddleware` as this PR does (slightly more surface, but discoverable directly from `addRoutingMiddleware()`). I prefer the configurable option for discoverability — most users hitting #3374 won't know the dispatcher is the extension point — but I'm happy to take whichever fits the 5.x direction better.

odan:
I guess you are talking about the logic in Slim 4, right? 
Because in the Slim 5 branch there should be no redundant `rawurldecode` anymore.

ErfanMomeniii:
You're right, I missed that the Security hardening commit (047b8d0d) on June 6 already removed the rawurldecode() from RoutingMiddleware::process(). My local merge happened around the same time, and I kept my own version of that line during the conflict resolution, which is why I was still seeing the duplicate in my branch. With 5.x as it stands, FastRouteDispatcher is the single decode point, and the DI-swap pattern you described works as expected for #3374. I close this MR. Thanks for the review.

PR #3452, opened 2026-06-06
5.x Middleware and routing security hardening

odan:
This PR improves middleware security by hardening JSON/XML body parsing, preventing XXE-style XML parsing risks and enforcing configured router base paths.

### Removed
- Removed `Slim\Middleware\CorsMiddleware` and the corresponding CORS middleware tests.

### Changed
- `BasePathMiddleware` now detects the base path and stores it on the router instead of passing it via a request attribute.
- `RoutingMiddleware` now reads base path only from the router.
- Requests outside a configured base path now return 404 instead of matching unprefixed routes.
- `MethodOverrideMiddleware` now only honors method overrides on original `POST` requests  because method override exists to support HTML-form style submissions, while allowing a GET or other method to become DELETE, PUT, or PATCH can bypass security assumptions in CSRF checks, proxies, caches, and route handling.
- `MethodOverrideMiddleware` now ignores unsupported override methods to prevent clients from injecting arbitrary or nonstandard HTTP verbs into downstream routing and middleware logic.
- JSON body parsing now always throws `HttpBadRequestException` for invalid JSON using `JSON_THROW_ON_ERROR`.
- XML body parsing now uses `LIBXML_NONET` and `LIBXML_NO_XXE` when available to make XML parsing safer by blocking external network access and, on PHP versions that support it, explicitly disabling XML External Entity processing.
That reduces the risk of XXE-style attacks, where a malicious XML payload tries to make the parser fetch remote resources or expose local files through external entities.
- Route registration APIs now return `RouteInterface` instead of concrete `Route`.
- 405 exception messages no longer include the allowed method list.
  - `Method not allowed.` instead of: `Method not allowed. Must be one of: GET, POST`

### Fixed
- Fixed base path auto-detection for subdirectory deployments.
- Fixed URL generation after base path auto-detection.
- Fixed routing behavior when the request URI does not start with the configured base path.
- Added regression coverage for invalid JSON, method override validation, and base path routing.


tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -25,29 +24,18 @@ final class BasePathMiddlewareTest extends TestCase
...
-            $basePath = $this->get(RouterInterface::class)->getBasePath();
-            $response->getBody()->write('basePath: ' . $basePath);
+        $app->get('/', function ($request, ResponseInterface $response) use ($app) {
+            $response->getBody()->write('basePath: ' . $app->getBasePath());
 
             return $response;
         });
 
         $serverParams = [
-            'REQUEST_URI' => '/',
```
llegaz, line 50: Hello @odan 

Are you really sure you want to get rid of these  `REQUEST_URI` server parameters ?
It seems a bit counterintuitive to me.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -63,28 +51,18 @@ public function testEmptyScriptName(): void
...
```
llegaz, line 87: Hello @odan 

Are you really sure you want to get rid of these  `REQUEST_URI` server parameters ?
It seems a bit counterintuitive to me.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -101,28 +79,18 @@ public function testScriptNameWithIndexPhp(): void
...
-            $basePath = $this->get(RouterInterface::class)->getBasePath();
-            $response->getBody()->write('basePath: ' . $basePath);
+        $app->get('/', function (ServerRequestInterface $request, ResponseInterface $response) use ($app) {
+            $response->getBody()->write('basePath: ' . $app->getBasePath());
 
             return $response;
         });
 
         $serverParams = [
-            'REQUEST_URI' => '/',
```
llegaz, line 125: Hello @odan 

Are you really sure you want to get rid of these  `REQUEST_URI` server parameters ?
It seems a bit counterintuitive to me.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -139,28 +107,18 @@ public function testScriptNameWithPublicIndexPhp(): void
...
-            $basePath = $this->get(RouterInterface::class)->getBasePath();
-            $response->getBody()->write('basePath: ' . $basePath);
+        $app->get('/', function ($request, ResponseInterface $response) use ($app) {
+            $response->getBody()->write('basePath: ' . $app->getBasePath());
 
             return $response;
         });
 
         $serverParams = [
-            'REQUEST_URI' => '/slim-hello-world/',
```
llegaz, line 163: Hello @odan 

Are you really sure you want to get rid of these  `REQUEST_URI` server parameters ?
It seems a bit counterintuitive to me.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -177,28 +135,18 @@ public function testSubDirectoryWithSlash(): void
...
-            $basePath = $this->get(RouterInterface::class)->getBasePath();
-            $response->getBody()->write('basePath: ' . $basePath);
+        $app->get('/foo', function ($request, ResponseInterface $response) use ($app) {
+            $response->getBody()->write('basePath: ' . $app->getBasePath());
 
             return $response;
         });
 
         $serverParams = [
-            'REQUEST_URI' => '/slim-hello-world/foo',
```
llegaz, line 201: Hello @odan 

Are you really sure you want to get rid of these  `REQUEST_URI` server parameters ?
It seems a bit counterintuitive to me.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
```diff
@@ -214,17 +162,11 @@ public function testSubDirectoryWithoutSlash(): void
...
     }
 
-    public function testSubDirectoryWithFooPath(): void
+    public function testStrictSubDirectoryWithFooPath(): void
     {
-        $definitions
-            = [
-                BasePathMiddleware::class => function (ContainerInterface $container) {
-                    $router = $container->get(RouterInterface::class);
+        $this->expectException(HttpNotFoundException::class);
```
llegaz, line 167: I don't get that "strict" and why it should get `OutsideOfBasePath` here with the base path still set to be `/slim-hello-world` why should this fire an exception ?

Sorry if I'm slow understanding but I don't get what you are trying to test. From my brief reviewal I don't get if what you are trying to achieve here is an upgrade or a downgrade of feature.

llegaz: Still I know you master your product and are probably right all the way I just want to point your attention here. Sometimes we tend to make the tests fit our code and won't follow what was philosophically contracted by the framework.

In all cases I'm pretty sure it serves URI rightfully and I would have prefer to test the code directly if I had a bit or more time.

It just hinted my curiosity while going through your PR that's why I'm asking.

llegaz (commented):
Just passing by and asking a few questions. Nice work.

Best regards,
Laurent

tests/Middleware/BasePathMiddlewareTest.php
odan: For the current implementation the BasePathMiddleware does not read `REQUEST_URI`, it reads the PSR request URI via: `$path = $request->getUri()->getPath();`.
In those tests, the path is already passed here: `->createServerRequest('GET', '/', $serverParams)` 
or ` ->createServerRequest('GET', '/slim-hello-world/?key=value', $serverParams)`
is **redundant** and could imply the middleware depends on `REQUEST_URI` which it not the case.

tests/Middleware/BasePathMiddlewareTest.php
odan: See my answer above.

tests/Middleware/BasePathMiddlewareTest.php
odan: This is because routing is strict about trailing slashes, `/foo/` does not match `/foo` so it becomes a 404.

llegaz: OK good sir so Slim 5 will forbid those URLs completely ? Compared to previous version that allowed to treat them as 2 separates and valid routes ? 

I didn't dive much in version 5 yet so sorry if I may ask questions on an obvious topic but it seems to me that this is a quite big move.

I'm not saying it's a bad move I just wonder why this choice ?

odan: No,  but this has been fixed now in the latest PR: https://github.com/slimphp/Slim/pull/3454

PR #3453, opened 2026-06-06
feat(middleware): add TrustedProxyMiddleware for reverse-proxy deploy…

ErfanMomeniii:
## Summary
 
Adds `TrustedProxyMiddleware` (PSR-15), which resolves the real client IP and external URI when Slim runs behind a reverse proxy. Without it, `REMOTE_ADDR` and the request URI reflect the proxy rather than the client, silently corrupting rate limiting, audit logs, geo-blocking, and absolute URL generation in any non-trivial deployment.

## Behavior

- Parses RFC 7239 `Forwarded` first; falls back to `X-Forwarded-For/Proto/Host/Port`.
- Walks the `for` chain right-to-left past trusted hops; returns the first untrusted address as the originating client.
- Publishes the result as the `client_ip` request attribute and rewrites the URI scheme/host/port via PSR-7 `withUri()`.
- Accepts trusted proxies as individual IPs or CIDR ranges, IPv4 or IPv6.

## Safety

Forwarded headers are honored only when `REMOTE_ADDR` matches the configured trusted-proxy whitelist. The list is empty by default, so the middleware is a no-op until explicitly opted in — it cannot introduce a spoofing vector through being installed.

## Usage

```php
$app->add(
    (new TrustedProxyMiddleware())->withTrustedProxies(['10.0.0.0/8', '172.16.0.0/12'])
);

// Downstream:
$ip  = $request->getAttribute('client_ip');

### DIFF
--- a/Slim/Middleware/TrustedProxyMiddleware.php
+++ b/Slim/Middleware/TrustedProxyMiddleware.php
@@ -1,0 +1,362 @@
+<?php
+
+/**
+ * Slim Framework (https://slimframework.com)
+ *
+ * @license https://github.com/slimphp/Slim/blob/5.x/LICENSE.md (MIT License)
+ */
+
+declare(strict_types=1);
+
+namespace Slim\Middleware;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Rewrites the request so it reflects the real client when the application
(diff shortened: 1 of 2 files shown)


ErfanMomeniii:
Hi @odan @akrabat, please review it. Thanks

odan:
At the moment, the v5 branch is still under development, and we do not plan to add more middleware to it for now. We try to keep the framework as lean as possible and let users install or implement additional middleware as needed.

That said this could certainly make a nice standalone composer package. I also like to mention that there is already a very similar package available: https://github.com/akrabat/ip-address-middleware

For these reasons I'm closing this PR. Thanks.

ErfanMomeniii:
understood @odan, I agree with you Thanks.

PR #3459, opened 2026-07-02
 Add StructArmed to QA 

samsonasik:
This PR adds StructArmed to github workflow static analysis for QA.

https://github.com/boundwize/structarmed

StructArmed is a configurable PHP architecture guard, that we can define layers and rules, then keep them enforced. 

For starter, this PR uses the PSR4 preset, and already found violations that fixed and included in this PR.

Since this repo support php 7.x as well, this only add on github workflow on php 8.2 👍

On more practical use case of how set the config layer -> ruleset for more complex use, you can take a look on how cakephp and codeigniter4 uses it :)

- https://github.com/cakephp/cakephp/blob/5.next/structarmed.php
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/structarmed.php

that can be in separate PR if this PR approved.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,4 +17,5 @@
 /phpstan.neon.dist  export-ignore
 /phpunit.xml.dist   export-ignore
 /psalm.xml          export-ignore
+/structarmed.php    export-ignore
 /tests              export-ignore
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,12 @@ jobs:
         if: matrix.analysis
         run: vendor/bin/phpstan
 
+      - name: Structarmed analysis
+        if: matrix.php == 8.2
+        run: |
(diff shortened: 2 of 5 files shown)


samsonasik:
Ready 👍

odan:
Thanks for the PR. 

We already have PHPStan (and Psalm), they cover PSR-4 just fine. Adding another QA tool that mostly overlaps isn't worth it.

> namespace Slim\Tests\Routing;

The fix is tiny, the only thing it caught was a namespace typo in a test file. Not enough reason to bring in a whole new tool.

If you want to send the namespace fix on its own, feel free to open a separate PR for that.

samsonasik:
Hi @odan , this PR is tiny on purpose for starter as I stated on the PR description, it can be more than that as architecture guard as I linked there as well that can be in next separate PR if you approve :)

PR #3461, opened 2026-07-29
Add query() route method for the HTTP QUERY verb (RFC 10008)

purwantoid:
## Summary

The HTTP `QUERY` method was recently published as [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008) on the IETF Standards Track. It is a safe, cacheable method that carries its query in the request body, enabling expressive queries without the URL-length and cacheability trade-offs of GET vs. POST. Support for `QUERY` is landing across the ecosystem, so this adds a `query()` helper as a sibling of the existing `get()`/`post()`/`put()`/`patch()`/`delete()`/`options()` route methods.

- Adds `query()` to `RouteCollectorProxyInterface` and its implementation in `RouteCollectorProxy`, equivalent to `map(['QUERY'], $pattern, $callable)`.
- Deliberately **not** added to `any()`'s method list, so existing `any()` routes are unaffected.
- Adds a CHANGELOG entry under `[Unreleased] / Added`.

```php
$app->query('/search', function (Request $request, Response $response) {
    $criteria = (string) $request->getBody();
    // ...run the query described in the request body...
    return $response;
});
```

## Test plan

- [x] `vendor/bin/phpunit` — full suite passes (437 tests)
- [x] `vendor/bin/phpcs` — clean (PSR-12)
- [x] `vendor/bin/phpstan --memory-limit=-1` — clean (level max)
- [x] Added `RouteCollectorProxyTest::testQuery()` (unit-level, mirrors `testOptions()`)
- [x] Added `AppTest::testQueryRoute()` (end-to-end, mirrors the GET/POST/etc. verb test)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Added
 
+- Add `query()` route helper for the HTTP `QUERY` method ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008))
+
 ### Changed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@
 
 ### Added
 
-- Add `query()` route helper for the HTTP `QUERY` method ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008))
-
 ### Changed
 
 ### Removed
--- a/Slim/Interfaces/RouteCollectorProxyInterface.php
+++ b/Slim/Interfaces/RouteCollectorProxyInterface.php
@@ -91,6 +91,16 @@ interface RouteCollectorProxyInterface
     public function options(string $pattern, $callable): RouteInterface;
 
     /**
+     * Add QUERY route
+     *
+     * @param string $pattern The route URI pattern
+     * @param callable|array{class-string, string}|string $callable The route callback routine
+     *
+     * @link https://www.rfc-editor.org/rfc/rfc10008
+     */
+    public function query(string $pattern, $callable): RouteInterface;
+
+    /**
      * Add route for any HTTP method
      *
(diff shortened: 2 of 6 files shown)


odan:
@akrabat I think we should add the query method for routing, but this would also be a breaking change in RouteCollectorProxyInterface. If we add it, we could use a new minor version (4.16.0). What do you think?

akrabat:
I agree we should support it. Do we have to put it in the interface? Can we use a dockblock for IDE autocomplete?

purwantoid:
@akrabat yes, we have add it in the interface.

akrabat:
> @akrabat yes, we have add it in the interface.

What I mean is that changing that interface breaks BC. Therefore we need another solution. 

Back in 4.2.0, we added `AdvancedCallableResolverInterface` which extends `CallableResolverInterface` as a way to add additional methods to an interface without breaking BC.

akrabat:
My guess is that we just need

```
@method RouteInterface query(string $pattern, callable|array{class-string, string}|string $callable) Add QUERY route, see https://www.rfc-editor.org/rfc/rfc10008
```

as a docbock entry on `RouteCollectorProxyInterface` for LLMs/IDE auto complete.

akrabat:
I have pushed my suggestion.

purwantoid:
@akrabat thanks.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Added
 
+- Add `query()` route helper for the HTTP `QUERY` method ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008))
```
odan: Not needed here.

purwantoid: @odan remove it

PR #3465, opened 2026-08-23
Substitute invalid UTF-8 in JSON error renderer output

iliaal:
`json_encode()` returns `false` when any string in the payload contains an invalid UTF-8 sequence, so a JSON error response for an exception whose message or title carries raw binary data (invalid-UTF-8 user input echoed into exception messages is the common path) was emitted as a zero-length body under `Content-Type: application/json`. Strict API clients then fail to parse the empty body and the actual error information is silently discarded.

This adds `JSON_INVALID_UTF8_SUBSTITUTE` so invalid byte sequences are replaced with the UTF-8 replacement character instead of failing the whole encode. The renderer output stays valid JSON in all cases.

Reproduction on 4.x before this change:

```php
$renderer = new \Slim\Error\Renderers\JsonErrorRenderer();
$output = $renderer(new \RuntimeException("bad \xB1\x31 bytes"), true);
var_dump(strlen($output)); // int(0)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Fixed
+ - Substitute invalid UTF-8 in JSON error renderer output (#811)
 
 ### Added
 
--- a/Slim/Error/Renderers/JsonErrorRenderer.php
+++ b/Slim/Error/Renderers/JsonErrorRenderer.php
@@ -16,6 +16,7 @@ use Throwable;
 use function get_class;
 use function json_encode;
 
+use const JSON_INVALID_UTF8_SUBSTITUTE;
 use const JSON_PRETTY_PRINT;
 use const JSON_UNESCAPED_SLASHES;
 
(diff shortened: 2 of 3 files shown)


akrabat:
I changed line 41 of JsonErrorRenderer back to `JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES` and the test didn't fail.

There's a couple of problems with the test.
 
First, the `false` in `$output = $renderer->__invoke($exception, false);` means that the error details are not rendered, and so only the title ("Slim Application Error") is encoded, so the JSON returned is:

    {
        "message": "Slim Application Error"
    }
    
which doesn't include the message we're looking for.

Secondly asserting that the key `message` exists in the array doesn't test that the invalid UTF-8 is correctly encoded in the JSON.

Can you update the test to prove that the change works?

iliaal:
`displayErrorDetails=false` only encoded the title, so the invalid UTF-8 never reached `json_encode`. I changed the test to pass `true` and assert the decoded exception message is `"Invalid \u{FFFD}1 UTF-8 sequence"`. That assertion fails if `JSON_INVALID_UTF8_SUBSTITUTE` is dropped.

akrabat (approved):
LGTM

PR #3466, opened 2026-08-23
Use consistent HTML escaping in error details fragment

iliaal:
`HtmlErrorRenderer::renderExceptionFragment()` escapes the exception message and stack trace with bare `htmlentities()`, while the title and description are escaped with explicit `htmlspecialchars(..., ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8')`. The two escape sites therefore behave differently across the supported PHP range: `htmlentities()` default flags gained `ENT_QUOTES` only in PHP 8.1, so on 7.4/8.0 single quotes in the message pass through unescaped, and the charset is inherited from the `default_charset` ini rather than pinned to UTF-8.

This aligns the fragment escaping with the same explicit call used everywhere else in the renderer, making behavior identical on every supported PHP version.

### DIFF
--- a/Slim/Error/Renderers/HtmlErrorRenderer.php
+++ b/Slim/Error/Renderers/HtmlErrorRenderer.php
@@ -32,11 +32,7 @@ class HtmlErrorRenderer extends AbstractErrorRenderer
             $html .= '<h2>Details</h2>';
             $html .= $this->renderExceptionFragment($exception);
         } else {
-            $description = htmlspecialchars(
-                $this->getErrorDescription($exception),
-                ENT_QUOTES | ENT_SUBSTITUTE,
-                'UTF-8'
-            );
+            $description = $this->escapeHtml($this->getErrorDescription($exception));
             $html = "<p>{$description}</p>";
         }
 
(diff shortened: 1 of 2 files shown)


Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -51,14 +50,20 @@ private function renderExceptionFragment(Throwable $exception): string
         $code = $exception->getCode();
         $html .= sprintf('<div><strong>Code:</strong> %s</div>', $code);
 
-        $html .= sprintf('<div><strong>Message:</strong> %s</div>', htmlentities($exception->getMessage()));
+        $html .= sprintf(
+            '<div><strong>Message:</strong> %s</div>',
+            htmlspecialchars($exception->getMessage(), ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8')
+        );
 
         $html .= sprintf('<div><strong>File:</strong> %s</div>', $exception->getFile());
```
odan: File is still emitted without HTML escaping

Slim/Error/Renderers/HtmlErrorRenderer.php
```diff
@@ -51,14 +50,20 @@ private function renderExceptionFragment(Throwable $exception): string
...
 
         $html .= sprintf('<div><strong>Line:</strong> %s</div>', $exception->getLine());
```
odan: Is still emitted without HTML escaping

Slim/Error/Renderers/HtmlErrorRenderer.php
iliaal: Agreed. `getFile()` is `string` and a path can contain `<>&"'`. I'll escape it with the same `htmlspecialchars(..., ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8')` call as message and trace.

Slim/Error/Renderers/HtmlErrorRenderer.php
iliaal: `getLine()` is `int`, so escaping is a no-op. I'll still add it for uniformity.

The real sibling is Type: anonymous class names embed the file path. I'll escape Type (and Code, which has no return type on `Throwable`) in the same pass.

PR #3467, opened 2026-08-23
v5 - Add query() route method for HTTP QUERY (RFC 10008)

odan:
Add a `query()`  method on `App`, `Router`, and `RouteGroup` for the HTTP `QUERY` verb ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html)).

`QUERY` is a safe, idempotent method that may include a request body, so it is a better fit than `GET` for searches and other parameterized lookups.

## Usage

```php
$app->query('/search', function ($request, $response) {
    return $response;
});


odan:
@akrabat Please check and approve for v5.

akrabat (approved):

PR #3468, opened 2026-08-26
Cache only the last getAllowedMethods() lookup

kaii-k:
`FastRouteDispatcher::getAllowedMethods()` memoizes its result per requested URI in an instance property with no cap, TTL, or eviction. On dispatchers that outlive a single request — persistent-worker runtimes (FrankenPHP worker mode, RoadRunner, Swoole) or any long-lived process reusing one `App`/`Dispatcher` instance (queue workers, CLI daemons, test/load harnesses) — an attacker can grow this cache without bound by requesting distinct, unmatched URIs, exhausting process memory over time.

This caps the cache at 1000 entries with oldest-first eviction. It's a no-op for the standard php-fpm-per-request model, since the dispatcher never survives past one request there.

Reported privately via security@slimframework.com — this PR is the fix discussed there.

### Test plan
- Added `testGetAllowedMethodsCacheIsBounded` reproducing unbounded growth against the cap.
- Full test suite passes (436 tests).
- phpcs / phpstan clean on changed files.

### DIFF
--- a/Slim/Routing/FastRouteDispatcher.php
+++ b/Slim/Routing/FastRouteDispatcher.php
@@ -15,18 +15,23 @@ use FastRoute\Dispatcher\GroupCountBased;
 class FastRouteDispatcher extends GroupCountBased
 {
     /**
-     * Maximum number of URIs to memoize in $allowedMethods.
+     * The URI that $allowedMethods was computed for.
      *
-     * Dispatcher instances can outlive a single request (e.g. persistent-worker
-     * runtimes such as FrankenPHP worker mode, RoadRunner, Swoole, or long-running
-     * CLI/queue processes that reuse one App/Dispatcher). Without a cap, requesting
-     * distinct URIs indefinitely grows this cache with no eviction, allowing
-     * unbounded memory consumption from client-controlled input.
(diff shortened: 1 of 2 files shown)


akrabat:
Looking at this, I'm not sure that we actually need the array of allowedMethods at all. We only actually need one URI and one list of allowed methods.

kaii-k:
Good point you're right, we don't need a history of URIs at all. `getAllowedMethods()` is only called once per `dispatch()` for the current URI, and across dispatches the cache only pays off when the same URI repeats back-to-back, so a single-entry memo covers that with no cap/eviction logic needed.

Pushed a commit replacing the bounded map with a single `$allowedMethodsUri` / `$allowedMethods` pair (last-URI memo). It's bounded by construction rather than by a configurable cap, which is simpler and removes the `MAX_ALLOWED_METHODS_CACHE_SIZE` constant and eviction branch entirely. Updated the regression test accordingly (asserts the memo holds only the most recent URI after many distinct lookups, plus a test that re-requesting an earlier URI recomputes rather than returning a stale hit). Full suite still green (437 tests), phpcs/phpstan clean.

akrabat (approved):
Thanks for the update.

